PR #6, opened 2022-04-15
Bind super+/ and super+a to launcher/applibrary

ids1024:
Also adds these packages to `Recommends` of Debian package.

These bindings don't give keyboard focus the created window, which is awkward. We also want to somehow make the bindings close the window if open.


Drakulix:
Initial focus has been fixed on master.
Toggling the window is out-of-scope for cosmic-comp. This should be handled through global shortcuts, handle by another component (likely something like cosmic-session), through e.g. this upcoming portal: https://github.com/flatpak/xdg-desktop-portal/pull/711

Drakulix (approved):
Thanks!

PR #16, opened 2022-06-09
Fix setuid

13r0ck:


Drakulix (approved):
Good catch! likely a copy&paste error

13r0ck:
A copy and paste solution too!

PR #20, opened 2022-06-27
allow keyboard grab in x11 mode

h1771t:


h1771t:
This PR depends on this one landing too: https://github.com/pop-os/smithay/pull/2

h1771t:
Closing this since resetting modifiers is not the right fix as noted here: https://github.com/Smithay/smithay/issues/646#issuecomment-1168633910

PR #23, opened 2022-07-15
feat: use applet host to activate launcher and app library

wash2:


wash2:
oops didn't see the log

Issue #29, opened 2022-08-07
[lemp10] Right-click with touchpad doesn't work


jacobgkau:
Version: [d493b90](https://github.com/pop-os/cosmic-comp/commit/d493b9067c45e304649cf9cba07f0077f1253fee)

On lemp10 running with Wayland, two-finger click does not right-click within apps (e.g. Nautilus or Firefox). Right-click with a USB mouse does work.

Three-finger click for middle click also does not work; middle-clicking with a USB mouse works.

(I'm making an educated guess this is the correct repo for an input issue like this, feel free to transfer if not.)

PolyMeilex:
Sounds like a lack of `libinput`'s [ClickMethod::Clickfinger](https://docs.rs/input/latest/input/enum.ClickMethod.html#)

I suppose setting it with [Device::config_click_set_method](https://docs.rs/input/latest/input/struct.Device.html#method.config_click_set_method) would do the job.

Drakulix:
This is more a lack of a configuration-utility (so [cosmic-settings](https://github.com/pop-os/cosmic-settings), than an issue for cosmic-comp. Though cosmic-comp currently lacks a way to configure this at runtime (the configuration is currently only read once on launch).

There is already a config file at `{XDG_STATE_DIR}/cosmic-comp/inputs.ron` (on my system that is `~/.local/state/cosmic-comp/inputs.ron`, that get populated with default settings on the first run.

Here you should be able to find your touchpad and set `click_method` to `Some(Clickfinger)` to enable two-finger right click.
@jacobgkau Can you confirm that works for your system?

jacobgkau:
I have `~/.local/state/cosmic-comp/inputs.ron` and there are two `click_method` lines (one for `PNP0C50::00 04F3:311D Touchpad` and one for `ETPS/2 Elantech Touchpad`). Both were set to `Some(ButtonAreas)` by default.

Changing them both to `Some(Clickfinger)` doesn't seem to work; I tried logging out and back in, then rebooting, and while my changes to the file were preserved, the touchpad still right-clicks with the bottom right and not with two-finger click.

Drakulix:
Hmm, that is weird, it does work on my system..
Is there anything in your logs by chance? Otherwise I will have to add some more debug-printing for applying the input configuration.

jacobgkau:
[...]

PR #30, opened 2022-08-09
config: Set click_method for input devices

Drakulix:
Fixes https://github.com/pop-os/cosmic-comp/issues/29

While the setting was correctly read-out from the device, changes were not applied as the necessary code for this particular setting was just missing..

@jacobgkau Can you confirm this fixes your problem?

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -431,6 +431,16 @@ impl Config {
                         );
                     }
                 }
+                if let Some(method) = config.click_method {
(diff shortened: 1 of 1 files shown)


jacobgkau (approved):
Please don't include slashes or underscores in branch names (or just use `_jammy`/`_nobuild` as necessary), it's easier to test things when CI builds them.

Confirmed this gets two-finger click working as configured on lemp10.

PR #37, opened 2022-09-20
Update smithay and implement shortcuts inhibit protocol

ids1024:
We may want to restrict what applications can use this protocol in some way? I guess the ecosystem is still figuring out how to restrict which Wayland protocols sandboxed (Flatpak) apps can use. For now, always allow.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -20,7 +20,10 @@ use smithay::{
     output::Output,
     reexports::wayland_server::{protocol::wl_surface::WlSurface, DisplayHandle},
     utils::{Buffer, Logical, Point, Rectangle, Size, SERIAL_COUNTER},
-    wayland::shell::wlr_layer::Layer as WlrLayer,
+    wayland::{
+        keyboard_shortcuts_inhibit::KeyboardShortcutsInhibitorSeat,
+        shell::wlr_layer::Layer as WlrLayer,
+    },
 };
 
 use std::{cell::RefCell, collections::HashMap};
@@ -173,6 +176,19 @@ impl State {
 
                 let device = event.device();
                 for seat in self.common.seats.clone().iter() {
+                    let current_output = active_output(seat, &self.common);
+                    let workspace = self.common.shell.active_space_mut(&current_output);
+                    let shortcuts_inhibited = workspace
+                        .focus_stack(seat)
+                        .last()
+                        .and_then(|window| {
+                            seat.keyboard_shortcuts_inhibitor_for_surface(
+                                &window.toplevel().wl_surface(),
+                            )
+                        })
+                        .map(|inhibitor| inhibitor.is_active())
+                        .unwrap_or(false);
+
                     let userdata = seat.user_data();
                     let devices = userdata.get::<Devices>().unwrap();
                     if devices.has_device(&device) {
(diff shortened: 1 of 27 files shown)


src/input/mod.rs
```diff
@@ -249,15 +265,22 @@ impl State {
...
+                                                userdata
+                                                    .get::<SupressedKeys>()
+                                                    .unwrap()
+                                                    .add(&handle);
+                                                return FilterResult::Intercept(Some(
+                                                    action.clone(),
+                                                ));
+                                            }
                                         }
                                     }
```
Drakulix, line 285: Maybe we want to have an `else` here in the meantime? Just one complicated (maybe even hard-coded) shortcut to release the lock for the user?

As stated in the protocol:
> The Wayland compositor is however under no obligation to disable all of its shortcuts, and may keep some special key combo for its own use, including but not limited to one allowing the user to forcibly restore normal keyboard events routing in the case of an unwilling client. The compositor may also use the same key combo to reactivate an existing shortcut inhibitor that was previously deactivated on user request.
>  When the compositor restores its own keyboard shortcuts, an "inactive" event is emitted to notify the client that the keyboard shortcuts inhibitor is not effectively active for the surface and seat any more, and the client should not expect to receive all keyboard events.
>
> When the keyboard shortcuts inhibitor is inactive, the client has no way to forcibly reactivate the keyboard shortcuts inhibitor.
> 
> The user can chose to re-enable a previously deactivated keyboard shortcuts inhibitor using any mechanism the compositor may offer, in which case the compositor will send an "active" event to notify the client.

src/wayland/handlers/keyboard_shortcuts_inhibit.rs
```diff
@@ -0,0 +1,22 @@
...
+    },
+};
+
+impl KeyboardShortcutsInhibitHandler for State {
+    fn keyboard_shortcuts_inhibit_state(&mut self) -> &mut KeyboardShortcutsInhibitState {
+        &mut self.common.keyboard_shortcuts_inhibit_state
+    }
+
+    fn new_inhibitor(&mut self, inhibitor: KeyboardShortcutsInhibitor) {
+        // TODO: Restrict what apps can inhibit shortcuts
```
Drakulix, line 17: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/68 should allow hiding a protocol to specific apps.

But even then, I would argue it should be exposed by default for new applications, so we can show a permission dialog on first-use. We can decline any request anyway.

Drakulix (approved):
Looks good. Slight nitpick

src/wayland/handlers/keyboard_shortcuts_inhibit.rs
ids1024: For Flatpak, ideally "inhibit system shortcuts" could be listed as one of the permissions an app requires when you install it, just as permissions are shown on install on Android.

src/input/mod.rs
ids1024: Yeah, I was thinking it could have some shortcut that works anyway. I wonder if there is anything that would be obvious to a user trying to get out of a program with shortcuts inhibited... not sure what other compositors are doing.

PR #39, opened 2022-09-21
fix: return if output is none

wash2:
Sometimes recently when I moved a window to a different workspace it would cause cosmic-comp to crash, and I noticed that it was because of the unwrap here. I've just made it return instead of unwrapping, which seems to have solved the issue, but maybe there is a better fix.


Drakulix (approved):
Looks like a correct fix to me

PR #40, opened 2022-09-21
Fix position window is rendered during drag with multiple outputs

ids1024:
This has been annoying me.

Moving tiling windows between monitors seems to work nicely with this change. Floating windows are positioned incorrectly when moved to another monitor, but this is existing behavior.


ids1024:
Okay, the position of floating windows after dragging between monitors seems to be correct too now.

Drakulix (approved):
Thanks! Looks good

Issue #36, opened 2022-09-15
Panic when xkb configuration is wrong


ids1024:
```
xkbcommon: ERROR: Couldn't find file "symbols/en" in include paths
xkbcommon: ERROR: 1 include paths searched:
xkbcommon: ERROR:       /usr/share/X11/xkb
xkbcommon: ERROR: 3 include paths could not be added:
xkbcommon: ERROR:       /home/ian/.config/xkb
xkbcommon: ERROR:       /home/ian/.xkb
xkbcommon: ERROR:       /etc/xkb
xkbcommon: ERROR: Abandoning symbols file "(unnamed)"
xkbcommon: ERROR: Failed to compile xkb_symbols
xkbcommon: ERROR: Failed to compile keymap
[...]
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/input/mod.rs:187:30
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This should probably be handled more gracefully, perhaps defaulting to something that should work (as long as the necessary xkb data files exist at all).

Drakulix:
Where exactly does this crash happen?

On current master, the result of `add_keyboard` is simply [ignored](https://github.com/pop-os/cosmic-comp/blob/master_jammy/src/input/mod.rs#L127).
It is definitely very debatable, if that is a proper solution, but I fear this back trace might be a bit old?

PR #41, opened 2022-09-23
input: Don't ignore keyboard init failures

Drakulix:
Fixes https://github.com/pop-os/cosmic-comp/issues/36

@ids1024 Is this roughly, what you had in mind?
Instead of silently discarding the returned error, cosmic-comp tries to recover by loading a standard config now and afterwards crashes explicitly, instead of some time later due to a missing `KeyboardHandle`.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -13,7 +13,7 @@ use smithay::{
     },
     desktop::{layer_map_for_output, Kind, WindowSurfaceType},
     input::{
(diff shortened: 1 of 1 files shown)


ids1024 (approved):
Yep, this seems better. This should be good enough at least until we figure out how we're going to handle keyboard configuration. (It should also be possible to change the keymap of the compositor while it's running, etc.)

PR #46, opened 2022-10-20
new shell

Drakulix:
Finally compiles again and displays a window, after ripping out half of it and putting some reworked parts back in.

Roadmap from here (will get filled out as more problems arise and can be tested):

Needs a lot of debugging some immediately noticeable problems:
- [x] Sometimes it gets `EGL_BAD_DISPLAY` in the logs, corresponds with damage getting broken.
- [x] Focus doesn't shift to new windows.
- [x] New tiled windows seem to be always added at the beginning of the tree, no matter where there is focus.
- [x] Pointer location in tiled windows seems to be off.
- [x] Kms backend damage issues
- [x] layer-shell popups don't appear to be clickable - merged https://github.com/Smithay/smithay/pull/805
- [x] ext-workspace activate request ~~does nothing~~ is broken in multi-monitor setups
- [x] sometimes monitors don't initialize correctly
  - [x] on startup
  - [x] on hotplugging

Missing features for parity with current master:
- [x] No grabs yet. Needs move and resize grabs at least
  - [x] Resize floating
  - [x] Resize tiling
  - [x] Move floating
  - [ ] Move tiling
- [x] Maximize/unmaximize is not handled
- [x] No support for changing orientation for tiling
- [x] Moving window requests are not handled
- [x] No capture support, needs to wire up new screencopy protocol
- [x] Debug-View feature doesn't compile yet and needs some rework anyway - first steps in smithay-egui done, blocked by https://github.com/Smithay/smithay/pull/807

Things to look out for:
- [x] Test new workspace code for both Dynamic-Workspace sets and static sized ones:
  - [ ] Test different workspace modes, see if swapping them at runtime produces desired results.
  - [x] Test if unplugging outputs handles workspaces as expected
- [ ] Focus In/Out requests are likely broken currently, wire up and check how they behave
- [ ] Tiling Tree updates should be waiting for client acknowledgements of any new sizes. 
- [ ] ...

... probably a lot more.


ids1024:
The workspaces protocol doesn't seem to be working correctly. All the workspaces it returns have the name `"1"`.

Drakulix:
*Should* be fixed by the latest commit (untested though).

ids1024:
https://github.com/pop-os/cosmic-comp/pull/46/commits/b0715ae3bee3a84c4f854f0331f51d0aae1afaae seems to cause the cursor to leave a trail on the screen.

Running with the kms backend it seems to sometimes hang on start before it shows anything on the screen. This isn't related to the latest commits.

Drakulix:
> https://github.com/pop-os/cosmic-comp/pull/46/commits/b0715ae3bee3a84c4f854f0331f51d0aae1afaae seems to cause the cursor to leave a trail on the screen.
> 
> 
> 
> Running with the kms backend it seems to sometimes hang on start before it shows anything on the screen. This isn't related to the latest commits.

I am aware, but thanks for reporting. This whole branch is still in a bit of flux, I am sorry. But the nested modes should be good enough for testing.

ids1024:
The screencopy implementation doesn't seem to handle shm pools correctly. The compositor crashes if the pool size isn't the exact length of the buffer. This can be seen by modifying `screenshot-screencopy.rs` increasing the size passed to `set_len()` and `create_pool()`.

ids1024:
Also, it seems the shm screencopy format it advertises as `Argb8888` is actually given `Rgba` image data?

Drakulix:
> The compositor crashes if the pool size isn't the exact length of the buffer.
> Also, it seems the shm screencopy format it advertises as Argb8888 is actually given Rgba image data?

Should both be fixed by [3662050](https://github.com/pop-os/cosmic-comp/pull/46/commits/3662050a9861058ab28c5a95666d84c72d86b112).

The correct format is now advertised and multiple buffers in a single pool (no matter how large) should be handled correctly now.
The example still crashes, when simply increasing the pool size, but on the client-side as the image-encoder library is not expecting a larger image slice and the example just copies the whole pool file descriptor.

ids1024:
Well, it now reports `ShmFormat::Abgr8888`, but still seems to be RGBA data.

With Pipewire we can reports a video format of ABGR or ARGB but OBS at least doesn't seem to connect to the stream when it reports those formats, but does with RGBA. So presumably we want to at least support RGBA here.

Repeatedly capturing frames with `wl_shm` seems to sometimes lead to laggy behavior? Somehow that isn't entirely consistent.

ids1024:
Anyway, https://github.com/pop-os/xdg-desktop-portal-cosmic/tree/screencopy_jammy has shm screencopy with OBS working. Need to properly integrate with Pipewire to negotiate dmabuf if supported by the client, dmabuf modifiers supported by client, mouse cursor metadata, etc.

Drakulix:
> Well, it now reports `ShmFormat::Abgr8888`, but still seems to be RGBA data.


That is correct. wl_shm formats are given in little endian.
Given on obs interface you might need to convert them.

See e.g. https://github.com/Smithay/smithay/blob/master/src/backend/renderer/gles2/mod.rs#L795-L801

ids1024:
Ah, okay. Seems a bit unusual to do it that way, but converting is easy enough.

And I see https://wayland.app/protocols/wayland#wl_shm:enum:format does indeed define these as "little endian" unlike a lot of things in Wayland that are "native endian" so it won't pose a complication for big endian systems. (Wii U port of Cosmic? :laughing: )

ids1024:
Compositor seems to segfault if the memfd buffer isn't large enough. For instance, removing the `set_len` call in `screenshot-screencopy.rs`.

Drakulix:
> Compositor seems to segfault if the memfd buffer isn't large enough. For instance, removing the `set_len` call in `screenshot-screencopy.rs`.

I guess there is validation missing in smithay here. cosmic-comp crashes during the copy call, at this point the request has passed two asserts, checking that the length reported by smithay (which happily created the slice for us) is indeed large enough. So without access to the underlying file descriptor I am not sure what the compositor could do here. Please fill an issue at smithay so we don't forget to (try and) fix this.

ids1024:
It seems `wl_buffer:Release` is never sent with screencopy (is that expected to be send with screencopy, or is it just redundant with ready/failed?). And shm screencopy leaks memory until the compositor crashes if I don't explicitly call `WlBuffer::destroy()`.

ids1024:
`OnDamage` seems to only sometimes work?

PR #53, opened 2022-12-31
Add Nix flake.

wsx-udscbt:
First of all, many thanks to the maintainers for this exciting project!
This PR adds a Nix flake, which allows Nix/NixOS users to quickly test out and hack on `cosmic-comp`. The Nix flake currently supports:
1. Building the `cosmic-comp` binary with `nix build`.
2. Running `cosmic-comp` with `nix run`.
3. Use `nix develop` to enter a development environment with the Rust toolchain and all the build-time/run-time dependencies installed.

I expect the maintenance burden to be small, as the flake still utilize the `cargo` build system underneath, and hence reusing the dependencies declared in `Cargo.toml` and `Cargo.lock`.
Updating other dependencies amounts to running `nix flake update`.

If it aligns with the project, I am also willing to add flakes to all the other components of COSMIC, with the goal of providing a nice integration with NixOS in the future (https://github.com/NixOS/nixpkgs/issues/199563).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ vendor
 vendor.tar
 .vscode
 *.log
+result
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,162 @@
+{
+  "nodes": {
+    "crane": {
+      "inputs": {
+        "flake-compat": "flake-compat",
+        "flake-utils": "flake-utils",
+        "nixpkgs": [
+          "nixpkgs"
+        ],
+        "rust-overlay": "rust-overlay"
+      },
+      "locked": {
+        "lastModified": 1672095661,
+        "narHash": "sha256-7NTsdCn3qsvU7A+1/7tY8pxbq0DYy1pFYNpzN6he9lI=",
+        "owner": "ipetkov",
+        "repo": "crane",
+        "rev": "98894bb39b03bfb379c5e10523cd61160e1ac782",
+        "type": "github"
+      },
+      "original": {
+        "owner": "ipetkov",
+        "repo": "crane",
+        "type": "github"
+      }
+    },
+    "fenix": {
+      "inputs": {
+        "nixpkgs": [
+          "nixpkgs"
+        ],
+        "rust-analyzer-src": "rust-analyzer-src"
+      },
+      "locked": {
+        "lastModified": 1672468395,
+        "narHash": "sha256-eIrQGHZub98IiLqtFXa00ooHBCjtKY+rfKJs5lhyF/c=",
+        "owner": "nix-community",
+        "repo": "fenix",
+        "rev": "d6f9bdfb4a25395b15f6c6feba3d0bba5e62b3ce",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nix-community",
+        "repo": "fenix",
+        "type": "github"
+      }
+    },
+    "flake-compat": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1668681692,
+        "narHash": "sha256-Ht91NGdewz8IQLtWZ9LCeNXMSXHUss+9COoqu6JLmXU=",
+        "owner": "edolstra",
+        "repo": "flake-compat",
+        "rev": "009399224d5e398d03b22badca40a37ac85412a1",
+        "type": "github"
+      },
+      "original": {
+        "owner": "edolstra",
+        "repo": "flake-compat",
+        "type": "github"
+      }
+    },
+    "flake-utils": {
+      "locked": {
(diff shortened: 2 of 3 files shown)


colemickens:
very cool, was able to launch cosmic-comp under Hyprland. I'd love to see more packages, a hm module, but I don't want to be too greedy :D.

Drakulix:
> This PR adds a Nix flake, which allows Nix/NixOS users to quickly test out and hack on cosmic-comp.

Super cool, thanks a lot for the PR. But I have to admit, that I know nothing about nix build system.

> I expect the maintenance burden to be small, as the flake still utilize the cargo build system underneath, and hence reusing the dependencies declared in Cargo.toml and Cargo.lock.

Sounds good. What would require a nix-flake update? Any Cargo.toml/lock dependency change? In that case it might be nice to automate this somehow? Is an outdated nix-flake useful for anyone?

> If it aligns with the project, I am also willing to add flakes to all the other components of COSMIC, with the goal of providing a nice integration with NixOS in the future (https://github.com/NixOS/nixpkgs/issues/199563).

In general we definitely want to try our best at supporting packaging efforts of COSMIC in general for other distributions. However we are not yet sure, how we will handle versioning and QA testing that.

We might end up doing git-builds only for Pop-OS and do frequent versioned releases on the [`cosmic-epoch`](https://github.com/pop-os/cosmic-epoch) repository, which should always contain a set of commits for all the different components, that were tested together. In that case (again nothing final), it might make more sense to centralize build files for other distros there instead. Those would probably also not need to be updated as frequently as the master-branches of all these projects.

Anyhow, given that cosmic-comp definitely can be developed and tested on its own, I am willing to merge this, if I can refer to you for updates and general maintenance on nix-support.

wsx-udscbt:
> Sounds good. What would require a nix-flake update? Any Cargo.toml/lock dependency change? In that case it might be nice to automate this somehow? Is an outdated nix-flake useful for anyone?

Rust crate dependencies are still handled entirely in `Cargo.toml`/`lock`, and changing them does not require updating the flake.
The flake takes cares of the dependencies outside of Cargo (`libEGL`, `libxkbcommon`, etc, that one typically `apt install` before building).
Hence we need to update `flake.nix` accordingly if any such dependency changes, or the build process itself changes, and I expect those to be not so often.

> In general we definitely want to try our best at supporting packaging efforts of COSMIC in general for other distributions. However we are not yet sure, how we will handle versioning and QA testing that.
> 
> We might end up doing git-builds only for Pop-OS and do frequent versioned releases on the [`cosmic-epoch`](https://github.com/pop-os/cosmic-epoch) repository, which should always contain a set of commits for all the different components, that were tested together. In that case (again nothing final), it might make more sense to centralize build files for other distros there instead. Those would probably also not need to be updated as frequently as the master-branches of all these projects.

Thanks for the bigger picture.
My understanding is that `flake.nix` here is for those that wish to test out the latest upstream changes.
And for a more stable versioned release, the build files typically will be included and maintained in the https://github.com/NixOS/nixpkgs repo, where better integrated testing can be done by the nixpkgs maintainers.
That said, it is still convenient to setup a CI build here based on `flake.nix`, which can perform some basic testing (compiling and `cargo test`).

> Anyhow, given that cosmic-comp definitely can be developed and tested on its own, I am willing to merge this, if I can refer to you for updates and general maintenance on nix-support.

Thank you! I am definitely willing to maintain the Nix support.
Is it OK for me to add a CI workflow building and testing the Nix flake in a separate PR?
Or are you using a CI system outside of Github?

Drakulix:
> Is it OK for me to add a CI workflow building and testing the Nix flake in a separate PR?

Totally fine. I am gonna merge this then.

> Or are you using a CI system outside of Github?

We do, but that just builds nightly deb-packages for the repo, so that has nothing to do and should not conflict with any additional github actions.

colemickens:
I just looked, hoping maybe you'd added one or two more but I'm seeing now you've added 9+ more. That is awesome @wsx-udscbt , I can't say thank you enough! I'd be curious to see where/how you're using this and experimenting...

Thank you again!

wsx-udscbt:
I'm running a sway on NixOS setup and am pretty new to cosmic.
All the flakes here are currently what I have and they are only for building the binaries for each components of cosmic.
I haven't figured out how to wire up all those components yet, and we still get a long way to go for integrating the entire cosmic environment in NixOS.

colemickens:
If you haven't, please consider opening a Discourse thread or GitHub Issue (nixpkgs or home-manager, or both) I suspect some others would chip in! I've been surprised at the number of users that nixpkgs-wayland gets, and I suspect there's... at least a few of us excitedly waiting for Cosmic.

PR #55, opened 2023-01-03
Setup Github CI.

wsx-udscbt:
The `build` workflow currently just compile the `cosmic-comp` binary according to `flake.nix`.
Currently it is set to trigger when there are new commits on `master_jammy` or any pull requests targeting it.

A potential thing to do is to setup a Cachix cache (cachix.org), which will cut down CI time and allow Nix users to avoid building the project locally if they choose to use the cache.

I've setup a similar cache on my other project (https://github.com/cosette-solver/cosette-prover), but the setup requires creating an account on cachix.org and setting an API key in the repository secrets.
Usage of the Cachix service is free up to a certain quota, and if any maintainer is willing to do this, I can update the workflow file accordingly to make use of the cache.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,18 @@
+name: "Build"
+on:
+  pull_request:
+    branches:
+    - master_jammy
+  push:
+    branches:
+    - master_jammy
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: cachix/install-nix-action@v18
+      with:
+        nix_path: nixpkgs=channel:nixpkgs-unstable
+    - run: nix build
+    - run: nix flake check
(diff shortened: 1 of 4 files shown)


Drakulix (approved):

Drakulix:
Lets see how bad the build times are first. Given that this is mostly intended for development and not for productive use (for now), I don't think caching is of top priority. I will revisit this decision later, when we get ready for proper releases. Thanks for the links and the PR! :)

wsx-udscbt:
In my forked repo, the CI runs for `cosmic-comp` is about 10 mins.
The `cosmic-applets` component seems to take a lot of ram when building locally though, so caching for the entire cosmic environment might be desirable in the future.
But I agree that we can revisit this matter when everything is settled down.
Thank you!

Drakulix (approved):

PR #69, opened 2023-02-10
Exclude "workspace-overview" from workspace screencopy

ids1024:
Not especially elegant, but simple enough. And works, mostly.

This doesn't capture correctly for the current workspace. I guess it's rendered only once, and the render for display on the screen has the layer shell surface? (There's a comment about this for cursors, which are also captured when they shouldn't be.)

### DIFF
--- a/src/wayland/handlers/screencopy.rs
+++ b/src/wayland/handlers/screencopy.rs
@@ -5,6 +5,7 @@ use std::{
 };
 
 use anyhow::anyhow;
+use calloop::LoopHandle;
 use cosmic_protocols::screencopy::v1::server::zcosmic_screencopy_session_v1::{
     FailureReason, InputType,
 };
@@ -24,7 +25,7 @@ use smithay::{
             Bind, Blit, BufferType, ExportMem, ImportAll, ImportMem, Offscreen, Renderer,
         },
     },
-    desktop::space::SpaceElement,
+    desktop::{layer_map_for_output, space::SpaceElement},
     output::Output,
     reexports::wayland_server::{
         protocol::{wl_buffer::WlBuffer, wl_shm::Format as ShmFormat, wl_surface::WlSurface},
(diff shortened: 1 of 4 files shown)


ids1024:
> Basically the if condition on line 1177 would have to check if layer_map_for_output(o).layers() contains a LayerSurface with the excluded workspace.

Ah, just changing that if condition is easy enough.

But changing that, `on_damage` screencopy of the current workspace ends up flickering between capturing the layer shell surface and capturing what's behind it? It seems to have no issues without `on_damage`.

src/wayland/handlers/screencopy.rs
```diff
@@ -1144,8 +1149,30 @@ impl State {
         output: &Output,
     ) -> Option<Vec<(Session, BufferParams)>> {
         let workspace = self.common.shell.active_space_mut(output);
-        if !workspace.pending_buffers.is_empty() {
-            Some(std::mem::take(&mut workspace.pending_buffers))
+        let mut pending_buffers = std::mem::take(&mut workspace.pending_buffers);
+
+        let mut i = 0;
```
Drakulix, line 1154: Looks like the `drain_filter`-pattern? :sweat_smile: 

Could you add a `TODO: drain_filter` comment here? I have not yet given up on this feature becoming stable at some point!

Drakulix (approved):
Looks good now!

src/wayland/handlers/screencopy.rs
ids1024: Maybe it would be good to have a `utils::drain_filter` function.

Or there's the the micro-crate solution of just depending on https://crates.io/crates/drain_filter_polyfill.

Drakulix: yes, I have thought about that. But that should be done in another PR imo.

PR #74, opened 2023-02-15
wip: layer surface popup positioning improvements

wash2:
Layer surface popups should probably only be unconstrained if the popup is not contained in the output geometry. 

Replacing the offset check with `output_contains_popup` seems to prevent the bottom panel's popups from being flipped / re-positioned off-screen. Popups which still are not contained in the output geometry are still unconstrained, as can be seen in the right panel popups, some of which may be flipped off-screen while unconstraining them. I'll work on some more improvements so that doesn't happen.


Drakulix:
For reference, fixed by https://github.com/pop-os/cosmic-comp/commit/dead320252f5e51ccaedcfbb12e3be9495e8dac4

PR #76, opened 2023-02-17
Draft: initial implementation for ext-foreign-toplevel-list-v1

i509VCB:
Pending on https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/187

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -429,6 +429,7 @@ dependencies = [
  "png",
  "puffin",
  "puffin_egui",
+ "rand",
  "regex",
  "renderdoc",
  "ron 0.7.1",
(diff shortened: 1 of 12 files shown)


ids1024:
The protocol was merged in wayland-protocols, so I suppose `src/wayland/protocols/ext_foreign_toplevel_list.rs` here could be moved into Smithay?

i509VCB:
How you implement this protocol is probably going to be quite compositor dependent. Hence why I am implementing directly in cosmic

ids1024:
We will need this for https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/124.

ids1024:
This was added in https://github.com/pop-os/cosmic-comp/pull/788.

PR #82, opened 2023-02-28
fix(init_logger): Directive::from_str expects warn

wash2:
I've also changed the log level from info to warn if there are no debug assertions for cosmic_comp & smithay. It seemed to have performance issues in electron apps for me otherwise.


Drakulix (approved):
A decent temporary work around. I already have an idea, what is causing the performance problems, I have been seeing as well.

Issue #62, opened 2023-01-24
No devices found when using QXL graphics driver


jackpot51:
Using `kvm -vga qxl`... and trying to run cosmic-comp fails.

ids1024:
I think it currently requires accelerated graphics, and worked when I tried it in virt-manager with 3D-enabled virtio video. But not QXL.

It needs to not fail without a drm renderer and instead fall back to llvmpipe.

PR #83, opened 2023-03-03
kms: Use primary node if no render node is found

ids1024:
It seems this is all that's needed to make cosmic-comp run in a VM without graphics acceleration.

Fixes https://github.com/pop-os/cosmic-comp/issues/62.


Drakulix:
Hmm, I would like to use a more explicit approach to initializing llvmpipe, than just letting gbm do it, but this is a good fallback for older drivers nevertheless.

PR #87, opened 2023-03-08
shell: Fix position of floating windows before/after drag

ids1024:
This fixes two issues that seem to have been introduced by 98dc342:

* During drag, CSD windows being offset from their original location and the frames rendered around them.
* Windows on additional monitors being positioned incorrectly when dragged.

This seems to work well in floating mode with `OutputBound` workspaces. With `Global` workspaces it seems position of dragged windows on additional monitors isn't right before or after this...

### DIFF
--- a/src/shell/layout/floating/grabs/moving.rs
+++ b/src/shell/layout/floating/grabs/moving.rs
@@ -66,7 +66,7 @@ impl MoveGrabState {
         elements.extend(AsRenderElements::<R>::render_elements::<I>(
             &self.window,
             renderer,
-            render_location.to_physical_precise_round(scale),
+            (render_location - self.window.geometry().loc).to_physical_precise_round(scale),
             scale,
         ));
         elements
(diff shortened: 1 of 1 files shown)


src/shell/layout/floating/grabs/moving.rs
```diff
@@ -66,7 +66,7 @@ impl MoveGrabState {
         elements.extend(AsRenderElements::<R>::render_elements::<I>(
             &self.window,
             renderer,
-            render_location.to_physical_precise_round(scale),
+            (render_location - self.window.geometry().loc).to_physical_precise_round(scale),
```
Drakulix, line 69: Right, we need that for the indicator to align, but does that actually work out with the map call in `ungrab` to the same location?

Drakulix (approved):

src/shell/layout/floating/grabs/moving.rs
ids1024: The previous version of the code was subtracting `self.window.geometry().loc`, and after this change dragging the window seems to work as expected, without anything jumping position at the start or end of drag. (Except with `Global` workspaces as mentioned; not sure what needs to be done there.)

Issue #58, opened 2023-01-10
Application unresponse due to not getting frame callback when moved on hotplug


ids1024:
* Run `cosmic-comp` with more than one monitor
* Open `gnome-terminal`, `foot`, etc. on one monitor
* Disable the monitor it is open on in `wdisplays`
* The workspace the application was on is moved to the display that is still enabled, but switching to that, the application seems slow to respond to input

Looking at `WAYLAND_DEBUG` logs of `foot`, it seems it isn't receiving a regular `frame` callback when this issue occurs, unlike when the application is running normally.

Looking at the code involved in sending frame callbacks, maybe the "primary scanout output" for the surface isn't being changed correctly?

Drakulix:
Do they get an output_enter event when the workspace is moved? In any case, super useful observation, I guess the workspace move logic is flawed here.

ids1024:
Hm, turning off my external monitor (while the internal monitor isn't enabled) the internal monitor starts working, but all my workspaces disappear into the void instead of moving. The applications seem to still be running though.

wash2:
I think this may also be affecting cosmic-panel. When I unplug a monitor with the panel running, it seems to get stuck when swapping buffers.

ids1024:
It is expected that occluded windows will not receive frame callbacks. `EglSwapBuffers` thus blocks potentially indefinitely, since it blocks on as frame callback (as implemented in Mesa), unless `EglSwapInterval` is set to `0`.

I guess the layer shell surface should receive `closed` from the compositor if the output it's on is disconnected? Not sure how that impacts Mesa's blocking on the frame callback. Probably best to set the swap interval to 0 and handle frame callbacks to drive rendering.

Drakulix:
> I guess the layer shell surface should receive closed from the compositor if the output it's on is disconnected? 

I pushed a commit, that does this, given it is definitely a good idea and a miss on my side.
But I am still investigating the original issue more.

ids1024:
I seem to be consistently seeing applications not receive frame callbacks when dragged to another monitor as well now, which wasn't previously happening.

ids1024:
It seems `default_primary_scanout_output_compare` is returning the `current_output`, since it has a larger `visible_area`. It seems to work better replacing that callback with one that always returns the `new_output`. Cosmic-comp seems to only show windows on the new output when moved, so I think that would be correct?

[...]

PR #88, opened 2023-03-10
Update primary scanout output when not visible on previous output

ids1024:
This seems to fix https://github.com/pop-os/cosmic-comp/issues/58. Though in the future the default behavior in Smithay may need some improvement.


Drakulix (approved):
Thanks for tracking this down!

ids1024:
I'm just happy if I don't have to restart applications or use them at 1 fps. :laughing:

PR #97, opened 2023-03-26
Fix latest window not getting focus when switching from a "fullscreen" workspace

### LINKED ISSUE #96: Latest window is not getting focus when switching from a "fullscreen" workspace
Scenario:
1. Open a window on workspace 1
2. Switch to workspace 2 and open another window
3. Go back to workspace 1 and make the window full screen
4. Switch to workspace 2 again
5. Try to type something
6. Keyboard focus is still on the window of workspace 1

Off topic: aside from a couple of bugs the compositor is already almost usable for my workflow. Great job, guys!

xDarksome:
Fixes #96

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -676,10 +676,7 @@ impl State {
                     0 => 9,
                     x => x - 1,
                 };
-                let _ = self
-                    .common
-                    .shell
-                    .activate(&current_output, workspace as usize);
+                Shell::activate_and_focus(self, seat, &current_output, workspace as usize)
             }
             Action::NextWorkspace => {
                 let current_output = seat.active_output();
@@ -690,7 +687,7 @@ impl State {
                     .active_num(&current_output)
                     .saturating_add(1);
                 // TODO: Possibly move to next output, if idx to large
-                let _ = self.common.shell.activate(&current_output, workspace);
+                Shell::activate_and_focus(self, seat, &current_output, workspace)
             }
             Action::PreviousWorkspace => {
                 let current_output = seat.active_output();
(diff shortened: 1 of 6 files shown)


Drakulix:
Hi!

Thanks for the issue and contribution.

While this fixes the error, I am a little concerned, that this is just covering a deeper problem, because we have logic to restore the focus, but it is clearly failing in this particular case.

More specifically this happens in [`Shell::refresh_focus`](https://github.com/pop-os/cosmic-comp/blob/master_jammy/src/shell/focus/mod.rs#L182).
Looking at it the reason for your issue seems pretty obvious. While we do compare the active window with the last element of the focus stack, we only do so for the `KeyboardFocusTarget:Element` variant. The `KeyboardFocusTarget::Fullscreen` variant ignores the focus stack, so it needs separate logic, but instead of unconditionally skipping, we should check if the surface in question is fullscreen on the current output and also fixup the focus if that is not the case.

Could you adjust the PR and check, if that solves the issue as well?

> Off topic: aside from a couple of bugs the compositor is already almost usable for my workflow. Great job, guys!

Thanks for kind words and detailed bug report btw. Much appreciated. 😊

xDarksome:
@Drakulix 
> we should check if the surface in question is fullscreen on the current output and also fixup the focus if that is not the case

Checking this condition wasn't enough, it was still breaking when switching from one fullscreen window to another.

Eventually I came up with the current solution. Could you look at the diff again, please? 
Also, at this point, maybe it would be better just to always "fix up" focus inside `Fullscreen` match arm.

xDarksome:
While testing this I also found another unrelated focusing bug: the compositor were losing keyboard focus when switching from empty to non-empty workspace.

I tried to fix this in this PR as well, but I'm not sure how optimal is the solution. Again, check it up please, the last commit in particular.

Drakulix:
> Eventually I came up with the current solution. Could you look at the diff again, please?

Looks good!

> Also, at this point, maybe it would be better just to always "fix up" focus inside Fullscreen match arm.

Nah, the check is less expensive as possibly sending an excess of focus events to clients. The solution is fine :+1: 

> I tried to fix this in this PR as well, but I'm not sure how optimal is the solution.

Looks good to me as well. Thanks for working on this!

PR #116, opened 2023-05-24
fix zombie processes from Action::Spawn

mmstick:
Each invocation of `Action::Spawn` was spawning a process without waiting on the child; resulting in an accumulation of zombie sh processes. This will spawn them in a background thread which waits on the child so that it can be reaped.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ endif
 
 VENDOR ?= 0
 ifneq ($(VENDOR),0)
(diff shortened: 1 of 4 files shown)


mmstick:
There's still one zombie process remaining, but I believe it's from [here](https://github.com/Smithay/smithay/blob/master/src/xwayland/xserver.rs#L502).

PR #122, opened 2023-05-25
Enable necessary protocols for supporting IMEs

ids1024:
I'm not entirely sure how text input focus should be managed: https://github.com/Smithay/smithay/pull/1038.

This "works" but is missing a bit of code to actually render the IME surface, and the code in anvil calling `set_point` (I'm not sure exactly what that is for or where/how it should be set here).


ids1024:
Updated with current API, and using `client_should_see_privileged_protocols`.

Drakulix (approved):

Issue #125, opened 2023-05-26
Unmaximized windows are not resized when tiled


mmstick:
With tiling enabled, unmaximizing a window moves it into the position of its node and is clipped to fit, but the clipped regions are still clickable. It's most noticeable if you toggle maximization of the left-most window and then try to click on the contents of any window to the right of it. It's fixable by triggering a resize by moving the window around.

Drakulix:
Input regions extending beyond tiles has been addressed in https://github.com/pop-os/cosmic-comp/commit/8dd8ac0492bcea102ba7b543f9bb9df21c438f07.
Missing resize has been added in https://github.com/pop-os/cosmic-comp/commit/e95bfeab53d83c40e9d244ba53419358cffe0e47.

Drakulix:
Firefox sometimes resets it's size when restoring a session, but that is likely not, what  you are seeing, if gedit and alacritty are also misbehaving. I'll investigate.

PR #131, opened 2023-06-08
shell: fix window size when unmaximize from tiling mode

13r0ck:
Steps to reproduce bug:
1. Enter tiling mode
2. Open 2+ windows
2. Maximize window (Super + m)
3. UnMaximize window (Super + m)
4. (Bug) the unmaximized window will render at the full size of the display, but be cropped to the size of the tiled area.

After this the window will render to the size of the tiling.

### DIFF
--- a/src/shell/workspace.rs
+++ b/src/shell/workspace.rs
@@ -300,6 +300,9 @@ impl Workspace {
             self.floating_layer.refresh();
             self.tiling_layer.recalculate(output);
             self.tiling_layer.refresh();
+            if let Some((_, _, rect)) = self.tiling_layer.windows().find(|(_, w, _)| w == window) {
+              window.set_geometry(rect);
+            };
             window.send_configure();
             self.fullscreen.retain(|_, w| w != window);
         }


Drakulix:
While this does indeed fix the issue (thanks for figuring that out!), the previous calls to the tiling layer (`recalculate` and `refresh`) should really fix this up and reset the geometry.

I will investigate, why this is not the case.

13r0ck:
I am confused?
Floating layout has it's own unmaximize function, and that too calls `window.set_geometry()`, so it isn't fair for tiling to do the same?

Drakulix:
> I am confused?
> 
> Floating layout has it's own unmaximize function, and that too calls `window.set_geometry()`, so it isn't fair for tiling to do the same?

Yes, but that is happening *inside* the floating layer.

I want the restoring to be handled inside the tiling layer as well and not fixed from the outside from the workspace.

The workspace is not responsible for sizing and layout, that is of concern of the layers. And to me the existing `recalculate` and `refresh` methods seem like a good fit.

13r0ck:
So the idea here is correct, just in the wrong location? would passing down `window` as one of the arguments be an acceptable solution? Or is there some other way to update the window's geometry?

mmstick:
This fixes #125

Drakulix:
> So the idea here is correct, just in the wrong location? would passing down `window` as one of the arguments be an acceptable solution? Or would you prefer `update_positions()` to iterate though `window`s and update there if they do not match?

The idea is correct, but `update_positions` should already iterate through the whole try and call `set_geometry` on every window, that doesn't match. So why is this fix helping? That is the true question.

Drakulix:
Figured it out, the `is_fullscreen`-check inside `recalculate` still returns true, because it is looking and the current_state, when it should be looking at the pending state, while the surface has not yet been configured again. I'll fix that in another PR.

Drakulix:
Fixed via https://github.com/pop-os/cosmic-comp/pull/132 instead.

PR #132, opened 2023-06-09
shell: Allow querying current vs pending state

### LINKED ISSUE #125: Unmaximized windows are not resized when tiled

Drakulix:
Superseeds #131, should fix #125.

Fixes the intent of the `is_maximized || is_fullscreen`-checks of the tiling layer, causing the geometry to be correctly restored by `recalculate`.

### DIFF
--- a/src/shell/element/mod.rs
+++ b/src/shell/element/mod.rs
@@ -247,14 +247,14 @@ impl CosmicMapped {
         }
     }
 
-    pub fn is_resizing(&self) -> Option<bool> {
+    pub fn is_resizing(&self, pending: bool) -> Option<bool> {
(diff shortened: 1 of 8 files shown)


Drakulix:
@13r0ck @mmstick Can you check, if this fixes #125 for you as well?

mmstick:
This definitely fixes #125.

Issue #17, opened 2022-06-21
Key bindings sometimes stop working


ids1024:
I'm not sure how exactly to figure out what's causing this, but after running cosmic-comp for a while, I sometimes am no longer able to switch workspaces. Super+num just types the number into the current window.

Drakulix:
Maybe the Super key gets stuck somehow? Maybe by swapping focus while holding it or something like that? (Just an educated guess, I did not check the code.)

ids1024:
Using cosmic-comp for several days now, I haven't been seeing this behavior for a while. So it appears to be resolved?

Closing.

ids1024:
Re-opening, since I seem to occasionally see this.

ids1024:
Huh, I see this currently, and wev shows `Mod2` as a locked modifier?

PR #135, opened 2023-06-29
Ignore caps/num lock in key bindings

ids1024:
I noticed https://github.com/pop-os/cosmic-comp/issues/17 again, but this time I realized num lock was on. Hopefully this is the only cause of that issue (and caps lock, but that is easier to notice).

I don't think we would event want either of these states to be part of a key binding. And `KeyModifiers` doesn't seem to be used anywhere else this would be relevant. So they can be ignored. This should be more similar to how keybindings are handled elsewhere.

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -903,8 +903,6 @@ pub enum KeyModifier {
     Alt,
     Shift,
     Super,
-    CapsLock,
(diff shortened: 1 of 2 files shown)


Drakulix (approved):
Nice catch!

PR #140, opened 2023-07-28
Use cosmic-config for input config; changeable at runtime

ids1024:
With this, it's possible to change `xkb-config` and `input-devices` and immediately see them effect the compositor.

In an earlier draft for input configuration I had a method:

```rust
pub fn register_handler<T: serde::de::DeserializeOwned + 'static>(
    &mut self,
    name: &'static str,
    cb: fn(T, &mut State),
)
```

But I'm not sure that makes the code less entangled without more changes.

https://github.com/pop-os/libcosmic/pull/126 would make a few things cleaner, but I'm still not sure on that design.

Leaving as a draft until cosmic-settings is updated to make use of this. And `input-devices` may not be the best way to configure input, but we'll need to think about the design there.

After input settings, settings from `config.ron` can also be migrated to cosmic-config and made changeable at runtime. `outputs.ron` can be dealt with later since we already have dynamic persistent output configuration that more or less works with `wdisplays`.


Drakulix:
Looks good to me code-wise.

> But I'm not sure that makes the code less entangled without more changes.

Me neither, seems clean enough to me anyway.

> And input-devices may not be the best way to configure input, but we'll need to think about the design there.

I think this is the biggest issue here. Looking at the cosmic-settings design, we need a way to configure defaults for a device type. I would like to retain the ability to special case the configuration of specific devices, though updating the defaults should probably also update all known devices, otherwise cosmic-settings UI wouldn't work as expected.

Thinking about this we might want to make a `HashMap<DeviceCapability, InputConfig>`?

> outputs.ron can be dealt with later since we already have dynamic persistent output configuration that more or less works with wdisplays.

Though this depends on the wlr-output-configuration protocol, which doesn't necessarily expose everything we want to make configurable in the long term. E.g. it only recently added variable refresh rate, so for cosmic-settings I think we want to use the cosmic-config, while keeping wlr-output-configuration for compatibility with other apps.

ids1024:
> Thinking about this we might want to make a HashMap<DeviceCapability, InputConfig>?

Would that group mouse and touchpad acceleration? I think the designs have those separate, as is commonly done.

Drakulix:
Right... So maybe we need to match on some property of the `UdevDevice` to figure out the device "type".

src/config/key_bindings.rs
```diff
@@ -0,0 +1,295 @@
...
+
+pub fn add_default_bindings(
+    key_bindings: &mut HashMap<KeyPattern, Action>,
+    workspace_layout: WorkspaceLayout,
+) {
+    let (workspace_previous, workspace_next, output_previous, output_next) = match workspace_layout
+    {
+        WorkspaceLayout::Horizontal => (
+            [KeySyms::KEY_Left, KeySyms::KEY_h],
+            [KeySyms::KEY_Right, KeySyms::KEY_j],
```
Drakulix, line 198:
```suggestion
            [KeySyms::KEY_Right, KeySyms::KEY_l],
```

Copy paste issue from my original code

src/config/key_bindings.rs
```diff
@@ -0,0 +1,295 @@
...
+            [KeySyms::KEY_Up, KeySyms::KEY_k],
+            [KeySyms::KEY_Down, KeySyms::KEY_j],
+        ),
+        WorkspaceLayout::Vertical => (
+            [KeySyms::KEY_Up, KeySyms::KEY_k],
+            [KeySyms::KEY_Down, KeySyms::KEY_j],
```
Drakulix, line 204:
```suggestion
            [KeySyms::KEY_Down, KeySyms::KEY_l],
```

See above

Drakulix:
So the solution is to just treat touchpads differently from everything else?

ids1024:
Yes. It's awkward, but I don't see a better way to do it with the current design we have for settings. Unless we want to just have individual settings like `touchpad-speed` instead of this general mechanism.

We can't really have settings per capability because a device can (in principle) have any number of capabilities. Libinput doesn't have "device types" even if things like Gnome Shell try to act like it does. I guess settings are largely tied to a particular capability, looking at libinput I didn't see any explicit association.

ids1024:
I still need to make use of this for mouse/touchpad settings in cosmic-settings and use that to test that its working as expected, but otherwise this PR should be good. Moving other settings to cosmic-settings can be done after this is merged.

Drakulix:
> I still need to make use of this for mouse/touchpad settings in cosmic-settings and use that to test that its working as expected, but otherwise this PR should be good. Moving other settings to cosmic-settings can be done after this is merged.

Sounds good. Please give me a ping once this is ready for final review and merge. I haven't found any glaring issues with the approach. Quite the opposite, I think this version looks quite clean! :)

ids1024:
After a couple small changes to `cosmic-comp-config`, and making it not spew `DeviceConfigError::Unsupported` trying to set default config for devices that don't support it, (and remembering to delete an old `input-devices` that was overriding the defaults), this seems to be working and suitable for use in cosmic-settings.

So probably no need to do more on this branch, and we can merge it.

Drakulix (approved):

PR #151, opened 2023-08-11
Send `SCANOUT` flag in dmabuf feedback when target_node != render_node

ids1024:
The Intel driver (at least) pads the stride to meet the requirements for scanout on AMD or Nvidia cards, when this flag is set.

This works around the bug I've seen in the Nvidia driver with corruption importing an unaligned dmabuf. This presumably is an issue in the Nvidia driver, but presumably it the fix would be for the import to fail, so using a stride that allows it to succeed is better anyway.


Drakulix:
Let it be known, that I hate doing this.

For reference: https://github.com/ids1024/egl-dmabuf-test
nvidia drivers (at least up to 535.104.05) don't reject buffers of strides they don't support.

PR #153, opened 2023-08-14
prelim-tablet-support

Vixea:
It's pretty basic, tested in another compositor to make sure nothing crashes. Probably shouldn't be merged anyway. Until I get the Inputs hooked up, probably going to ask some questions since I don't know this codebase perfectly.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -20,7 +20,8 @@ use smithay::{
     backend::input::{
         Axis, AxisSource, Device, DeviceCapability, GestureBeginEvent, GestureEndEvent,
         GesturePinchUpdateEvent as _, GestureSwipeUpdateEvent as _, InputBackend, InputEvent,
-        KeyState, PointerAxisEvent,
+        KeyState, PointerAxisEvent, TabletToolEvent, TabletToolButtonEvent, TabletToolTipEvent,
+        TabletToolTipState,
     },
     desktop::{layer_map_for_output, space::SpaceElement, WindowSurfaceType},
     input::{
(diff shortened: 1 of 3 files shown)


Vixea:
@Drakulix how do you update keyboard_focus in cosmic?

Drakulix:
> @Drakulix how do you update keyboard_focus in cosmic?

`Common::set_focus`. So if you have the state somewhere, you would call `data.common.set_focus` with a `KeyboardFocusTarget`, which will likely be a `CosmicMapped` in most cases (the latter implements `Into` for the former).

Vixea:
I now notice something wrong with the commit "implement TabletToolButton and TabletToolTip" but I noticed some inconsistencies with the has_device check sometimes it's before the debug stuff other times it's after does that matter or is it design choices between different sessions and or people?

Vixea:
yea as a whole this file seems inconsistent in style somethings have scoped uses then other InputEvents use global what would you like me to do @Drakulix

Drakulix:
> yea as a whole this file seems inconsistent in style

Oh yeah, it is also in need to be split into multiple files.

> inconsistencies with the has_device check sometimes it's before the debug stuff other times it's after does that matter or is it design choices between different sessions and or people?

Likely different sessions. From what I have been seeing in all those cases the code also could have been moved after the `has_device`-check. Which is a bit unnecessary anyway, given we add every device right now, but that might change in the future.

Debug stuff might be useful before the check to be always able to control the debug interface. But for example the shortcut-inhibitor code (while not harmful) shouldn't be.

>  somethings have scoped uses then other InputEvents use global what would you like me to do

Is this answered by the comment above? If not, could you go into more detail? Looking at your current code, that looks like a good approach to me.

Vixea:
Ok I'll continue to use global uses

Vixea:
Also for TabletToolProximity what focus is expecting is a little different it expects (WlSurface, Point,i32, Logical>)
Do you know exactly if anything that I noticed something with one of the previous comments? I'm not sure

Drakulix:
> Also for TabletToolProximity what focus is expecting is a little different it expects (WlSurface, Point,i32, Logical>)

What is that point used for?

Sounds like we could have a separate tablet focus, smithay doesn't really model that right now, but I think that is an oversight, because the tablet code is quite old. (Just like it doesn't model touch-input very well.) So maybe we want to add a `TabletTarget` to the `Seat` first.

> Do you know exactly if anything that I noticed something with one of the previous comments? I'm not sure

Sorry, I am not really parsing this sentence. Could you elaborate?

Vixea:
yea, I should bring that up within the matrix chat

Vixea:
Just to make sure I done it right force pushes after a rebase won't mess up any use of git blame?

Drakulix:
> Just to make sure I done it right force pushes after a rebase won't mess up any use of git blame?

No, that is totally fine and expected to be necessary after a rebase.

Vixea:
closing due to some interesting comments about pointer and tablet interactions that need to be figured out and a complete lack of interest from me

PR #154, opened 2023-08-30
WIP Security Context protocol

ids1024:
Requires https://github.com/Smithay/smithay/pull/1106.


ids1024:
Incorporated this into https://github.com/pop-os/cosmic-comp/pull/156 since it requires a Smithay update, which has non-trivial breaking API changes.

PR #156, opened 2023-09-05
Update smithay, add pointer gestures and security context

ids1024:
The boilerplate to add pointer gestures is a bit verbose, but not that complicated. And it works, for instance for pinch zooming in Firefox.

For building with the `debug` feature to work, https://github.com/Smithay/smithay-egui/pull/12 will be needed.


Drakulix (approved):
LGTM!

PR #165, opened 2023-09-07
Use cosmic-config for workspace settings

ids1024:
`WorkspaceAmount` now can be changed dynamically, which seems to basically work. We'll see how that works with `WorkspaceMode` and `WorkspaceLayout`.

If `WorkspaceLayout` can be changed at runtime, we'll have to update default keybindings in some way.

### DIFF
--- a/cosmic-comp-config/src/workspace.rs
+++ b/cosmic-comp-config/src/workspace.rs
@@ -1,0 +1,43 @@
+// SPDX-License-Identifier: GPL-3.0-only
+
+use serde::{Deserialize, Serialize};
+
+fn default_workspace_layout() -> WorkspaceLayout {
+    WorkspaceLayout::Vertical
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct WorkspaceConfig {
+    pub workspace_mode: WorkspaceMode,
+    pub workspace_amount: WorkspaceAmount,
+    #[serde(default = "default_workspace_layout")]
+    pub workspace_layout: WorkspaceLayout,
+}
+
+impl Default for WorkspaceConfig {
+    fn default() -> Self {
+        Self {
+            workspace_mode: WorkspaceMode::OutputBound,
+            workspace_amount: WorkspaceAmount::Dynamic,
+            workspace_layout: WorkspaceLayout::Vertical,
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub enum WorkspaceAmount {
+    Dynamic,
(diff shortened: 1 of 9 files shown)


src/shell/mod.rs
```diff
@@ -435,6 +429,12 @@ impl WorkspaceSet {
             }
         }
     }
+
+    pub fn update_amount(&mut self, amount: WorkspaceAmount) {
+        self.amount = amount;
+        // TODO
+        // - what if there are more workspace than amount?
```
Drakulix, line 436: Merge the last couple of workspaces. There should be code somewhere in this file, that does this (perhaps for the mode change? not sure).

src/config/mod.rs
```diff
@@ -530,6 +514,16 @@ fn config_changed(config: cosmic_config::Config, keys: Vec<String>, state: &mut
...
                 update_input(state);
             }
+            "workspaces" => {
+                let value = get_config::<WorkspaceConfig>(&config, "workspaces");
+                state
+                    .common
+                    .shell
+                    .workspaces
+                    .update_amount(value.workspace_amount);
+                // TODO workspace mode
```
Drakulix, line 524: Oh I was fearing the day, we would be able to change the mode at runtime. There is a completely untested `Shell::set_mode`-function. It will be fun to see how broken that is. :rofl:

cosmic-comp-config/src/workspace.rs
```diff
@@ -0,0 +1,43 @@
...
+    pub workspace_mode: WorkspaceMode,
+    pub workspace_amount: WorkspaceAmount,
+    #[serde(default = "default_workspace_layout")]
+    pub workspace_layout: WorkspaceLayout,
+}
+
+impl Default for WorkspaceConfig {
+    fn default() -> Self {
+        Self {
+            workspace_mode: WorkspaceMode::Global,
```
Drakulix, line 20: WorkspaceMode::OutputBound is much better tested right now. Maybe we want to default to that while Global mode gets more fixes.

Drakulix (commented):
Looks good on first glance.

Please don't forget to update the `config.ron` in the repo to not include these keys anymore.

ids1024:
Updating for the latest workspaces changes, this now works pretty well, once I found a way to make sure cosmic-workspace-unstable-v1 get notified when the number of workspaces is reduced.

The only limitation to dynamic changes is that `WorkspaceLayout` is used for setting the default key bindings, but that is only considered on start. I think that's fine for now, and this can be changed later when the key bindings are made dynamic and moved to cosmic-config.

Drakulix (approved):
LGTM!

PR #174, opened 2023-09-14
Pointer constraints

ids1024:
Depends on https://github.com/Smithay/smithay/pull/872.
    
For some reason, pointer confinement isn't working correctly here at the moment, even though the code matches Anvil...
    
May need solution to https://github.com/Smithay/smithay/issues/1126 for behavior to be correct.

### DIFF
--- a/src/shell/element/stack.rs
+++ b/src/shell/element/stack.rs
@@ -1169,6 +1169,27 @@ impl PointerTarget<State> for CosmicStack {
         }
     }
 
+    fn frame(&self, seat: &Seat<State>, data: &mut State) {
+        if let Some((location, serial, time)) = self
+            .0
+            .with_program(|p| p.last_location.lock().unwrap().clone())
+        {
+            self.pointer_leave_if_previous(seat, data, serial, time, location);
+        }
+
+        match self.0.with_program(|p| p.current_focus()) {
+            Focus::Header => PointerTarget::frame(&self.0, seat, data),
+            Focus::Window => self.0.with_program(|p| {
+                PointerTarget::frame(
+                    &p.windows.lock().unwrap()[p.active.load(Ordering::SeqCst)],
+                    seat,
+                    data,
+                )
+            }),
+            _ => {}
+        }
+    }
+
     fn leave(&self, seat: &Seat<State>, data: &mut State, serial: Serial, time: u32) {
         if let Some((location, serial, time)) = self
             .0
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3761,7 +3761,7 @@ checksum = "62bb4feee49fdd9f707ef802e22365a35de4b7b299de4763d44bfea899442ff9"
 [[package]]
 name = "smithay"
 version = "0.3.0"
-source = "git+https://github.com/smithay//smithay?rev=58d5bdc#58d5bdca0d65b5f45408201770da0ce86b1fa057"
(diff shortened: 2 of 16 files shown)


ids1024:
Updated to make use of https://github.com/Smithay/smithay/pull/1137. Seems to be working fairly well now. In earlier versions I had an issue with applications that I think the `frame` changes fix.

ids1024:
Hm, looks like this introduces an issue with moving the pointer between monitors...

ids1024:
Okay, just needed to move the `output_geometry` clamping to use the new output when the cursor has moved.

src/shell/element/stack.rs
```diff
@@ -1169,6 +1169,27 @@ impl PointerTarget<State> for CosmicStack {
...
+    fn frame(&self, seat: &Seat<State>, data: &mut State) {
+        if let Some((location, serial, time)) = self
+            .0
+            .with_program(|p| p.last_location.lock().unwrap().clone())
+        {
+            self.pointer_leave_if_previous(seat, data, serial, time, location);
+        }
+
+        match self.0.with_program(|p| p.current_focus()) {
+            Focus::Header => PointerTarget::frame(&self.0, seat, data),
```
Drakulix, line 1181: I am not convinced this works as expected. So if the focus swaps from the application to the header/stack-elements, we send a leave to the window, but a frame to the iced-element, which will just discard it.

So the wayland client won't get a frame for this leave event. I think we need to re-work this, so that the window always gets a leave event, if it got any pointer events. Same applies to decorated windows (so `src/shell/element/window.rs`).

Drakulix (commented):
Looks pretty good, just some interactions around compositor-drawn/managed objects, where I am not certain this is a 100% correct.

Vixea:
I'm 100% certain that it's not correct I tested the pr for the lols, it crashes whenever the cursor interacts with compositor related objects(not always in all cases though)

Vixea:
Ok figured out one reproduction case moving the pointer to the edge of the primary screen on the right side or in this case the side that doesn't have a second monitor

ids1024:
@Vixea Good catch. Looks like these a possible panic in the `clamp` call since I didn't put things in the right order there... that part is easy enough, anyway.

Vixea:
Huh, whenever clicking on windows that use pointer constraints when constrained the cursor jumps to look straight down(apps in question are games though maybe that is a relative pointer bug... hmm.

ids1024:
Hm, I had some weird behavior like that before, but it seemed to not be a problem in this PR. But I just tested OpenMW. I'll have to try some other things.

Vixea:
I used Minecraft though I've found some regressions that I think were caused by the fullscreen/maximize refractors

ids1024:
I guess I was just moving around and didn't even end up clicking in my quick test... seems there's an issue in Smithay here: https://github.com/Smithay/smithay/issues/1148.

That wasn't happening in anvil since it's missing a call to `frame`.

src/shell/element/stack.rs
ids1024: `leave` on the `WlSurface` still sends a `frame` in https://github.com/Smithay/smithay/blob/509030a151ad440f8406229f32d674811efd9007/src/wayland/seat/pointer.rs#L187-L190.

This is necessary since https://github.com/Smithay/smithay/blob/509030a151ad440f8406229f32d674811efd9007/src/desktop/wayland/window.rs#L330-L334 calls `take` on the surface in `leave`, so `frame` will not be delivered to surfaces that are part of a window after that.

So focusing window decorations sends a `frame` correctly after `leave`, though this ideally could be handled in a better way.

PR #176, opened 2023-09-21
wayland: map windows with correct geometry

alyssais:
Consider a simple client that doesn't use `set_window_geometry`, committing after attaching a buffer to a window surface.

`Window::geometry` falls back to `Window::bbox`, but when `Shell::map_window` called `Window::geometry`, `Window::bbox` hadn't been set yet, because that only happened later in `<State as CompositorHandler>::commit`, so the window would be mapped with a size of (0, 0).

This resulted in server-side decorations not being drawn for such clients until they were re-focused or resized.

The fix is to add an extra call to `Window::on_commit` before mapping it, so that its bbox is set correctly when `Shell::map_window` asks for its geometry.


Drakulix (approved):
Nice catch! Thanks

PR #184, opened 2023-09-28
Fix NextOutput and PreviousOutput when Shell::activate returns None

### LINKED ISSUE #160: Focus switching shortcuts do not traverse monitors
In the OutputBound workspaces mode, with three monitors arranged horizontally, I cannot use super+hjkl to move focus to windows on other monitors. However, the move shortcuts (super+shift+hjkl) work correctly

jackpot51:
This fixes #160, but I am not sure when Shell::activate is supposed to return a position versus when it is not.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -1187,19 +1187,25 @@ impl State {
                     .cloned()
                 {
                     let idx = self.common.shell.workspaces.active_num(&next_output).1;
-                    if let Ok(Some(new_pos)) = self.common.shell.activate(&next_output, idx) {
-                        seat.set_active_output(&next_output);
-                        if let Some(ptr) = seat.get_pointer() {
-                            ptr.motion(
-                                self,
-                                None,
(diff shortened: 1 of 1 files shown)


jackpot51:
When using the focus shortcuts, this activates whatever window was last active on that output, not the window in the focus direction. Fixing this probably requires not falling back to next/previous output in the focus shortcuts.

Drakulix:
> Shell::activate is supposed to return a position versus when it is not.

The position is used to actually move the cursor to the new output (which can't happen inside the Shell module).
So this sounds more like an oversight to be, that `Shell::activate` is not returning a position, when it is clearly switching outputs and not just a workspace on the same output.

I'll merge this for the meantime though, as I am rewriting large parts of the workspace-managing code in `shell/mod.rs` anyway, which should also fix this issue.

> Fixing this probably requires not falling back to next/previous output in the focus shortcuts.

Yes.

PR #186, opened 2023-09-29
chore: Update deps

Drakulix:
Looks scarier, than it is, but the update had some larger impact:

- Changes due to how the `Display` is registered to the event loop with the latest wayland-rs version, let us get rid of `Data`. Finally making the state of the EventLoop and of our wayland components the same type
- xkbcommon now uses proper types instead of aliases, changing the input and config code around a bit.

Draft because waiting on https://github.com/pop-os/cosmic-protocols/pull/19

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,16 +3,10 @@
 version = 3
 
 [[package]]
-name = "Inflector"
-version = "0.11.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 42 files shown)


ids1024 (approved):

Drakulix:
Now also includes an update to libcosmic, due to vendoring otherwise not being very happy about the duplicate `cosmic-config` dependency.

Issue #5, opened 2022-04-15
Support "super" key binding, for launcher (or other functionality)


ids1024:
As far as I can tell, key bindings currently can't support a binding using only "modifiers". Which is needed so that "super" can open the launcher.

PR #187, opened 2023-09-30
Modifier-only keybindings

PixelDoted:
This PR adds support for modifier-only keybindings  
and adds a new keybinding to open the launcher when releasing `Super`  

any combination of modifiers should work (e.g. `Super` and `Ctrl+Shift+alt+Super`)  

fixes: #5


src/config/key_bindings.rs
```diff
@@ -90,11 +90,11 @@ pub struct KeyPattern {
     pub modifiers: KeyModifiers,
     /// The actual key, that was pressed
     #[serde(deserialize_with = "deserialize_Keysym")]
-    pub key: u32,
+    pub key: Option<u32>,
```
Drakulix, line 93: I think we could just annotate `key` with `#[serde(default  = "Option::None")]` and keep the old deserialization code (see below).

src/config/types.rs
```diff
@@ -56,13 +56,18 @@ where
...
 {
     use serde::de::{Error, Unexpected};
 
-    let name = String::deserialize(deserializer)?;
+    let name: Option<String> = Option::deserialize(deserializer)?;
+    if name.is_none() {
+        return Ok(None);
+    }
+
+    let name = name.unwrap();
```
Drakulix, line 70: If we keep the `String::deserialize` code here and just wrap it in the end with `Some` + an appropriate `default` tag (see above), we don't need to change the config file. We can just omit `key` for entries that don't need/want it.

Drakulix (commented):
I added a couple of suggestions to prevent the whole config file getting filled with `Some`.

(If those don't work out for some reason, there is also [`implicit_some`](https://github.com/ron-rs/ron/blob/master/docs/extensions.md#implicit_some) as a last resort.)

Otherwise nice work, I like the design of the Shortcut queue. :+1:

PR #189, opened 2023-10-04
use xdg-open for default browser and file explorer in config.ron

wash2:
I guess we need another method for the default terminal. Maybe something like `i3-reasonable-terminal` or `x-terminal-emulator`.

### DIFF
--- a/config.ron
+++ b/config.ron
@@ -65,13 +65,11 @@
         (modifiers: [Super], key: "r"): Resizing(Outwards),
         (modifiers: [Super, Shift], key: "r"): Resizing(Inwards),
 
-        //TODO: ability to select default web browser
-        (modifiers: [Super], key: "b"): Spawn("firefox"),
-        //TODO: ability to select default file browser
-        (modifiers: [Super], key: "f"): Spawn("nautilus"),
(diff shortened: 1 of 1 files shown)


wash2:
> On debian, at least, x-terminal-emulator is supposed to mark the system's preferred terminal. `www-browser` is the same for the web browser.

Weirdly, I had to manually add firefox as an alternative using update-alternatives for `x-www-browser` to open it, while chrome and firefox-trunk were already added by default (chrome with the highest priority). But ya, I was avoiding adding `x-terminal-emulator` or `x-www-browser` because they may not work on non-debian distributions.

Drakulix (approved):

hojjatabdollahi:
Just noticed that pressing `Super+f` (or running `xdg-open ~`) opens `RawTherapee` instead of `Nautilus` on my system.  The Default Applications in Gnome Control Center doesn't have an option to change the default file browser, we should make sure that cosmic-settings has that option.

The fix was easy though.
```
$ xdg-mime query default inode/directory
com.rawtherapee.RawTherapee.desktop
$ xdg-mime default org.gnome.Nautilus.desktop inode/directory
```

Issue #179, opened 2023-09-22
Floating window height changes when grabbed and moved


WatchMkr:
- Open an app
- Make the app taller (does not happen when making the app shorter)
- Drag and drop the app from the headerbar

PR #193, opened 2023-10-06
Fixes floating window resizing after being moved

PixelDoted:
This PR fixes a bug where a floating window resizes after being moved

fixes: #179

### DIFF
--- a/src/shell/layout/floating/mod.rs
+++ b/src/shell/layout/floating/mod.rs
@@ -110,9 +110,12 @@ impl FloatingLayout {
                 mapped.min_size().unwrap_or((0, 0).into()),
                 mapped.max_size().unwrap_or((0, 0).into()),
             );
-            if win_geo.size.w > geometry.size.w / 3 * 2 {
+            if win_geo.size.w > max_size.w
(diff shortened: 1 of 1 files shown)


src/shell/layout/floating/mod.rs
```diff
@@ -110,9 +110,12 @@ impl FloatingLayout {
                 mapped.min_size().unwrap_or((0, 0).into()),
                 mapped.max_size().unwrap_or((0, 0).into()),
             );
-            if win_geo.size.w > geometry.size.w / 3 * 2 {
+            if win_geo.size.w > max_size.w
+                || win_geo.size.w < min_size.w
+                || win_geo.size.w > geometry.size.w
+            {
```
Drakulix, line 116: I don't think this code is the issue here, given it is supposed to be applied for freshly created windows, where we need to find a new fitting size. And that mostly works fine. What problematic here, is that with `last_geometry` set we should just always use that. So there is basically an `else` missing in lines 107/108.

Drakulix (approved):

Drakulix:
Thanks for working on this and bringing this to my attention :)

PR #196, opened 2023-10-10
feat: theme integration

wash2:
Depends on https://github.com/pop-os/libcosmic/pull/178 and should make cosmic-comp theme variables customizable via cosmic-settings after https://github.com/pop-os/cosmic-settings/pull/78

### DIFF
--- a/src/shell/element/mod.rs
+++ b/src/shell/element/mod.rs
@@ -759,6 +759,22 @@ impl CosmicMapped {
             popup_elements.into_iter().map(C::from).collect(),
         )
     }
+
+    pub(crate) fn set_theme(&self) {
+        match &self.element {
+            CosmicMappedInternal::Window(w) => w.set_theme(),
+            CosmicMappedInternal::Stack(s) => s.set_theme(),
+            CosmicMappedInternal::_GenericCatcher(_) => {}
+        }
+    }
+
+    pub(crate) fn force_redraw(&self) {
+        match &self.element {
+            CosmicMappedInternal::Window(w) => w.force_redraw(),
+            CosmicMappedInternal::Stack(s) => s.force_redraw(),
+            CosmicMappedInternal::_GenericCatcher(_) => {}
+        }
+    }
 }
 
 impl IsAlive for CosmicMapped {
--- a/src/shell/layout/tiling/mod.rs
+++ b/src/shell/layout/tiling/mod.rs
@@ -1,10 +1,7 @@
 // SPDX-License-Identifier: GPL-3.0-only
 
 use crate::{
-    backend::render::{
-        element::AsGlowRenderer, BackdropShader, IndicatorShader, Key, Usage, ACTIVE_GROUP_COLOR,
-        GROUP_COLOR,
-    },
+    backend::render::{element::AsGlowRenderer, BackdropShader, IndicatorShader, Key, Usage},
     shell::{
         element::{
             resize_indicator::ResizeIndicator,
@@ -25,6 +22,7 @@ use crate::{
         CosmicSurface, Direction, FocusResult, MoveResult, OutputNotMapped, OverviewMode,
         ResizeDirection, ResizeMode, Trigger,
     },
+    theme::group_color,
     utils::{prelude::*, tween::EaseRectangle},
     wayland::{
         handlers::xdg_shell::popup::get_popup_toplevel,
@@ -3843,6 +3841,10 @@ impl TilingLayout {
 
         Ok((window_elements, popup_elements))
     }
+
+    pub(crate) fn update_gaps(&mut self, gaps: (i32, i32)) {
+        self.gaps = gaps;
+    }
 }
 
 const GAP_KEYBOARD: i32 = 8;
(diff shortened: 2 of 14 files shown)


src/utils/iced.rs
```diff
@@ -251,13 +253,16 @@ impl<P: Program + Send + 'static> IcedElement<P> {
             })
             .ok();
 
+        let theme_guard = THEME.read().unwrap();
+        let theme = theme_guard.clone();
+        drop(theme_guard);
```
Drakulix, line 258: Can't this simply be?

```suggestion
        let theme = THEME.read().unwrap().clone();
```

src/utils/iced.rs
```diff
@@ -251,13 +253,16 @@ impl<P: Program + Send + 'static> IcedElement<P> {
...
         let mut internal = IcedElementInternal {
             outputs: HashSet::new(),
             buffers: HashMap::new(),
             pending_update: None,
             size,
             cursor_pos: None,
-            theme: Theme::dark(), // TODO
+            theme: theme.clone(), // TODO
```
Drakulix, line 265:
```suggestion
            theme: theme.clone(),

File: src/shell/element/mod.rs [hunk 3]
```diff
@@ -759,6 +759,22 @@ impl CosmicMapped {
             popup_elements.into_iter().map(C::from).collect(),
         )
     }
+
+    pub(crate) fn set_theme(&self) {
```
Drakulix, line 763: Since this has no parameter, could we maybe rename this to something like `reload_theme` or `update_theme`? (Same then for `CosmicStack`, `CosmicWindow`, `IcedElement`..)

src/shell/layout/tiling/mod.rs
```diff
@@ -3843,6 +3841,10 @@ impl TilingLayout {
 
         Ok((window_elements, popup_elements))
     }
+
+    pub(crate) fn update_gaps(&mut self, gaps: (i32, i32)) {
+        self.gaps = gaps;
+    }
```
Drakulix, line 3847: This is somewhat inconsistent. The `IcedElement` reloads the theme, but although `theme::gaps()` exists, this gets set instead of a similar `update_gaps`-function. (Or more generally `update_theme`.)

Not sure what solution would be cleanest, but I would like it to be consistent.

src/shell/layout/tiling/mod.rs
```diff
@@ -4086,7 +4088,7 @@ where
                                     if render_active_child { 16 } else { 8 },
                                     alpha * if render_potential_group { 0.40 } else { 1.0 },
                                     output_scale,
-                                    GROUP_COLOR,
+                                    group_color(),
```
Drakulix, line 4091: So the `IcedElement` caches the theme, while the `render_output` function of the `TilingLayout` just locks the `THEME` many times through `*_color`-functions? (See overall review comment.)

Drakulix (commented):
I am a little torn reviewing this..

- Either locking isn't cheap and we want to avoid the runtime cost. Then every element should cache the theme after creation. One-off elements like the `IndicatorShader`-elements should instead have a cached theme (or parts of it) in their parent (so `TilingLayout` for example). In those cases we would only read from the global theme at startup and when it gets updated.
- Or locking is negligible and we could just access the global state everywhere. Then we don't need `set_theme` on the `Cosmic*`-elements, as they could just access the theme directly whenever they need it.

Honestly I would love to have the best of both worlds. No locking, so it would probably need to be stored in `State`/`Common`, but passed by argument, so we don't have caching. But the `TilingLayout` needs `gaps` for `arrange`, which is called as part of almost every method, so I get, that we have to cache some parts here.

So sorry for requesting quite a refactor, but thinking about this, I would prefer:
- The `Theme` to be part of the `Common`-struct, so that it can be accessed through `state.common.theme`, instead of as a global
- The theme should be passed through the `render_output` functions, so that it is readable when creating the `RenderElements`.
- The only non-color property are the `gaps` and the `TilingLayout` should indeed cache that.
- The `IcedElement` also needs the theme outside of `render_elements`, since for some reason `update` needs it... So we also keep a copy there and initialize that on `IcedElement::new`. This means, that all `Cosmic*`-elements also need a new theme argument on creation. This might be the most annoying consequence of this design, but I already need an `EventLoopHandle` because of the async-runtime required for IcedElements. So `State` should already be available in some form in all places to get an event-loop handle and the theme could be cached/stored/cloned the same way.
- To update gaps and iced-elements, we introduce `shell.update_theme(&Theme)`, which can call `update_gaps` for all workspaces/tiling_layout and `reload_theme` for all `mapped`-elements.

I hope I didn't miss anything, that caused you to go the global-variable route instead, but to me this seems quite sound. If you feel stuck, I can also finish this off, since I have a pretty good idea, how I would like to see this integrated. But should you want to tackle it, I hope the idea has become clear.

Also in case this wasn't clear: Thanks for all the effort that already went into this. I am very much looking forward to this feature and I am quite happy I have not had to figure all of this out on my own.

wash2:
Ok thanks for the review. I'll try refactoring this as you've suggested and let you know if I run into any additional issues.

PR #101, opened 2023-04-16
Implement wlr_data_control protocol

xDarksome:

Issue #99, opened 2023-04-01
Using system clipboard inside a fullscreen Helix editor makes the Alacritty window to reposition itself


xDarksome:
So, I have the following setup: running [helix](https://github.com/helix-editor/helix) editor inside an [alacritty](https://github.com/alacritty/alacritty) terminal window.

Copy-pasting commands inside a fullscreen Alacritty window works fine. (`Ctrl-Shift-C` -> `Ctrl-Shift-V` inside the prompt)

Copy-pasting text via Alacritty shortcuts inside a fullscreen Alacritty window running Helix works as well.

However, when I try to use bultin Helix commands which manipulate the system clipboard inside a _fullscreen_ Alacritty window, for some reason, the window gets shifted towards the center of the screen. 

It looks like an issue with `wl-copy`, but I have no idea where to look for the root cause of this. Do you have any insight on this issue?

Drakulix:
The root cause of this is likely coming from the fact, that `wl-copy` will spawn a transparent window to obtain window focus, which is required to read clipboard contents.

In more details:
1. `wl-clipboard` will try to use the `wlr-data-control` protocol to access or set clipboard contents.
  - This protocol basically enables every application to freely access the clipboard as long as it has access to the wayland-socket.
  - While access can be restricted, there is currently no infrastructure to limit it for example to *only* `wl-copy` and `wl-paste`. As such cosmic-comp decides against supporting this protocol for the associated security risk at the moment.
2. Failing to bind `wlr-data-control` it can access the clipboard like any other client, but that requires a handle to a window, that has focus.
  - As such `wl-copy` will spawn a window with `xdg-shell`, hoping the compositor gives new windows focus immediately. So even this approach relies on implementation-defined behaviour and we should actually re-consider "*supporting*" focus stealing in cosmic-comp.
  - Given a new window is spawned cosmic-comp will take the currently active fullscreen window out of that state. Which is desired behaviour on our side. This is the bug you are seeing.
  - The transparent/invisible `wl-copy` window will get focus, at which point `wl-copy` will store the new contents in the clipboard and unmaps the window again.
  - Alacritty regains focus - being the previously focused window - but is not fullscreen anymore.

**Sidenote**: If cosmic-comp would never give focus to wl-copy, the program would give up and not work at all.

[...]

PR #197, opened 2023-10-10
Update `smithay`, with cursor/selection changes

ids1024:
We'll probably want to add support for `cursor-shape-v1`. Not sure about `wlr-data-control-unstable-v1`. But this just updates to work with the latest smithay commit for new.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -737,6 +737,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "cursor-icon"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


Drakulix (approved):
LGTM!

Drakulix:
> We'll probably want to add support for cursor-shape-v1.

Yes we do. Luckily cosmic-comp is already loading a bunch of different cursor shapes for internal use, so loading all of them in an organized manner should hopefully not prove to difficult.

> Not sure about wlr-data-control-unstable-v1.

Yes we do, see https://github.com/pop-os/cosmic-comp/issues/99 and https://github.com/pop-os/cosmic-comp/pull/101.

But not by default and even once there is a setting to explicitly expose it, it still shouldn't be exposed to sandboxed clients (security-context).

PR #198, opened 2023-10-16
Theme integration jammy

wash2:


src/shell/layout/tiling/mod.rs
```diff
@@ -3088,6 +3110,7 @@ impl TilingLayout {
                     Some(None),
                     None,
                     None,
+                    theme,
```
Drakulix, line 3113: Couldn't we use `self.theme` here and skip passing the theme to `element_under` (through all of the stack up to `surface_under`)? (That kinda feels wrong to me.)

src/utils/iced.rs
```diff
@@ -257,7 +258,7 @@ impl<P: Program + Send + 'static> IcedElement<P> {
             pending_update: None,
             size,
             cursor_pos: None,
-            theme: Theme::dark(), // TODO
+            theme, // TODO
```
Drakulix, line 261:
```suggestion
            theme,
```

(This is what that `TODO` was about)

Drakulix (commented):
I am much more happy with this draft, that looks really good. :)

Would you mind to rebase this on top of #194? That should actually simplify the code in `shell/workspace.rs` a bit, since you don't need the theme any more for maximizing/fullscreening (I think) and I hope I can merge that PR soon anyway. (If not we can still merge this into the branch and have me maintain it.)

wash2:
Ok, I'll rebase on #194 today

src/backend/render/mod.rs
```diff
@@ -913,7 +914,12 @@ where
     }
 
     renderer.bind(target).map_err(RenderError::Rendering)?;
-    let res = damage_tracker.render_output(renderer, age, &elements, CLEAR_COLOR);
+    let res = damage_tracker.render_output(
+        renderer,
+        age,
+        &elements,
+        crate::theme::clear_color(state.theme.cosmic()),
```
ids1024, line 921: Testing an implementation of ext-session-lock, I  noticed if I passed a different color to `damage_tracker.render_output` when the session lock was enabled, it ended up flickering between that and the previous color. The damage tracker I guess doesn't automatically recognize a change to the clear color as damage.

So I think  to change `clear_color` dynamically without possible damage tracking issues, Smithay or cosmic-comp may need some change here to mark everything as damaged when this changes?

src/theme.rs
```diff
@@ -0,0 +1,77 @@
...
+
+pub(crate) fn active_window_hint(theme: &Theme<Srgba>) -> palette::Srgba {
+    if let Some(hint) = theme.window_hint {
+        palette::Srgba::from(hint)
+    } else {
+        theme.accent_color()
+    }
+}
+
+pub fn watch_theme(handle: LoopHandle<'_, State>) -> Result<(), cosmic_config::Error> {
```
ids1024, line 36: Hm. We could use `ConfigWatchSource` from libcosmic here, but I guess that would mean 3 channels and calloop sources instead of 1...

I guess we end up spawning a thread here for each `watch` call? Probably the overhead doesn't matter that much.

So that should be fine for now, though it would be nice to figure out how to use epoll+inotify and kqueue to monitor for changes without threads and channels.

src/backend/render/mod.rs
wash2: ok, I can make the clear color static for now.

Drakulix (approved):
LGTM!

PR #199, opened 2023-10-17
fix(shell): lagging tab animations

mmstick:
Fixes lagging tab animations by discarding the expired animations when adding new animations, or doing a cleanup of animations.


Drakulix (approved):
Can you drop the last commit? The shortened duration doesn't align with the design files for the feature. But other than that this is amazing and improves the experience significantly.

It still would be nice to eventually synchronize this with the actual wayland buffer size for resizes, but this this more than good enough for now.

PR #203, opened 2023-10-25
Add `ext-session-lock` protocol

ids1024:
This is now usable, though some details may need to be adjusted.


src/backend/winit.rs
```diff
@@ -52,6 +52,8 @@ pub struct WinitState {
 
 impl WinitState {
     pub fn render_output(&mut self, state: &mut Common) -> Result<()> {
+        println!("Render output");
```
Drakulix, line 55: Isn't this super spammy? :sweat_smile:

src/input/mod.rs
```diff
@@ -281,7 +284,12 @@ impl State {
                     let serial = SERIAL_COUNTER.next_serial();
                     let time = Event::time_msec(&event);
                     let keyboard = seat.get_keyboard().unwrap();
-                    let current_focus = keyboard.current_focus();
+                    let mut current_focus = keyboard.current_focus();
+                    if self.common.session_lock.is_some() {
+                        if !matches!(current_focus, Some(KeyboardFocusTarget::LockSurface(_))) {
+                            current_focus = None;
+                        }
+                    }
```
Drakulix, line 292: Could we rather fix this in `Shell::refresh_focus` and force the focus to be on the lock surface?

src/input/mod.rs
```diff
@@ -1153,6 +1169,10 @@ impl State {
         pattern: KeyPattern,
         direction: Option<Direction>,
     ) {
+        if self.common.session_lock.is_some() {
+            return;
+        }
```
Drakulix, line 1174: We probably should at least allow `Action::Terminate`. There is also a point to be made to allow `Action::Debug`.

Drakulix (commented):
Looks good for a first draft, disregarding the couple of TODOs still in there.

src/backend/winit.rs
ids1024: Lol. Naturally I added that when I was trying to figure out why nothing was showing up, and I guess I didn't remove it.

src/input/mod.rs
ids1024: I'm not entirely sure about the best way to handle keyboard focus, but there are a few things to consider:
* If the session is locked but there is are no lock surfaces (the session lock client has not created them yet, or has crashed), no surface should receive keyboard events.
* If there are multiple outputs, each of them have lock surfaces that can receive keyboard input, so it makes sense to handle changes to keyboard focus normally, but only allow lock surfaces to have focus.
* The session is locked and lock surfaces are created, the lock surface on the current output should gain initial focus.

Looking at `refresh_focus`, that could be a good place to handle this. I think the first and last of these things could be handled there, while it could avoid messing with the surface by no changing focus when it's already an alive lock surface.

src/input/mod.rs
ids1024: `Debug` makes sense if normal releases won't be built with the `debug` feature anyway. Terminating the compositor is a sensitive operation if the compositor was started manually from a tty, in which case it will drop you to a logged in console. If that's a concern. Otherwise it should go back to the login screen, but will kill the user's processes.

Drakulix: > Terminating the compositor is a sensitive operation if the compositor was started manually from a tty

Right, can we somehow determine this or at least allow this, when running cosmic-greeter as a display manager? I feel like you want to be able to kill a misbehaving session.

src/state.rs
```diff
@@ -459,6 +476,36 @@ impl Common {
...
+                            select_dmabuf_feedback(
+                                surface,
+                                render_element_states,
+                                &feedback.render_feedback,
+                                &feedback.scanout_feedback,
+                            )
+                        },
+                    )
+                }
+            }
```
Drakulix, line 506: I guess given the fixed output we can also easily send the fractional scale here. (See the window code a few lines below.)

Drakulix (approved):
Almost! One little nitpick.

Drakulix (approved):

PR #205, opened 2023-10-27
Allow session lock if existing session lock is no longer valid

ids1024:
This makes it possible to handle a crash in the session lock client by restarting it, for instance. This is similar to how Sway behaves.

Depends on Smithay change.


ids1024:
Updated now that the Smithay dependency is updated.

Drakulix (approved):

PR #207, opened 2023-10-31
xdg-activation: Initial implementation

### LINKED ISSUE #192: Clicking a Link In a App on A Different Workspace Does Not Move to the Browser

Drakulix:
This provides an initial implementation of the xdg-activation protocol to prevent focus stealing, passing of focus by client (e.g. required to fix #192) voluntarily, additional context for new apps (e.g. workspace they were launched from) and urgent state for workspaces.

Draft until this is used by cosmic-launcher/cosmic-applibrary.


Drakulix:
Changed so that it still focused new windows, but still includes the rest of the xdg-activation logic, so that it can be merged without being potentially too disruptive.

PR #208, opened 2023-10-31
wayland: Privileged protocol security

Drakulix:
Draft until everything is updated to not break. Perhaps this should be split in two commits. One that enables security-contexts to inherit the privileged flag and one that enables all the enforcing, that can be merged later.

In any case this breaks parts of the desktop today, but once cosmic-panel makes use of security-context applets could even enjoy more privileges. E.g. always creating valid xdg-activiation tokens via #207.

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -279,10 +279,32 @@ impl BackendData {
     }
 }
 
-pub fn client_has_security_context(client: &Client) -> bool {
+pub fn client_has_no_security_context(client: &Client) -> bool {
     client
         .get_data::<ClientState>()
         .map_or(true, |client_state| client_state.security_context.is_none())
+}
+
+pub fn client_is_privileged(client: &Client) -> bool {
+    client
+        .get_data::<ClientState>()
+        .map_or(false, |client_state| client_state.privileged)
+}
+
+pub fn client_should_see_privileged_protocols(client: &Client) -> bool {
+    if std::env::var("COSMIC_ENABLE_WAYLAND_SECURITY")
+        .map(|x| {
+            x == "1"
+                || x.to_lowercase() == "true"
+                || x.to_lowercase() == "yes"
+                || x.to_lowercase() == "y"
+        })
+        .unwrap_or(false)
+    {
+        client_is_privileged(client)
+    } else {
+        client_has_no_security_context(client)
+    }
 }
 
 impl State {
@@ -306,14 +328,14 @@ impl State {
         let keyboard_shortcuts_inhibit_state = KeyboardShortcutsInhibitState::new::<Self>(dh);
         let output_state = OutputManagerState::new_with_xdg_output::<Self>(dh);
         let output_configuration_state =
-            OutputConfigurationState::new(dh, client_has_security_context);
+            OutputConfigurationState::new(dh, client_should_see_privileged_protocols);
         let presentation_state = PresentationState::new::<Self>(dh, clock.id() as u32);
         let primary_selection_state = PrimarySelectionState::new::<Self>(dh);
         let screencopy_state = ScreencopyState::new::<Self, _, _>(
             dh,
             vec![CursorMode::Embedded, CursorMode::Hidden],
-            client_has_security_context,
-        ); // TODO: privileged
+            client_should_see_privileged_protocols,
+        );
         let shm_state =
             ShmState::new::<Self>(dh, vec![wl_shm::Format::Xbgr8888, wl_shm::Format::Abgr8888]);
         let seat_state = SeatState::<Self>::new();
@@ -322,11 +344,11 @@ impl State {
         let kde_decoration_state = KdeDecorationState::new::<Self>(&dh, Mode::Client);
         let xdg_decoration_state = XdgDecorationState::new::<Self>(&dh);
         let session_lock_manager_state =
-            SessionLockManagerState::new::<Self, _>(&dh, client_has_security_context);
+            SessionLockManagerState::new::<Self, _>(&dh, client_should_see_privileged_protocols);
         XWaylandKeyboardGrabState::new::<Self>(&dh);
         PointerConstraintsState::new::<Self>(&dh);
         PointerGesturesState::new::<Self>(&dh);
-        SecurityContextState::new::<Self, _>(&dh, client_has_security_context);
+        SecurityContextState::new::<Self, _>(&dh, client_has_no_security_context);
 
         let shell = Shell::new(&config, dh);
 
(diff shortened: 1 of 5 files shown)


src/wayland/handlers/security_context.rs
```diff
@@ -17,10 +19,37 @@ impl SecurityContextHandler for State {
...
+                            .and_then(|data| data.downcast_ref::<XWaylandClientData>())
+                            .and_then(|data| data.user_data().get::<DrmNode>().cloned())
+                    });
+
                 if let Err(err) = state.common.display_handle.insert_client(
                     client_stream,
                     Arc::new(ClientState {
                         security_context: Some(security_context.clone()),
+                        privileged: privileged
+                            && security_context.sandbox_engine.as_deref() == Some("cosmic-panel"),
```
ids1024, line 51: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/253

We should be using a reverse DNS name. Which seems better anyway, since it can match the app id for cosmic-panel.

ids1024:
At some point we'll likely want to not just have a binary `privileged` flag, but a way to give permission for a protocol to specific apps. Not sure how that would work though, since Flatpak itself would not be privileged. And likewise dealing with apps that aren't started by a sandbox engine or by cosmic-session/cosmic-panel and thus have no security context...

For development, would it make sense to have a command line argument or env var to disable enforcement of privileged protocols?

I'd still note that we can't expect this to actually be secure for apps that aren't sandboxed, but it's good not to encourage apps to use sensitive compositor-specific protocols that aren't available in Flatpak, etc.

src/wayland/handlers/security_context.rs
Drakulix: good call.

Drakulix:
> At some point we'll likely want to not just have a binary `privileged` flag, but a way to give permission for a protocol to specific apps. Not sure how that would work though, since Flatpak itself would not be privileged. And likewise dealing with apps that aren't started by a sandbox engine or by cosmic-session/cosmic-panel and thus have no security context...

Right, I think the only reasonable thing to do here is [get_client_credentials](`https://docs.rs/wayland-server/latest/wayland_server/backend/struct.Handle.html#method.get_client_credentials`) and try to match the binary. (Sadlyy this is not portable.)

If we are reasonably sure it is the same the user originally allow-listed, we can allow additional protocols. Check should e.g. include the binary is not user-writable imo.

If we figure out the flatpak-sandbox is actually the flatpak binary, we can also match on the app-id to allow specific protocols.

> For development, would it make sense to have a command line argument or env var to disable enforcement of privileged protocols?

Yeah I also thought about just generally enabling all protocols for debug builds. Though those might be too slow for debugging, so an environment variable sounds like a good idea.

ids1024:
Yeah. The BSDs all seem to support `getpeereid` to get the uid/gid of a Unix socket peer, but unlike Linux or FreeBSD, I'm not sure there's a way to get the pid on NetBSD and OpenBSD. It *may* be possible to identify the binary path through `/proc` from pid. But that is theoretically racy.

Restricting it to debug builds also isn't ideal because someone working on the panel, lock screen, or an applet doesn't necessarily want to have to rebuild the compositor.

Drakulix:
> Restricting it to debug builds also isn't ideal because someone working on the panel, lock screen, or an applet doesn't necessarily want to have to rebuild the compositor.

Probably this should also be a toggle from the debug interface. Maybe we should make a "debug-interface wishlist" issue at some point.

Anyway `COSMIC_DISABLE_WAYLAND_SECURITY` exists now and can be easily added to our `start-cosmic` script or , if desired. Though just as I write this, I realize that we also source the user's profile, so we still need to make sure we don't ship this flag eventually.

Maybe just in the debug-interface would be enough until we have proper configuration for security aspects?

Drakulix:
Ok, this PR now only includes inheriting the privileged flag for security-context clients already having it (and allowlisted for cosmic-panel for now), as well as a `COSMIC_ENABLE_WAYLAND_SECURITY` flag for testing, so this can be merged early before all pieces for proper security handling are in place.

src/state.rs
```diff
@@ -322,11 +344,11 @@ impl State {
         let kde_decoration_state = KdeDecorationState::new::<Self>(&dh, Mode::Client);
         let xdg_decoration_state = XdgDecorationState::new::<Self>(&dh);
         let session_lock_manager_state =
-            SessionLockManagerState::new::<Self, _>(&dh, client_has_security_context);
+            SessionLockManagerState::new::<Self, _>(&dh, client_should_see_privileged_protocols);
         XWaylandKeyboardGrabState::new::<Self>(&dh);
         PointerConstraintsState::new::<Self>(&dh);
         PointerGesturesState::new::<Self>(&dh);
-        SecurityContextState::new::<Self, _>(&dh, client_has_security_context);
+        SecurityContextState::new::<Self, _>(&dh, client_should_see_privileged_protocols);
```
ids1024, line 351: This would prevent `flatpak` from using the security context protocol if `COSMIC_ENABLE_WAYLAND_SECURITY` is set, since it wouldn't be started with a privileged socket. Would it be better to use `client_has_no_security_context` here?

Though it doesn't make a difference until we start taking into account the security context of Flatpak apps.

src/wayland/handlers/security_context.rs
```diff
@@ -17,10 +19,38 @@ impl SecurityContextHandler for State {
...
+                            .and_then(|data| data.user_data().get::<DrmNode>().cloned())
+                    });
+
                 if let Err(err) = state.common.display_handle.insert_client(
                     client_stream,
                     Arc::new(ClientState {
                         security_context: Some(security_context.clone()),
+                        privileged: privileged
+                            && security_context.sandbox_engine.as_deref()
+                                == Some("com.system76.CosmicPanel"),
```
ids1024, line 52: We may want to further restrict which privileged protocols an applet has access to (define required permissions in desktop files?). But this is reasonable for now.

(And for that to really be secure, we'd also want to sandbox applets.)

src/state.rs
Drakulix: > Would it be better to use client_has_no_security_context here?

Right, that wasn't on purpose, changing it back.

> Though it doesn't make a difference until we start taking into account the security context of Flatpak apps.

Well it should already block all "privileged" protocols. Though once we enforce security there is indeed no difference here, but given there might be in the future this is still a good change.

src/wayland/handlers/security_context.rs
Drakulix: Defining necessary protocols in the desktop file would already be a good idea, I agree.

PR #209, opened 2023-10-31
Add support for profiling using tracy

ids1024:
I'm not sure entirely how to deal with `profiling::finish_frame!` when there are multiple outputs rendering frames, and some more functions could be annotated, but Tracy seems generally useful for profiling with these changes. So they might as well be merged.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -761,6 +761,7 @@ dependencies = [
  "once_cell",
  "ordered-float",
  "png",
+ "profiling",
  "puffin",
  "puffin_egui",
  "regex",
@@ -1889,6 +1890,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "generator"
+version = "0.7.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5cc16584ff22b460a382b7feec54b23d2908d858152e5739a120b949293bd74e"
+dependencies = [
+ "cc",
+ "libc",
+ "log",
+ "rustversion",
+ "windows 0.48.0",
+]
+
+[[package]]
 name = "generic-array"
 version = "0.14.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2038,7 +2052,7 @@ dependencies = [
  "presser",
  "thiserror",
  "winapi",
- "windows",
+ "windows 0.51.1",
 ]
 
 [[package]]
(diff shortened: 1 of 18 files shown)


ids1024:
Looks like `profiling` has support for `puffin` as well? So presumably if we want we should be able to support both fairly easily without having explicit `puffin::profile_function!` calls everywhere.

Drakulix:
> Looks like `profiling` has support for `puffin` as well? So presumably if we want we should be able to support both fairly easily without having explicit `puffin::profile_function!` calls everywhere.

Right. It doesn't hurt to also have puffin-egui available inside the compositor I guess. But we definitely don't want special puffin calls, where we could just use profiling.

ids1024:
Seems `profiling` doesn't currently support the latest version of `puffin`: https://github.com/aclysma/profiling/issues/58.

ids1024:
`profiling` has been updated. But to use `puffin` 0.18 (which it now uses), we'd need to use `puffin_egui` 0.24, which also requires `egui` 0.24, which I ran into an issue updating to: https://github.com/Smithay/smithay-egui/pull/16

ids1024:
Rebased on `master`, and added a change I had stashed to work around a panic in `UtcOffset::current_local_offset().expect("No yet multithreaded")`. That one is annoying, still not sure what to do about it...

Drakulix:
> Rebased on `master`, and added a change I had stashed to work around a panic in `UtcOffset::current_local_offset().expect("No yet multithreaded")`. That one is annoying, still not sure what to do about it...

I mean we do have cosmic-screenshot now, so maybe we can drop the code for creating window screenshots now? That is the only thing that needs the local clock for formatting the file name.

ids1024:
Oh, actually that part can be addressed by temporarily using `time::util::local_offset:set_soundness`. A bit verbose but it's correct even from a paranoid perspective, as long as tracy's thread isn't engaging in any `setenv` shenanigans.

I've committed more `profiling` annotations I've found useful, as well as added them to any function using `puffin::profile_function!`. `profiling` won't work with puffin here until various packages are released and updated. But anyway, if we care about puffin this doesn't change anything. If we don't we can remove the puffin code, but that's somewhat independent.

I also `cargo update` the tracy deps so it's on Tracy 0.10.

This should be fine to merge, and would be useful to have in `master`.

Drakulix (approved):

PR #215, opened 2023-11-03
kms: Filter bad nvidia targets

Drakulix:


Drakulix:
Seems to not improve consistency, while this helps on some setups, it does the opposite on others.

PR #217, opened 2023-11-07
Initial `wl_touch` support

ids1024:
A very rough implementation, that is sufficient to get a touchscreen device working.

I'm not sure how we should detect which output is associated with a touchscreen device.

The Smithay APIs for touch events could probably be improved a bit.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -41,7 +41,12 @@ use smithay::{
         Seat, SeatState,
     },
     output::Output,
-    reexports::wayland_server::DisplayHandle,
+    reexports::{
+        input::event::touch::{
+            TouchDownEvent as LibinputTouchDownEvent, TouchMotionEvent as LibinputTouchMotionEvent,
+        },
+        wayland_server::DisplayHandle,
+    },
     utils::{Point, Serial, SERIAL_COUNTER},
     wayland::{
         keyboard_shortcuts_inhibit::KeyboardShortcutsInhibitorSeat,
@@ -217,6 +222,8 @@ impl State {
         needs_key_repetition: bool,
     ) where
         <B as InputBackend>::PointerAxisEvent: 'static,
+        <B as InputBackend>::TouchDownEvent: 'static,
+        <B as InputBackend>::TouchMotionEvent: 'static,
     {
         use smithay::backend::input::Event;
         match event {
@@ -1174,9 +1181,23 @@ impl State {
             }
             InputEvent::TouchDown { event, .. } => {
                 if let Some(seat) = self.common.seat_with_device(&event.device()).cloned() {
-                    // TODO: Configuration option for mapping touch device to output
-                    // Is it possible to determine mapping for external touchscreen?
-                    let Some(output) = self.common.shell.builtin_output().cloned() else {
+                    // TODO Is it possible to determine mapping for external touchscreen?
+                    let map_to_output = if let Some(event) =
+                        <dyn Any>::downcast_ref::<LibinputTouchDownEvent>(&event)
+                    {
+                        self.common
+                            .config
+                            .map_to_output(&event.device())
+                            .and_then(|name| {
+                                self.common
+                                    .shell
+                                    .outputs()
+                                    .find(|output| output.name() == name)
+                            })
+                    } else {
+                        None
+                    };
+                    let Some(output) = map_to_output.or_else(|| self.common.shell.builtin_output()).cloned() else {
                         return;
                     };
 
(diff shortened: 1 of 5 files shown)


src/input/mod.rs
```diff
@@ -1139,7 +1140,93 @@ impl State {
...
+                        overview.0.clone(),
+                        workspace,
+                        self.common.session_lock.as_ref(),
+                    )
+                    .map(|(target, pos)| (target, pos.as_logical()));
+
+                    if let Some((target, pos)) = under {
+                        if let Some(wl_surface) = target.wl_surface() {
+                            let serial = SERIAL_COUNTER.next_serial();
+                            let mut touch = seat.get_touch().unwrap();
```
Drakulix, line 1169: We should check here, if the client has actually bound wl_touch and do pointer emulation otherwise.

src/input/mod.rs
```diff
@@ -1139,7 +1140,93 @@ impl State {
                     );
                 }
             }
-            _ => { /* TODO e.g. tablet or touch events */ }
+            InputEvent::TouchDown { event, .. } => {
+                if let Some(seat) = self.common.seat_with_device(&event.device()).cloned() {
+                    let current_output = seat.active_output();
```
Drakulix, line 1145: I think most compositors just assume, that any touchscreen will be part of the internal screen, so we could match on the DRM connector. I am not aware of any other link, though I would hope `event.device()` could give us a reference to the udev device, which somehow *might* tell us more?

Drakulix (commented):
> The Smithay APIs for touch events could probably be improved a bit.

Really? :rofl: /s

While this is a good starting point, there is no good way to implement touch grabs right now. So this is really not usable for tablet-like devices, just kinda usable for laptops with touchscreens, I guess.

So I am not sure, if we want to merge this in the meantime or first rework the smithay api to be more like the Pointer and Keyboard apis.

src/input/mod.rs
ids1024: Sway has a `map_to_output` configuration option for absolute input devices, which matches `xinput --map-to-output`. Not sure how if determines default mappings.

We could add a heuristic for determining which screen is internal, at least.

ids1024:
> While this is a good starting point, there is no good way to implement touch grabs right now. So this is really not usable for tablet-like devices, just kinda usable for laptops with touchscreens, I guess.
>
> So I am not sure, if we want to merge this in the meantime or first rework the smithay api to be more like the Pointer and Keyboard apis.

Yeah, as is this works well enough on a laptop with a touch screen (maybe with some issues in edge cases) but wouldn't actually be usable on a tablet unless SSDs and other in-compositor UI handles touch events, an on screen keyboard is added, etc.

It may be reasonable to merge this once the major hacks here (using `seat.active_output()`) are fixed, but we will need to update Smithay's APIs.

src/input/mod.rs
ids1024: Do any other compositors emulate `wl_pointer` for touch events like that? Or is it generally assumed clients are responsible for that? That should be safe to do though, for any clients that don't.

Drakulix: Honestly I am not sure. I never looked into this a lot, but at least Xwayland will never bind wl_touch, so we need to do emulation for some clients anyway or there is no way to interact with them.

src/input/mod.rs
Drakulix: Yeah we need to make this configurable eventually. I would also love to have the same thing for tablets. (Though more free-form than just assign to output. Assign to region (relative to output) or window would also be nice.

src/input/mod.rs
ids1024: I haven't tested it, but https://gitlab.freedesktop.org/xorg/xserver/-/blob/master/hw/xwayland/xwayland-input.c does call `wl_seat_get_touch` and have other `wl_touch` related code.

So I'm not sure how touch works in X11, but checking if the client binds `wl_touch` shouldn't change anything.

Drakulix: Oh interesting! That is new to me, but pretty cool.

ids1024: Presumably it needs `wl_touch` to generate the multi-touch events specified in Xinput 2.2: https://gitlab.freedesktop.org/xorg/proto/xorgproto/-/blob/master/specs/XI2proto.txt.

That protocol also has a section titled "Pointer emulation from multitouch events". Presumably that's handled in XWayland and doesn't require anything from us.

ids1024:
I've added a check for the "builtin monitor" similar to the default used in Sway.

This seems to be working fine with external monitors and fractional scaling. So I think it's good enough for initial touch support.

src/input/mod.rs
```diff
@@ -205,6 +205,7 @@ pub fn add_seat(
             .expect("Failed to load xkb configuration files");
     }
     seat.add_pointer();
+    seat.add_touch();
```
Drakulix, line 208: I am not sure we want to do that unconditionally or only, if we actually detect a touch device.

The problem with that is obviously, that I am not sure any clients bind this dynamically...

Drakulix (approved):
I kinda hate merging this would proper support for grabs an what not.. but I guess having it in this state is better than not having it.

Drakulix:
I wonder if we could make the mapped display a property of our input-config at least? So that users *could* override it and we just take the buildin display as a default?

src/input/mod.rs
ids1024: Yeah, it's trivial to add when we get a touch device, and not too hard to track how many touch devices we have and remove it when there aren't any more. But if clients don't handle that properly with keyboard/pointer, I expect that's an issue with touch too.

Drakulix: Likely a safe assumption..

ids1024:
added a `map_to_output` option, like Sway has.

Drakulix (approved):

PR #224, opened 2023-11-14
Handle session lock surfaces in `visible_outputs_for_surface`

ids1024:
Fixes issue with re-draw not being queued on initial surface commit until cursor is moved.


ids1024:
Added a commit to schedule a renderer when the session lock is created or destroyed, which also improves behavior in certain cases.

Drakulix (approved):
No need to block this on the open discussion though, the changes seem very sensible.

PR #225, opened 2023-11-16
fix: set the workspace of the activated token's surface to urgent if not focusing it

wash2:
This seems to fix an issue where the workspace for the activated settings wasn't appearing as urgent when spawning another instance of settings by the launcher from another workspace. Is this not intended?


Drakulix (approved):
Nice catch! No not intended, just a hole in the logic, I didn't think of.

PR #234, opened 2023-11-22
Update Smithay

ids1024:
For the `drm` protocol, we can't construct an `ImportNotifier`, since that is specifically based around `ZwpLinuxBufferParamsv1`. So we need a new method for importing with that protocol.

This could be improved, but should match current behavior.


Drakulix (approved):
LGTM

PR #256, opened 2023-12-28
Initial support for tablet input

ids1024:


Drakulix (approved):
Neat to re-use the touchscreen mapping like that, though I would love to support map_to_region at some point. (Or perhabs even map_to_window? Though for that we might need a more reliable way to identify windows than just the app_id.)

But this is a good start and can be refined once we have a proper settings interface.

ids1024:
> Or perhabs even map_to_window? Though for that we might need a more reliable way to identify windows than just the app_id.

Yeah, something like that would be useful for graphics tablets. I'm not sure what the best/normal way to handle it is.

`map_to_output` is good for handling tablet input integrated into screens, anyway.

PR #265, opened 2024-01-02
xwayland: Consider fullscreen requests for unmapped windows

### LINKED ISSUE #252: Fullscreen applications launch windowed.
In COSMIC, I'm seeing application that should be launching full-screen instead launching in windows. In my case, I'm seeing this with Focuswriter. 

If I launch Focuswriter using the Launcher, there is a high incidence of Focuswriter launching in a window. This seems more prevalent on a 3rd gen i5 system I tested than on a 12th gen i7 system I also looked at. If I launch from the command line, the application seems to successfully launch in fullscreen every time. 

On a related note, other members of QA reported seeing issues with games not correctly launching fullscreen. I tested that using two separate games I have through Steam. The first was Dwarf Fortress, which launched successfully every time. The second was Stellaris, which failed every time to launch correctly. 

I'm uncertain what the best debug procedure is for this, so let me know if you would like me to try and create these issues in a specific way for diagnostic purposes.

Drakulix:
Fixes #252

### DIFF
--- a/src/xwayland.rs
+++ b/src/xwayland.rs
@@ -443,6 +443,18 @@ impl XwmHandler for State {
             if let Some(workspace) = self.common.shell.space_for_mut(&mapped) {
                 workspace.fullscreen_request(&surface, None)
             }
+        } else {
+            let output = self.common.last_active_seat().active_output();
+            if let Some(o) = self
+                .common
+                .shell
+                .pending_windows
+                .iter_mut()
+                .find(|(s, _, _)| s == &surface)
+                .map(|(_, _, o)| o)
+            {
+                *o = Some(output);
+            }
         }
     }
 


gabriele2000:
I'm desperately trying to know what's the cause of one weird extreme stuttering issue that I have if a game goes under 60FPS...
Could this PR have modified how games in WINE (So in our case, running with XWayland) render stuff?

Recording this issue internally is pointless, since the output is smooth, while I see a lot of incosistent frame pacing with my eyes, but I really can't understand this issue.

I traced back my first discover to the 4th of January when I first experienced the issue.

Drakulix:
> I'm desperately trying to know what's the cause of one weird extreme stuttering issue that I have if a game goes under 60FPS... Could this PR have modified how games in WINE (So in our case, running with XWayland) render stuff?
> 
> Recording this issue internally is pointless, since the output is smooth, while I see a lot of incosistent frame pacing with my eyes, but I really can't understand this issue.
> 
> I traced the issue back to the 4th of January when I first experienced it.

No, please just open an issue about this and don't comment on random PRs.

All this code does is fix some games getting windowed windows, when they requested the fullscreen state a bit earlier than usual during window creation. It absolutely has nothing to do with frame pacing.

What you are likely seeing causing the flickering is the state the nvidia driver on linux due to missing implicit synchronization. This problem most prominently shows itself with Xwayland. We are trying our best to work around that issue, but we have a few devs with exactly the same problem.

This is bound to be fixed with the wp_linux_explicit_sync_v2 protocol and a new XWayland release making use of it, but that hasn't arrived yet. (We sincerely hope that will happen soon and in time for our alpha. Once possible we will add necessary support asap.)

Looking at your support files it seems like you also got an Intel GPU. The problem is side-stepped (at the cost of performance), when the nvidia-gpu is not used for outputting the image.
So possibly you were previously using an integrated display or a different port to connect your display? Or if you are running system76 hardware, have you switched from Hybrid to Nvidia mode?

gabriele2000:
> No, please just open an issue about this and don't comment on random PRs.

I'm sorry, I didn't mean to cause any disturbance.

> All this code does is fix some games getting windowed windows, when they requested the fullscreen state a bit earlier than usual during window creation. It absolutely has nothing to do with frame pacing.

Ok, that was I was thinking too. Still had to be sure.

> What you are likely seeing causing the flickering is the state the nvidia driver on linux due to missing implicit synchronization.

You see, I didn't have this problem before January.
That's exactly what the hybrid mode (Intel iGPU and Nvidia dGPU) fixes, since when I'm in hybrid mode the "half-presented frames" are fully synchronized thanks to the other GPU, like it's stated [here](https://gitlab.freedesktop.org/xorg/xserver/-/issues/1317#note_1589460).

The fact that everything worked before, and somehow broke again only in COSMIC DE, has spiked my curiosity and I'm just wondering what's the cause, that's why I'm trying to know what broke it.

> This is bound to be fixed with the wp_linux_explicit_sync_v2 protocol and a new XWayland release making use of it, but that hasn't arrived yet. (We sincerely hope that will happen soon and in time for our alpha. Once possible we will add necessary support asap.)

Yeah, that's because Nvidia has proposed, about a year ago, a solution that a lot of devs were just pointlessly debating (if I remember correctly from a lot of PR that I've read) and everything's moving again, finally.

> Looking at your support files it seems like you also got an Intel GPU. The problem is side-stepped (at the cost of performance), when the nvidia-gpu is not used for outputting the image. So possibly you were previously using an integrated display or a different port to connect your display? Or if you are running system76 hardware, have you switched from Hybrid to Nvidia mode?

I have an MSI laptop.
I have only one monitor so no, I've never even tried the HDMI adapter, haha!

Since the fix on PRIME system has been released inside an Nvidia driver update, I've been running "POP Wayland", until it broke with driver v540; it was at that point that I've decided to daily drive COSMIC DE, with great satisfaction.

PR #268, opened 2024-01-05
Workspace protocol updates

Drakulix:
Still missing reordering


ids1024:
It seems for this to work the protocol version of the global needs to be updated. It seems to be working with https://github.com/pop-os/cosmic-protocols/pull/23, and changing `create_global` to use version `2`.

(At least it's advertising a tiling state. Haven't tested everything yet.)

PR #269, opened 2024-01-05
floating: Maximize animation

Drakulix:
And a bunch of fixes for geometry restoring in floating mode.

### DIFF
--- a/src/shell/grabs/moving.rs
+++ b/src/shell/grabs/moving.rs
@@ -546,27 +546,6 @@ impl Drop for MoveGrab {
                     }
 
                     match previous {
-                        ManagedLayer::Tiling => {
-                            let (window, location) = state
-                                .common
-                                .shell
-                                .active_space_mut(&output)
-                                .tiling_layer
-                                .drop_window(grab_state.window);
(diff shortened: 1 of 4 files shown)


wash2:
This seems mostly good, but for some reason I see a compositor panic when I maximize a window in tiling mode, then switch to window mode, unmaximizing the window, and then drag the window.

wash2:
thread 'main' panicked at 'assertion failed: was_floating != was_tiled.is_some()': src/shell/mod.rs:2292

Drakulix:
> This seems mostly good, but for some reason I see a compositor panic when I maximize a window in tiling mode, then switch to window mode, unmaximizing the window, and then drag the window.

I assume "window mode" should have been "floating mode". I can't reproduce it with the last commit anymore, can you verify? Basically I think what has happened was switching to "floating mode" removed the window from the tiling_layer and added it to the floating, unmaximizing restored the previous state, so it re-added the window to the tiling layer and then the check breaks as the window is mapped twice (which should never happen).

I added a check if tiling-mode is still enabled to all cases, where we map stuff to the tiling_layer later again.

wash2 (approved):

PR #270, opened 2024-01-06
Add cosmic_comp.ftl PL translation

VandaLHJ:
Added Polish translation, there wasn't any so i thought it would be welcome.


Drakulix (approved):
Thanks!

PR #283, opened 2024-01-22
fix!: default config uses `pactl` instead of `amixer`

jokeyrhyme:
This might be controversial or unnecessary, so I understand completely if this PR isn't desirable

I assume that xdg-desktop-portal-cosmic already relies on pipewire for screensharing, therefore users will likely already have pipewire-pulse, and therefore are more likely to have `pactl` available than `amixer`

I've tested this and the keybindings do have the intended effect on my system

An alternative could be to send messages to `cosmic-settings-daemon`, and implement volume controls in a more portal manner over there...

### DIFF
--- a/config.ron
+++ b/config.ron
@@ -82,9 +82,9 @@
 
         (modifiers: [], key: "Print"): Spawn("cosmic-screenshot"),
 
-        (modifiers: [], key: "XF86AudioRaiseVolume"): Spawn("amixer sset Master 5%+"),
-        (modifiers: [], key: "XF86AudioLowerVolume"): Spawn("amixer sset Master 5%-"),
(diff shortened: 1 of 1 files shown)


Drakulix:
> An alternative could be to send messages to cosmic-settings-daemon, and implement volume controls in a more portal manner over there...

The current solution is definitely a stop-gap. Long term goal is for cosmic-settings-daemon to handle this shortcuts directly though the global-shortcuts portal, once that is finalized.

So this would entirely bypass cosmic-comp's config.

That said I am not very keen on merging this in it's current state. Distributions not offering pulseaudio/pipewire by default may choose to alter the default config, but imo right now this does what the user wants more often than going through `amixer`. E.g. when dealing with virtual output devices, equalizers, etc.

PR #287, opened 2024-01-24
fix: bind media playback keys to `playerctl`

jokeyrhyme:
- add default keybindings for `playerctl`: https://github.com/altdesktop/playerctl
- tested to confirm that playback in Spotify in Firefox in flatpak can be controlled with the keyboard (play/pause, stop, next, previous)
- alternatively, similar to https://github.com/pop-os/cosmic-comp/pull/283 , such functionality could be implemented across `xdg-desktop-portal-cosmic` (GlobalShortcuts portal) and `cosmic-settings-daemon` so that third-party `playerctl` is not required/involved

### DIFF
--- a/config.ron
+++ b/config.ron
@@ -85,6 +85,14 @@
         (modifiers: [], key: "XF86AudioRaiseVolume"): Spawn("amixer sset Master 5%+"),
         (modifiers: [], key: "XF86AudioLowerVolume"): Spawn("amixer sset Master 5%-"),
         (modifiers: [], key: "XF86AudioMute"): Spawn("amixer sset Master toggle"),
+
+        (modifiers: [], key: "XF86AudioNext"): Spawn("playerctl next"),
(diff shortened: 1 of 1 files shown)


Drakulix:
> such functionality could be implemented across xdg-desktop-portal-cosmic (GlobalShortcuts portal) and cosmic-settings-daemon so that third-party playerctl is not required/involved

exactly. cosmic-settings-daemon could use MPRIS directly here, which is what would be preferable. I don't mind people coming up with these configuration, maybe we should start a wiki-page, but I don't want to merge this in the official default config as it adds another dependency and the shortcuts will likely be moved into a different config file or daemon anyway.

jokeyrhyme:
Gotcha

Actually, doing this the proper way is probably still something I can manage (it's difficult/unpredictable for me to make larger contributions at the moment)

I'll see what I can come up with :)

Cheers!

PR #296, opened 2024-02-06
Implement kiosk mode for cosmic-greeter

jackpot51:
I've tested this with cosmic-greeter and some other cosmic apps and it appears to work. It always exits with the exit status of the command, or with `1` if the command is killed with a signal.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -9,7 +9,7 @@ use smithay::{
 };
 
 use anyhow::{Context, Result};
-use std::{ffi::OsString, sync::Arc};
+use std::{env, ffi::OsString, process, sync::Arc};
 use tracing::{error, info, warn};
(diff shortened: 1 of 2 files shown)


Drakulix (approved):
I feel like we could also not create a socket on the filesystem and just use an anonymous pipe + `WAYLAND_SOCKET` for the kiosk client, but this is more than good enough.

LGTM!

jackpot51:
Yeah, it may also make sense to not run Xwayland.

Drakulix:
> Yeah, it may also make sense to not run Xwayland.

On the other side the kiosk-mode is more versatile like this. Once Xwayland is lazily started this won't be an issue anymore anyway. :woman_shrugging:

PR #301, opened 2024-02-08
feat: add tiling variables to cosmic config

wash2:
Can be tested with https://github.com/pop-os/cosmic-applets/pull/214

### DIFF
--- a/cosmic-comp-config/src/lib.rs
+++ b/cosmic-comp-config/src/lib.rs
@@ -7,7 +7,7 @@ use std::collections::HashMap;
 pub mod input;
 pub mod workspace;
 
-#[derive(Clone, Debug, Default, PartialEq, CosmicConfigEntry)]
+#[derive(Clone, Debug, PartialEq, CosmicConfigEntry)]
 #[version = 1]
 pub struct CosmicCompConfig {
     pub workspaces: workspace::WorkspaceConfig,
@@ -17,11 +17,34 @@ pub struct CosmicCompConfig {
     pub xkb_config: XkbConfig,
     /// Autotiling enabled
     pub autotile: bool,
-    /// Tile windows on all workspaces
-    /// If set to false, autotile only applies to new windows
-    pub tile_all_windows: bool,
+    /// Determines the behavior of the autotile variable
+    /// If set to Global, autotile applies to all windows in all workspaces
+    /// If set to PerWorkspace, autotile only applies to new windows, and new workspaces
+    pub autotile_behavior: TileBehavior,
     /// Active hint enabled
     pub active_hint: bool,
+}
+
+impl Default for CosmicCompConfig {
+    fn default() -> Self {
+        Self {
+            workspaces: Default::default(),
+            input_default: Default::default(),
+            input_touchpad: Default::default(),
+            input_devices: Default::default(),
+            xkb_config: Default::default(),
+            autotile: Default::default(),
+            autotile_behavior: Default::default(),
+            active_hint: true,
+        }
+    }
+}
+
+#[derive(Debug, Default, Copy, Clone, PartialEq, Deserialize, Serialize)]
+pub enum TileBehavior {
+    #[default]
+    Global,
+    PerWorkspace,
 }
 
 #[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
--- a/cosmic-comp-config/src/lib.rs
+++ b/cosmic-comp-config/src/lib.rs
@@ -8,12 +8,20 @@ pub mod input;
 pub mod workspace;
 
 #[derive(Clone, Debug, Default, PartialEq, CosmicConfigEntry)]
+#[version = 1]
 pub struct CosmicCompConfig {
     pub workspaces: workspace::WorkspaceConfig,
     pub input_default: input::InputConfig,
     pub input_touchpad: input::InputConfig,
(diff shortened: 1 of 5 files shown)


src/backend/render/mod.rs
```diff
@@ -664,7 +668,7 @@ where
 
     let offset = match previous.as_ref() {
         Some((previous, previous_idx, start)) => {
-            let layout = state.config.workspace.workspace_layout;
+            let layout = state.config.config.workspaces.workspace_layout;
```
Drakulix, line 671: I am not really a big fan of the `config.config` throughout the code, I think that is confusing.

Couple of ideas to resolve this:
- Make a separate `TilingConfig` struct instead of querying `CosmicConfConfig` directly.
- Rename the field to `cosmic` (to differentiate from `static` and `dynamic`).

cosmic-comp-config/src/lib.rs
```diff
@@ -8,12 +8,20 @@ pub mod input;
...
     pub workspaces: workspace::WorkspaceConfig,
     pub input_default: input::InputConfig,
     pub input_touchpad: input::InputConfig,
     pub input_devices: HashMap<String, input::InputConfig>,
     pub xkb_config: XkbConfig,
+    /// Autotiling enabled
+    pub autotile: bool,
+    /// Tile windows on all workspaces
+    /// If set to false, autotile only applies to new windows
+    pub tile_all_windows: bool,
```
Drakulix, line 22: I am not sure I understand how this works. Is this a toggle that moves all windows to the tiling layer at runtime? Does this change behaviour, if it is already set at cosmic-comp's launch?

src/input/mod.rs
```diff
@@ -2048,10 +2056,28 @@ impl State {
                 }
             }
             Action::ToggleTiling => {
-                let output = seat.active_output();
-                let workspace = self.common.shell.workspaces.active_mut(&output);
-                let mut guard = self.common.shell.workspace_state.update();
-                workspace.toggle_tiling(seat, &mut guard);
+                if self.common.config.config.tile_all_windows {
+                    let autotile = !self.common.config.config.autotile;
```
Drakulix, line 2060: So the toggle tiling shortcut can now enabled/disables tiling globally? Is that dependant of the applet enabling "all workspaces" vs "per workspace"?

src/backend/render/mod.rs
wash2: Ok, I think I prefer the latter if that is alright with you?

src/input/mod.rs
wash2: Yes, as far as I've understood the design, and the text for the autotile toggle changes from "Automatically tile window" to "Automatically tile current workspace" when changing from "All workspaces" to "Per workspace". (I think the text may have a typo and it should probably say "Automatically tile all windows")

src/backend/render/mod.rs
Drakulix: Sure. I am just wondering how much this breaks current configs? Not something we need to worry about too much, but this would move all keys into a single cosmic-config file, right?

cosmic-comp-config/src/lib.rs
wash2: Ahh that is a good point. I completely forgot about static workspaces. On launch, this would set all workspaces to be the saved autotile value, but it should probably also include a list of static workspace names and their values, so they can be persisted.

src/input/mod.rs
Drakulix: AH with that typo explained, that name makes sense now. So we are using cosmic-comp for all the settings regarding the applet. Makes perfect sense to store this then, but I dislike the name for that setting as it was highly confusing to me. :sweat_smile:

This might be a bit too verbose, but could we name the key: `tiling_toggle` and not have a `bool` but something like: `enum ToggleBehaviour { Globally, PerWorkspace }` as the value?

src/backend/render/mod.rs
wash2: No, each of these keys would be a separate file, but ya, I think it breaks the current configs because it removes the "tiling_enabled" key.

cosmic-comp-config/src/lib.rs
Drakulix: > Ahh that is a good point. I completely forgot about static workspaces.

We dropped static workspaces for the alpha, so don't worry about it now. We will need a list of those eventually, but that can be added later and is only relevant for cosmic-workspaces though.

src/input/mod.rs
wash2: Ya, I'll double check I guess that my assumption was correct about the typo :sweat_smile: but I'm ok changing it to an enum

cosmic-comp-config/src/lib.rs
wash2: Ok, I will leave it out for now then. It should just be ok as an extra key  that can be added in the future.

src/backend/render/mod.rs
Drakulix: The static-config breaking isn't a big issue. Once we move over the key-bindings to something else (likely once dealing with the issue of global keybindings), I want to get rid of it anyway. So this is likely the last time we will break it and just before the alpha is the best time to do that. :laughing: 

I am fine with the second option then.

Drakulix (approved):

PR #245, opened 2023-12-14
Update libcosmic/iced

ids1024:
Seems to have a few issues currently. (Window menu not formatted correctly, panic what creating stack...)

PR #307, opened 2024-02-10
Update libcosmic/iced

ids1024:
Still has some build issues to fix.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -45,17 +45,6 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "ahash"
-version = "0.7.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5a824f2aa7e75a0c98c5a504fceb80649e9c35265d44525b5f94de4771a395cd"
-dependencies = [
- "getrandom",
- "once_cell",
- "version_check",
-]
-
-[[package]]
-name = "ahash"
 version = "0.8.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "77c3a9648d43b9cd48db467b3f87fdd6e146bcc88ab0180006cef2179fe11d01"
@@ -121,6 +110,12 @@ name = "android-properties"
 version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fc7eb209b1518d6bb87b283c20095f5228ecda460da70b44f0802523dea6da04"
+
+[[package]]
+name = "android-tzdata"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
 
 [[package]]
 name = "android_system_properties"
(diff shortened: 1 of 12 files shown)


ids1024:
This now builds and runs, but ends up with panics calling ` layout.children().next().unwrap()` when a window stack is created.

The widget logic there seems right...

ids1024:
Now it panics in `rounded_rectangle` because `draw_background` is called for `Tab` with an infinite width. Presumably `layout.bounds()` should be finite.

wash2 (approved):

Drakulix (changes requested):
This PR pins cosmic-comp to a cpu core with 100% usage when creating a stack.

ids1024:
Is that when testing a debug or release build?

When I test it, CPU usage seems fairly similar to the previous version in release builds, maybe a bit higher. (And already higher than it should be). While debug builds seem to have pretty high CPU usage and are less responsive.

Not sure what would have changed in iced here.

Drakulix:
> Is that when testing a debug or release build?
> 
> When I test it, CPU usage seems fairly similar to the previous version in release builds, maybe a bit higher. (And already higher than it should be). While debug builds seem to have pretty high CPU usage and are less responsive.


Ah yeah, debug build. Still that was *fine* before and is now quite unusable, you can't even resize a stack any more. I highly suspect something is running for every frame now, causing redraws.

ids1024:
Right. Software rendering without optimization is generally expected to perform poorly (we could configure cargo to optimize specific dependencies in debug builds to help), but it shouldn't really have gotten worse. And if there's no damage it shouldn't even need to do any rendering each frame.

ids1024:
Running `cargo flamegraph` with a dev build shows it's spending an obscene amount of time in tiny-skia `fill_path` and cosmic-text `shape_until_scroll`. Which is perhaps reasonable performance for those functions without any optimization, but they shouldn't be called so much when nothing is actually changing.

flamegraph

ids1024:
Actually, I seem to be seeing this performance issue on debug builds with `master` as well? Not sure this PR has changed this.

It seems damage tracking is broken for the stack `IcedElement`. `force_redraw` is being called every frame, clearing `last_primitives` so the software renderer has to redraw the whole bar even when nothing has changed. (I'm not sure if forcing redraws like that should be necessary with the way iced does damage tracking?)

Looking at a backtrace, `force_redraw` is called by the stack's `set_activate`, called by `CosmicMapped::set_activate`, called by `Shell::update_active`, called by `Common::refresh_focus`. Probably somewhere here it should be making sure nothing actually happens when the focus hasn't changed.

ids1024:
Bisecting, I seem to be seeing poor responsiveness in debug builds for a terminal in a stack starting with https://github.com/pop-os/cosmic-comp/pull/245.

Drakulix:
Pushed a fix for that, can you rebase and see if this PR is good now? https://github.com/pop-os/cosmic-comp/commit/7234a81672f90f81db67e67c119c6a4f8983ee73

ids1024:
Yep, that seems to help.

Using `opt-level = 2`  in dev builds for `tiny-skia`, `rustybuzz`, and `tty-parser` doesn't seem to add much to build time. So I've added a commit with that. This should help avoid excess CPU usage when there actually is damage. Though the this suggests the `iced_tiny_skia` renderer may need some optimization, particularly around text.

I think this branch should be good to merge. Other than dev build performance, I've been using this branch a fair amount in a release build, and haven't noticed any issues.

Drakulix:
> Using `opt-level = 2`  in dev builds for `tiny-skia`, `rustybuzz`, and `tty-parser` doesn't seem to add much to build time. So I've added a commit with that. This should help avoid excess CPU usage when there actually is damage. Though the this suggests the `iced_tiny_skia` renderer may need some optimization, particularly around text.

Oh neat, didn't know you could selectively optimize like that. Nice!

Drakulix (approved):

ids1024:
Yep. I saw something recently about how you can now even use cranelift to build your crate while having dependencies built and optimized with LLVM. I still need to experiment with that one.

PR #308, opened 2024-02-12
use memory for system cursor

cmeissl:
smithay only considers memory based cursor elements for scan-out on a cursor plane.
without this change the default system cursor won't be put on a cursor plane.


Vixea (approved):
not a maintainer or anything but with the patch one of the bugs I have with an app is fixed with this

Drakulix (approved):
LGTM. Thanks!

PR #309, opened 2024-02-13
config.ron: add cosmic-files

ahoneybun:
Adds cosmic-files for Super+F

### DIFF
--- a/config.ron
+++ b/config.ron
@@ -71,7 +71,7 @@
         (modifiers: [Super, Shift], key: "r"): Resizing(Inwards),
 
         (modifiers: [Super], key: "b"): Spawn("xdg-open http://"),
-        (modifiers: [Super], key: "f"): Spawn("xdg-open ~"),
+        (modifiers: [Super], key: "f"): Spawn("cosmic-files"),
         //TODO: ability to select default terminal
         (modifiers: [Super], key: "t"): Spawn("cosmic-term"),
 


XV-02:
I do not believe we should merge this. 

This works - with `cosmic-files` or the appropriate commands for nautilus, etc - as a temporary remedy.

There are two particular reasons for opposing merging.
1) At current, `cosmic-files` effectively works as a read-only file manager. For most users, something that looks like as complete as `cosmic-files` but does not yet allow for full functionality might be even more frustrating than `xdg-open` bringing the wrong application up. 
2) This doesn't resolve the [underlying issue](https://github.com/pop-os/cosmic-session/issues/39) of the wrong application taking default priority for certain mimetypes - in this case the `inode/directory` type. 

As such, this isn't so much a fix for the core issue as it is a bandaid. That said, I do think that users who run into this issue should consider modifying their config files for their own ease of use.

gabriele2000:
I agree with @XV-02 since as much as I like how `cosmic-files` is turning out, it's still incomplete and lacks basic features that most users would like to use:
- Opening a terminal in "this folder" (plus, using nautilus for that still opens `gnome-terminal` and not `cosmic-terminal`
- Cannot empty the recycle bin
- Cannot see detailes info about a file/folder
- Cannot open something using admin privileges
- Doesn't open most text files that aren't .txt
- "Open in new tab" is still not available

The best way would be to further wait for `cosmic-files` to be more complete, then implement the "keyboard shortcuts" settings page and finally the user would be able to change from the default "cosmic-files" to another file manager, if they like.

jackpot51:
We will be deploying a mimeapps.list file for cosmic to ensure the right defaults are selected. For now I am okay with nautilus being selected by default, as there is a lot of work to do on cosmic-files. Thankfully work is progressing quickly and this shouldn't be too long.

PR #310, opened 2024-02-13
config.ron: Add alternative screenshot hotkey (Super + Shift + P)

ryanabx:
This commit adds ~~Super + Shift + S~~ Super + Shift + P as a screenshot hotkey. Some computers (Macbooks, some ASUS laptops, etc) do not have a dedicated Print key. This ensures they can trigger a screenshot without opening a terminal to call cosmic-screenshot.

### DIFF
--- a/config.ron
+++ b/config.ron
@@ -81,6 +81,7 @@
         (modifiers: [Super]): Spawn("cosmic-launcher"),
 
         (modifiers: [], key: "Print"): Spawn("cosmic-screenshot"),
+        (modifiers: [Super, Shift], key: "P"): Spawn("cosmic-screenshot"),
 
         (modifiers: [], key: "XF86AudioRaiseVolume"): Spawn("amixer sset Master 5%+"),
         (modifiers: [], key: "XF86AudioLowerVolume"): Spawn("amixer sset Master 5%-"),


RyzenDew (approved):
tested and works nice job

Drakulix:
While I don't mind having another combination for taking screenshots, we also are still lacking a default combination for toggling sticky windows, which would also be a candidate for `Super+Shift+S` I assume.

This might be a question for @pop-os/ux as https://github.com/pop-os/cosmic-epoch/issues/46 doesn't contain default shortcuts for either of them. Open for suggestions.

ryanabx:
> This might be a question for @pop-os/ux as https://github.com/pop-os/cosmic-epoch/issues/46 doesn't contain default shortcuts for either of them. Open for suggestions.

I'm fine with another key combination, here's some ideas:
Super + Ctrl + S
Super + Alt + S
Super + Shift + any F Key (idk which ones are taken)
Super + Shift + P (P for print?)

Also, whenever people tag @pop-os/ux, I can't see the members of the group and it doesn't even show up as a group.

maria-komarova:
There is at least one member of the group at the moment. I agree that it seems a good idea to have an alternative key combination for screenshots. We might need to ponder the exact key combination just a little longer since we should make sure it doesn't come in contradiction with anything else. But I'm sure we can settle on something like Super+Shift+P or another alternative.

ryanabx:
> There is at least one member of the group at the moment.
I think the group must be invisible to non system76 employees, it shows up as plaintext to me ¯\_(ツ)_/¯

> I agree that it seems a good idea to have an alternative key combination for screenshots. We might need to ponder the exact key combination just a little longer since we should make sure it doesn't come in contradiction with anything else. But I'm sure we can settle on something like Super+Shift+P or another alternative. 

Yeah! Super+Shift+P is one that's nice, my only ask is that somewhere the screenshot keys and other hotkeys be documented so that other users can see

ryanabx:
I have changed the PR to add Super + Shift + P as the alt screenshot hotkey. Thoughts? @maria-komarova

Bahasnyldz:
I think you should able to take screenshot with one hand, P is so far from shift and super keys. 
Maybe consider Super + Shift + S for screenshot and Super+S for sticky windows?

maria-komarova:
We have discussed this more internally and rather than introducing an alternative shortcut right now we're wondering if typing "screenshot" (or a part of the word) in the launcher might be a good way to launch the screenshot tool. One can also add an second shortcut in the Settings once it is implemented or in the cosmic config file right now.

ryanabx:
> We have discussed this more internally and rather than introducing an alternative shortcut right now we're wondering if typing "screenshot" (or a part of the word) in the launcher might be a good way to launch the screenshot tool. One can also add an second shortcut in the Settings once it is implemented or in the cosmic config file now.

I think that'd be a great idea. The only problem I can envision is that sometimes you want to screenshot something while your mouse is focused on a certain position (i.e. if I were trying to get a screenshot of something highlighted) and opening the launcher causes those elements to become unfocused.

I'd still support an alternative screenshot hotkey in addition to the launcher shortcut, but since it's manually remappable by editing the config file it's not a high priority

PR #317, opened 2024-02-21
Update Smithay, with `Window` that supports X11 surfaces

ids1024:


Drakulix (approved):
Seems like a nice improvement for the code-base!

Issue #231, opened 2023-11-22
Fullscreen SDL2 applications not working properly on XWayland with `SDL_VIDEO_MINIMIZE_ON_FOCUS_LOSS`


ids1024:
I've been looking into an issue with XWayland windows unfullscreening when the cursor leaves the surface and moves to a different monitor.

It seems this happens with apps that enable `SDL_HINT_VIDEO_MINIMIZE_ON_FOCUS_LOSS`. Which was the default until SDL 2.0.14, but changed to "improve behavior with some WMs".

This can be overridden by setting the `SDL_VIDEO_MINIMIZE_ON_FOCUS_LOSS` env var.

On Cosmic, this behavior in SDL causes the surface to lose fullscreen status, but not be "minimized". It thus is never un-minimized and restored to fullscreen by SDL. On Gnome Shell X11, the setting behaves as its name suggests: the window minimizes to the dock, and is fullscreen again when re-opened. I haven't compared Sway or i3 yet.

I was hoping investigating these apps would also help fix similar issues with fullscreening wine applications. Perhaps wine does something similar? Or that may somehow be a different issue with similar symptoms.

ids1024:
Looks like Sway has a workaround for this, that also applies to wine. Something similar should work in cosmic-comp: https://github.com/swaywm/sway/commit/4f718e6c75d2168d417b4b41c0e3d5408e0afded

Clients probably should accept the WM ignoring minimize requests, but apparently clients expect it to work.

Drakulix:
> Looks like Sway has a workaround for this, that also applies to wine. Something similar should work in cosmic-comp: [swaywm/sway@4f718e6](https://github.com/swaywm/sway/commit/4f718e6c75d2168d417b4b41c0e3d5408e0afded)
> 
> Clients probably should accept the WM ignoring minimize requests, but apparently clients expect it to work.

xdg-shell has an explicit `minimize`-capability now. We should only apply such workarounds, when the bound version is lower or (which seems to be the more pressing issue here) when dealing with Xwayland windows at least.

I'll push a commit that makes sure, we don't accidentally expose the Minimize-Capability.

We probably should also think about how we want to integrate minimizing with the dock in the long run (and thus don't require such stupid workarounds any more).

ids1024:
This is about the handling of the EWMH `_NET_WM_STATE` property in the X WM. I don't think that `xdg-shell` feature would be applicable.

[...]

PR #319, opened 2024-02-23
Minimize

Drakulix:


ids1024:
https://github.com/pop-os/cosmic-comp/issues/231 still seems to be an issue, so minimization in XWayland must not be working quite the way X clients expect it too. (Which I expected; I had trouble getting that to work when I tried just modifying the XWayland code in the past to fix this.)

But that doesn't block this, anyway.

ids1024:
Not sure what I had tried previously.

It seems `X11Surface::set_mapped(false)` is defined to set `WM_STATE` to `IconicState`. But isn't called anywhere in cosmic-comp.

That also doesn't seem quite right. https://x.org/releases/X11R7.6/doc/xorg-docs/specs/ICCCM/icccm.html seems to describe both `IconicState` and `NormalState` as "mapped".

PR #322, opened 2024-02-24
WIP Input focus update

ids1024:
Not too much to see here code-wise at the moment (just an initial attempt at combining `LockSurface` and `Popup` focus targets into a `WlSurface` target). But I figure I may as well start a WIP PR now that I've been looking into what to change here.

Updating cosmic-comp for https://github.com/Smithay/smithay/pull/1326 is fairly complicated. Particularly trying to do things more correctly, and not just copying the logic from the trait impls removed there. Adjusting how pointer focuses are defined without updating to that commit seems like a good place to start iterating on things without have to address everything at once.

The Wayland protocol spec doesn't define how exactly input to subsurfaces works, but on KDE, Gnome Shell, And Sway:
* Subsurfaces the mouse moves over get pointer focus
* Subsurfaces never get keyboard focus

On cosmic-comp, it seems with a couple subsurfaces, some but not all subsurfaces can get pointer focus? So something isn't quite right.

Given how we were implementing `PointerTarget` for `Popup` and `LockSurface` to just call the `PointerTarget` methods on the `WlSurface`, we're presumably also not sending pointer input to subsurfaces of the `LockSurface` or `Popup`. So we want to have a `PointerFocusTarget::WlSurface` with the subsurface.

Meanwhile, for `KeyboardFocusTarget`, we don't want to have a subsurface as the focus. And it seems some code here does care that the keyboard focus is a popup... I wonder if condensing into a `KeyboardFocusTarget::WlSurface` makes as much sense as the simplification of combining surface types into a `PointerFocusTarget::WlSurface`.

### DIFF
--- a/src/shell/focus/target.rs
+++ b/src/shell/focus/target.rs
@@ -61,6 +61,13 @@ impl TryFrom<PointerFocusTarget> for KeyboardFocusTarget {
         match target {
             PointerFocusTarget::WlSurface(surface) => Ok(KeyboardFocusTarget::WlSurface(surface)),
             PointerFocusTarget::Element(mapped) => Ok(KeyboardFocusTarget::Element(mapped)),
+            PointerFocusTarget::X11(surface) => {
+                if let Some(wl_surface) = surface.wl_surface() {
+                    Ok(KeyboardFocusTarget::WlSurface(wl_surface))
+                } else {
+                    Err(())
+                }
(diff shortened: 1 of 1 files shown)


Drakulix:
Closing in favour of https://github.com/pop-os/cosmic-comp/pull/322. Thanks for starting this!

PR #323, opened 2024-02-24
Chore: Use Just instead of Make

ryanabx:
This PR moves the build automation system over to Just instead of Make.

Todo (but I may not have the time to do them):

[] Test using the justfile commands
[] Move the debian build/install commands over to just


justfile
```diff
@@ -0,0 +1,74 @@
...
+    cargo clippy --all-features {{args}} -- -W clippy::pedantic
+
+# Runs a clippy check with JSON message format
+check-json: (check '--message-format=json')
+
+# Installs files
+install:
+    install -Dm0755 {{bin-src}} {{bin-dst}}
+
+install-bare-session:
```
Drakulix, line 51: Should this maybe depend on `install`? The bare session without the binary isn't particular useful.

ryanabx: Updated

justfile
```diff
@@ -0,0 +1,74 @@
...
+    install -Dm0755 {{bin-src}} {{bin-dst}}
+
+install-bare-session: install
+    @just data/install-bare-session
+
+# Uninstalls installed files
+uninstall:
+    rm {{bin-dst}}
+
+uninstall-bare-session:
```
Drakulix, line 58: I am sorry to only notice this now, but I guess this should then also depend on `uninstall`. :sweat_smile:

ryanabx: Still needs testing, but also done!

justfile
```diff
@@ -0,0 +1,74 @@
...
+
+rootdir := ''
+prefix := '/usr'
+
+base-dir := absolute_path(clean(rootdir / prefix))
+
+export SHARE_DIR := sharedir
+export BIN_DIR := base-dir / 'bin'
+export LIB_DIR := libdir
+
```
dengelt, line 12: Suggestion: Add a line like `cargo-target-dir := env('CARGO_TARGET_DIR', 'target')` here to read the CARGO_TARGET_DIR environment variable.

justfile
```diff
@@ -0,0 +1,74 @@
...
+bin-src := 'target' / 'release' / name
```
dengelt, line 13: Suggestion: Change this line to `bin-src := cargo-target-dir / 'release' / name` to use the previously read environment variable instead of hardcoding the target directory.

PR #327, opened 2024-02-26
fix(stack): set correct colors for light theme

mmstick:
Updates colors for the stack tabs to match the current values from the design documentation.
This will noticeably improve visibility of tabs when using a light theme.

### DIFF
--- a/src/shell/element/stack.rs
+++ b/src/shell/element/stack.rs
@@ -861,7 +861,9 @@ impl Program for CosmicStackInternal {
                 theme::Container::custom(|theme| iced_widget::container::Appearance {
                     icon_color: Some(Color::from(theme.cosmic().background.on)),
                     text_color: Some(Color::from(theme.cosmic().background.on)),
-                    background: Some(Background::Color(theme.cosmic().palette.neutral_3.into())),
+                    background: Some(Background::Color(tab::primary_container_color(
+                        theme.cosmic(),
+                    ))),
                     border_radius: BorderRadius::from([8.0, 8.0, 0.0, 0.0]),
                     border_width: 0.0,
                     border_color: Color::TRANSPARENT,
--- a/src/shell/element/stack/tab_text.rs
+++ b/src/shell/element/stack/tab_text.rs
@@ -1,53 +1,63 @@
+use std::hash::{Hash, Hasher};
+
 use cosmic::{
-    iced::Element,
+    iced::{alignment, Point},
     iced_core::{
         gradient,
         layout::{Layout, Limits, Node},
         mouse::Cursor,
         renderer::{self, Renderer as IcedRenderer},
-        widget::{Tree, Widget},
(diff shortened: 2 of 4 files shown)


src/shell/element/stack.rs
```diff
@@ -861,7 +861,11 @@ impl Program for CosmicStackInternal {
                 theme::Container::custom(|theme| iced_widget::container::Appearance {
                     icon_color: Some(Color::from(theme.cosmic().background.on)),
                     text_color: Some(Color::from(theme.cosmic().background.on)),
-                    background: Some(Background::Color(theme.cosmic().palette.neutral_3.into())),
+                    background: Some(Background::Color(if theme.cosmic().is_dark {
+                        Color::from_rgba(0.149, 0.149, 0.149, 1.0)
```
Drakulix, line 865: Aren't these defined in the theme? If not, can we make these at least constants please, because I feel like we are using these in multiple places.

Additionally, we are hard-coding specific values for light and dark here. Can't we take the user-selected background color?

src/shell/element/stack/tab_text.rs
```diff
@@ -141,7 +141,7 @@ where
                 border_color: Color::TRANSPARENT,
             },
             Background::Gradient(Gradient::Linear(
-                gradient::Linear::new(Degrees(180.))
+                gradient::Linear::new(Degrees(90.))
```
Drakulix, line 144: nice catch!

src/shell/element/stack.rs
mmstick: In our figma designs, only the text color and borders inherit from the system theme.
I can make constants for them. Though I'm currently working on hiding the text overlay when text is not overflowing.

Drakulix: > I can make constants for them.

Thanks

> Though I'm currently working on hiding the text overlay when text is not overflowing.

Estimating the size of the widgets beforehand was quite complex for the various design. Any changes making this simpler are very welcome.

mmstick:
All that remains is fixing the gradient color on the light theme.

Drakulix (approved):
Looks great code-wise. I'll give it a quick test and merge once marked ready.

Drakulix (changes requested):
This is crashing for me, if I overflow the stack to the point it should show arrows left and right.

```
ERROR panic: thread 'main' panicked at 'called `Option::unwrap()` on a `None` value': /home/drakulix/.cargo/git/checkouts/libcosmic-b367e32ffc370f4f/efe4ce2/iced/widget/src/container.rs:287
```

mmstick:
Looking into the cause of the missing children in the container overlay.

mmstick:
I pushed a fixed for the layout issue. The `Tabs` widget had a misalignment of children in layout in one area. GitHub's UI  doesn't seem to be showing the commits though, but it is in the branch.

Drakulix (approved):
Thanks!

PR #332, opened 2024-03-06
Set keyboard LEDs (capslock, numlock, scrollock)

ids1024:
It doesn't seem there's currently a way to iterate over input devices, so this adds a `Vec` of libinput keyboard devices. When run with the kms backend.

Not the most elegant solution, but it doesn't seem practical to add a type generic to `Devices` given how backends are handled, and `Device` can't be made into a trait object. Another trait could be added, but this is fine, and technically should perform better (but if it actually had to iterate over many keyboards, something has gone very wrong).


Drakulix (approved):
LGTM

PR #341, opened 2024-03-07
tiling: Check for popups before tiles in `element_under`

Drakulix:
Draft while https://github.com/Smithay/smithay/pull/1354 isn't merged and cosmic-comp's smithay dependency isn't updated.


Drakulix:
Not needed with WlSurface pointer focus in https://github.com/pop-os/cosmic-comp/pull/380

PR #342, opened 2024-03-07
feat: workspace switching touchpad gestures

ryanabx:
This PR implements the next workspace and previous workspace touchpad gestures.

To-do:

- [x] Cleanup/refactoring needed. I'll need some feedback on how it's currently implemented
- [x] ~~Fix a visual bug where the screen turns gray when letting go of the gesture in between workspaces~~
- [x] Factor in velocity possibly? If the gesture is let go at a fast velocity then always continue the switch workspaces action
- [x] Spring animations when letting go of gesture

Left for a future PR:
- [ ] Invalid gestures (i.e. trying to switch to a workspace that doesn't exist, rubber banding, etc.)

Implementation details (OUTDATED):

* A new enum WorkspaceDelta is created (name can be changed) that can have either an Instant (with optional offset) or a delta from 0.0 to 1.0.
* A new tuple is added to State::Common called gesture_state which contains the number of fingers used, the total delta, and the direction the gesture is in.
* New constants GESTURE_MAX_LENGTH and GESTURE_THRESHOLD
  * GESTURE_MAX_LENGTH is the maximum length of the gesture that moves the screen
  * GESTURE_THRESHOLD is the percentage of the maximum length required to complete the workspace switch
* New functions in Shell that help update deltas for workspaces

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -3,6 +3,7 @@
 use crate::{
     backend::render::cursor::CursorState,
     config::{xkb_config_to_wl, Action, Config, KeyModifiers, KeyPattern},
+    input::gestures::{GestureState, SwipeAction},
     shell::{
         focus::{target::PointerFocusTarget, FocusDirection},
         grabs::{ResizeEdge, SeatMenuGrabState, SeatMoveGrabState},
@@ -10,7 +11,8 @@ use crate::{
             floating::ResizeGrabMarker,
             tiling::{SwapWindowGrab, TilingLayout},
         },
-        Direction, FocusResult, MoveResult, OverviewMode, ResizeDirection, ResizeMode, Trigger,
+        Direction, FocusResult, InvalidWorkspaceIndex, MoveResult, OverviewMode, ResizeDirection,
+        ResizeMode, Trigger, WorkspaceDelta,
     },
     state::Common,
     utils::prelude::*,
@@ -67,6 +69,8 @@ use std::{
 };
 
 crate::utils::id_gen!(next_seat_id, SEAT_ID, SEAT_IDS);
+
+pub mod gestures;
 
 #[repr(transparent)]
 pub struct SeatId(pub usize);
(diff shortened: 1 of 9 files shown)


ryanabx:
@Drakulix the functionality is now here, but it needs some cleanup and review! There's also a bug in my code that's causing the screen to go gray when you let go of the gesture in between workspaces. Take a look at the changes when you can! :)

XV-02:
So, I just want to make a note of a few things I am noticing for when our engineers look at this:

1. When you finish changing workspaces, the display flickers once.

2. When (as noted by ryanabx) you stop mid-way through the change animation, the screen will display grey. Interestingly enough, for me, both the panel and the dock were still correctly rendered. Additionally, any event which causes a screen redraw (moving the mouse, or having a flashing text cursor) seems to resolve this. I'm not certain of the cause of this, but it reminds me somewhat of the occasional grey screen when attempting to lock the screen using `cosmic-greeter` 

3. Currently, this workspace gesture implementation is leveraging "natural" scrolling. While we support "natural" scrolling as an option, it is not our default. We should decide how we want to handle gesture scroll directions and if we want to expose any kind of settings page for it in future.

4. We should check the animation speed. The animation on this implementation feels quite a bit quicker than our current equivalent in Gnome. This is most noticeable when "flicking" between workspaces, as the change doesn't quite feel animated or visually intuitive on account of how short it is.

ryanabx:
v2:

* Fixed the gray screen bug when swiping in-between workspaces and letting go before going all the way
* Changed GESTURE_MAX_LENGTH to 150 (up from 50, should be a longer swipe to activate)
* Some refactoring

@Drakulix Now would be a good time for an initial review and to ask ~~a few questions~~ (actually I could only think of one question so far):

* How should the threshold to switch workspaces be calculated? Currently it's just if the threshold is > 0.5, then go to the next workspace, else previous one. I think we should factor in velocity somehow, for the people who swipe quickly. I don't know yet how we do that though.
  * (This part could be implemented in a future PR)

src/shell/mod.rs
```diff
@@ -1151,7 +1193,124 @@ impl Shell {
...
+                            WorkspaceDelta::Gesture(delta) => {
+                                if delta.abs() >= GESTURE_THRESHOLD {
+                                    set.activate(
+                                        set.active,
+                                        Some(WorkspaceDelta::Shortcut(Instant::now(), delta)),
+                                        &mut self.workspace_state.update(),
+                                    )?;
+                                } else {
+                                    set.activate_previous(
+                                        Some(WorkspaceDelta::Shortcut(Instant::now(), 1.0 - delta)),
```
Drakulix, line 1306: Clever trick to make the compositor run the rest of the animation itself. But this would result in the rest always taking 200ms (or whatever `ANIMATION_DURATION` ends up being) even if the remaining movement is just a tiny portion, right?

So we kinda have to either forge an `Instant` in the past (possible by substracting `Duration`) or add a `Duration` parameter to the `Shortcut` variant.

Also regarding my previous comment, I see now, why `activate` takes an `Option` to be able to insert these `Shortcut`-variants. So replacing this with a boolean wouldn't work, but I still would prefer getting rid of the `Option`.

Drakulix (commented):
Really liking a lot about this implementation, except for a couple of comments/nitpicks.


Feeling is much more difficult to judge, but my first impression is, that it is definitely not *there* yet. I feel like the current system of just easing the values isn't enough here for a satisfying feeling.

See what niri has recently implemented for gestures (essentially by stealing a bunch of what libadwaita is doing):

- https://github.com/YaLTeR/niri/releases/tag/v0.1.3
- https://github.com/YaLTeR/niri/commit/ae89b2e514fd4c22a38c2dce258707c369ca944a
- https://github.com/YaLTeR/niri/commit/f9127616b02eb12b462aca3e48fa27bf0479a4b0

Given niri's code base is GPL-3.0 and cosmic-comp is as well, we can take liberal "inspiration" as long as credit is properly attributed.

We currently also can't scroll where there is no workspace. Meaning gestures in the wrong direction on the first or last workspace simply do nothing. We probably want (with proper rubber banding) to allow this, which might be best represented by another `WorkspaceDelta` variant, as this would require some more changes to the rendering code. We probably want to move all elements, **but** the `background` and `bottom` layer-shell layers in that case.

src/shell/mod.rs
ryanabx: > Clever trick to make the compositor run the rest of the animation itself. But this would result in the rest always taking 200ms (or whatever ANIMATION_DURATION ends up being) even if the remaining movement is just a tiny portion, right?

Nope, Shortcut takes a f64 value for how far in the animation we should be starting from (from 0.0 to 1.0) and some math is done to make sure we're at the right place in the animation.

src/input/mod.rs
```diff
@@ -1094,6 +1096,44 @@ impl State {
                         },
```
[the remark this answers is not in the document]
ryanabx: Fixed in 35ae0cbe9bf5875ab69116d0e0acb1535d7f0856

src/input/mod.rs
```diff
@@ -1081,6 +1082,7 @@ impl State {
                             fingers: event.fingers(),
                         },
                     );
+                    self.common.gesture_state = Some((event.fingers(), Direction::Right, 0.0));
```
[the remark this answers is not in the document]
ryanabx: Fixed in 35ae0cbe9bf5875ab69116d0e0acb1535d7f0856

src/input/mod.rs
```diff
@@ -1107,6 +1147,43 @@ impl State {
                             cancelled: event.cancelled(),
                         },
                     );
```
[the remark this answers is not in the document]
ryanabx: Fixed in 35ae0cbe9bf5875ab69116d0e0acb1535d7f0856

src/shell/mod.rs
ryanabx: > Also regarding my previous comment, I see now, why activate takes an Option to be able to insert these Shortcut-variants. So replacing this with a boolean wouldn't work, but I still would prefer getting rid of the Option.

Would maybe having WorkspaceDelta impl Default be a good compromise? Then you could just insert WorkspaceDelta::default or whatever instead of None

ryanabx:
Fixed some of the issues in your review. I'll work on the animations today as well!

> We currently also can't scroll where there is no workspace. Meaning gestures in the wrong direction on the first or last workspace simply do nothing. We probably want (with proper rubber banding) to allow this, which might be best represented by another `WorkspaceDelta` variant, as this would require some more changes to the rendering code. We probably want to move all elements, **but** the `background` and `bottom` layer-shell layers in that case.

Phew, I'll look into this too, that'll be a little more advanced :sweat_smile:

src/shell/mod.rs
Drakulix: > Would maybe having WorkspaceDelta impl Default be a good compromise? Then you could just insert WorkspaceDelta::default or whatever instead of None

I don't think default covers it well given the two variants.

And `WorkspaceDelta::shortcut()` is only marginally longer than `WorkspaceDelta::default()` and tells you much more.

src/state.rs
```diff
@@ -136,6 +136,33 @@ pub fn advertised_node_for_surface(w: &WlSurface, dh: &DisplayHandle) -> Option<
...
+    }
+}
+
+impl Default for GestureState {
+    fn default() -> Self {
+        GestureState {
+            fingers: 0,
+            direction: Direction::Right,
+            delta: 0.0,
+        }
```
Drakulix, line 162:
```suggestion
        GestureState::new(0)
```

src/state.rs
```diff
@@ -159,6 +159,7 @@ pub struct Common {
     pub clock: Clock<Monotonic>,
     pub should_stop: bool,
     pub local_offset: time::UtcOffset,
+    pub gesture_state: Option<(u32, Direction, f64)>,
```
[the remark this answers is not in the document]
Drakulix: Good stuff!

src/state.rs
ryanabx: Fixed in 7385acea4cf6bc50834d253a190a30d532feaa64

src/shell/mod.rs
```diff
@@ -383,15 +392,42 @@ impl WorkspaceSet {
...
             state.add_workspace_state(&self.workspaces[idx].handle, WState::Active);
-
-            self.previously_active = Some((old_active, Instant::now()));
+            match workspace_delta {
+                Some(x) => self.previously_active = Some((old_active, x)),
+                None => {
+                    self.previously_active =
+                        Some((old_active, WorkspaceDelta::Shortcut(Instant::now(), 0.0)))
+                }
+            }
```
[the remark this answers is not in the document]
ryanabx: Fixed in 7385acea4cf6bc50834d253a190a30d532feaa64, you're right it's much cleaner this way, thanks! :p

src/input/mod.rs
```diff
@@ -1094,6 +1096,44 @@ impl State {
...
                 }
+                if let Some((fingers, mut dir, mut delta)) = self.common.gesture_state {
+                    if let Some(seat) = self.common.seat_with_device(&event.device()).cloned() {
+                        if fingers == 3 {
+                            // TODO: Implement 3 finger swipe gestures
+                        } else if fingers == 4 {
+                            if delta == 0.0 {
+                                // Find direction
+                                if event.delta().x.abs() > event.delta().y.abs() {
+                                    // Horizontal, do nothing (for now)
```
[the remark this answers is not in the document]
ryanabx: Fixed/implemented horizontal gesture in 7385acea4cf6bc50834d253a190a30d532feaa64

ryanabx:
I've resolved most of the comments. I believe the last thing to figure out is whether we need to support a config option to change to three finger gestures, and how to implement the better-feeling animation described above.

I'm currently still confused with Niri's Spring animations, and how they work. Currently, we ease with a value from 0 to 1, but Niri uses a Duration on some of these Spring functions: https://github.com/YaLTeR/niri/commit/ae89b2e514fd4c22a38c2dce258707c369ca944a#diff-2af07d519be6bbc5de22a3324ac74473f74136c70d620d9869f8919c465bee3fR38-R62

Drakulix:
> I've resolved most of the comments. I believe the last thing to figure out is whether we need to support a config option to change to three finger gestures, and how to implement the better-feeling animation described above.
> 
> I'm currently still confused with Niri's Spring animations, and how they work. Currently, we ease with a value from 0 to 1, but Niri uses a Duration on some of these Spring functions: [YaLTeR/niri@ae89b2e#diff-2af07d519be6bbc5de22a3324ac74473f74136c70d620d9869f8919c465bee3fR38-R62](https://github.com/YaLTeR/niri/commit/ae89b2e514fd4c22a38c2dce258707c369ca944a#diff-2af07d519be6bbc5de22a3324ac74473f74136c70d620d9869f8919c465bee3fR38-R62)

From what I understand (I could be wrong, I am not super familiar with that code base), the spring animation is used to transition to the respective workspace, after the gesture was released. (During the gesture only rubber-banding applies I believe.) So the spring animation probably has a dynamic duration (depending on the velocity of the gesture for example).

ryanabx:
> From what I understand (I could be wrong, I am not super familiar with that code base), the spring animation is used to transition to the respective workspace, after the gesture was released. (During the gesture only rubber-banding applies I believe.) So the spring animation probably has a dynamic duration (depending on the velocity of the gesture for example).

Would it be okay to work on the spring animations in a separate PR? And land preliminary touchpad gesture support before then?

ryanabx:
Velocity is factored in now! Currently velocity is just only to determine the workspace to move to, but will be used for more things probably tonight since I'm on spring break and what else am I gonna do? Swipe tracking is implemented with heavy inspiration from Niri. I plan on looking more into how Niri does things like its rubber band smoothing on the gesture and how it reacts when the gesture ends.

src/input/mod.rs
```diff
@@ -675,6 +676,8 @@ impl State {
 
                     position += event.delta().as_global();
 
+                    // RYANABX LOOK HERE
+
```
Drakulix, line 680: Well, if I am supposed to merge this, you might need to cleanup some left-over comments. ;)

src/state.rs
```diff
@@ -136,6 +136,136 @@ pub fn advertised_node_for_surface(w: &WlSurface, dh: &DisplayHandle) -> Option<
...
+        }
+    }
+}
+
+impl Default for GestureState {
+    fn default() -> Self {
+        GestureState::new(0)
+    }
+}
+
```
Drakulix, line 268: I think this code deserves it's own module it this point.

Drakulix (commented):
I don't know, it doesn't feel difficult to review, but what would make it easier, would be you using separate commits.

I personally rather merge something complete, rather than in-progress.

src/input/mod.rs
ryanabx: oops :laughing:

src/state.rs
ryanabx: done in 26b8cd8ec581870c8e194c08745734dac96d493a

src/input/mod.rs
ryanabx: fixed in 0fd688dae601937a723586e37bc1d90359a47e9e

ryanabx:
The PR is (once again) ready for review, now with spring gestures! @Drakulix this time I'm sure it's ready.

I've decided to leave "invalid gestures" i.e. attempting to swipe to a workspace that doesn't exist for another PR. I'll probably work on that next week on and off because my school will be starting back up. Some of the infrastructure is already in this PR, for example the Rubber Band struct and impl from Niri.

Drakulix (approved):
Code look good!

Feels much better than before, really smooth now.

It is a bit unfortunate, that you can not reverse the direction after it has detected one, but that is also something a future PR *might* address.

Thanks for all the time and patience working on this.

ryanabx:
> That is not what I mean. My problem is you can't scroll a little in one direction and change your mind and go the other way without releasing the touchpad in between.

Ohh yeah, I can look into the possibility of making that possible in a follow up PR

PR #361, opened 2024-03-17
fix: Avoid crash on malformed keymap config

LouisDISPA:
I was testing cosmic DE and while configuring the keymaps I accidentally broke the keymap config file.
This makes cosmic-comp to crash and you must go through another desktop environment to change the config and fix it.
I think it's better to log the error and skip the file.

Also I was having a specific issue when installing from cosmic-epoch, they are no keymaps by default when no config is found. The second commit adds the default `config.ron` keymaps if no config file is found. It is a specific issue so I don't know if you want that.

Thanks for all of your amazing work, can't wait to see the alpha.
Also love the blog posts, they are really well made.

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -207,13 +207,23 @@ impl Config {
             debug!("Trying config location: {}", path.display());
             if path.exists() {
                 info!("Using config at {}", path.display());
-                let mut config: StaticConfig =
-                    ron::de::from_reader(OpenOptions::new().read(true).open(path).unwrap())
-                        .expect("Malformed config file");
-
-                key_bindings::add_default_bindings(&mut config.key_bindings, workspace_layout);
-
-                return config;
+                let Ok(file) = OpenOptions::new().read(true).open(path) else {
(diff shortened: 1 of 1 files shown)


src/config/mod.rs
```diff
@@ -207,20 +207,39 @@ impl Config {
...
+        }
 
-                key_bindings::add_default_bindings(&mut config.key_bindings, workspace_layout);
+        info!("No config found, consider installing a config file. Using default mapping.");
 
-                return config;
+        let mut config = ron::from_str(include_str!("../../config.ron")).unwrap_or_else(|err| {
+            debug!("Failed to load internal default config: {}", err);
+            StaticConfig {
+                // Small usefull keybindings by default
```
Drakulix, line 235:
```suggestion
                // Small useful keybindings by default
```

Drakulix (approved):
Thanks, this is a good change.

Eventually all these config stuff shall be moved into cosmic-config, where malformed or missing configs are handled much better, but in the meantime this is much better than the current state of things. :+1:

PR #364, opened 2024-03-18
feat: drag regions for maximize and half tiling

ryanabx:
this pr enables dragging the mouse to the output edges to maximize, or half tile in any direction

* ~~right now only maximize works, can't figure out why the half tiling isn't working~~ **fixed**
* ~~currently rendering an outline, not a shaded region, I'm looking for how to render a shaded region instead~~
* ~~the shaded region covers the whole output, not just the area between panels~~ **fixed**
* ~the drag regions don't extend into the panels, making it kind of awkward to maximize when there is a top panel for example~ **fixed**

https://github.com/pop-os/cosmic-comp/assets/56272643/32f0d6a0-c164-4f3b-91af-60db6c0b43aa

### DIFF
--- a/src/shell/grabs/moving.rs
+++ b/src/shell/grabs/moving.rs
@@ -12,7 +12,8 @@ use crate::{
             CosmicMappedRenderElement,
         },
         focus::target::{KeyboardFocusTarget, PointerFocusTarget},
-        CosmicMapped, CosmicSurface, ManagedLayer,
+        layout::floating::TiledCorners,
+        CosmicMapped, CosmicSurface, Direction, ManagedLayer,
     },
     utils::prelude::*,
 };
@@ -27,7 +28,7 @@ use smithay::{
             ImportAll, ImportMem, Renderer,
         },
     },
-    desktop::space::SpaceElement,
+    desktop::{layer_map_for_output, space::SpaceElement},
     input::{
         pointer::{
             AxisFrame, ButtonEvent, GestureHoldBeginEvent, GestureHoldEndEvent,
@@ -55,6 +56,7 @@ pub struct MoveGrabState {
     indicator_thickness: u8,
     start: Instant,
     previous: ManagedLayer,
+    snapping_zone: Option<SnappingZone>,
     stacking_indicator: Option<(StackHover, Point<i32, Logical>)>,
 }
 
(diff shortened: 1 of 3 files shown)


src/shell/grabs/moving.rs
```diff
@@ -200,6 +223,69 @@ impl MoveGrabState {
...
+            ),
+            SnappingZone::Bottom => (
+                Point::from((
+                    output_geometry.loc.x,
+                    output_geometry.loc.y + output_geometry.size.h - SNAPPING_RANGE,
+                )),
+                Size::from((output_geometry.size.w, SNAPPING_RANGE)),
+            ),
+        };
+        Rectangle::from_loc_and_size(loc, size).as_local()
```
Drakulix, line 273: I think it would probably be good, if this took panels into account. Meaning imo the zone should be all of the `exclusive_zone` at the specific border + SNAPPING_RANGE.

So dragging onto a panel would still activate the zone, but there would always be an additional range next to every dock/panel/etc.

src/shell/grabs/moving.rs
```diff
@@ -560,6 +673,22 @@ impl Drop for MoveGrab {
...
+                                    state.common.shell.maximize_toggle(&window, &seat);
+                                } else {
+                                    let direction = match sz {
+                                        SnappingZone::TopSnap => Direction::Up,
+                                        SnappingZone::Bottom => Direction::Down,
+                                        SnappingZone::Left => Direction::Left,
+                                        SnappingZone::Right => Direction::Right,
+                                        _ => Direction::Up,
+                                    };
+                                    window.handle_move(direction);
```
Drakulix, line 688: Just calling that on the window, won't do what you want. That handles moves e.g. inside a stack.

You want to split the `FloatingLayout::move_current_element` into `FloatingLayout::move_element` and `move_current_element`.

ryanabx: fixed in bee0b664cc83aec19f5da5100570c8f17af3c987

src/shell/grabs/moving.rs
ryanabx: fixed in bee0b664cc83aec19f5da5100570c8f17af3c987

ryanabx:
Everything is ready @Drakulix 

Do we want the render to be a solid color instead of an outline? If so, how do I do that?

Drakulix:
> Do we want the render to be a solid color instead of an outline? If so, how do I do that?

For that we would need to use the `BackdropShader` instead of the `IndicatorShader`. But I actually like the outline, I wouldn't put any time into that until we have a design for this feature.

cc @pop-os/ux, for the video in the first post of the current implementation.

> Also, when the mouse is at the very edge of the screen, I don't think it's counting that as inside the output geometry rectangle

Yeah indeed the very right and bottom edge will probably not be included, given the logic of `contains`: https://github.com/Smithay/smithay/blob/master/src/utils/geometry.rs#L1068

But you are also flooring the cursor location, so imo this should be correct?

ryanabx:
> Also, when the mouse is at the very edge of the screen, I don't think it's counting that as inside the output geometry rectangle

We discussed this in the Mattermost, and it's a separate bug that likely affects other things, so that will be resolved in a separate PR.

Otherwise the PR is ready!

maria-komarova:
Looks like a good start to me. I'm working on the design, including the multi-monitor question but it shouldn't block merging this PR. I can then test and see how we can iterate on what's here.

PR #358, opened 2024-03-15
cosmic-screencopy-v2

Drakulix:
Fixes https://github.com/pop-os/cosmic-comp/issues/338

PR #365, opened 2024-03-18
ext-image-source and ext-screencopy

Drakulix:
Implementation of **unmerged** upstream protocol ext-image-source and ext-screencopy (https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/124) based on #358.


russelltg:
It's merged now! 🎉

Drakulix:
> It's merged now! 🎉 

We know. Waiting for a new wayland-protocols release before we can update wayland-rs.

AlexProgrammerDE:
wayland-protocols 1.37 was released. Seems like wayland-rs can be updated after this PR is merged: https://github.com/Smithay/wayland-rs/pull/754

AlexProgrammerDE:
`wayland-rs` now merged the PR and the crate `wayland-protocols` from the project has these two protocols now in version `0.32.4`. I guess this PR can be updated now to use this new version that contains `ext-image-capture-source-v1` and `ext-image-copy-capture-v1`.

Quackdoc:
It would be nice if this could be supported as we are now seeing applications that can utilize this

Quackdoc:
It might be worth noting that sway has merged support for these protocols, so we may start to see more applications use them.

ids1024:
It may be easiest to support first on the client side. I've started working on client code to abstract over the cosmic-screencopy-v2 protocol and the final version of the ext protocol (which is only slightly different). If we merge support in clients first, we could just drop support for cosmic-screencopy-v2 in the compositor and replace it with ext-image-copy-capture.

ids1024:
Client-side, I was prototyping a potentially better and higher-level API for screen capture to share between the cosmic components that use screencopy, but for now in https://github.com/pop-os/cosmic-protocols/pull/46 I'm incrementally improving the API we have now, and making it work with ext-image-copy-capture (with a few more changes I still need to make, that branch should work for output/toplevel capture with ext-image-copy-capture).

So we need to decide about workspace capture (https://github.com/pop-os/cosmic-protocols/pull/47), potentially using ext-workspace-v1 (which we don't currently implement), but other than that we should be able to have our clients updated to support ext-image-copy-capture quite soon.

I'll be happy to be able to test against Sway to see if there are differences in the implementations, help identify which issues are compositor bugs, etc.

Drakulix:
Handled in other PRs, transition is done on cosmic-comp's side.

Issue #355, opened 2024-03-13
xwayland: activating a window in a different workspace doesn't focus the window


ryanabx:
To reproduce:

* Open Discord (xwayland) in one workspace (could be any xwayland app)
* Open OnlyOffice (xwayland) in another workspace (could be any xwayland app)
* Focus Discord, then click on OnlyOffice using the app list applet
* Workspace switches, but the app itself cannot be interacted with (this can be fixed by clicking the icon in the app list applet again)

This could be an app-list issue, or a cosmic-comp issue

coryc96:
This can be reproduced without the applist also. It appears to happen a lot if the xwayland apps are "overlapping" each other on two different workspaces. So for example, 
Workspace 1: Cosmic Term | Spotify
Workspace 2: Cosmic Files | Ferdium
Changing workspaces between these mean that both Spotify and Ferdium will not be interactable (until clicking it in the app list, as you mentioned)

But if you swap the position so they aren't overlapping, then they both work.
Workspace 1: Cosmic Term | Spotify
Workspace 2: Ferdium | Cosmic Files

This works as expected. I think there's another piece of the puzzle that I'm missing because I've seen extra unpredictable behavior similar to this, but I can't reproduce it. Either way, it seems to be workspace switching is not properly or fully changing the workspace context the user is in.

ryanabx:
A fun way to debug this is by playing an xwayland game with audio, that stops playing audio when unfocused. Surprisingly a lot of the time the xwayland window is focused, even when on different workspaces.

ryanabx:
Another way to activate focus when it shouldn't be activated:

* Open steam and a windowed xwayland game
* Focus xwayland game, then focus steam
* Move steam window around by clicking and dragging titlebar
* While the steam window is moving around, focus will be given to the game

Issue #329, opened 2024-03-04
toplevel remains activated when the window is no longer available


ryanabx:
The easiest example for me to repro:
1. Open Visual Studio Code, select "Open Folder"
2. Open any project folder
3. The toplevel remains active even when the window is gone.

![screenshot-2024-03-04-05-35-37](https://github.com/pop-os/cosmic-comp/assets/56272643/5f9e130d-0bc5-40f8-b037-82d1b148e128)

In the screenshot, we can see that cosmic-app-list thinks that there are two windows open, and running the `toplevel-list` example client, it shows two windows open still.

I don't know enough about wayland protocols yet to diagnose how this happened, but I assume the toplevel tells the compositor when it should be deactivated? Maybe Visual Studio Code does it in a weird way, and this is an edge case that needs to be figured out.

Issue #280, opened 2024-01-18
Mouse input for app components outside window borders doesn't register


git-f0x:
If an application running through Xwayland has a drop-down menu long enough to go outside window borders, any menu items beyond the border aren't clickable.
In the image below, the last selectable item is **SMB**.
![Screenshot_2024-01-17_22-59-29](https://github.com/pop-os/cosmic-comp/assets/150025636/902cda7f-9f8d-4bde-92c9-6e56fb84707e)

In the case of a non-resizable window being added to a stack (manually), any tabs (or parts of tabs) outside the border (active hint) aren't clickable and all windows in the stack become non-resizable while that window is in the stack.
![Screenshot_2024-01-17_23-01-12](https://github.com/pop-os/cosmic-comp/assets/150025636/9bf2e98f-9358-4334-835c-72797ba22775)

When dragging a scroll bar with a mouse, if the cursor leaves the window borders while scrolling, the scrolling is stopped, and when the cursor returns, the scroll bar jumps to the ingress point. This also happens in a maximized window (e.g. Firefox) when the cursor touches the right screen edge while dragging the scroll bar.

P.S. Unrelated to this but in the case it wasn't noticed, left-clicking on the stack symbol in the upper-left corner of a floating **non-sticky** stack makes the stack disappear and the apps stay open in the background.

git-f0x:
After some more testing, it seems the drop-down menu issue is exclusive to Xwayland apps, and after testing with that branch (running the built binary from a tty), the behavior remains the same.
The non-resizable window in a stack issue seems to persist, just that sometimes clicking on the tabs outside causes the entire stack to shrink and become unresizable.
Scrolling behavior is also the same. It would be nice if it behaved in the same way as COSMIC GNOME (when a scroll bar is being held, the scrolling follows the cursor regardless of its position on the screen).

Drakulix:
> After some more testing, it seems the drop-down menu issue is exclusive to Xwayland apps, and after testing with that branch (running the built binary from a tty), the behavior remains the same.

That this is exclusive to Xwayland apps is a very useful hints. Thanks I will further look into all the issues you listed. I had just fixed some behaviour with popups overlapping on that PR, but probably missed parts special to Xwayland.

git-f0x:
Yeah, I tested with Firefox (by pressing Alt to reveal menu options) and Inkscape, by shrinking the window, and all menu options were selectable.
[...]

Issue #279, opened 2024-01-17
Issues when attempting to run multiple electron apps in stacked windows.


XV-02:
When attempting to run multiple electron applications in stacked windows, only the most recently launched application is responsive. 

For example, I frequently use four electron-based applications:
1. The stand-alone `Slack` client.
2. The stand-alone `Spotify` client.
3. `VSCodium` - which I use multiple instances of.
4. The `Joplin` desktop client, which I use sporadically for note-taking. 

Most often, I will launch Slack first. Slack will be responsive (though other issues, like a comically large cursor in the Slack window, may be present.)

Later, I will launch  VSCodium, at which point slack will become unresponsive. The new application will be responsive however. I will quit Slack, and the new application will remain responsive. I will relaunch Slack. Slack will be responsive, but VSCodium will stop being responsive. 

If I do not relaunch Slack, but instead launch Joplin, VSCodium will still become unresponsive. As it will if I launch subsequent instances of VSCodium. 

In any configuration I have tried, this is the case. The second I launch a subsequent Electron application, the first one I launched becomes unresponsive. I don't think it is specific to any given electron application.

PR #374, opened 2024-03-21
Various xwayland/focus fixes

Drakulix:
Should address (in no particular order):
- https://github.com/pop-os/cosmic-comp/issues/355
- https://github.com/pop-os/cosmic-comp/issues/329
- https://github.com/pop-os/cosmic-comp/issues/280
- https://github.com/pop-os/cosmic-comp/issues/279
- https://github.com/pop-os/cosmic-comp/issues/162


ryanabx:
I've been waiting for this one!! Great job 👏 

If you need someone to review the code I'd be happy to after lunch 🙂

Drakulix:
> I've been waiting for this one!! Great job 👏
> 
> If you need someone to review the code I'd be happy to after lunch 🙂

I mostly need somebody running Xwayland apps more frequently to test this more.

I have deployed this on a couple of machines and will let it sit to see, if I introduced new bugs. So far everything seems to be good, but more testing definitely is necessary. So please comment with anything (positive or negative) you can find.

ryanabx:
> I mostly need somebody running Xwayland apps more frequently to test this more.

I can definitely test it! I'll test it on xwayland electron apps, and some games and keep you posted

git-f0x:
I've tested #280 and it seems to be fixed. Though the clamp mouse position PR by Ryan doesn't seem to be included (though it might be because I'm running this from a tty).

src/shell/element/surface.rs
```diff
@@ -73,6 +73,22 @@ impl From<X11Surface> for CosmicSurface {
...
+impl PartialEq<WlSurface> for CosmicSurface {
+    fn eq(&self, other: &WlSurface) -> bool {
+        self.wl_surface().map(|s| &s == other).unwrap_or(false)
+    }
+}
+
+impl PartialEq<X11Surface> for CosmicSurface {
+    fn eq(&self, other: &X11Surface) -> bool {
+        if self.0.is_x11() {
+            self.x11_surface().map(|s| s == other).unwrap_or(false)
```
ids1024, line 85: Calling `is_x11` and `x11_surface` seems redundant. `x11_surface()` should return `Some` if and only if `is_x11()`.

This could also use `map_or`.

src/input/mod.rs
```diff
@@ -2241,7 +2251,9 @@ impl State {
                 }
             }
             if let Some(or) = shell.override_redirect_windows.iter().find(|or| {
-                or.is_in_input_region(&(global_pos.as_logical() - or.geometry().loc.to_f64()))
+                or.is_in_input_region(
+                    &(global_pos.as_logical() - X11Surface::geometry(*or).loc.to_f64()),
```
ids1024, line 2255: Another issue caused by the wrong `geometry` method being called? And I guess the fact `or` is a double reference makes it not use the implementation one would expect. Smithay probably needs more coordinate space types, and maybe fewer overloaded names.

src/input/mod.rs
```diff
@@ -942,7 +941,18 @@ impl State {
...
-                                    if let Some((target, _)) =
+                                    if let Some(or) =
+                                        self.common.shell.override_redirect_windows.iter().find(
+                                            |or| {
+                                                or.is_in_input_region(
+                                                    &(pos.as_logical()
+                                                        - or.geometry().loc.to_f64()),
+                                                )
+                                            },
+                                        )
```
ids1024, line 952: This seems to be the same as the code that was changed later in this file (in e8a9a373e82a4d35e8e611c79b741bae7c4ac1be). Presumably the same fix is needed here?

Maybe the redundancy could also be improved somehow.

src/input/mod.rs
Drakulix: Yep. Only stepping through it I noticed the problem..

src/xwayland.rs
```diff
@@ -136,6 +136,104 @@ impl State {
...
+                                }),
+                        )
+                })
+                .collect::<Vec<_>>();
+
+            // we don't include the popup elements, which contain the OR windows, because we are not supposed to restack them.
+            // Which is also why we match upwards, to not disturb elements at the top.
+            //
+            // But this also means we need to match across all outputs and workspaces at once, to be sure nothing that shouldn't be on top of us is.
+            if let Err(err) = xwm.update_stacking_order_upwards(order.iter().rev()) {
```
ids1024, line 230: Is `update_stacking_order_upwards()` with `rev()` different from `update_stacking_order_downwards()`?

It could be good to pass the iterator here without `collect, but I guess that wouldn't be possible with `update_stacking_order_*` taking `impl Iterator<Item = &'a W>`. (Maybe it could use a `Borrow` generic.)

ids1024: (I guess it is different, based on the "Which is also why we match upwards, to not disturb elements at the top" comment.)

Drakulix: Yes, sadly it is different. Essentially we don't want to touch the top-most elements of the stacking order, which `_downwards` could potentially do.

src/input/mod.rs
Drakulix: Actually it's not a `&&X11Surface` here.

src/input/mod.rs
Drakulix: Actually it's not a &&X11Surface here.

Drakulix: fixed. good catch

ryanabx:
Seems like this PR causes a regression in cosmic-greeter. When I build this PR, replace my native cosmic-comp with a usroverlay that has the new bin, and log out, cosmic-greeter only has a blank screen with a mouse. In this state, I can open up only certain apps (I was able to open steam, but not a terminal) and I can interact with cosmic-launcher

git-f0x:
I've had similar issues previously (had to reinstall the OS to fix somewhat), and occasionally now, so the issue might be more on the cosmic-greeter side than here.

ryanabx:
> I've had similar issues previously (had to reinstall the OS to fix somewhat), and occasionally now, so the issue might be more on the cosmic-greeter side than here (maybe). Though it is weird that you can open Steam when logged out.

Yeah it is weird. It even had my profile signed in and everything

Drakulix:
> Seems like this PR causes a regression in cosmic-greeter. When I build this PR, replace my native cosmic-comp with a usroverlay that has the new bin, and log out, cosmic-greeter only has a blank screen with a mouse. In this state, I can open up only certain apps (I was able to open steam, but not a terminal) and I can interact with cosmic-launcher

Are you sure you were running cosmic-greeter as login manager here and not just a bare cosmic-comp on your user-session? Because cosmic-greeter is a completely separate user account that should not have access to your .steam folder.

Also cosmic-greeter as a login manager always had this problem of not starting up correctly all the time. There is some race condition here, so this PR likely didn't regress anything in that respect.

ryanabx:
Yeah, there's separate regressions to cosmic-greeter that make me doubt the origin of the issue. If anything, I doubt what's being done here is wrong

ryanabx:
@Drakulix I spotted a regression:

In visual studio code, I can't click on items in a dropdown, it will instead click whatever is under the pop-up

https://github.com/pop-os/cosmic-comp/assets/56272643/4006bcd6-35b3-452f-81a5-632d36ceb782

Drakulix:
> In visual studio code, I can't click on items in a dropdown, it will instead click whatever is under the pop-up

> Another regression [...]

Should both be fixed. Thanks for spotting!

ryanabx (approved):
Can confirm this fixes the regressions I mentioned above, as well as the issues mentioned in the original post, AND this issue I had https://github.com/pop-os/cosmic-comp/issues/359 . Great work, thanks so much!

x64code:
This seems to fix issues ive been having with window focus on spotify, mattermost, rubymine and warp

Drakulix:
Looks like this is working pretty well after the last round of fixes. Merging.

gabriele2000:
I don't know if I'm tripping or something, but opening `winecfg` causes a weird behaviour: the window appears for a split second and it just closes itself, weird.

There's this error though:
`0088:err:hid:udev_bus_init UDEV monitor creation failed`

git-f0x:
I'm running it from Bottles, but winecfg opens fine for me, just that attempting to resize the window makes it shrink to almost just the window decoration (haven't opened it before now, so unsure if it's caused by this PR).

gabriele2000:
> just that attempting to resize the window makes it shrink to almost just the window decoration

Everything's normal, happened even before that PR.

gabriele2000:
Ok everyone, I'm sorry.
It was a temporary error, restarting the computer fixed everything.

PR #380, opened 2024-03-25
input: Rework PointerFocus to operate directly on WlSurface

Drakulix:
Continuation of https://github.com/pop-os/cosmic-comp/pull/322 to eventually bump smithay again.

This takes about half of @ids1024 original approach, by leaving the `KeyboardFocusTarget` as is and adjusting the `PointerFocusTarget` to go all the way and split up `CosmicMapped` focus into iced-UI and surfaces.

Reasoning:
- The smithay changes actually don't require us to adjust the keyboard focus and I don't think consolidating into a single `WlSurface` variant is a good idea, as we are loosing context. The old PR spent quite a bit of code to reconstruct toplevel-surfaces. Let's simply not do that.
- The new pointer focus logic is meant to fix inconsistencies when wrapping multiple elements, regarding grabs, which suddently can't differentiate anymore between the individual elements (which in particular problematic for touch input). We have our fair share of weird workarounds and hacks in `CosmicWindow` and especially in `CosmicStack` to make the window unaware of the decorations and provide proper wl_poiinter events. All of this is brittle and in some parts even straight up buggy (for admittedly rare input combinations). Lets split the focus in the actual window surface(s) and the decorations to fix this.

Remaining TODOs:
- [x] Rebase
- [x] Currently this comments out the screencopy stuff inside the window-types, as there is no central location to get pointer events for a single window anymore. Lets fix this (somehow..)
- [x] Actually upgrade smithay
- [x] Incorporate https://github.com/pop-os/cosmic-comp/pull/341

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -548,6 +548,20 @@ version = "0.12.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fba7adb4dd5aa98e5553510223000e7148f621165ec5f9acd7113f6ca4995298"
 dependencies = [
+ "bitflags 2.4.2",
+ "log",
+ "polling 3.3.2",
+ "rustix 0.38.30",
+ "slab",
+ "thiserror",
+]
+
+[[package]]
+name = "calloop"
+version = "0.13.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b99da2f8558ca23c71f4fd15dc57c906239752dd27ff3c00a1d56b685b7cbfec"
+dependencies = [
  "async-task",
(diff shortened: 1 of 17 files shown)


ids1024:
SSDs still don't handle touch events, but it looks like that's expected. That can be added later.

ids1024 (approved):
The changes all seem reasonable, and everything that I've tested is working fine.

Drakulix:
> SSDs still don't handle touch events, but it looks like that's expected. That can be added later.

Yeah, that requires implementing the TouchTarget for them and properly supporting TouchGrabs. Left that out for now, as I don't have a touchscreen to test that anyway.

PR #386, opened 2024-03-26
WIP Touch support for SSDs (And some refactoring of Window pointer input handling)

ids1024:
This should be directing touch events to the Iced program, but for some reason it doesn't seem to be working currently.

Providing good support for resizing (and moving) windows with the touch screen may be a little more complicated.


ids1024:
The `PointerTarget` implementations for `CosmicWindow`/`ComsicStack` seems to be handling window geometry inconsistently. Not sure what the right way is.

Testing sctk's `simple_window` modified with an unusual geometry, I'm not sure Smithay is handling geometry correctly in general.

ids1024:
Or maybe not a Smithay issue per-se, just that Anvil's SSDs have similar problems. We do need to do something about how overloaded the term "geometry" is.

This PR should improve handling of window geometry a bit. Though there's probably more to improve. Likewise there's plenty more to go for optimal touch handling, but this is a start.

Drakulix (approved):
Thanks. Awesome cleanup.

PR #396, opened 2024-04-01
draft: gesture configuration backend

ryanabx:
This is an implementation of a backend gesture configuration based on a gist I made a few weeks ago: https://gist.github.com/ryanabx/e18b70d58f651c1a6946a5bca7a11165

This method of configuration would allow the frontend settings to do whatever they'd like with the gesture configuration without having to do something complex (unless they wanted to!)

Basically, gesture configurations are split per-finger and can be one of two types:

* Workspace dependent (depends on the direction of workspaces (horizontal/vertical) (for example, workspace switching gestures would need to be workspace dependent)
* Directional (doesn't depend on the direction of workspaces) (for example, window focus switching gestures need to be workspace independent)


ryanabx:
Rebased to include the RWLock changes. @Drakulix can you look at this at some point? I think this is a good direction to go for gesture configurations but I need to know if you and others agree

cosmic-comp-config/src/input.rs
```diff
@@ -46,6 +46,43 @@ pub struct ScrollConfig {
...
+    WorkspaceForward,
+    WorkspaceBackward,
+    WorkspaceOverviewEnable,
+    WorkspaceOverviewDisable,
+    WindowUp,
+    WindowDown,
+    WindowLeft,
+    WindowRight,
+    Custom(String),
+}
```
Drakulix, line 85: Do we really need another enum for this? We already have the `Action`-enum that is used for key-bindings. Can't we just use that and special case some actions like `NextWorkspace`/`PreviousWorkspace` when using gestures? (To animate those properly.)

That would make configuration much more flexible and allow binding more things to gestures.

ryanabx: Sure, I'll look into doing that

ryanabx: After looking into it, I'll hold onto this PR until https://github.com/pop-os/cosmic-comp/pull/400 is finished, because it moves the Actions enum into cosmic-comp-config, where I need it

ryanabx:
Going to start working on this again. But it looks like actions and keybinds are defined in c-s-d, so I'm going to make the proper gesture type there first!

ryanabx:
I think this should be ready now, pending https://github.com/pop-os/cosmic-settings-daemon/pull/63

cosmic-comp-config/src/input.rs
ryanabx: Should be fixed. Needs re-review.

ryanabx:
CI failed because I was using Cargo 1.82.0 to generate the lock file. Thanks WSL/Ubuntu

PR #400, opened 2024-04-03
feat: runtime configurable keybindings

mmstick:
Work in progress. Adds key bindings to `cosmic-comp-config` to enable configuring it at runtime in the settings application.


Drakulix:
Looks good on first glance, however I was hoping we could eventually move this out of cosmic-comp, see https://github.com/pop-os/xdg-desktop-portal-cosmic/issues/4#issuecomment-1908272094 .

Essentially given global-shortcuts will need to be managed somehow (including permissions, remember configuration, etc) by our portal, I would prefer another process to be responsible for this and cosmic-comp just forwarding registered events. (And also not call things like `amixer`, but let the audio applet register it's key-bind or cosmic-workspaces/applibrary/etc.)

But I suppose we can move that code later.

However if the only thing the config still contains in the end is `data_control_enabled` we could drop the old config. (Though I don't want to make that variable hot-reloadable and easily user-writable as that is a security hole. So we probably would want to make that an environment variable.)

Cargo.toml
```diff
@@ -10,6 +10,33 @@ members = [
...
+  "desktop",
+  "use_system_lib",
+  "renderer_glow",
+  "renderer_multi",
+  "wayland_frontend",
+  "xwayland",
+]
+git = "https://github.com/smithay/smithay.git"
+rev = "74ef59a3f"
+version = "0.3"
```
Drakulix, line 38: See https://github.com/pop-os/cosmic-settings-daemon/pull/29#pullrequestreview-2131064839. I would be very cautious to make this a workspace dependency and would very much prefer to avoid that.

Makefile
```diff
@@ -23,6 +22,8 @@ endif
 
 TARGET_BIN="$(DESTDIR)$(bindir)/$(BINARY)"
 
+KEYBINDINGS_CONF="$(DESTDIR)$(sharedir)/cosmic/com.system76.CosmicSettings.Shortcuts/v1/defaults"
```
Drakulix, line 25: I wonder if we should just codify a default set of shortcuts. (Especially since some of them depend on the workspace layout choosen, no?)

Distros could still create this file to override them.

src/config/mod.rs
```diff
@@ -179,79 +152,56 @@ impl Config {
...
-                    Err(err) => {
-                        error!("Malformed config file (skipping): {}", err);
-                        continue;
+        // Source key bindings from com.system76.CosmicSettings.Shortcuts
+        let settings_context = shortcuts::context().unwrap();
+        let system_actions = shortcuts::system_actions(&config);
+        let mut shortcuts = shortcuts::shortcuts(&settings_context);
+
+        // Add any missing default shortcuts recommended by the compositor.
+        key_bindings::add_default_bindings(&mut shortcuts, workspace.workspace_layout);
```
Drakulix, line 170: Could this override shortcuts explicitly set by the user? If so, should we not just apply these to some kind of default config, before merging the configuration? Though a merge might also be complicated, if the user wants to unbind default shortcuts. Perhaps the direction-specific shortcuts need some kind of `Auto`-variant for the default behavior?

Drakulix (commented):
Really liking this except for the one issue outlined below.

Especially how the separation between shortcuts and system-actions will hopefully make moving that responsibility into another component much easier!

src/config/mod.rs
mmstick: This function will avoid inserting bindings that have already been set, so it won't override settings from the user. I'm open to ideas for moving this into the defaults config.

Makefile
mmstick: We could do that in addition to the config, but cosmic-settings needs a way to get the on-disk config for displaying defaults.

Drakulix: My gut feeling was having a `Default`-impl in `cosmic-settings-config`, is there a reason why this wouldn't work out? What happens currently if this file gets missing somehow?

mmstick: If the file is not defined, only the default bindings set in cosmic-comp will be defined right now.

mmstick: I'll look into implementing a default function

mmstick:
@Drakulix I believe this to be ready now. cosmic-settings is in a good state with the current setup.

Makefile
Drakulix: Has this proven to be too much of a hassle?

Drakulix (approved):
LGTM. Feel free to merge once cosmic-settings is ready as well.

PR #402, opened 2024-04-03
main: Increase soft file limit

Drakulix:
~~Draft as this still needs testing, but~~ Seems to work well, this should increase our client limit and avoid crashing from buggy clients (though ddos-ing the server with large amounts of file descriptors is still possible).

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -74,6 +74,7 @@ use std::{
     any::Any,
     cell::RefCell,
     collections::HashMap,
+    os::unix::process::CommandExt,
     thread,
     time::{Duration, Instant},
 };
@@ -2285,6 +2286,7 @@ impl State {
                     .env("XDG_ACTIVATION_TOKEN", &*token)
                     .env("DESKTOP_STARTUP_ID", &*token)
                     .env_remove("COSMIC_SESSION_SOCK");
+                unsafe { cmd.pre_exec(|| Ok(crate::utils::rlimit::restore_nofile_limit())) };
 
                 std::thread::spawn(move || match cmd.spawn() {
                     Ok(mut child) => {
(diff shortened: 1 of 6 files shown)


src/input/mod.rs
```diff
@@ -2285,6 +2286,7 @@ impl State {
                     .env("XDG_ACTIVATION_TOKEN", &*token)
                     .env("DESKTOP_STARTUP_ID", &*token)
                     .env_remove("COSMIC_SESSION_SOCK");
+                unsafe { cmd.pre_exec(|| Ok(crate::utils::rlimit::restore_nofile_limit())) };
```
ids1024, line 2289: https://doc.rust-lang.org/std/os/unix/process/trait.CommandExt.html#notes-and-safety

I'm not sure if something like `tracing::warn!` is technically "async-signal-safe" and sound to use after forking... though it's probably fine...

ids1024:
This seems reasonable in general, given sway, kwin, and gnome-shell seem to all modify this rlimit.

src/input/mod.rs
Drakulix: oh, good call. I already changed the initial version of this, which used OnceCell to use Atomics for the globals instead to avoid any locks. Not sure if tracing introduces locks, so maybe we should just use a plain old println! here?

ids1024: The docs for `pre_exec` specifically mention "Panicking in the closure is safe only if all the format arguments for the panic message can be safely formatted". So an `eprintln!` should be safe.

Not sure if `tracing` is using locks, but it's generally doing non-trivial things since we have tracing configured to direct errors to journald.

Drakulix: Exactly that was my thinking. We also can never be sure a dependency might not add locks in future semver-compatible locks, so only relying on std seems a good fit here.

Drakulix: fixed

PR #404, opened 2024-04-05
shell: Touch support for move and resize grabs

ids1024:
Moving Gnome Terminal with touch now works, except the window is shown frozen in place even as the screen indicates tiling or a stack being created, which happens on release?

Also need to add touch to other grabs, and have SSDs (and libcosmic) start move/etc. based on touch events.


ids1024:
Move grabs are working pretty nicely with touch now. And this has resize grabs.

Though I see I've broken the rendering for moving on other monitors. And resizing with `ResizeSurfaceGrab`  breaks moving between monitors.

ids1024:
https://github.com/pop-os/cosmic-comp/pull/415 fixes one of those issues. Added a commit fixing the issue moving between monitors.

There does seem to be another issue when the pointer isn't on the same output as touch events.

src/shell/focus/target.rs
```diff
@@ -497,8 +497,8 @@ impl TouchTarget<State> for PointerFocusTarget {
                 TouchTarget::down(window, seat, data, event, seq)
             }
             PointerFocusTarget::StackUI(stack) => TouchTarget::down(stack, seat, data, event, seq),
-            // TODO: implement TouchTarget for ResizeFork/Grabs
-            PointerFocusTarget::ResizeFork(_fork) => {}
+            // TODO: implement TouchTarget for Grabs
```
Drakulix, line 500: Is this comment still relevant with the latest changes?

Drakulix (approved):
LGTM!

src/shell/focus/target.rs
ids1024: Now that all the grabs should have touch support, I guess that comment isn't really applicable anymore.

Though handling touch events on SSDs to start drags is still needed.

Drakulix: > Though handling touch events on SSDs to start drags is still needed.

Is that something you want to address before I go ahead and merge this?

ids1024: It's probably not that hard to add, but it can be done later.

ids1024: My cosmic-comp changes for https://github.com/Smithay/smithay/pull/1386 are also based on this PR.

PR #407, opened 2024-04-05
Use loginctl lock-session instead of running cosmic-greeter

jackpot51:
Requires https://github.com/pop-os/cosmic-greeter/pull/27


Drakulix:
Note: The power applet also needs to be changed regarding calling the lock-screen.

jackpot51:
I am pretty sure the power applet already tries to use logind and this is why it does not work: https://github.com/pop-os/cosmic-applets/blob/32b7012b79a4c0f7857fba93f3245c8a59b244f4/cosmic-applet-power/src/lib.rs#L397

Drakulix (approved):

Issue #211, opened 2023-11-01
Performance with Intel surfaces rendered with Intel GPU to Nvidia output


ids1024:
It seems good to document what I've seen with performance somewhere.

I've been seeing about 20fps with `es2gears` on a 1440p monitor. On a 1650 mobile with the 545.23.06 driver.

Looking at Tracy profiling, using https://github.com/Smithay/smithay/pull/1134 with some changes:

![2023-10-30T17:08:39,229664208-07:00](https://github.com/pop-os/cosmic-comp/assets/2263150/2b899b0b-ac26-4247-97a8-2c13ce7ed046)

Context 0 in this case is the Nvidia GPU. The portion of `render` before `clear` seems to be the wait at https://github.com/Smithay/smithay/blob/4f9480e5e02d05379e14ec49e1135cc8a57275d1/src/backend/renderer/multigpu/mod.rs#L1144. Without that performance is improved but still less that 60fps (and there's artifacting, of course).

Both GPUs spend about 10ms here drawing in `render_texture_from_to`.

With Intel rendering to an Intel target (but just at 1080p), it only takes about 967μs. Forcing a linear modifier raises that to around 2-5ms, seemingly with more variation frame to frame. I guess that's just (multi-level) caching? So rendering to a linear modifier is part of the performance difference here, though presumably that should be fine, and normal for reverse PRIME like this...

It is strange though to see `draw_solid` in `GlesFrame::clear` taking 3.85ms. That should be fast and just involve the Nvidia framebuffer, which should be efficient to render to...

ids1024:
Also seeing bad performance with Nvidia surfaces on an Nvidia output, strangely. Even when modifying the compositor to only use the Nvida GPU. It blocks on `EGLFence::wait`, but removing the `wait` call doesn't improve fps significantly, only changing where it's blocked.

Performance seem to vary in OpenMW depending on what the camera is pointed at, but not just based on how much of the screen in changing. So maybe render calls are slow on the client for some reason? (So the impact depends on what it's rendering.)

I don't think another GPU or the CPU should be accessing the buffer in this case. Using 8 bit color instead of 10 bit doesn't help. It is using an Nvidia modifier, though I don't know if it's the most optimal Nvidia specific modifier. But presumably Nvidia's egl-wayland would choose a reasonable one out of those offered...

flukejones:
[...]

PR #411, opened 2024-04-09
Add `linux-drm-syncobj-v1` protocol

ids1024:
Based on https://github.com/Smithay/smithay/pull/1356.

I think the blocker logic should be correct for handling acquire points (if I properly understand the transaction system in Smithay). Though I don't see rendering issues with Mesa git when the blocker is removed... maybe it needs to be tested with something heavier than vkcube. (Or is there something still forcing implicit sync?).

The logic I added in Smithay for signaling releases may be a little less correct. Though maybe not more incorrect that how buffer releases are currently handled? (If I understand DRM correctly, with direct scanout we should make sure not to release until we've committed a new buffer and are sure the display controller won't want to read the buffer.)

We'll be able to test more when the next Nvidia driver is released. This at least gives us a way to test the explicit sync support they're adding.

Presumably we should test if `drmSyncobjEventfd` is supported... maybe just creating a syncobj and calling that to see if it works? I'm also still a little unsure how this ends up working with multiple GPUs... particularly if one is Nvidia.

### DIFF
--- a/src/backend/kms/mod.rs
+++ b/src/backend/kms/mod.rs
@@ -35,7 +35,11 @@ use smithay::{
         wayland_server::{Client, DisplayHandle},
     },
     utils::{Clock, DevPath, Monotonic, Size},
-    wayland::{dmabuf::DmabufGlobal, relative_pointer::RelativePointerManagerState},
+    wayland::{
+        dmabuf::DmabufGlobal,
+        drm_syncobj::{supports_syncobj_eventfd, DrmSyncobjState},
+        relative_pointer::RelativePointerManagerState,
+    },
 };
 use surface::GbmDrmOutput;
 use tracing::{error, info, trace, warn};
@@ -70,6 +74,8 @@ pub struct KmsState {
 
     session: LibSeatSession,
     libinput: Libinput,
+
+    pub syncobj_state: Option<DrmSyncobjState>,
 }
 
 pub fn init_backend(
@@ -136,6 +142,8 @@ pub fn init_backend(
 
         session,
         libinput: libinput_context,
+
+        syncobj_state: None,
     });
 
     // start x11
@@ -145,6 +153,23 @@ pub fn init_backend(
     for (dev, path) in udev_dispatcher.as_source_ref().device_list() {
         if let Err(err) = state.device_added(dev, path.into(), dh) {
             warn!("Failed to add device {}: {:?}", path.display(), err);
+        }
+    }
+
+    let kms = match &mut state.backend {
+        BackendData::Kms(kms) => kms,
+        _ => unreachable!(),
+    };
+    if let Some(primary_node) = kms
+        .primary_node
+        .and_then(|node| node.node_with_type(NodeType::Primary).and_then(|x| x.ok()))
+    {
+        if let Some(device) = kms.drm_devices.get(&primary_node) {
+            let import_device = device.drm.device().device_fd().clone();
+            if supports_syncobj_eventfd(&import_device) {
+                let syncobj_state = DrmSyncobjState::new::<State>(&dh, import_device);
+                kms.syncobj_state = Some(syncobj_state);
+            }
         }
     }
 
--- a/src/wayland/handlers/compositor.rs
+++ b/src/wayland/handlers/compositor.rs
@@ -14,6 +14,7 @@ use smithay::{
             CompositorHandler, CompositorState, SurfaceAttributes,
         },
         dmabuf::get_dmabuf,
+        drm_syncobj::DrmSyncobjCachedState,
         seat::WaylandFocus,
         shell::{
             wlr_layer::LayerSurfaceAttributes,
@@ -100,7 +101,14 @@ impl CompositorHandler for State {
 
     fn new_surface(&mut self, surface: &WlSurface) {
         add_pre_commit_hook::<Self, _>(surface, move |state, _dh, surface| {
+            let mut acquire_point = None;
             let maybe_dmabuf = with_states(surface, |surface_data| {
+                acquire_point = surface_data
+                    .cached_state
+                    .get::<DrmSyncobjCachedState>()
+                    .pending()
+                    .acquire_point
+                    .clone();
                 surface_data
                     .cached_state
                     .get::<SurfaceAttributes>()
@@ -113,6 +121,25 @@ impl CompositorHandler for State {
                     })
             });
             if let Some(dmabuf) = maybe_dmabuf {
+                if let Some(acquire_point) = acquire_point {
+                    if let Ok((blocker, source)) = acquire_point.generate_blocker() {
+                        let client = surface.client().unwrap();
+                        let res = state.common.event_loop_handle.insert_source(
+                            source,
+                            move |_, _, state| {
+                                let dh = state.common.display_handle.clone();
+                                state
+                                    .client_compositor_state(&client)
+                                    .blocker_cleared(state, &dh);
+                                Ok(())
+                            },
+                        );
+                        if res.is_ok() {
+                            add_blocker(surface, blocker);
+                            return;
+                        }
+                    }
+                }
                 if let Ok((blocker, source)) = dmabuf.generate_blocker(Interest::READ) {
                     let client = surface.client().unwrap();
                     let res =
--- a/src/wayland/handlers/drm_syncobj.rs
+++ b/src/wayland/handlers/drm_syncobj.rs
@@ -1,0 +1,19 @@
+// SPDX-License-Identifier: GPL-3.0-only
+
+use crate::state::{BackendData, State};
+use smithay::{
+    delegate_drm_syncobj,
+    wayland::drm_syncobj::{DrmSyncobjHandler, DrmSyncobjState},
+};
+
+impl DrmSyncobjHandler for State {
+    fn drm_syncobj_state(&mut self) -> &mut DrmSyncobjState {
+        let kms = match &mut self.backend {
+            BackendData::Kms(kms) => kms,
+            _ => unreachable!(),
+        };
+        kms.syncobj_state.as_mut().unwrap()
+    }
+}
+
+delegate_drm_syncobj!(State);
--- a/src/wayland/handlers/mod.rs
+++ b/src/wayland/handlers/mod.rs
@@ -10,6 +10,7 @@ pub mod decoration;
 pub mod dmabuf;
 pub mod drm;
 pub mod drm_lease;
+pub mod drm_syncobj;
 pub mod foreign_toplevel_list;
 pub mod fractional_scale;
 pub mod idle_inhibit;


Drakulix:
> I think the blocker logic should be correct for handling acquire points (if I properly understand the transaction system in Smithay). Though I don't see rendering issues with Mesa git when the blocker is removed... maybe it needs to be tested with something heavier than vkcube. (Or is there something still forcing implicit sync?).

For that you probably have to remove the old `dmabuf.generate_blocker` logic, which pulls a fence out of the dmabuf to do essentially the same thing. We basically should check, if the client uses explicit-sync and then use the acquire fence and otherwise fallback to polling the dmabuf directly.


> The logic I added in Smithay for signaling releases may be a little less correct. Though maybe not more incorrect that how buffer releases are currently handled? (If I understand DRM correctly, with direct scanout we should make sure not to release until we've committed a new buffer and are sure the display controller won't want to read the buffer.)

Yeah, I am pretty sure smithay's code isn't correct in what it does today, but given direct-scanout mostly works, that is probably using implicit-sync in the background.

What needs to happen is storing the fence generated by compositing **or** in case of direct scanout, we need to get the `OUT_FENCE_PTR` property of the respective plane. Then we can wait for that fence, once the buffer is replaced and won't be used for new rendering/scanout operations and signal release once that fence is done. (Maybe there is even kernel api to "signal once this other fence is signalled"?)


> Presumably we should test if `drmSyncobjEventfd` is supported... maybe just creating a syncobj and calling that to see if it works? I'm also still a little unsure how this ends up working with multiple GPUs... particularly if one is Nvidia.

Yeah, imo this needs a compile test similar to what we do for gbm in smithay: https://github.com/Smithay/smithay/blob/master/build.rs#L99-L125

So if the local kernel/drm headers of the system support it, we enable the feature and assume the kernel does as well. I don't know how well runtime detection would work, but we have to make sure to not advertise the global, if this function isn't supported.

ids1024:
> For that you probably have to remove the old dmabuf.generate_blocker logic, which pulls a fence out of the dmabuf to do essentially the same thing. We basically should check, if the client uses explicit-sync and then use the acquire fence and otherwise fallback to polling the dmabuf directly.

This should already be doing that. If there's an acquire point, it adds a `DrmSyncPointBlocker` and skips adding the `DmabufBlocker`.

> Yeah, I am pretty sure smithay's code isn't correct in what it does today, but given direct-scanout mostly works, that is probably using implicit-sync in the background.

I'm not sure if implicit sync does does something to help with releases (blocking client writes to the buffer until the display controller is done with the buffer... but yeah, it does seem to mostly work. If implicit sync isn't involved, this won't be more problematic with explicit sync and the same limitation.

>  (Maybe there is even kernel api to "signal once this other fence is signalled"?)

It should be possible to do something with `drmSyncobjTransfer` and such. Probably the other implementations of the protocol do something like that, so we can look at those.

We just have to make sure to only do that once the buffer is no longer used elsewhere in the compositor.

ids1024:
> I am pretty sure, that is what our current release logic does, with the exception of handling direct-scanout. Which would be handled by the fence anywhere in this case, so that should be correct. We just need to adjust the DrmCompositor to extract the out-fence and expose that in the RenderResult.

I mean if we want to have `OUT_FENCE_PTR` directly signal the release point:
* We can't do that if we are using the same buffer for direct scanout or rendering elsewhere (on a different monitor; screencopy)
* We can't do that if we might use the buffer again. Which could even be the case if another buffer has been committed, since the new buffer may still be blocked when we want to render the next frame.

So I'm not sure when we could actually do that? Maybe with `commit-queue-v1`, where we might know the next buffer is ready, but aren't using it until the next frame.

So at least for now I think we need to stick to signaling the release point from CPU? But should still track when `OUT_FENCE_PTR` has signaled scanout is done with the buffer.

Drakulix:
> I mean if we want to have `OUT_FENCE_PTR` directly signal the release point:
> 
>     * We can't do that if we are using the same buffer for direct scanout or rendering elsewhere (on a different monitor; screencopy)

Right, so this rather needs to be a list of fences to wait for. Meaning we probably have to wait and signal ourselves instead of relying on `drmSyncobjTransfer`.

>     * We can't do that if we might use the buffer again. Which could even be the case if another buffer has been committed, since the new buffer may still be blocked when we want to render the next frame.

But merge of the state should only happen, once all blockers are resolved. And only then we release, so I believe that issue is already handled correctly. Nothing would be able to use that buffer for rendering any more at that point.

But we still need to track the buffer to be able to signal later, so we might as well unify the approach and handle the `release`-event the same. I feel like this could benefit from some infrastructure and refactoring in smithay.

> So I'm not sure when we could actually do that? Maybe with `commit-queue-v1`, where we might know the next buffer is ready, but aren't using it until the next frame.

I think we can implement both fifo and commit-queue with blockers as well.

> So at least for now I think we need to stick to signaling the release point from CPU? But should still track when `OUT_FENCE_PTR` has signaled scanout is done with the buffer.

Yeah, I am coming to the same conclusion, but that isn't too bad, as that is just another fd in the loop and a very small action.

src/backend/kms/mod.rs
```diff
@@ -450,6 +451,9 @@ pub fn init_backend(
     // Create relative pointer global
     RelativePointerManagerState::new::<State>(&dh);
 
+    // TODO check if main device supports syncobj eventfd?
```
Drakulix, line 454: I guess worst case we can still fail the `import_timeline`-request, right?
Is Xwayland/mesa able to handle this / fallback to implicit sync?

ids1024: I don't think there's any way to fallback once we expose the global. The protocol doesn't make `import_timeline` failable, except as a protocol error.

It looks like Mutter uses `drmSyncobjEventfd (drm_fd, 0, 0, -1, 0) != -1 || errno != ENOENT)` to check if the call is supported. So we probably want to do the same, and only create the global if it is supported.

ryzendew:
Can we get this rebased on the newest master please

gabriele2000:
Is this going to be implemented soon?
It's the only thing that forces me to use vsync in every game, something that I'd rather turn off since Wayland doesn't need it.

ids1024:
> Is this going to be implemented soon?

I've just marked https://github.com/Smithay/smithay/pull/1356 as ready for review, so hopefully if no issues come up this should be merged soon.

I'm not sure if explicit sync will help with whatever issues you're seeing, but if you'd like to test this PR and report how it impacts behavior (for a certain game / graphics card / driver), that could be helpful.

gabriele2000:
> I'm not sure if explicit sync will help with whatever issues you're seeing, but if you'd like to test this PR and report how it impacts behavior (for a certain game / graphics card / driver), that could be helpful.

I think it will.
The issue that I'm seeing is that with vsync disabled anything less than 60FPS feels like 15FPS, maybe 20FPS, it's definitely the sync issue that I began to encounter half a year ago.

I will report back.

ids1024:
FPS being lower than expected (often around 30 on a 60 fps monitor) with vsync is something I've noticed with Intel-rendered windows on a 1650 mobile (later NVIDIA cards don't seem to have this issue). In my testing Gnome Wayland seemed similar, so I assume it's something in the driver.

https://github.com/pop-os/cosmic-comp/issues/211 has some previous testing I've done with that.

gabriele2000:
> FPS being lower than expected (often around 30 on a 60 fps monitor) with vsync is something I've noticed with Intel-rendered windows on a 1650 mobile (later NVIDIA cards don't seem to have this issue). In my testing Gnome Wayland seemed similar, so I assume it's something in the driver.
> 
> #211 has some previous testing I've done with that.

In my case the inverse is true.
Internal framerate is good with vsync disabled and enabled, but with vsync disabled what I see isn't what the computer sees.

Personally I see extreme lag, the game renders fine though because while watching a recording it's all good.

Tipcat-98:
Tried this briefly with NVIDIA 560.35.03.
Firefox would frequently crash with:
"MozCrashReason":"Error flushing display: Broken pipe"

Other things seemed to work fine though.

Using Pop!_os 22.04 with popdev:master branch.

gabriele2000:
> Tried this briefly with NVIDIA 560.35.03. Firefox would frequently crash with: "MozCrashReason":"Error flushing display: Broken pipe"
> 
> Other things seemed to work fine though.
> 
> Using Pop!_os 22.04 with popdev:master branch.

Can you read my comment and please tell me if you don't have the issue?
Are you on a hybrid setup?

Tipcat-98:
I'm using a desktop, with my intel integrated gpu disabled in BIOS.
I couldn't really see any frame rate oddities with or without v-sync.

ids1024:
Hm, are there still issues with Firefox? nvidia/egl-wayland 1.1.15 was supposed to fix some issues like that. Not sure if Firefox also needed fixes.

Tipcat-98:
Appears so.
Can confirm that I'm on egl-wayland 1.1.16

I could send the firefox crash log if you think it could help.

gabriele2000:
> I'm using a desktop, with my intel integrated gpu disabled in BIOS. I couldn't really see any frame rate oddities with or without v-sync.

So my theory could still stand: it must be a problem with hybrid setups

ids1024:
Yeah, I think the issue you're seeing is related in some way to copies between GPUs. Not sure why later Nvidia cards seem to have much less trouble with that.

If you start a program with an env var like `WAYLAND_DISPLAY=wayland-1-renderD128` or `WAYLAND_DISPLAY=wayland-1-renderD129` (depending on which is the Nvidia GPU) you can make sure the program is started on that GPU, and Cosmic knows which GPU it's running on. If you set it to render with the Nvidia GPU that way, performance should be good on Nvidia outputs.

If it's easy for you to test Gnome Wayland on the same setup, it would be good to know if it has similar issues. (If Gnome is doing better on any systems, we should try to figure out what they're doing differently.)

gabriele2000:
> Yeah, I think the issue you're seeing is related in some way to copies between GPUs. Not sure why later Nvidia cards seem to have much less trouble with that.

Figures

> If you start a program with an env var like `WAYLAND_DISPLAY=wayland-1-renderD128` or `WAYLAND_DISPLAY=wayland-1-renderD129` (depending on which is the Nvidia GPU) you can make sure the program is started on that GPU, and Cosmic knows which GPU it's running on. If you set it to render with the Nvidia GPU that way, performance should be good on Nvidia outputs.

Both commands results in the issue still existing, so far I've tried to launch games in steam with that variable.
I'll try to launch steam directly with those commands, then launch a random game

ids1024:
Oh yeah, that command won't change the behavior of XWayland clients, which would probably include most Steam games.

We may need to make some changes to make sure unnecessary buffer copies are avoided for XWayland clients with multiple GPUs. It's all a bit awkward currently since there isn't a reliable way for the compositor to actually know which GPU the buffers it gets from the client are allocated on. (https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/268 aims to address that; though I guess X would also need a protocol change.)

gabriele2000:
> Oh yeah, that command won't change the behavior of XWayland clients, which would probably include most Steam games.

Wasn't this protocol required to basically fix gaming on nvidia?
If I understood what you said correctly, wine/proton needs wayland support or we need Yet Another Protocol to fix XWayland, right?

Why is this issue only present on PRIME systems?

Drakulix:
> Wasn't this protocol required to basically fix gaming on nvidia even from hybrid setups?

Explicit sync was required to fix nvidia mostly on desktops, not really on hybrid setups.

> If I understood what you said correctly, wine/proton needs wayland support or we need Yet Another Protocol to fix XWayland, right?

No, we don't need a new protocol to fix Xwayland, while the protocol @ids1024 mentioned would certainly be helpful, it doesn't in it of itself fix the core issue.

The issue is, that copying from one GPU to another takes time with 1000-series cards and lower. There is nothing to fix this (apart from maybe inside the nvidia-driver, but we don't know that and it doesn't appear to be particularly badly optimized for this use case).

The only thing we *can* fix, is rendering games/apps on the nvidia-gpu and displaying them on a display directly connected to the nvidia gpu (some of your external displays *might* be connected to it directly). That will require multiple Xwayland instances, but can theoretically work and is on our TODO-list.

But there is nothing fixing the delay, if you want to game on e.g. your internal display, which is likely hard-wired to your integrated gpu (if you don't have a switch in your BIOS). The internal FPS will always be fine in that case, as your nvidia gpu can keep up just fine with the application, but you will likely see stutter or at least not reach those same frame counts because of the copy operation, that needs to happen to actually show it on your display.

gabriele2000:
> Explicit sync was required to fix nvidia mostly on desktops, not really on hybrid setups.

That explains a lot.
Still, why this problem occurs only with VSync off?
Keep in mind that, at first, before explicit sync, nvidia added a hybrid fix for the "jumpback", so that you could actually game on wayland, with vsync off, with no problems.

One day something changed and this "slow" synchronization problem happened.

ids1024:
Hm, I can reproduce the Firefox crash with Nvidia, but it doesn't seem to be a protocol error.

ids1024:
Okay, looks like this one was my fault, not Nvidia's.

https://github.com/Smithay/smithay/pull/1547 appears to fix the Firefox crash.

ids1024:
I think everything should be good now. So this can be merged if no more issues are occurring.

ptr1337:
> I think everything should be good now. So this can be merged if no more issues are occurring.

Ive tested this MR on my 4070 Super with 560 Drivers. I was not able to open discord, nor cosmic settings or equal.
I will check tomorrow to gather some logs.

ids1024:
Hm. `cosmic-settings` should be using Vulkan (via wgpu), and Nvida's Vulkan doesn't even use explicit sync yet in the 560 driver (while their EGL implementation does). Unless it's a multi-gpu system and is running on the integrated GPU.

Hopefully the logs provide more context.

Tipcat-98:
Since the smithay update, cosmic settings seems to be working on nvidia.

ids1024:
Yep, https://github.com/Smithay/smithay/pull/1554 should fix that issue on Nvidia, at least.

Not sure about the issue on the rx7900xtx, which does look like a real synchronization issue. I assume that's using vkd3d on wine on xwayland for rendering? The only RDNA hardware I have is a Steam Deck, and I'm not seeing that issue running a couple games on it under cosmic-comp.

StayBlue:
I'm using this with the latest Nvidia driver ([560.35.03](https://www.nvidia.com/en-us/drivers/details/230918/)), which has worked great for me. If any other NixOS users would like to try this, you may try applying the overlay I made for it, which is available [here](https://codeberg.org/StayBlue/nix-personal/commit/e928331a9de558fb52583320f51a849adf8a0526).

Tipcat-98:
Does this require any testing still or is there anything else an end-user can do to help?

ids1024:
I need to look into this more. But there are a couple reports here of visual issues due to synchronization issues (so a bug in cosmic-comp/smithay, or the driver), which I haven't so far reproduced on hardware I have. It's hard to debug without having a way to reproduce the issue, though I could try a few more things to see if I can.

More testing on different hardware and software and specific details about what does and does not work is always welcome.

ptr1337:
Hey,

would you mind to rebase the patchset on latest master?

ids1024:
So the only known issue with this, other than the Nvidia issue that was fixed a while ago, seems to be https://github.com/pop-os/cosmic-comp/pull/411#issuecomment-2379147723 on the RX7900 XTX. That was a while ago, but does look like a synchronization issue, and I can't recall that we've merged anything that would fix it since (unless it was a bug in Mesa/XWayland/DXVK or VKD3D/etc and fixed there).

@skygrango How much have you tested this on the RX7900 XTX lately? It seems to be working now?

ids1024:
The errors about popup positioning are unrelated to the rendering code, so they won't be dependent on driver or on drm-syncobj. So that's probably unrelated. Though that's interesting, I don't think I've seen issues like that with menus, on native Wayland clients.

https://github.com/pop-os/cosmic-comp/pull/1122 may be relevant, and has some improvements to popup positioning.

Drakulix:
Merging for more widespread testing after 5.1 and before alpha 6. This seems to be good on a bunch of our test machines. Thanks everybody else as well for providing test results.

Issue #408, opened 2024-04-05
Snapping floating windows > Top location intitiates early


WatchMkr:
When there is no panel or dock on the top edge, snapping initiates within an area that users expect to drop a floating window. The sides work well.

Let's reduce the top initiation points (half/maximize) by 30% and see how well it works.

PR #412, opened 2024-04-09
fix: reduce size of top snap range

ryanabx:
Reduces top snap range to 38, from 56, reducing top size to approximately 67.8% of the old size.

cc: @WatchMkr

fixes: https://github.com/pop-os/cosmic-comp/issues/408

I haven't tested this for usability yet, but it should be about 67.8% as usable /s

We could alternatively make the total size 32, like the other sides


Drakulix (approved):

WatchMkr:
@ryanabx Thanks! This feels much better so far.

PR #415, opened 2024-04-11
moving: Show `snapping_indicator` only on current output

ids1024:
Without this check, a floating window that is dragged near an edge between monitors causes snapping indicators to show on both, since the window intersects both.


ryanabx (approved):
LGTM

Drakulix (approved):

Issue #403, opened 2024-04-03
DnD Icons stuck under cursor after DnD is cancelled with the Escape key


wash2:
My version of cosmic-comp is:
```
apt info cosmic-comp
Package: cosmic-comp
Version: 0.1~1712071350~22.04~fa7926f
Priority: optional
Section: x11
Maintainer: Victoria Brekenfeld <victoria@system76.com>
Installed-Size: 21.9 MB
Depends: libc6 (>= 2.35), libgbm1 (>= 21.3.0~rc1), libgcc-s1 (>= 4.2), libinput10 (>= 1.19.1), libseat1 (>= 0.5.0), libudev1 (>= 183), libxkbcommon0 (>= 0.5.0), libegl1, libwayland-server0
Recommends: cosmic-session, libgl1-mesa-dri
Homepage: https://github.com/pop-os/cosmic-comp
Download-Size: 7,027 kB
APT-Sources: http://apt.pop-os.org/staging/master jammy/main amd64 Packages
Description: Wayland compositor of pop-os cosmic shell
```

This can be recreated by dragging a file in nautilus and pressing the Escape key, or in firefox by dragging something and pressing the escape key. I first noticed the issue implementing DnD for cosmic-files though.

ids1024:
I see this is also an issue with cosmic-workspaces.

ids1024:
It looks like in https://github.com/pop-os/cosmic-comp/blob/master/src/wayland/handlers/data_device.rs, `ClientDndGrabHandler::dropped` isn't getting called, so the DnD icon isn't getting unset. Though if `get_dnd_icon()` returns a value, that suggests `is_alive` also is true for the surface.

ids1024:
The escape key is handled in https://github.com/pop-os/cosmic-comp/blob/895ea6aec11485b8f142a41b2c05618c571edd69/src/input/mod.rs#L513-L533. `Action::Escape` calls `unset_grab` on the pointer and keyboard.

I guess calling `unset_grab` with the DnD grab doesn't actually cancel the DnD operation. So the server and client both think it's still going on, but the grab has been removed.

Maybe the grab could have a `Drop` impl to clean up state (like cosmic's `MoveGrab` uses `Drop`). Or if the grab traits had a dedicated destructor method they could have access to the `State` when being removed.

Cancelling a DnD operation on escape doesn't seem to be how Sway at least works. But seems valid. Whether or not we want that, being able to remove the grab without breaking Smithay should probably be possible.

PR #420, opened 2024-04-12
Add `unset` method to grab traits

ids1024:
Requires https://github.com/Smithay/smithay/pull/1386. Based on https://github.com/pop-os/cosmic-comp/pull/404, since that adds/changes grabs.

Fixes https://github.com/pop-os/cosmic-comp/issues/403.


Drakulix (approved):
Needs to be rebased though.

PR #424, opened 2024-04-16
Fix support for cursor planes

ids1024:
I've been wondering for a why cosmic-comp wasn't using cursor planes, despite seemingly having doing everything needed to support them.

It seems there was just a missing implementation of the `kind()` method.


Drakulix (approved):
Nice catch!

PR #435, opened 2024-04-19
Support for idle notify and idle inhibit

snaggen:
Initial support for idle notify and idle inhibit.

### DIFF
--- a/src/wayland/handlers/mod.rs
+++ b/src/wayland/handlers/mod.rs
@@ -9,6 +9,8 @@ pub mod dmabuf;
 pub mod drm;
 pub mod drm_lease;
 pub mod fractional_scale;
+pub mod idle_inhibit;
+pub mod idle_notify;
 pub mod image_source;
 pub mod input_method;
 pub mod keyboard_shortcuts_inhibit;
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -198,6 +198,7 @@ impl State {
                     .for_device(&event.device())
                     .cloned();
                 if let Some(seat) = maybe_seat {
+                    self.common.idle_notifier_state.notify_activity(&seat);
                     let current_output = seat.active_output();
                     let shortcuts_inhibited = self
                         .common
@@ -512,6 +513,7 @@ impl State {
 
                 let mut shell = self.common.shell.write().unwrap();
                 if let Some(seat) = shell.seats.for_device(&event.device()).cloned() {
+                    self.common.idle_notifier_state.notify_activity(&seat);
                     let current_output = seat.active_output();
 
                     let mut position = seat.get_pointer().unwrap().current_location().as_global();
@@ -718,6 +720,7 @@ impl State {
                     .for_device(&event.device())
                     .cloned();
                 if let Some(seat) = maybe_seat {
+                    self.common.idle_notifier_state.notify_activity(&seat);
                     let output = seat.active_output();
                     let geometry = output.geometry();
                     let position = geometry.loc.to_f64()
@@ -785,6 +788,7 @@ impl State {
 
                 let mut shell = self.common.shell.write().unwrap();
                 if let Some(seat) = shell.seats.for_device(&event.device()).cloned() {
+                    self.common.idle_notifier_state.notify_activity(&seat);
                     #[cfg(feature = "debug")]
                     if shell.seats.iter().position(|x| x == &seat).unwrap() == 0
                         && self.common.egui.active
@@ -994,6 +998,7 @@ impl State {
                     .for_device(&event.device())
                     .cloned();
                 if let Some(seat) = maybe_seat {
+                    self.common.idle_notifier_state.notify_activity(&seat);
                     #[cfg(feature = "debug")]
                     if self
                         .common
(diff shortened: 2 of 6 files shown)


src/wayland/handlers/mod.rs
```diff
@@ -9,6 +9,7 @@ pub mod dmabuf;
 pub mod drm;
 pub mod drm_lease;
 pub mod fractional_scale;
+pub mod idle;
```
Drakulix, line 12: Can you split this into two modules for `idle_notify` and `idle_inhibit` please?
I know they are related in how this implementation currently works, but they don't have to be.

Drakulix (commented):
Looks mostly good! Thanks for working on this :)

snaggen:
Inhibit is broken... seems the surface_primary_scanout_output function is not hooked up. And since I shamelessly stole this from niri, I must investigate this.

snaggen:
Ok, By thinking about this over night, I got to the conclusion that niri is doing this wrong. There are many use case where you want to inhibit idle without being in full screen. So, I will modify this to allow any surface that is alive to inhibit idle.

Drakulix:
> Tested with swayidle and idle seems to work as expected. Also, inihibit seems to work. I first thought it didn't but it seems the scanout check in the refresh only works for fullscreen.... so inhibit only works for fullscreen. Not sure if that is desired? I shamelessly stole that part of the code from niri, so I'm not sure what to think about it.

Hmm, it definitely shouldn't only work for fullscreen. The check basically makes sure the surface is actually visible, otherwise any client could just create a wl_surface and never map it any screen, which is undesirable.

So imo the check was fine, but it should only work for windows on the active workspace (or sticky windows). How were you testing?

snaggen:
I was using https://github.com/stwa/wayland-idle-inhibitor since the video player i tested doesn't seem to inhibit. But, I'll add back the check then and see if it was my testing that was flawed.

snaggen:
Ok, seems that it was flawed testing. The current MR works as expected when tested with Clapper, which uses GTK4 builtin Video playout. That seems to have proper support for the wayland inhibit protocol. I notice that firefox is supposed to have inhibit support, but I couldn't get that to work. But since clapper works properly, I assume the firefox issue is a client issue.

snaggen:
About the failure with firefox, maybe @stransky can cast some light on this. Does firefox require some specific conditions besides just playing video, to inhibit idle?

snaggen:
Also, a simple way to test this is to use swayidle like:
```bash
swayidle -w timeout 5 "echo idle" resume "echo resume"
```

stransky:
> About the failure with firefox, maybe @stransky can cast some light on this. Does firefox require some specific conditions besides just playing video, to inhibit idle?

You can run Firefox on terminal with MOZ_LOG="LinuxWakeLock:5" env variable and you'll get info when Firefox sends inhibit/uninhibit. AFAIK the inhibit event is generated by DOM HTML video element and depends on visibility and so on, it may request more source investigation but it's possible to find it.

YaLTeR:
hey, just a few drive by comments

> I got to the conclusion that niri is doing this wrong. There are many use case where you want to inhibit idle without being in full screen

Not really related to this PR, but in niri it works whenever the surface is anywhere on screen, not just fullscreen.

> firefox

When I was implementing idle, I found that Firefox always tries the portal inhibit first, and the portal inhibit literally never reports a failure (there are various bugs reported about this into the portals), so Firefox never got to the Wayland inhibit. I ended up also implementing the FDO D-Bus inhibit endpoint.

snaggen:
Yes, my niri comment was wrong. I had misunderstood what it did. So, I went back to the same logic in this PR

snaggen:
> When I was implementing idle, I found that Firefox always tries the portal inhibit first, and the portal inhibit literally never reports a failure (there are various bugs reported about this into the portals), so Firefox never got to the Wayland inhibit. I ended up also implementing the FDO D-Bus inhibit endpoint.

I'll see of I get some time to test Firefox a bit more. But @stransky , would it be possible to prefer Wayland inhibit, and fallback to dbus? Would that be more reliable?

stransky:
> > When I was implementing idle, I found that Firefox always tries the portal inhibit first, and the portal inhibit literally never reports a failure (there are various bugs reported about this into the portals), so Firefox never got to the Wayland inhibit. I ended up also implementing the FDO D-Bus inhibit endpoint.
> 
> I'll see if I get some time to test Firefox a bit more. But @stransky , would it be possible to prefer Wayland inhibit, and fallback to dbus? Would that be more reliable?

DBus is what most of the apps use and it's preferred. It may be possible to add config to force-select particular inhibit type - please file a bug for it at https://bugzilla.mozilla.org/ and cc me there.

Drakulix:
I think the proper solution for the firefox prooblem is to implement the dbus interface in xdg-desktop-portal-cosmic (utilizing this protocol), I don't think firefox needs to be adjusted here.

@snaggen Anything else you want to add to this PR? Or is this ready for review and potentially merge?

snaggen:
This PR is ready for review.

Drakulix (approved):
Thanks for working on this!

PR #438, opened 2024-04-20
feat: floating window tiling gaps

ryanabx:
> **Note:** the images here are slightly old, the maximize state no longer has gaps, as @git-f0x pointed out that the maximized state in tiling mode doesn't have gaps

screenshot-2024-04-20-04-23-45

screenshot-2024-04-20-03-57-42

https://github.com/pop-os/cosmic-comp/assets/56272643/044df604-215c-4f2e-a222-946ae47ffd23

**Friendly reminder** tiling gaps can be configured in cosmic settings > appearance !

### DIFF
--- a/src/shell/grabs/moving.rs
+++ b/src/shell/grabs/moving.rs
@@ -42,7 +42,7 @@ use smithay::{
         Seat,
     },
     output::Output,
-    utils::{IsAlive, Logical, Point, Rectangle, Scale, Serial, SERIAL_COUNTER},
+    utils::{IsAlive, Logical, Point, Rectangle, Scale, Serial, Size, SERIAL_COUNTER},
 };
 use std::{cell::RefCell, collections::HashSet, sync::atomic::Ordering, time::Instant};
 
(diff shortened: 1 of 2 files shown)


git-f0x:
While I'm not certain, I don't think maximize should have gaps, since it should probably match the behavior from clicking the maximize button (since it doesn't have gaps when maximized in auto tiling).

ryanabx:
> While I'm not certain, I don't think maximize should have gaps, since it should probably match the behavior from clicking the maximize button (it doesn't have gaps when maximized in auto tiling).

Thanks for pointing that out to me! I never realized because I never maximize in tiling mode 😅

wiiznokes:
I know its not the case for tilling, but shouldn't the focus hint size be taken into account to calculate the gaps ?

Something like `gaps = focus_hint + windows_gaps` or

```rust
gaps = if focus_hint > windows_gaps {
    focus_hint
} else {
     windows_gaps
};
```

git-f0x:
That would make sense aesthetically, but it might not be expected behavior to have gaps (e.g. if gaps are set to 0) because of the focus hint.

wiiznokes:
That currently the case in the implementation. But `gaps = windows_gaps + focus_hunt` should be considered. But it might take a lot of space in some case

wiiznokes:
That currently the case in the implementation. But `gaps = windows_gaps + focus_hunt` should be considered. But it might take a lot of space in some case.

Edit: I see what you meant, if there is multiple windows, you will end up have gaps on non focused windows. The problem is that when the windows is on a edge on the screen, his focus hint get cropped

git-f0x:
Well, in that case, setting so that the gap ≥ focus hint would resolve that (e.g. focus hint and gap set to 2). Anything else might lead to unexpected behavior for end users.

PR #445, opened 2024-04-24
Update `smithay`

ids1024:


src/wayland/handlers/input_method.rs
```diff
@@ -32,6 +32,8 @@ impl InputMethodHandler for State {
             .map(|e| e.geometry())
             .unwrap_or_default()
     }
+
+    fn popup_repositioned(&mut self, _: PopupSurface) {}
```
Drakulix, line 36: We probably should handle this similarly to how niri does it, no?

ids1024: Probably. I haven't looked at how niri implements it.

Vixea:
Not to rush this or anything but can the git commit be updated to include the pointer fix while I don't know for sure I have a feeling it might fix a recent regression with pointers in fullscreen applications

Vixea:
Umm now xwayland is completely broken. Probably do to this?
https://github.com/Smithay/smithay/commit/8773a74210214556b028ebce3f9c5e57582851e6
Though not sure if it's a smithay problem or cosmic problem

ids1024:
Hm, at least in some quick testing XWayland works fine for me. (`xterm` and `gimp` start and generally seem to work.)

Vixea:
Have you checked steam it's the only application I really opened because of the test case

Vixea:
Yea it's saying display is not set even on xterm

ids1024:
Seems like a timing thing. When I start with cosmic-session, I can't start xterm from the launcher. But if I open a terminal with

Vixea:
Huh that's funny because I tried the terminal too, but it didn't work

src/wayland/handlers/input_method.rs
Drakulix: https://github.com/YaLTeR/niri/pull/295

Drakulix:
Added some changes to lazy initialize the rest of the session, once we got the DISPLAY from xwayland. Seems to work well in my testing. @Vixea does this fix your issues?

ryanabx:
This will be necessary (and probably the final hurdle) to getting cosmic-comp on Fedora Rawhide :partying_face:

ids1024:
Updated to fix minor merge conflict.

Vixea:
I'm sadly unable to test btw

Drakulix:
Since Xwayland's DISPLAY seems to propagate correctly now and I don't see other regressions, I am going ahead and merge this. Everything else can be handled later.

PR #451, opened 2024-04-29
screencopy: Fix reference cycle, that leaks buffer

ids1024:
Before this, `/proc/$(pidof cosmic-comp)/maps` quickly expands in length when cosmic-workspaces is opened and closed a bunch of time, preventing the GPU memory from being freed. Which on my Intel system can lead to OOM eventually.

There may be other leaks to deal with, but `maps` no longer shows this issue.


Drakulix (approved):
Nice catch. Thanks

PR #459, opened 2024-05-04
Inhibit workspace animations when workspaces view is open

ids1024:
Fixes https://github.com/pop-os/cosmic-workspaces-epoch/issues/27.

We want this to apply to changes to workspace either through keybindings or the cosmic-workspaces UI, so it adding a check here seems reasonable. In principle it could be good to have some kind of privileged protocol for setting things like this.

We may also want a configuration option to disable animations at some point.

### DIFF
--- a/src/shell/mod.rs
+++ b/src/shell/mod.rs
@@ -44,8 +44,8 @@ use crate::{
     utils::prelude::*,
     wayland::{
         handlers::{
-            toplevel_management::minimize_rectangle, xdg_activation::ActivationContext,
-            xdg_shell::popup::get_popup_toplevel,
+            screencopy::WORKSPACE_OVERVIEW_NAMESPACE, toplevel_management::minimize_rectangle,
+            xdg_activation::ActivationContext, xdg_shell::popup::get_popup_toplevel,
         },
         protocols::{
             toplevel_info::{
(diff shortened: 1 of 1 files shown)


ids1024:
Unrelated to this specific change, I'm not sure exactly what the `else` condition in `activate` is meant to help with.

It does cause weird behavior if you call `super+<num>` twice in rapid succession, so it animates but then jumps back to the start of the animation.

ryanabx:
> It does cause weird behavior if you call `super+<num>` twice in rapid succession, so it animates but then jumps back to the start of the animation.

This can also be triggered with the other workspace switching methods:

Super + ctrl + up/down same way in rapid succession

Four finger swipe same direction in rapid succession

src/shell/mod.rs
```diff
@@ -408,13 +408,23 @@ impl WorkspaceSet {
             return Err(InvalidWorkspaceIndex);
         }
 
+        // Animate if workspaces overview isn't open
+        let layer_map = layer_map_for_output(&self.output);
+        let animate = !layer_map
+            .layers()
+            .any(|l| l.namespace() == WORKSPACE_OVERVIEW_NAMESPACE);
```
Drakulix, line 415: How weird is this interaction with the touchpad gestures? I imagine it feels quite bad when the workspace abruptly changes. We probably just want to disable the gestures for now, when the overview is active?

Drakulix:
> It does cause weird behavior if you call super+<num> twice in rapid succession, so it animates but then jumps back to the start of the animation.

Yes that should reset the animation, but rather just change the target workspace, because I didn't want to implement an animation queue just yet. :sweat_smile: 

> In principle it could be good to have some kind of privileged protocol for setting things like this.

Yeah absolutely, I don't like hard-coding all these exceptions for the `app_id`. It might make an incentive for any alternative projects to re-use it.

Lets first figure out, what the overview needs first though and then design a protocol. So short-term these kinda changes are totally fine for me.

But the protocol could also help with excluding the view from rendering and possibly also letting cosmic-comp handle the background rendering.

ids1024:
This was merged in https://github.com/pop-os/cosmic-comp/pull/609.

PR #465, opened 2024-05-09
WIP Ei protocol support

ids1024:
Uses https://github.com/Smithay/smithay/pull/1388 to support EI sender contexts.

This isn't meant to be suitable to merge at present. I just wanted to test this with cosmic-comp, and maybe prototype support in the portal.

We may not want to expose an ei socket in the filesystem, but then the portal will need some private Wayland or DBus protocol to get an ei FD from the server.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3136,9 +3136,9 @@ dependencies = [
 
 [[package]]
 name = "log"
-version = "0.4.28"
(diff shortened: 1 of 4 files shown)


ids1024:
Superseded by https://github.com/pop-os/cosmic-comp/pull/2442.

PR #481, opened 2024-05-15
Fix tablet input

ids1024:
We need to call `add_tablet` for each tablet device, not just if the capability is new. Also, the tool needs to be added with `add_tool`.

`xournalpp` seems to be working now.


Drakulix (approved):

Vixea:
Hmm still no luck getting anything even said xournalpp working on my Wacom tablet

PR #496, opened 2024-05-21
feat: configurable keyboard repeat rate and delay

lazytanuki:
Hi !
This PR updates the `xkb_config` file structure and handling to allow for setting the keyboard repeat rate and delay.
Thanks !

### DIFF
--- a/cosmic-comp-config/src/lib.rs
+++ b/cosmic-comp-config/src/lib.rs
@@ -54,6 +54,10 @@ pub struct XkbConfig {
     pub layout: String,
     pub variant: String,
     pub options: Option<String>,
+    #[serde(default = "default_repeat_delay")]
+    pub repeat_delay: u32,
+    #[serde(default = "default_repeat_rate")]
+    pub repeat_rate: u32,
 }
 
 impl Default for XkbConfig {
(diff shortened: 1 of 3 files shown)


cosmic-comp-config/src/lib.rs
```diff
@@ -54,6 +54,8 @@ pub struct XkbConfig {
     pub layout: String,
     pub variant: String,
     pub options: Option<String>,
+    pub repeat_delay: u32,
+    pub repeat_rate: u32,
```
Drakulix, line 58: This will reset existing config files, that don't contain these values yet, which I would like to avoid. Can you add `#[serde(default = ..)]` tags please?

(Those take a function, so you need to add `default_repeat_delay` and `default_repeat_rate` functions as well. (And best use them in the default implementation as well, so that these values are only declared once. See https://serde.rs/field-attrs.html#default--path)

Drakulix (commented):
Thanks! Looks good, just a small nitpick

cosmic-comp-config/src/lib.rs
lazytanuki: Done!

Issue #468, opened 2024-05-09
[Games] Window in windowed mode gets smaller & smaller when a window is moved


onlyreportingissues:
[Video_2024-05-09_15-33-03.webm](https://github.com/pop-os/cosmic-comp/assets/103491514/f8ac528e-ea5b-449a-bf96-1f80483415a0)

Issue #502, opened 2024-05-23
Reduce SSD height and account for height in X11 dialogs


WatchMkr:
Reduce height to 36 pixels. Design is [here](https://www.figma.com/design/SkAtS5qlVOGQC0n895AHzX/Design-System-(WIP)?node-id=8251-18187&t=vR9HL5oxuu6o5sVr-0) at the bottom.

Account for the SSD height in X11 dialogs
![image (5)](https://github.com/pop-os/cosmic-comp/assets/3430283/3ba9c6b7-9dda-4c13-ba51-7281d0d39c51)

> Looking at xprop and xwininfo, I see the geometry of the window has 48 pixels less height than the minimum/maximum size requested by the client. So yeah, that's SSD_HEIGHT.
> 
> https://github.com/pop-os/cosmic-comp/blob/dfb3bea595425201cc00643e60f2827ffd33ac61/src/shell/element/surface.rs#L417 is supposed to account for that, but I guess it's somehow not working for Xwayland surfaces.

WatchMkr:
SSD height is also mentioned in issue https://github.com/pop-os/cosmic-comp/issues/395 and adds a screenshot of Spotify being cut off. I haven't been able to reproduce normal X11 app windows... just dialogs.

PR #513, opened 2024-05-28
Fixes for handling of SSDs in window geometry

ids1024:
This seems to fix the issue described in https://github.com/pop-os/cosmic-comp/issues/502 with XWayland dialogs not being allocated enough height.

This may also fix https://github.com/pop-os/cosmic-comp/issues/468, or that could be a similar issue. That will need more testing. If comic-comp gets the `geometry` of a window then calls `set_geometry`, if everything isn't perfectly symmetric that may lead to the sizing shrinking or growing like that.

### DIFF
--- a/src/shell/element/surface.rs
+++ b/src/shell/element/surface.rs
@@ -415,9 +415,9 @@ impl CosmicSurface {
         }
         .map(|size| {
             if self.is_decorated(false) {
-                size + (0, SSD_HEIGHT).into()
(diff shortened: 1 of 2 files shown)


ryanabx:
Can you also look into the first part of #502, if in another MR if needed?

jackpot51 (approved):
Since @Drakulix is off, I will approve.

PR #521, opened 2024-05-30
fix: apply activated state when updating maximized layout

wash2:


ryanabx (approved):
Logic is sound to me, and can confirm it fixes the 2 issues I mentioned on mattermost

mmstick (approved):

Issue #494, opened 2024-05-20
[cosmic-term] Copying & pasting text from & to cosmic-term doesn't work, if both tabs are stacked in the same window


onlyreportingissues:
E.g. copying a text from a stacked Firefox tab to the stacked cosmic-term tab & vice versa.

ids1024:
This also sometimes happens copying from cosmic-term to itself. Until it's moved out of the stack. But it works after moving it back.

I'm not sure exactly what the pattern of when it works and doesn't is.

ids1024:
It seems this is reproducible by opening two cosmic-term windows and putting them in a stack.

It is only possible to copy to one of the windows, including from itself. (Unlike Firefox, which handles that copy internally.) It is possible to paste into either window. The behavior is clearer to understand when Firefox isn't involved.

It looks like we're allowing any client in the focused stack to set the clipboard? Since that uses `client_of_object_has_focus` to test which client has keyboard focus, which uses `same_client_as`, which for `CosmicMappedInternal::Stack`, we implement as `s.surfaces().any(|w| w.same_client_as(object_id))`. Presumably we'll want to restrict that to only the active window in the stack. Not sure if anything else would rely on this `all` behavior of `same_client_as`.

This isn't the issue here, but could cause bugs or security problems, so it's something to address.

It seems the problem is simply that `set_data_device_focus` isn't called when the active element of the stack is changed, so the previous active element stays the same. Since `KeyboardFocusTarget` contains the whole stack, and not the single window, and isn't changed here, `SeatHandled::focus_changed` isn't called.

So the issue isn't really that complicated, but can have somewhat weird effects. Somehow we'll need to make sure that changes to the active member of a stack always set the data device focus, if the stack has keyboard focus.

PR #533, opened 2024-05-31
Set clipboard/primary focus in `refresh_focus`

ids1024:
This should fix https://github.com/pop-os/cosmic-comp/issues/494, and make clipboard and primary focus consistently correct.

Changing the active element of a stack needs to change the clipboard focus, but it wasn't being changed since the `KeyboardFocusTarget` was unchanged. The `CosmicStack` methods that change the active stack element also have no obvious way to change the keyboard focus. So we can set this in `refresh_focus`, which should be correct.

If the new focus `WlSurface` is `None`, this clears the focus instead of leaving it as the previous code did. I believe that is desirable.

Requires https://github.com/Smithay/smithay/pull/1442 to avoid repeated `offer`s, instead of only when focus changed.

(Perhaps this could better be solved by having a `WlSurface` variant of `KeyboardFocusTarget`, like pointer focus, or some mechanism for a stack of focus, which could help other things. But it's also unclear exactly how that would work with the code for setting the active stack element, among other questions.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4581,7 +4581,7 @@ checksum = "3c5e1a9a646d36c3599cd173a41282daf47c44583ad367b8e6837255952e5c67"
 [[package]]
 name = "smithay"
 version = "0.3.0"
-source = "git+https://github.com/smithay//smithay?rev=12bcefc#12bcefc2009c4648bb5e9ffde6e03a4cd4c02d90"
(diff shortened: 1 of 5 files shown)


ids1024:
Added a commit changing how `same_client_as` works for stacks, which should be necessary to prevent a background window in a stack from setting the clipboard.

Not sure if anything else uses `same_client_as` that might for some reason need the previous behavior, but I wouldn't expect it.

Drakulix (approved):
LGTM

PR #561, opened 2024-06-19
Update to latest Smithay version

ids1024:
Changing pointer/touch events to use `f64` is annoying, but not too complicated.


Drakulix (approved):
Seems all sensible to me

PR #565, opened 2024-06-20
Maximise window when Up and not tilled 

wiiznokes:
Same of #563, but simpler 

> This PR change the behaviour when moving a window Upward that is not tilled. It will maximize it.


Drakulix:
Sorry, but is not the intended behaviour. We have a dedicated shortcut for maximize.

wiiznokes:
oh, didn't know there was a shortcut

Issue #397, opened 2024-04-01
xwayland apps not clickable


wiiznokes:
I noticed that after some time, vs code become not clickable, at least some part of the windows. This happen in both floating and tilling.
It's only the bottom left part that is not clickable.
Curiously, the bug seems to appears only when the window is on the left side of my screen.

wiiznokes:
Maybe it only with nvidia driver, i don't have this issue on my laptop

ryanabx:
Nope, I'm not running nvidia

mmstick:
You have to unstack and restack XWayland windows when they do this.

wiiznokes:
> You have to unstack and restack XWayland windows when they do this.

That don't fix it for me

mmstick:
Here's a workaround for VS Code to enable Wayland support.

1. Copy VS Code's `code.desktop` file your your home folder

```
mkdir -p ~/.local/share/applications
cp /usr/share/applications/code.desktop ~/.local/share/applications/
```

2. Edit the `Exec` line to the following

```
Exec=/usr/share/code/code --enable-features=UseOzonePlatform --enable-features=WaylandWindowDecorations --ozone-platform-hint=auto --unity-launch %F
```

3. Create an alias for code in your `~/.bashrc`

```
alias code='code --enable-features=UseOzonePlatform --enable-features=WaylandWindowDecorations --ozone-platform-hint=auto'
```

Issue #484, opened 2024-05-15
Strange mouse/cursor behaviour in Xwayland apps.


Tipcat-98:
This started when #448 and related commits were added.

In one game, V Rising.
The cursor is offset from it's actionable position.
As seen in the screenshot the option box is highlighted when the cursor is moderately to the left of it.
This only happens in "Exclusive Fullscreen" mode, where in "Windowed" it works fine.

https://media.discordapp.net/attachments/633653387723931659/1240331053198217216/IMG_2177.jpg?ex=66462bcf&is=6644da4f&hm=6ed2bf35c8961e1520daea2d50bf350bf7d8d1ce1820a1a5d1f401db045a7ece&=&format=webp&width=895&height=671

In another game, World of Warcraft.
The cursor position matches where it actually is but you cannot move it to the very right of the screen.
As seen in the screenshot, I cannot move it further to the right.
This only happens in "Fullscreen (window)" mode, in "Windowed" mode it is fine.

https://media.discordapp.net/attachments/633653387723931659/1240331052250173602/IMG_2178.jpg?ex=66462bcf&is=6644da4f&hm=785a6fbe5937073b46aead7a76c4a7cb90b7764613c7704ee8ef30cd4b1d0ca0&=&format=webp&width=895&height=671

I have two 1920x1080 monitors and I'm on Nvidia.

System Info:
Cosmic: Updated to latest git as of today
OS: Pop!_OS 22.04 LTS x86_64
Kernel: 6.8.0-76060800daily20240311-g
DE: GNOME 42.5
CPU: Intel i7-6700 (8) @ 4.000GHz
GPU: NVIDIA GeForce RTX 2060 Rev. A
GPU Driver: 550.67
Memory: 32024MiB

Tipcat-98:
I can't really confirm if this only happens with xwayland apps, I tried native wayland terraria and it does exhibit the same instability with nvidia as my other issue - #419.

I'll rename to say fullscreen apps.

Tipcat-98:
![screenshot-2024-06-11-15-28-41](https://github.com/pop-os/cosmic-comp/assets/73800942/14226c9d-2464-46cd-b847-96d2ece94773)

Tipcat-98:
This seems to have worsened as of late.
With the Nvidia legacy code path fix you can now get in-game without crashing but in most xwayland games you can't click anything on the screen

ltsymes23:
[...]

PR #575, opened 2024-06-26
output: Store position as u32 and offset bad configs

Drakulix:
Should at least partially fix https://github.com/pop-os/cosmic-comp/issues/397.

Fixes the exact symptoms observed in:
- https://github.com/pop-os/cosmic-epoch/issues/394
- https://github.com/pop-os/cosmic-epoch/issues/391

Will reset output configuration for already bad settings (where parsing now with u32 will fail), but should simply correct any new bad configs, by offsetting everything.

### DIFF
--- a/src/backend/kms/device.rs
+++ b/src/backend/kms/device.rs
@@ -22,7 +22,7 @@ use smithay::{
         rustix::fs::OFlags,
         wayland_server::{protocol::wl_buffer::WlBuffer, DisplayHandle, Weak},
     },
-    utils::{DevPath, DeviceFd, Transform},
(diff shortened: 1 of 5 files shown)


Tipcat-98:
Probably fixes #484 too?

Drakulix:
> Probably fixes #484 too?

Maybe, I am not sure that is related to the same issue. But please test and report!

Tipcat-98:
It seems to have partially fixed it, the wow issue is fixed with this patch but the v rising one seems to persist , so maybe one of the issues I had was really just this.

PR #582, opened 2024-06-27
Configurable mouse focus

### LINKED ISSUE #564: [Feature Request] Focus follows mouse for tiling mode
This should be an option that users can enable and disable to suit their workflow.

(I hope i created this in the correct project)

skewballfox:
- resolves #564 

Still learning the codebase, and right now this is just some boilerplate. As mentioned in 564 there would be 3 mouse focus policies:
- focus requires click (current and default)
- focus changes when mouse moves to a new output
- focus follows mouse (sloppily or lazily)

The first requires 0 changes and the code for click to focus still needs to execute when following one of the other policies (if there is a delay before window focusing, clicking a window should still focus it.). So the only place code needs to be added is in the two `todo!()`s currently in `src/input/mod.rs`. I think  I can probably extract the contents of [this conditional](https://github.com/pop-os/cosmic-comp/blob/713ac470aa992712b105116e44c338986d591782/src/input/mod.rs#L755) into a function, with some minor changes to reuse the existing code for moving keyboard focus to a window or region.

### DIFF
--- a/cosmic-comp-config/src/input.rs
+++ b/cosmic-comp-config/src/input.rs
@@ -5,6 +5,8 @@
 pub use input::{AccelProfile, ClickMethod, ScrollMethod, TapButtonMap};
 use serde::{Deserialize, Serialize};
 
+// Note: For the following values, None is used to represent the system default
+// Configuration for input devices
 #[derive(Clone, Debug, Default, PartialEq, Deserialize, Serialize)]
 pub struct InputConfig {
     pub state: DeviceState,
--- a/cosmic-comp-config/src/lib.rs
+++ b/cosmic-comp-config/src/lib.rs
@@ -23,6 +23,12 @@ pub struct CosmicCompConfig {
     pub autotile_behavior: TileBehavior,
     /// Active hint enabled
     pub active_hint: bool,
+    /// Enables changing keyboard focus to windows when the cursor passes into them
+    pub focus_follows_cursor: bool,
+    /// Enables warping the cursor to the focused window when focus changes due to keyboard input
+    pub cursor_follows_focus: bool,
+    /// The delay in milliseconds before focus follows mouse (if enabled)
+    pub focus_follows_cursor_delay: u64,
     /// Let X11 applications scale themselves
     pub descale_xwayland: bool,
 }
@@ -50,6 +56,9 @@ impl Default for CosmicCompConfig {
             autotile: Default::default(),
             autotile_behavior: Default::default(),
             active_hint: true,
+            focus_follows_cursor: false,
+            cursor_follows_focus: false,
+            focus_follows_cursor_delay: 250,
             descale_xwayland: false,
         }
     }
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -27,7 +27,10 @@ use crate::{
         },
     },
 };
-use calloop::{timer::Timer, RegistrationToken};
+use calloop::{
+    timer::{TimeoutAction, Timer},
+    RegistrationToken,
+};
 use cosmic_comp_config::{workspace::WorkspaceLayout, TileBehavior};
 use cosmic_config::ConfigSet;
 use cosmic_settings_config::shortcuts;
@@ -85,6 +88,17 @@ use std::{
 };
 
 pub mod gestures;
+
+/// Used for debouncing focus updates due to pointer motion, if after the focus change is
+/// triggered the event will cancel if the pointer moves to the original target
+#[derive(Debug)]
+pub struct PointerFocusState {
+    //the window under the cursor prior to it's movement
+    originally_focused_window: Option<KeyboardFocusTarget>,
+    //the window under the cursor after it's movement
+    scheduled_focused_window: Option<KeyboardFocusTarget>,
+    token: RegistrationToken,
+}
 
 #[derive(Default)]
 pub struct SupressedKeys(RefCell<Vec<(Keycode, Option<RegistrationToken>)>>);
@@ -203,24 +217,19 @@ impl State {
                     .cloned();
                 if let Some(seat) = maybe_seat {
                     self.common.idle_notifier_state.notify_activity(&seat);
-                    let current_output = seat.active_output();
-                    let shortcuts_inhibited = self
-                        .common
-                        .shell
-                        .read()
-                        .unwrap()
-                        .active_space(&current_output)
-                        .focus_stack
-                        .get(&seat)
-                        .last()
-                        .and_then(|window| {
-                            window.wl_surface().and_then(|surface| {
-                                seat.keyboard_shortcuts_inhibitor_for_surface(&surface)
-                            })
-                        })
-                        .map(|inhibitor| inhibitor.is_active())
-                        .unwrap_or(false);
 
+                    let current_focus = seat.get_keyboard().unwrap().current_focus();
+                    //this should fall back to active output since there may not be a focused output
+                    let focused_output = seat.focused_or_active_output();
+
+                    let shortcuts_inhibited = current_focus.is_some_and(|f| {
+                        f.wl_surface()
+                            .and_then(|surface| {
+                                seat.keyboard_shortcuts_inhibitor_for_surface(&surface)
+                                    .map(|inhibitor| inhibitor.is_active())
+                            })
+                            .unwrap_or(false)
+                    });
                     let keycode = event.key_code();
                     let state = event.state();
                     trace!(?keycode, ?state, "key");
@@ -266,7 +275,7 @@ impl State {
                                             shell.set_overview_mode(None, data.common.event_loop_handle.clone());
 
                                             if let Some(focus) = current_focus {
-                                                if let Some(new_descriptor) = shell.workspaces.active(&current_output).1.node_desc(focus) {
+                                                if let Some(new_descriptor) = shell.workspaces.active(&focused_output).1.node_desc(focus) {
                                                     let mut spaces = shell.workspaces.spaces_mut();
                                                     if old_descriptor.handle != new_descriptor.handle {
                                                         let (mut old_w, mut other_w) = spaces.partition::<Vec<_>, _>(|w| w.handle == old_descriptor.handle);
@@ -275,7 +284,7 @@ impl State {
                                                                 if let Some(focus) = TilingLayout::swap_trees(&mut old_workspace.tiling_layer, Some(&mut new_workspace.tiling_layer), &old_descriptor, &new_descriptor) {
                                                                     let seat = seat.clone();
                                                                     data.common.event_loop_handle.insert_idle(move |state| {
-                                                                        Shell::set_focus(state, Some(&focus), &seat, None);
+                                                                        Shell::set_focus(state, Some(&focus), &seat, None,true);
                                                                     });
                                                                 }
                                                                 old_workspace.refresh_focus_stack();
@@ -288,7 +297,7 @@ impl State {
                                                                 std::mem::drop(spaces);
                                                                 let seat = seat.clone();
                                                                 data.common.event_loop_handle.insert_idle(move |state| {
-                                                                    Shell::set_focus(state, Some(&focus), &seat, None);
+                                                                    Shell::set_focus(state, Some(&focus), &seat, None,true);
                                                                 });
                                                             }
                                                             workspace.refresh_focus_stack();
@@ -296,7 +305,7 @@ impl State {
                                                     }
                                                 }
                                             } else {
-                                                let new_workspace = shell.workspaces.active(&current_output).1.handle;
+                                                let new_workspace = shell.workspaces.active(&focused_output).1.handle;
                                                 if new_workspace != old_descriptor.handle {
                                                     let spaces = shell.workspaces.spaces_mut();
                                                     let (mut old_w, mut other_w) = spaces.partition::<Vec<_>, _>(|w| w.handle == old_descriptor.handle);
@@ -306,7 +315,7 @@ impl State {
                                                                 if let Some(focus) = TilingLayout::move_tree(&mut old_workspace.tiling_layer, &mut new_workspace.tiling_layer, &new_workspace.handle, &seat, new_workspace.focus_stack.get(&seat).iter(), old_descriptor.clone()) {
                                                                     let seat = seat.clone();
                                                                     data.common.event_loop_handle.insert_idle(move |state| {
-                                                                        Shell::set_focus(state, Some(&focus), &seat, None);
+                                                                        Shell::set_focus(state, Some(&focus), &seat, None,true);
                                                                     });
                                                                 }
                                                                 old_workspace.refresh_focus_stack();
@@ -559,7 +568,7 @@ impl State {
                             _ => {}
                         });
                     }
-
+                    let original_position = position;
                     position += event.delta().as_global();
 
                     let output = shell
@@ -587,16 +596,111 @@ impl State {
                         return;
                     }
 
-                    if ptr.is_grabbed()
-                        && seat
+                    if ptr.is_grabbed() {
+                        if seat
                             .user_data()
                             .get::<ResizeGrabMarker>()
                             .map(|marker| marker.get())
                             .unwrap_or(false)
-                    {
-                        if output != current_output {
-                            ptr.frame(self);
-                            return;
+                        {
+                            if output != current_output {
+                                ptr.frame(self);
+                                return;
+                            }
+                        }
+                        //If the pointer isn't grabbed, we should check if the focused element should be updated
+                    } else if self.common.config.cosmic_conf.focus_follows_cursor {
+                        let shell = self.common.shell.read().unwrap();
+                        let (old_keyboard_target, _) =
+                            shell.keyboard_target_from_position(original_position, &seat);
+                        let (new_keyboard_target, _) =
+                            shell.keyboard_target_from_position(position, &seat);
+
+                        if old_keyboard_target != new_keyboard_target
+                            && new_keyboard_target.is_some()
+                        {
+                            let create_source = if self.common.pointer_focus_state.is_none() {
+                                true
+                            } else {
+                                let PointerFocusState {
+                                    originally_focused_window,
+                                    scheduled_focused_window,
+                                    token,
+                                } = self.common.pointer_focus_state.as_ref().unwrap();
+
+                                if &new_keyboard_target == originally_focused_window {
+                                    //if we moved to the original window, just cancel the event
+                                    self.common.event_loop_handle.remove(*token);
+                                    //clear the state
+                                    self.common.pointer_focus_state = None;
+                                    false
+                                } else if &new_keyboard_target != scheduled_focused_window {
+                                    //if we moved to a new window, update the scheduled focus
+                                    self.common.event_loop_handle.remove(*token);
+                                    true
+                                } else {
+                                    //the state doesn't need to be updated or cleared
+                                    false
+                                }
+                            };
+
+                            if create_source {
+                                // prevent popups from being unfocusable if there is a gap between them and their parent
+                                let delay = calloop::timer::Timer::from_duration(
+                                    //default to 250ms
+                                    std::time::Duration::from_millis(
+                                        self.common.config.cosmic_conf.focus_follows_cursor_delay,
+                                    ),
+                                );
+                                let seat = seat.clone();
+                                let token = self
+                                    .common
+                                    .event_loop_handle
+                                    .insert_source(delay, move |_, _, state| {
+                                        let target = state
+                                            .common
+                                            .pointer_focus_state
+                                            .as_ref()
+                                            .unwrap()
+                                            .scheduled_focused_window
+                                            .clone();
+                                        //clear it prior in case the user twitches in the microsecond it
+                                        //takes this function to run
+                                        state.common.pointer_focus_state = None;
+
+                                        Shell::set_focus(
+                                            state,
+                                            target.as_ref(),
+                                            &seat,
+                                            Some(SERIAL_COUNTER.next_serial()),
+                                            false,
+                                        );
+
+                                        TimeoutAction::Drop
+                                    })
+                                    .ok();
+                                if token.is_some() {
+                                    let originally_focused_window =
+                                        if self.common.pointer_focus_state.is_none() {
+                                            old_keyboard_target
+                                        } else {
+                                            // In this case, the pointer has moved to a new window (neither original, nor scheduled)
+                                            // so we should preserve the original window for the focus state
+                                            self.common
+                                                .pointer_focus_state
+                                                .as_ref()
+                                                .unwrap()
+                                                .originally_focused_window
+                                                .clone()
+                                        };
+
+                                    self.common.pointer_focus_state = Some(PointerFocusState {
+                                        originally_focused_window,
+                                        scheduled_focused_window: new_keyboard_target,
+                                        token: token.unwrap(),
+                                    });
+                                }
+                            }
                         }
                     }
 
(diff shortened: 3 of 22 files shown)


Drakulix:
Thanks for working on this!

Though after some internal discussions we have now a better idea of how this feature should exactly look like: https://github.com/pop-os/cosmic-comp/issues/564#issuecomment-2197140658

skewballfox:
> Thanks for working on this!
> 
> Though after some internal discussions we have now a better idea of how this feature should exactly look like: [#564 (comment)](https://github.com/pop-os/cosmic-comp/issues/564#issuecomment-2197140658)

no worries. Were you guys planning on doing the refactors separately or should I attempt to do those in this PR?

specifically these changes: 

> To facilitate this, we will need a bit of refactoring, as to what previously was the active_output of a seat. Parts of the code assume that output will contain a focused element. But with this change conceptionally the "focused output" and the "active output" (the one with the cursor on) will now be different things.

Drakulix:
> no worries. Were you guys planning on doing the refactors separately or should I attempt to do those in this PR?
> 
> specifically these changes:
> 
> > To facilitate this, we will need a bit of refactoring, as to what previously was the active_output of a seat. Parts of the code assume that output will contain a focused element. But with this change conceptionally the "focused output" and the "active output" (the one with the cursor on) will now be different things.

If you want to, you may very well give this a go. Basically you can just `grep` the whole code base for `seat.active_output()` and see where it afterwards makes the assumption, that the returned output has a focused element.

I don't think most cases qualify, but those that do should instead do something like `seat.get_keyboard().current_focus()` to get the focused element and figure out the output by querying the `Shell`.

Otherwise I'll get around to it eventually.

skewballfox:
One thing I'm fuzzy on is the places where I've encountered `active_output()` being used to get the current focused window, they generally involve filtering to a CosmicMapped instance(example: [handle_shortcut_action](https://github.com/pop-os/cosmic-comp/blob/f02520c194624d2bcd31c7d37e442754dbbd59a1/src/input/mod.rs#L1637))

Are variants like popups coercable to `CosmicMapped` or `CosmicSurface`?

Asking because right now I'm trying implement a function to call during `Action::Close` on `KeyboardFocusTarget` but `PopupKind::InputMethod(PopupSurface)` doesn't have a useful method for this.

Drakulix:
> One thing I'm fuzzy on is the places where I've encountered `active_output()` being used to get the current focused window, they generally involve filtering to a CosmicMapped instance(example: [handle_shortcut_action](https://github.com/pop-os/cosmic-comp/blob/f02520c194624d2bcd31c7d37e442754dbbd59a1/src/input/mod.rs#L1637))
> 
> Are variants like popups coercable to `CosmicMapped` or `CosmicSurface`?

Popups might be related to their parent: https://github.com/pop-os/cosmic-comp/blob/f02520c194624d2bcd31c7d37e442754dbbd59a1/src/shell/focus/target.rs#L163

> Asking because right now I'm trying implement a function to call during `Action::Close` on `KeyboardFocusTarget` but `PopupKind::InputMethod(PopupSurface)` doesn't have a useful method for this.

However I don't think that method works for input-method. What you probably would need to do there (and what cosmic-comp likely fails to do in other places, IME support isn't well-tested yet..) is using [`get_parent`](https://smithay.github.io/smithay/smithay/wayland/input_method/struct.PopupSurface.html#method.get_parent) and then `shell.element_for_surface`.

skewballfox:
hey for the commands affecting a workspace, such as `MigrateWorkspaceToNextOutput`, those should also be switched to get the focused output rather than the active output correct?

skewballfox:
I'm going to assume this is the case, though if I need to roll back changes for those specific actions let me know

skewballfox:
Give me a few, tried to squash the commits and now it's including changes from other commits in the list of changes

Drakulix:
> hey for the actions affecting a workspace, such as `MigrateWorkspaceToNextOutput`, those should also be switched to get the focused output rather than the active output correct?

I think for that specific one it might make more sense to use the active output, though there might not be much point into moving an empty workspace? :sweat_smile: 
But my gut feeling would be to use the active_output, whenever an action concerns a whole workspace and use the output with the focused element, whenever the action concerns mostly that element or focus in some way.

> I'm going to assume this is the case, though if I need to roll back changes for those specific actions let me know

Sure. Do you want me to review what's already there?

I am usually busy with other stuff and will only check comments every now and then, but if you feel like you want some input on the actual code, please explicitly ping me any time. Otherwise it isn't clear to me, what is just a rebase or warrants a full review.

skewballfox:
@Drakulix If you could review primarily the changes to `handle_shortcut_actions` (starting on line 1645 of `src/input/mod.rs` I'd appreciate it. mainly to see which actions should be rolled back or if I missed any that need to be switched.

There are a few functions I'm not sure whether to convert or not: 
- [`map_window` in the shell module](https://github.com/pop-os/cosmic-comp/blob/11914e97ea5e5bc1d856ccf01f70d9a2ff0c500f/src/shell/mod.rs#L1741). I suspect it should be.
- [update for `ContextMenu` in `src/shell/graphs/mod.rs`](https://github.com/pop-os/cosmic-comp/blob/11914e97ea5e5bc1d856ccf01f70d9a2ff0c500f/src/shell/grabs/menu/mod.rs#L200), mainly because I'm lacking a bit of context of for what this is
- the stuff related to layers, but mainly because I'm not sure what a layer is in this context. I'm assuming it has to do with the z order or windows, but I can't see how a layer would have keyboard focus

Also I'm starting to think there might be a better, or at least more idiomatic, way of doing this. Instead of repeatedly getting the keyboard focus target, and then using that to determine the output with the shell, add a new `FocuedOutput` entry to user data, add a getter (`seat.focused_output()`) but not a setter, and set/update the  entry within `ActiveFocus::set()`, this would minimize the code changes necessary and make better use of the existing focus related stucts like `FocusStack`. plus, I can set it up to fall back to ActiveOutput when there are no focus targets which seems less error prone

src/shell/seats.rs
```diff
@@ -229,13 +245,31 @@ impl SeatExt for Seat<State> {
...
+            Some(
+                self.user_data()
+                    .get::<FocusedOutput>()
+                    .map(|x| x.0.lock().unwrap().clone())
+                    .unwrap(),
+            )
+        } else {
+            None
+        }
+    }
```
Drakulix, line 271: I wonder, if we really want to have this function or just always query `Shell` with the seat instead. My issue is, that this is storing the same state in multiple ways. If we ever forget to update `focused_output` when changing focus, this will break behaviour quite drastically.

src/input/mod.rs
```diff
@@ -2035,199 +2045,202 @@ impl State {
                     _ => unreachable!(),
                 };
 
-                let current_output = seat.active_output();
-                let mut shell = self.common.shell.write().unwrap();
-                let next_output = shell.next_output(&current_output, direction).cloned();
+                if let Some(focused_output) = seat.focused_output() {
```
Drakulix, line 2048: Also instead of wrapping everything in an `if` we could just do something like:
`let Some(focused_output) = seat.focused_output() else { return };`

These kind of actions don't do anything without focus anyway.

src/input/mod.rs
```diff
@@ -2350,7 +2366,8 @@ impl State {
                 &self.common.config,
                 self.common.event_loop_handle.clone(),
             ),
-
+            // NOTE: implementation currently assumes actions that apply to outputs should apply to the active output
+            // rather than the output that has keyboard focus
```
Drakulix, line 2370: Yes, this seems right to me, as `map_window` should still happen on the active_output.

Drakulix (commented):
- `ToggleWindowFloating` should also use the focused output. (Like ToggleStacking and ToggleSticky, but that is delegated to `Shell`.)

> `map_window`. I suspect it should be.

No windows should spawn where the cursor is, so `active_output`.

> [update for ContextMenu in src/shell/graps/mod.rs](https://github.com/pop-os/cosmic-comp/blob/11914e97ea5e5bc1d856ccf01f70d9a2ff0c500f/src/shell/grabs/menu/mod.rs#L200), mainly because I'm lacking a bit of context of for what this is

This is the context menu you get by right-clicking header-bars of apps. So the actions in there should always apply to the window, which the menu belongs to. As such that shouldn't need changing.

> the stuff related to layers, but mainly because I'm not sure what a layer is in this context. 

Those are wlr-layer-shell components, essentially shell-components like the panel or launcher. Those can have focus, but that should also remain on the active-output, so probably no changes necessary. (Though we should double check.)

src/shell/seats.rs
skewballfox: It's up to you, though the places where this is being set is the same as when keyboard focus target is updated. 

In the meantime, I've reverted the changes to `input/mod.rs` on my end and will force push once I implement changes from the other reviews.

src/input/mod.rs
skewballfox: do you want me to change the existing instances or just do that going forward?

src/shell/seats.rs
skewballfox: we could wrap all 3 actions (set Active focus, set focused output, set keyboard focus) inside a new private function that is called internally by `setfocus` and `refresh_focus` to make sure that those actions always happen together. That would probably be the best way to implement focus follows pointer if set anyway

skewballfox:
would it be okay if I added a bool argument to `set_focus` to indicate whether the pointer should also be updated if focus changes?

If a focus change is caused by the pointer(movement or click), then even if pointer follows focus is set it shouldn't change position. 

the alternative would be to check if the keyboard focus target is different from the active focus at the current call sites, prior to calling `set_focus`(along with filtering for positions that would cause set_focus to return early)

src/input/mod.rs
skewballfox: Would this comment be useful for other contributors or should I go ahead and remove it?

src/shell/seats.rs
Drakulix: > we could wrap all 3 actions (set Active focus, set focused output, set keyboard focus) inside a new private function that is called internally by `setfocus` and `refresh_focus` to make sure that those actions always happen together. That would probably be the best way to implement focus follows pointer if set anyway

Sounds good to me.

src/input/mod.rs
Drakulix: I would like you to change the existing instances as well for consistency and readability. The extra level of indentation also makes the git-diff far more annoying to parse.

src/input/mod.rs
Drakulix: Please feel free to add these kinds of comments that explain design decisions!

Drakulix:
> would it be okay if I added a bool argument to `set_focus` to indicate whether the pointer should also be updated if focus changes?
> 
> If a focus change is caused by the pointer(movement or click), then even if pointer follows focus is set it shouldn't change position.

Makes sense. This is the kinda behaviour that isn't readable from the config, because it depends on the type of focus change. So imo a new argument is a good solution.

> Edit: Also, would it be okay if I moved the setting of focus data (active focus, keyboard focus) inside a check to make sure the keyboard focus has changed (if keyboard focus target != previous value of active focus)

Setting the focus to the same value is harmless and shouldn't cause any issues.

src/input/mod.rs
skewballfox: implemented

src/shell/focus/mod.rs
```diff
@@ -219,9 +223,28 @@ fn update_focus_state(
...
-                todo!("update cursor position");
+                // get the top left corner of the target element
+                let element = state
+                    .common
+                    .shell
+                    .read()
+                    .unwrap()
+                    .focused_element(target.unwrap())
+                    .unwrap();
+                let top_left = element.active_window_geometry().loc.to_f64();
```
Drakulix, line 235: This is probably not in global coordinates, which would be what you need.
You need to use something similar to [this](https://github.com/pop-os/cosmic-comp/blob/master/src/shell/mod.rs#L2740-L2760) for that (perhabs that should be a new `Shell::element_geometry` method).

skewballfox: if I just need the location as global coordinates, would calling `element.active_window_geometry.as_global()` work? Also if I need globabl coordinates, I'm guessing `PointerFocusTarget.motion` is not the function I need to call given MotionEvent want a logical coordinate. What should I use instead?

skewballfox:
also, I got focus follows mouse working, at least on the local system I've been using for testing

skewballfox:
so it turns out the issue *might* not be related to layer order, and instead a lack of delay for refocus. I can only hover over an overlay panel if I drag the mouse down the side of the output where the window isn't present, then onto the panel. I think the gap between the panel and the popup is stealing focus. I'll roll back that latest commit locally to retest and confirm in the morning.

src/shell/mod.rs
```diff
@@ -1782,6 +1782,8 @@ impl Shell {
         };
     }
 
+    /// Determine whether a Surface that has changed should be focused, and if so, return the target. Only called when the
+    /// Compositor is handling visibile state changes.
```
Drakulix, line 1786: This isn't entirely correct. This method does "map" windows somewhere onto a workspace/output. Otherwise they aren't visible, as they are not part of the internal tree representation of the shell.
This can effect focus, which is why it might return a new focus target, but it isn't concerned with only focus.

skewballfox:
So I looked a bit more into how to handle delaying focus enough to allow time to move the cursor to windows floating over the a focusable window. If we do go with the option to use a delay, we can likely do so with `self.common.event_loop_handle.insert_source` and a `calloop::timer::Timer`.

An alternative (and apparently preferred way is [this comment about gnome's implementation](https://superuser.com/a/614078) is still accurate) is to delay changing window focus until the pointer has stopped moving. I haven't figured out how to detect that though.

src/shell/seats.rs
skewballfox: I'll mark this as resolved since I've moved setting these to an internal function

skewballfox:
@Drakulix This is ready for review. both settings are working. Thanks for all the help

juarezr:
Adding a third possible mode:

- [Mouse follows focus](https://extensions.gnome.org/extension/4642/mouse-follows-focus/)
- Useful especially with multiple monitors and keyboard-centric workflows (Alt+tab and Focus prev/next/last keys)

skewballfox:
@juarezr that's actually what "cursor follows focus" is: the pointer warps to the top left corner of the newly focused window if that focus change was caused by a keyboard event. 

it's been implemented in the PR.

src/input/mod.rs
```diff
@@ -2379,7 +2461,8 @@ impl State {
                     .unwrap()
                     .toggle_stacking_focused(seat);
                 if let Some(new_focus) = res {
-                    Shell::set_focus(self, Some(&new_focus), seat, Some(serial));
+                    //TODO: verify whether cursor follows focus should be checked when toggling stacking
+                    Shell::set_focus(self, Some(&new_focus), seat, Some(serial), false);
```
Drakulix, line 2465: Drop the TODO please. I don't think it should, the window / group won't visually move when creating a stack, so if the user moved the pointer outside of it, that was probably intentional.

cosmic-comp-config/src/input.rs
```diff
@@ -29,6 +31,10 @@ pub struct InputConfig {
     pub tap_config: Option<TapConfig>,
     #[serde(skip_serializing_if = "Option::is_none", default)]
     pub map_to_output: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none", default)]
+    pub focus_follows_cursor: Option<bool>,
+    #[serde(skip_serializing_if = "Option::is_none", default)]
+    pub cursor_follows_focus: Option<bool>,
 }
```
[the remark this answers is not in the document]
skewballfox: This is unused. When I initially started I wasn't sure if it should where the setting should go. I left it here in case it needed to be input device or seat specific. I'll remove it in my next push

src/input/mod.rs
```diff
@@ -1574,6 +1475,133 @@ impl State {
         }
     }
 
+    /// Used with cursor triggered events to check whether the keyboard focus should be updated.
+    fn maybe_update_focus(
```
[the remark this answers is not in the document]
skewballfox: I think I named it that because under was initially set to None, and it looked as though there were some branches where it was never set to a value. Since the focus doesn't change everytime I thought it best to prefix it with some kind of conditional

cosmic-comp-config/src/lib.rs
```diff
@@ -23,6 +23,10 @@ pub struct CosmicCompConfig {
     pub autotile_behavior: TileBehavior,
     /// Active hint enabled
     pub active_hint: bool,
+    /// Enables changing keyboard focus to windows when the cursor passes into them
+    pub focus_follows_cursor: bool,
+    /// Enables warping the cursor to the focused window when focus changes due to keyboard input
+    pub cursor_follows_focus: bool,
```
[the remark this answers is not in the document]
skewballfox: it's not, it's only read from `CosmicCompConfig`. It was a bit of scaffolding left over from when I started the PR and was still figuring things out.

src/input/mod.rs
```diff
@@ -561,16 +559,55 @@ impl State {
...
+                                let keyboard_focus_target = self
+                                    .common
+                                    .shell
+                                    .read()
+                                    .unwrap()
+                                    .element_under(loc.as_global(), &output);
+
+                                // prevent panel windows from collapsing when they occlude an empty workspace
+                                if keyboard_focus_target.is_some() {
+                                    // prevent popups from being unfocusable if there is a gap between them and their parent
```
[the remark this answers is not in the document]
skewballfox: A better solution might be to trigger the check on the end of pointer movement,(apparently that's how it's [handled with gnome](https://superuser.com/a/614078)). Do you have an idea how I might trigger it only on the final PointerMotion event?

I'm guessing I need to swap the delay in `insert_source` for something else

src/shell/focus/mod.rs
```diff
@@ -144,6 +145,7 @@ impl Shell {
         // update FocusStack and notify layouts about new focus (if any window)
         let workspace = self.space_for_mut(&mapped);
         let workspace = if workspace.is_none() {
+            //should this be the active output or the focused output?
```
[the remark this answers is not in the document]
skewballfox: I changed this to use `focused_or_active_output` which falls back to active output if focused output is none

src/wayland/handlers/compositor.rs
```diff
@@ -271,7 +271,7 @@ impl State {
                     if let Some(target) = res {
                         let seat = shell.seats.last_active().clone();
                         std::mem::drop(shell);
-                        Shell::set_focus(self, Some(&target), &seat, None);
+                        Shell::set_focus(self, Some(&target), &seat, None, false);
```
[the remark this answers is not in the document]
skewballfox: for this, and the menu grab stuff, I changed them to update the cursor position. I guess just confirm that's the behavior you want next time you test this.

skewballfox:
@Drakulix This is ready to be tested again. Let me know if I should keep or revert the cursor_follows_focus changes. No rush though.

cosmic-comp-config/src/input.rs
```diff
@@ -5,6 +5,8 @@
 pub use input::{AccelProfile, ClickMethod, ScrollMethod, TapButtonMap};
 use serde::{Deserialize, Serialize};
 
+// Note: For the following values, None is used to represent the system default
+/// Configuration for input devices
```
Drakulix, line 9:
```suggestion
// Configuration for input devices
```
(not a doc-comment)

src/input/mod.rs
```diff
@@ -1574,6 +1510,227 @@ impl State {
...
+                    };
+                }
+            }
+        }
+        std::mem::drop(shell);
+        Shell::set_focus(self, under.as_ref(), seat, Some(serial), false);
+    }
+
+    /// Derives a keyboard focus target from a global position. This is a variant of maybe_update_focus
+    /// that doesn't require write lock on the shell.
```
Drakulix, line 1641: The focus-update code gets increasingly complicated and I am not a big fan of this, given we duplicate a bunch of that logic now. From what I am seeing, this doesn't need `Common`/`self`, but only the shell.

Couldn't this be a method of the shell instead, that we could also use inside `maybe_update_focus`?

skewballfox: I moved it the shell, and use it inside of `maybe_update_focus`. I wasn't sure if I should leave the modifier key check in maybe update, given it was checked immediately prior to that in `keyboard_focus_from_position` function. 

I could see an argument for keeping it or leaving it.

src/input/mod.rs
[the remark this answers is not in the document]
skewballfox: would `cursor_focus_event` or something similar be a better name?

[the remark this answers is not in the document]
skewballfox: changed it to `cursor_focus` since it's only called for cursor clicks and cursor following

Drakulix:
A couple of issues I am still noticing using this PR:
- All three config options don't reload at runtime. This needs some changes here: https://github.com/pop-os/cosmic-comp/blob/master/src/config/mod.rs#L571
- Toggling tiling/floating doesn't warp the cursor (should it?)
- Switching between workspaces doesn't warp the cursor, if the new target isn't below it (should it?)

On the latter two, I am not sure how aggressive people would expect "cursor_follows_focus" to be, but both scenarios feel like a significant enough change to me, that I would expect it to warp.

However I am certainly not the target audience of any of these config options, so I am very much open to discuss those.

skewballfox:
> Changing the active output, doesn't take away focus for empty outputs (as intended!), but the focus indicator still vanishes, this probably needs some changes in the rendering code.

when did you test that? How were you confirming focus stayed in place? I ask because I just tried to replicate that(with/without changes to `src/backend/render/mod.rs`) but keyboard focus is stolen when the mouse moves to the empty workspace(keystrokes don't show in last focused window, or anywhere)

> On the latter two, I am not sure how aggressive people would expect "cursor_follows_focus" to be, but both scenarios feel like a significant enough change to me, that I would expect it to warp.

If I'm being honest, also not the target audience for `cursor_follows_focus`, or at least I've never used it/seen it in another DE. I don't think it should trigger for events caused by the mouse (such as clicking to launch something from the doc), but that's something I could live with if it's the desired behavior. 

I can try to extract, add a function to fire when the focus doesn't change, but for events like the focused window warps into another position

> Switching between workspaces doesn't warp the cursor, if the new target isn't below it (should it?)

you mean below the cursor right? as in two windows in the new workspace, the last focused on that workspace was window y, but window x is now under the cursor after the switch?

theshatterstone:
Hello. I am a Hyprland user, where I have mouse_follows_focus enabled and when it comes to the last 2 bullets by Draculix, toggling between tiling and floating does NOT warp the cursor or focus, and switching between workspaces also doesn't affect or change the focus. The only case in which mouse_follows_focus affects the mouse is when the focus changes via a keybind. So in terms of the way it should work, I don't think such things like points 3 and 4 should be the case.

Drakulix (approved):
LGTM! Could you do a `cargo fmt` run and squash that into the commit? I see a bunch of little style nitpicks, that I am sure that would catch. Otherwise this is good to go!

skewballfox:
It's good to go now

Drakulix:
Thanks again for all the work and feedback that went into this :tada:!

KaKi87:
Hi,
How to know when this will be available from APT updates ?
Thanks

leviport:
I can fit it into today's update release.

KaKi87:
Wow, wonderful, thanks !

theshatterstone:
Hello, I know this is far from the best place for it, but does anyone know where/how should I report that cosmic-comp at the COPR repo failed to build, so these don't work yet?

Drakulix:
> Hello, I know this is far from the best place for it, but does anyone know where/how should I report that cosmic-comp at the COPR repo failed to build, so these don't work yet?

https://pagure.io/fedora-cosmic/SIG

theshatterstone:
I emailed the COPR maintainer and they fixed it, thanks! Though it does feel like there's a slight delay in focus switching.

skewballfox:
@theshatterstone the delay is adjustable. the default is 250(milliseconds), you could set it to 100ms like this:
```
echo 100 > ~/.config/cosmic/com.system76.CosmicComp/v1/focus_follows_cursor_delay
```

KaKi87:
Hi,

> ```bash
> echo true > ~/.config/cosmic/com.system76.CosmicComp/v1/cursor_follows_focus
> ```

This one seems to be already enabled for me even though the file didn't exist, so I tried the following :

```bash
echo false > ~/.config/cosmic/com.system76.CosmicComp/v1/cursor_follows_focus
```

Without success.

So, how to actually *disable* the cursor following the focus, please ?

Thanks

Issue #48, opened 2022-12-09
Compostor segfaults on shm screencopy if file isn't large enough


ids1024:
This can be reproduced by modifying https://github.com/pop-os/cosmic-protocols/blob/main/examples/screenshot-screencopy.rs to remove the `set_len` call.

Interestingly attaching too small a buffer to a window also crashes the compositor, but it isn't a segfault; it's a SIGBUS that is caught. So maybe a somewhat different issue? But related: https://github.com/Smithay/smithay/issues/834

(I noticed this again since I was running `screenshot-screencopy` to take a screenshot, wondering why the compositor crashed, then seeing this was the version I had modified to crash the compositor :laughing:.)

Drakulix:
The SIGBUS is expected and should be caught by the shm-handler in smithay. If it is not, that is also a smithay bug.

ids1024:
https://github.com/Smithay/smithay/pull/1475 fixes the segfault, but then it panics in the `.unwrap()` in `submit_buffer`.

A bit awkward since it isn't easy to convert the error here to a `<R as Renderer>::Error`. Or to change the error type to `anyhow` or such.

PR #598, opened 2024-07-03
screencopy: Handle error from with_buffer_contents_mut 

ids1024:
Fixes https://github.com/pop-os/cosmic-comp/issues/48.

https://github.com/Smithay/smithay/pull/1475 fixes the segfault so this instead produces an error. Then we need to handle the error here.

### DIFF
--- a/src/wayland/handlers/screencopy/render.rs
+++ b/src/wayland/handlers/screencopy/render.rs
@@ -294,7 +294,11 @@ pub fn render_workspace_to_buffer(
             let size = buffer_dimensions(buffer).unwrap();
             let format =
                 with_buffer_contents(buffer, |_, _, data| shm_format_to_fourcc(data.format))
-                    .map_err(|_| DTError::OutputNoMode(OutputNoMode))? // eh, we have to do some error
+                    .map_err(|err| {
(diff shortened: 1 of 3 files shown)


Drakulix (approved):
LGTM

src/wayland/handlers/screencopy/render.rs
```diff
@@ -113,7 +114,8 @@ where
             }
             Ok(())
         })
-        .unwrap()
+        .map_err(|err| <R as Renderer>::Error::from_gles_error(GlesError::BufferAccessError(err)))
+        .and_then(|x| x)
         {
             frame.fail(FailureReason::Unknown);
```
Drakulix, line 120: Hmm, it is debatable if this shouldn't be a client constraints error instead.

ids1024: Actually, `frame.fail` may not really be needed for the `BufAccessError`, since Smithay sends a `wl_buffer` protocol error when that fails.

Drakulix: Well then it doesn't matter, because the client dies anyway. Better to send this rather than not imo.

Issue #601, opened 2024-07-04
Games with xwayland break in fullscreen when direct scanout is triggered


Dekomoro:
Wine Games repeatedly minimize and maximize, get a black screen when the mouse cursor leaves the window, or start with a blackscreen when in fullscreen

Native titles will repeatedly maximize and minimize, freeze, or both depending on cursor focus.
This is exhibited in both wine, and native titles made with FNA but not ones made with the source engine or godot

Native titles made with source 2 will stay minimized , if you try to open it from the application dock it will minimize itself automatically, unless you open it from the dock on the other monitor. However this may be another issue entirely as source 2 games exhibit this issue when using the Wayland back end(requires removing the SDL_VIDEO_DRIVER=x11 variable from cs2.sh)

Games that either work with native wayland or don't trigger direct scan-out do not exhibit this issue.

For wine it is at least not likely an issue with dxvk as pre-source 2 games that use it for Vulkan don't exhibit this issue

Dekomoro:
Forgot to mention but in fna's case the console spews out the error `RGBA8 swapchain unsupported, falling back to BGRA8 with swizzle`. Unsure if it's relevant at all however.

Dekomoro:
When minimized I also get this error `Failed to acquire swapchain image, not presenting`

Drakulix:
> Wine Games repeatedly minimize and maximize, get a black screen when the mouse cursor leaves the window, or start with a blackscreen when in fullscreen

The minimize issue isn't related to direct scanout, I wouldn't be so sure about the black screen as well.



> Forgot to mention but in fna's case the console spews out the error `RGBA8 swapchain unsupported, falling back to BGRA8 with swizzle`. Unsure if it's relevant at all however.

That is interesting, but shouldn't really concern the compositor. It might affect the gpu's ability to do direct scanout though. Could you post the outputs of `eglinfo` and `drm_info` on your system?

> When minimized I also get this error Failed to acquire swapchain image, not presenting

Expected, minimized windows are throttled.

Dekomoro:
Yeah.
drm_info: https://gist.github.com/Dekomoro/da97aeac6f0e450837cfbcb2a958ac97
eglinfo: https://gist.github.com/Dekomoro/7c494e5ced5ac1f3701b01a61ca9921a

I get the same results on Arch, POP_OS with cosmic master branch, and using the fedora copr if that matters

Drakulix:
[...]

Issue #475, opened 2024-05-10
Fullscreen games running in xwayland become non-functional once focus is lost


Dekomoro:
This includes freezing, black screens, other artifacts.

Tested with various games that I've found to work with Wayland and X11(using the SDL_VIDEODRIVER variable) including Terraria, Factorio, and Team Fortress 2. Wine games seem to always exhibit this behavior including ones that lose focus on launch.

Typically is gauranteed to happen if you move your mouse off the fullscreen game onto another monitor.

Seems related to https://github.com/pop-os/cosmic-comp/issues/442 and https://github.com/pop-os/cosmic-comp/issues/460

ids1024:
This could be related to https://github.com/pop-os/cosmic-comp/issues/231. I fixed one issue related to that, but there seem to be more problems.

For Linux-native SDL games, does setting `SDL_VIDEO_MINIMIZE_ON_FOCUS_LOSS=0` help here?

Dekomoro:
Appears it does not fix it

Dekomoro:
It does not

Dekomoro:
It appears it does

Dekomoro:
Would there be any way to get this to apply to games using wine?

Dekomoro:
It definitely seems to be related to fullscreen apps minimizing. Which happens to function when wayland apps do it.

Dekomoro:
Also happens both in tiling and floating modes if that info helps

ids1024:
Hm, odd that disabling `SDL_VIDEO_MINIMIZE_ON_FOCUS_LOSS` doesn't help, since I've seen that help.

We have several issues with Xwayland still. There's a lot of subtle details in getting the X window manager logic to play along with all clients.

Dekomoro:
Testing on Arch Linux it exhibits different behavior, Terraria freezes on focus lost with SDL_VIDEO_MINIMIZE_ON_FOCUS_LOSS set to 0, but crashes if it's set to 1

onlyreportingissues:
This sounds super similar: https://github.com/swaywm/sway/pull/5566

Dekomoro:
> Testing on Arch Linux it exhibits different behavior, Terraria freezes on focus lost with SDL_VIDEO_MINIMIZE_ON_FOCUS_LOSS set to 0, but crashes if it's set to 1

Apparently the crashing was just an issue with the game, games just freeze in fullscreen in general now, but not in borderless windowed

[...]

Issue #383, opened 2024-03-26
XWayland issues while changing workspace with a fullscreen game open and blank games


gabriele2000:
Just gonna copy-paste this from [here](https://github.com/pop-os/cosmic-comp/issues/316#issuecomment-2019822468) since it's a separate issue even if it's similar (not related to Cyberpunk 2077)

- Here's the log of journalctl 
[journalctl.log](https://github.com/pop-os/cosmic-comp/files/14754908/journalctl.log)

- First issue: The Crew 1 is just a black screen since yesterday's update to cosmic- packages [Video 1](https://youtu.be/A50kTFik9HA)

- Second issue: Switching workspaces or doing something remotely related to changing focus leads to the first issue [Video 2](https://youtu.be/0PReL5F2VG8)

Context on the first issue:
Proton's FSR patch, using FSHack, makes the game render at whatever resolution you want, then the game's DPI are scaled as if you were using the native resolution of your screen.

I use 720p that FSR will upscale at 1080p.
That leads to the game loading then quickly resizing the window or something like that, you can see a quickly changing window before the game renders (in the second video, in the first video the game just fails to render)

Drakulix:
I wonder if this is somewhat related to https://github.com/pop-os/cosmic-comp/issues/231, given that we are dealing with focus here. Could you try setting `SDL_VIDEO_MINIMIZE_ON_FOCUS_LOSS=0` and see if that changes the behaviour in any way?

gabriele2000:
Upon further testing, I don't quite understand the culprit of it.
Some games mess with "stealing focus" or something like that.
There's a unity game that once unfocused it would enter an unrecoverable state, it becomes a blank screen, but it only happens while it's in fullscreen.

This kind of issues puzzles me, since I've been told that "exclusive fullscreen doesn't exist in wayland" and XWayland it an X11 session inside Wayland (that means that the parent session doesn't have fullscreen).

So... what?!

Drakulix:
Should be fixed with https://github.com/pop-os/cosmic-comp/pull/603

PR #603, opened 2024-07-05
Fix xwm games

Drakulix:
This seems to fix the majority of minimize-bugs happening with a lot of games. (Where maximizing the window immediately minimizes it again.)

The culprit was the `wl_keyboard::leave` event, causing the app to register a focus loss, when we switched between `KeyboardTarget::Element` and `KeyboardTarget::Fullscreen`. (Confusingly most games seem to unfullscreen themselves before minimizing and restore fullscreen state after unminimizing... causing this weird transition in our code.)

This skips the set of events (thanks to https://github.com/Smithay/smithay/pull/1479), when the new target represents the same surface, which should be fine?!

Dunno, this probably needs some more testing to see, if it doesn't break other apps, but first tests seem promising.

Ref:
- https://github.com/pop-os/cosmic-comp/issues/601
- https://github.com/pop-os/cosmic-comp/issues/475
- https://github.com/pop-os/cosmic-comp/issues/383
- https://github.com/pop-os/cosmic-comp/issues/231

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4856,7 +4856,7 @@ checksum = "3c5e1a9a646d36c3599cd173a41282daf47c44583ad367b8e6837255952e5c67"
 [[package]]
 name = "smithay"
 version = "0.3.0"
-source = "git+https://github.com/smithay//smithay?rev=4171247#41712470abede74dab8a039ce048c70d56115ce6"
+source = "git+https://github.com/smithay//smithay?rev=e2b233f#e2b233ff588747310c8c4aa6bbb6d56c877351a8"
 dependencies = [
  "appendlist",
  "ash 0.38.0+1.3.281",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -119,4 +119,4 @@ inherits = "release"
 lto = "fat"
 
 [patch."https://github.com/Smithay/smithay.git"]
-smithay = {git = "https://github.com/smithay//smithay", rev = "4171247"}
+smithay = {git = "https://github.com/smithay//smithay", rev = "e2b233f"}
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -597,7 +597,7 @@ impl State {
                                     (0, 0),
                                     WindowSurfaceType::ALL,
                                 )
-                                .is_some()
+                                .is_none()
                                 {
                                     ptr.frame(self);
                                     return;
--- a/src/shell/element/surface.rs
+++ b/src/shell/element/surface.rs
@@ -378,8 +378,11 @@ impl CosmicSurface {
             .get_or_insert_threadsafe(Minimized::default)
             .0
             .store(minimized, Ordering::SeqCst);
-        if let WindowSurface::X11(surface) = self.0.underlying_surface() {
-            let _ = surface.set_mapped(!minimized);
+        if !minimized {
+            if let WindowSurface::X11(surface) = self.0.underlying_surface() {
+                let _ = surface.set_mapped(false);
+                let _ = surface.set_mapped(true);
+            }
         }
     }
 
--- a/src/shell/element/surface.rs
+++ b/src/shell/element/surface.rs
@@ -365,28 +365,21 @@ impl CosmicSurface {
     }
 
     pub fn is_minimized(&self) -> bool {
-        match self.0.underlying_surface() {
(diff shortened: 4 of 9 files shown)


Vixea:
Good news fixes minecraft from my testing bad news it seems all my steam games are broken. :| I'm blaming steam

Drakulix:
https://github.com/pop-os/cosmic-comp/issues/434 should now also be fixed

Drakulix:
More updates that slightly fix behaviour, but still doesn't restore most windows properly...

Draft until it works completely to not break anything currently working by merging a subset of fixes.

ids1024:
This `replace` method seems like a nice and simple solution to the leave/enter issue with fullscreening.

Though it does mean that when the `WlSurface` isn't changed, `enter`/`leave`/`modifiers` also aren't called on the targets that delegate to `WlSurface`, which could change behavior if they do anything other than call `enter`/`leave`/`modifiers` on the `WlSurface`.

I think that's not a problem with the current targets, but is something to keep in mind.

Vixea:
Tested, it again this time steam games work. Alt tab, super+a / shash. (I need to find a better way to write this)
Also please tell me if you wouldn't like me to clutter prs with testinf reports as i assume you at least do a little bit.

Drakulix:
Okay the latest commit should fix the majority of game breaking issues.

Unfortunately it also breaks the minimized window preview for all Xwayland apps, as we now unmap Xwayland windows on minimize (as most of them seem to expect). As a result, we don't have a wl_surface anymore, that we could render from. So when the applet later requests a buffer via screencopy it gets nothing..

Drakulix:
> Also please tell me if you wouldn't like me to clutter prs with testinf reports as i assume you at least do a little bit.

Yes please, especially any previously unknown issues are very valuable, as is confirming that the changes are indeed fixing stuff.

Dekomoro:
Tested, and seems to fix my issues so far. I'll test further and let you know my findings.

Dekomoro:
Only issue I'm recieving is some games fail to go fullscreen after minimizing again. other than that it's functional.

Drakulix:
> Only issue I'm recieving is some games fail to go fullscreen after minimizing but still function in a windowed state. other than that it's functional.

Does that specific game behave differently on other desktops? Which one is it, so I can maybe try to reproduce this?

Dekomoro:
yes, it's team fortress 2.

Drakulix:
> yes, main game exhibiting this issue is Team Fortres 2.

Thanks! Will try to reproduce this.

ryanabx:
> Okay the latest commit should fix the majority of game breaking issues.
> 
> 
> 
> Unfortunately it also breaks the minimized window preview for all Xwayland apps, as we now unmap Xwayland windows on minimize (as most of them seem to expect). As a result, we don't have a wl_surface anymore, that we could render from. So when the applet later requests a buffer via screencopy it gets nothing..

We could maybe fix this by changing the applet to screen grab when the minimize is pending? Just thinking out loud

Drakulix:
> > Okay the latest commit should fix the majority of game breaking issues.
> > Unfortunately it also breaks the minimized window preview for all Xwayland apps, as we now unmap Xwayland windows on minimize (as most of them seem to expect). As a result, we don't have a wl_surface anymore, that we could render from. So when the applet later requests a buffer via screencopy it gets nothing..
> 
> We could maybe fix this by changing the applet to screen grab when the minimize is pending? Just thinking out loud

Decided on a more "creative" approach to fix this: https://github.com/pop-os/cosmic-comp/pull/603/commits/e47e9fff8660a55dcd7dedafd464755e562df332

So far from 16 reported issues listing various games, no game is complaining about this hack. If it turns out to be a problem for any specific game, we will just need to add an idle-task to capture a preview and then unmap it afterwards.

Drakulix:
Alright, going ahead and merge this. This PR improves the situation significantly without any regressions I can find.

There are still a couple of open problems.

Some games don't correctly return to fullscreen state after minimizing (but they also don't request fullscreen state, so I wonder what is missing for them).

Some other games still struggle to work correctly when toggling between borderless-windowed and fullscreen modes, but both modes generally do seem to work with everything I have tried, the games just needed restarts.

More fixes will come at some point, but debugging these has proven to be more complex. Especially as solutions for some games seem to mysteriously break others, so I am going with this for now until I have time to investigate this further.

PR #609, opened 2024-07-10
Render only layer shell surfaces when `cosmic-workspaces` is open, and add a `SplitRenderElements`

ids1024:
This helps with issues like https://github.com/pop-os/cosmic-workspaces-epoch/issues/59.

I'm not entirely happy with the growing complexity of `workspace_elements`. I was thinking of a couple ways to split it up, noticed the most redundant part was the repeated code extending `p_elements` and `w_elements`. And the cleanest way to de-duplicate that out is with a dedicated `SplitRenderElements` type, which could be good in general. I have a commit adding that.

This also adds the commit from https://github.com/pop-os/cosmic-comp/pull/459, which makes more of a difference now. (If we wanted an animation between workspaces with this, we'd need to make sure `cosmic-comp` and `cosmic-comp` animate in sync.)

This seems to generally work with cosmic-workspaces modified to leave the background transparent. But I still need to do a bit more testing with it. (And I guess make this change on apply to the X11/winit backends.)

### DIFF
--- a/src/backend/render/mod.rs
+++ b/src/backend/render/mod.rs
@@ -464,6 +464,13 @@ where
 #[cfg(not(feature = "debug"))]
 pub type EguiState = ();
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum ElementFilter {
+    All,
+    ExcludeWorkspaceOverview,
+    LayerShellOnly,
+}
+
 #[derive(Clone, Debug)]
 pub struct SplitRenderElements<E> {
     pub w_elements: Vec<E>,
@@ -528,7 +535,7 @@ pub fn workspace_elements<R>(
     previous: Option<(WorkspaceHandle, usize, WorkspaceDelta)>,
     current: (WorkspaceHandle, usize),
     cursor_mode: CursorMode,
-    exclude_workspace_overview: bool,
+    element_filter: ElementFilter,
     _fps: Option<(&EguiState, &Timings)>,
 ) -> Result<Vec<CosmicElement<R>>, RenderError<R>>
 where
@@ -556,7 +563,7 @@ where
         now,
         output,
         cursor_mode,
-        exclude_workspace_overview,
+        element_filter == ElementFilter::ExcludeWorkspaceOverview,
     ));
 
     #[cfg(feature = "debug")]
@@ -642,8 +649,7 @@ where
         .as_ref()
         .filter(|f| !f.is_animating())
         .is_some();
-    let overlay_elements =
-        split_layer_elements(renderer, output, Layer::Overlay, exclude_workspace_overview);
+    let overlay_elements = split_layer_elements(renderer, output, Layer::Overlay, element_filter);
 
     // overlay is above everything
     elements
@@ -655,7 +661,7 @@ where
 
     if !has_fullscreen {
         elements.extend_from_workspace_elements(
-            split_layer_elements(renderer, output, Layer::Top, exclude_workspace_overview),
+            split_layer_elements(renderer, output, Layer::Top, element_filter),
             (0, 0).into(),
         );
     };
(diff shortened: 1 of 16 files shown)


src/backend/render/mod.rs
```diff
@@ -586,20 +642,22 @@ where
...
-    let mut window_elements = if !has_fullscreen {
-        let (top_elements, top_popups) =
-            split_layer_elements(renderer, output, Layer::Top, exclude_workspace_overview);
-        elements.extend(top_popups.into_iter().map(Into::into));
-        top_elements.into_iter().map(Into::into).collect()
-    } else {
-        Vec::new()
+    elements
+        .p_elements
+        .extend(overlay_elements.p_elements.into_iter().map(Into::into));
```
Drakulix, line 657: Why not `overlay_elements.join` or `chain` at least? (Which should probably still yield an `ExactSizeIterator`, which should be good for `extend`.

src/shell/mod.rs
```diff
@@ -416,13 +416,23 @@ impl WorkspaceSet {
...
             state.remove_workspace_state(&self.workspaces[old_active].handle, WState::Active);
             state.remove_workspace_state(&self.workspaces[old_active].handle, WState::Urgent);
             state.remove_workspace_state(&self.workspaces[idx].handle, WState::Urgent);
             state.add_workspace_state(&self.workspaces[idx].handle, WState::Active);
-            self.previously_active = Some((old_active, workspace_delta));
+            self.previously_active = if animate {
+                Some((old_active, workspace_delta))
+            } else {
+                None
+            };
```
Drakulix, line 435: Same question as previously, how does this interact with gestures?

Looks like it should just immediately swap the workspace, which seems a bit jarring. We probably want the windows from the overview to react to the gesture, so do we need a way further down the line to capture the gesture in the overview-client? I guess for now it might make sense to simply disable the gesture, when the overview is visible?

src/backend/render/mod.rs
```diff
@@ -799,23 +807,25 @@ where
...
+                    (!move_active && is_active_space).then_some(&last_active_seat),
+                    overview,
+                    resize_indicator,
+                    active_hint,
+                    theme,
+                )
+                .map_err(|_| OutputNoMode)?,
+            offset.to_physical_precise_round(output_scale),
+        );
+    }
```
Drakulix, line 824: Shouldn't the workspace elements from the "previous" workspace also be ignored the same? Or does this rely on animations being disabled for the overview?

Drakulix (commented):
LGTM on first glance. The `SplitRenderElements` are definitely an improvement as is the `ElementFilter`, even if we should definitely investigate minimizing allocations at some point.

src/backend/render/mod.rs
ids1024: Yeah, it doesn't make a difference with the animations disabled, though it wouldn't hurt to add another if condition for that.

ids1024:
> even if we should definitely investigate minimizing allocations at some point.

Yeah. Can't really use iterators currently, since something like `background_layer_elements` would error due to having to call `split_layer_elements` twice with an `&mut R`. And it would generally make things a mess... Maybe internal iteration where everything including `AsRenderElements` takes a callback to call with each element could work.

Need to come up with a good benchmark for this.

src/shell/mod.rs
Drakulix: Conclusion from mattermost: We should disabled gestures while the workspace overview is open.

src/input/mod.rs
```diff
@@ -1000,7 +1000,7 @@ impl State {
                     .cloned();
                 if let Some(seat) = maybe_seat {
                     self.common.idle_notifier_state.notify_activity(&seat);
-                    if event.fingers() >= 3 {
+                    if event.fingers() >= 3 && workspace_overview_is_open(&seat.active_output()) {
```
Drakulix, line 1003: Isn't that reversed? If we have a gesture with three or more fingers, we handle it in the compositor. So when the workspace overview is **not** open, we want to skip this behaviour and send the gesture to the focused application. (The overview.)

ids1024: Ugh, I actually had fix this already when I tested it but hadn't amended the commit.

Drakulix (approved):
LGTM now :)

PR #611, opened 2024-07-14
support CARGO_TARGET_DIR

dengelt:
Previously failed to build from source when CARGO_TARGET_DIR is set, because the Makefile assumed the target dir was './target'.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ libdir = $(prefix)/lib
 sharedir = $(prefix)/share
 
 BINARY = cosmic-comp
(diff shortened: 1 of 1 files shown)


dengelt:
I noticed there is another PR that replaces the Makefile with a justfile, but that one does not support CARGO_TARGET_DIR yet. When that is fixed and the PR gets merged, this PR would of course be obsolete and could be closed.

Drakulix (approved):

Drakulix:
Thanks!

PR #643, opened 2024-07-22
fix: only suppress keys when an action is on press

### LINKED ISSUE #640: Super key to activate the launcher often requires two presses

wash2:
Actions which are on release don't have anything to suppress, and end up suppressing the next release. fixes #640 . I'm not the most familiar with the input code, but I don't think that this fix should break anything.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -464,7 +464,10 @@ impl State {
                                                 ) || modifiers_bypass
                                             {
                                                 modifiers_queue.clear();
-                                                seat.supressed_keys().add(&handle, None);
+                                                // only suppress if the action is on Press
+                                                if !modifiers_bypass {
+                                                    seat.supressed_keys().add(&handle, None);
+                                                }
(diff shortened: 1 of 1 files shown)


ids1024:
It does seems https://github.com/pop-os/cosmic-comp/commit/4b2082f72a1902914e8cd09820456f2ce851cc33 caused https://github.com/pop-os/cosmic-comp/issues/640, and this change fixes it.

I was going to say this looks correct, but with a bit more testing it seems suppression still isn't working quite right. Testing with `wev` shows a key press event for `Super_L`, modifiers, then a leave event when the launcher opens, since that gets keyboard focus.

The last part can be changed by remapping `super` to something that doesn't spawn a window. Then we see the client is getting the key press and modifiers when the super key is pressed, but not getting a key release and modifiers when it is released. The asymmetry clearly isn't correct. The client then incorrectly thinks the key is being held down.

I guess the `Super_L` key press needs to be suppressed?

Drakulix (approved):
LGTM

PR #647, opened 2024-07-23
feat: double click to maximzie

### LINKED ISSUE #333: server-side decorations: double click to maximize window
Here's the related client-side issue on libcosmic: https://github.com/pop-os/libcosmic/issues/323

It'd be nice to be able to double click the title bar in server side decorations to maximize the window.

leb-kuchen:
should fix #333 
This updates libcosmic to https://github.com/pop-os/libcosmic/commit/8e67f3ce1c0e953c444a2ec43e70a78abcf17400.
Text has a new field wrap.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -912,7 +912,7 @@ dependencies = [
 [[package]]
 name = "cosmic-config"
 version = "0.1.0"
-source = "git+https://github.com/pop-os/libcosmic/#928b714d14c4973a8c9467370864f8adc1c69d7c"
+source = "git+https://github.com/pop-os/libcosmic/#a5996b4e90f6aad943b7c61b961fae5edacd7697"
 dependencies = [
  "atomicwrites",
  "calloop 0.14.0",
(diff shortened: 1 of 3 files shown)


Drakulix (commented):
while this does look good code-wise, have you confirmed this works well not only with server-side decorations, but the other ways in which we use libcosmic in `cosmic-comp`?

Please make sure there are no obvious visual issues with the window header context menu (shown on right-click for either csd or ssd windows, stacks, resize indicators (shown on Super+R), swap indicators (shown on Super+X, when moving to a different window), stack indicators (shown when dragging a window on top of another window in tiling mode).

In any case thanks for working on this!

leb-kuchen:
I have not noticed any obvious regressions, but I have not used this patch on a daily basis, so I can only give you information from a few minutes of testings.

Drakulix:
> I have not noticed any obvious regressions, but I have not used this patch on a daily basis, so I can only give you information from a few minutes of testings.

That is fine. I'll do some further testing.

PR #656, opened 2024-07-26
Fix build with `debug` and `profile-with-tracy`; add CI test

ids1024:
Having a CI test should make it easy to catch this sort of thing.

Lets see if CI passes here...


ids1024:
Hm, complaining about not finding `libseat.pc` after installing `libseat-dev`?

Drakulix (approved):

PR #657, opened 2024-07-27
Update cosmic_comp.ftl PL translation

VandaLHJ:
The idea behind sticky don't work well in my language and is probably misleading about the functionality. I opted for zakrywające okna, because it's always on top, but maybe someone will find better allegory for that kind of functionality later on. Also, stack isn't really good to translate, i opted for karta(tab), because it's what the stack is. It might be not the best translation, but i feel it's more natural.


resources/i18n/pl/cosmic_comp.ftl
```diff
@@ -1,20 +1,20 @@
...
 window-menu-screenshot = Zrób zrzut ekranu
 window-menu-move = Przesuń
 window-menu-resize = Zmień rozmiar
 window-menu-move-prev-workspace = Przenieś do poprzedniego obszaru roboczego
 window-menu-move-next-workspace = Przenieś do następnego obszaru roboczego
-window-menu-stack = Włącz zakrywanie okien
-window-menu-unstack-all = Usuń zakrywanie okien
-window-menu-unstack = Usuń zakrycie okna
-window-menu-sticky = Przylepne okno
+window-menu-stack = Stwórz okna w kartach
```
Kartonrealista, line 14: Zgrupuj okna w karty

resources/i18n/pl/cosmic_comp.ftl
```diff
@@ -1,20 +1,20 @@
...
+window-menu-unstack-all = Rozdziel okna w kartach
```
Kartonrealista, line 15: Rozgrupuj okna

resources/i18n/pl/cosmic_comp.ftl
```diff
@@ -1,20 +1,20 @@
...
+window-menu-unstack = Rozdziel okno od kart
```
Kartonrealista, line 16: Oddziel okno od grupy

Kartonrealista: alt: Wydziel okno z grupy

resources/i18n/pl/cosmic_comp.ftl
VandaLHJ: Yep, feels better than my fast pick.

Kartonrealista:
I've made so suggestions based on terminology used in Polish versions of graphics and spreadsheet programs, where joining separate entities is mostly referred to as grouping. I don't think "Stwórz" sounds right, you're not creating something new, and a more typical word in computing is "Utwórz".

resources/i18n/pl/cosmic_comp.ftl
VandaLHJ: I'm not too sure about this one, perhaps Rozdziel grupę okien would feel better?

resources/i18n/pl/cosmic_comp.ftl
VandaLHJ: Yep, alt suggestion feels imo bit better.

resources/i18n/pl/cosmic_comp.ftl
Kartonrealista: Sure

Drakulix (approved):
Thanks!

PR #658, opened 2024-07-29
input: Send modifiers of modifier-only bindings to clients

Drakulix:
I feel like this is the better approach rather then pretending the modifier event never happened. Potentially fixes missed inputs and cleans up this method a bit.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -386,6 +386,8 @@ impl State {
                                             )));
                                         }
                                     }
+                                    
+                                    std::mem::drop(shell);
 
                                     // cancel grabs
                                     if is_grabbed
@@ -435,8 +437,7 @@ impl State {
                                     }
 
                                     // handle the rest of the global shortcuts
-                                    let mut intercepted = false;
-                                    let mut can_clear_modifiers_shortcut = true;
+                                    let mut clear_queue = true;
                                     if !shortcuts_inhibited {
                                         let modifiers_queue = seat.modifiers_shortcut_queue();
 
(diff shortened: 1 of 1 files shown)


ids1024:
Testing on Gnome, it seems to not send the super key press event. But then if it's followed with another key that doesn't have a binding, it sends `modifiers` first.

...but it doesn't actually end up sending the key press event for `Super_L`, while it does send the key release event for it, so Gnome doesn't seem to be doing this correctly either (based on the version in 24.04).

Overall I don't know it matters much whether or not we suppress the modifier, as long as make it behave consistently. (Send both key press and release, or neither; probably send a modifiers change if and only if the key press/down are sent.)

Drakulix:
> Testing on Gnome, it seems to not send the super key press event. But then if it's followed with another key that doesn't have a binding, it sends `modifiers` first.
> 
> ...but it doesn't actually end up sending the key press event for `Super_L`, while it does send the key release event for it, so Gnome doesn't seem to be doing this correctly either (based on the version in 24.04).

That is exactly what I wanted to avoid with this.

Sending the pressed event later would also work, but feels much more annoying to track and to get right (as we have to figure out the correct keycode), which is precisely why I opted for this as most application will probably not react to a modifier being pressed anyway. And for those that do, I can see how filtering that might even be more confusing to users.

ids1024:
Yeah. Deferring sending the modifier would probably be more confusing to users if the modifier impacts an application in some way. So it doesn't really sound better in practice even if it theoretically makes sense.

ids1024 (approved):
I haven't confirmed any issues that this may fix in clients, but the changes here look good, and the behavior seems correct based on `wev`.

PR #663, opened 2024-07-30
Update Turkish translation

Beryesa:
Fix little mistakes and bring it closer to the established translations by the GNOME and KDE Turkish teams for consistency.


Drakulix (approved):
Thanks!

Issue #429, opened 2024-04-17
Shortcut: Hold Super+ Right Click to resize window


Quackdoc:
As an appendage to this https://github.com/pop-os/cosmic-comp/issues/384 It would be really nice right click would resize the window, and left click would move the window. It would also be great if the button combination for this could be configured.

Drakulix:
I am not super familiar with this feature on other desktops. Which corner does this usually resize? bottom-right?

Quackdoc:
It is most typically dependent on which quadrant is grabbed in.

bartv42:
See also https://github.com/pop-os/cosmic-epoch/issues/363

russelltg:
Definitely agree with this, use this regularly on sway/hyprland/kde.

Drakulix:
This will be part of alpha 2, if you want to express support for feature requests, please simply use the thumbs-up-reaction.

PR #664, opened 2024-07-30
Resize window with Super + Right-Click Drag

Koranir:
Relevant issue: #429 

This PR adds the ability to resize windows with super + right-click, based on the quadrant that the mouse is in at the time of the initial click.

This modifies the behaviour of the super + click to move window feature to only work with left clicks, as that previously worked with any sort of click, including middle or right clicks (which is probably not the intended behaviour).

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -386,7 +386,7 @@ impl State {
                                             )));
                                         }
                                     }
-                                    
+
                                     std::mem::drop(shell);
 
                                     // cancel grabs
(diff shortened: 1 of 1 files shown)


src/input/mod.rs
```diff
@@ -850,42 +850,97 @@ impl State {
...
+                                                                            (true, false) => ResizeEdge::BOTTOM_LEFT,
+                                                                            (false, false) => ResizeEdge::BOTTOM_RIGHT
+                                                                        };
+                                                                        shell.resize_request(
+                                                                            &surface,
+                                                                            &seat_clone,
+                                                                            serial,
+                                                                            edge
+                                                                        )
+                                                                    })();
```
Drakulix, line 916: While I certainly see the appeal of the `?` operator, I don't think we need this rather ugly closure here. `let`-`else` and `and_then` work perfectly fine here, so I would rather prefer something like this:

```suggestion
                                                                    let Some(target_elem) = shell.element_for_surface(&surface) else { return };
                                                                    let Some(geom) = shell
                                                                            .space_for(target_elem).and_then(|space| space.element_geometry(target_elem).to_f64()) else { return };
                                                                    let center = geom.loc + geom.size.downscale(2.0);
                                                                    let offset = center.to_global(&output) - pos;
                                                                    let edge = match (offset.x > 0.0, offset.y > 0.0) {
                                                                        (true, true) => ResizeEdge::TOP_LEFT,
                                                                        (false, true) => ResizeEdge::TOP_RIGHT,
                                                                        (true, false) => ResizeEdge::BOTTOM_LEFT,
                                                                        (false, false) => ResizeEdge::BOTTOM_RIGHT
                                                                    };
                                                                    let res = shell.resize_request(
                                                                        &surface,
                                                                        &seat_clone,
                                                                        serial,
                                                                        edge
                                                                    );
```

Drakulix (commented):
Thanks for picking this up! Looks pretty good apart from some minor nitpicks.

Quackdoc:
If a app window has another window above it, you cannot do a horizontal drag in the top quandrants, likewise if there is a window below it, a bottom quadrant drag does not work.

if you have applications surrounding an app on all sides, horizontal adjustment seems impossible.

Koranir:
> If a app window has another window above it, you cannot do a horizontal drag in the top quandrants, likewise if there is a window below it, a bottom quadrant drag does not work.
> 
> if you have applications surrounding an app on all sides, horizontal adjustment seems impossible.

Is this in the tiling mode? I'm not sure how to reproduce this.

On a maybe-related note, it looks like dragging tiling windows only works on one axis at a time (this also applies to dragging directly with the cursor on the corner).

Quackdoc:
Correct, it is in tiling mode, also can verify the exact same behavior when clicking the corner, it only effecting vertical resizing.

Koranir:
Yep, I can confirm now. That is a separate issue, which affects behaviour before this PR as well.

Drakulix (approved):
Thanks! The code looks much more concise and cleaner now.

Drakulix:
> Yep, I can confirm now. That is a separate issue, which affects behaviour before this PR as well.

Yes, this is a limitation of the `ResizeForkGrab`, which really needs to track two nodes to be able to handle corner resizing correctly. I don't think that is a deal-breaker for merging this though.

PR #665, opened 2024-07-31
chore(config): enable tap-to-click and disable-while-typing by default for touchpads

mmstick:
Closes https://github.com/pop-os/iso/issues/332


Drakulix (approved):
LGTM

PR #668, opened 2024-07-31
toplevel-info: Fix race between binding `wl_ouput` and `output_enter`

ids1024:
This fixes an issue with `cosmic-panel` where, when a workspace is moved back to an output after a monitor is disconnected and reconnected, the panel doesn't hide because `cosmic-panel` thinks no toplevel is open on that monitor.

After some testing, it seems `output_enter` isn't being sent here. In particular, the `output_leave` call happens before the client binds the `wl_output`, so there is no `wl_output` to send in an event yet.

This is addressed by keeping track of a set of `wl_output`s that we have sent the event to. So if an output is bound, `refresh` can add it to this list and send the event.

This is not needed for workspaces (though it could be done similarly) since the handle objects are created by server events. So no race should occur as long as the workspaces global is bound before the toplevel info one.

### DIFF
--- a/src/wayland/protocols/toplevel_info.rs
+++ b/src/wayland/protocols/toplevel_info.rs
@@ -1,12 +1,12 @@
 // SPDX-License-Identifier: GPL-3.0-only
 
-use std::sync::Mutex;
+use std::{collections::HashSet, sync::Mutex};
 
 use smithay::{
     output::Output,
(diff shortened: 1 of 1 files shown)


wash2 (approved):

Drakulix (approved):
Thanks for tracking this down!

PR #682, opened 2024-08-05
Make `RenderElement` implementations generic to reduce duplication

ids1024:
It would make sense to have a bound like `for<'frame> R::Frame<'frame>: AsGlowFrame<'frame>`. But that appears to not behave properly due to current limitations of the borrow checker: https://blog.rust-lang.org/2022/10/28/gats-stabilization.html#implied-static-requirement-from-higher-ranked-trait-bounds

Instead, this makes `glow_frame` and `glow_frame_mut` associated functions of the `AsGlowRenderer` trait. Then it is pretty straightforward to make the `RenderElement` implementations generic using that and `FromGlesError`.

It would make sense to make `Self::Error: FromGlessError` a requirement of the `AsGlowRenderer` trait, but due to the lack of implied bounds support, that produces a bunch of errors about missing bounds. If Rustc improves that eventually, some bounds could be cleaned up a bit: https://github.com/rust-lang/rust/issues/44491

### DIFF
--- a/src/backend/render/element.rs
+++ b/src/backend/render/element.rs
@@ -11,12 +11,12 @@ use smithay::{
         gles::{GlesError, GlesTexture},
         glow::{GlowFrame, GlowRenderer},
         utils::{CommitCounter, DamageSet, OpaqueRegions},
-        Frame, ImportAll, ImportMem, Renderer,
+        ImportAll, ImportMem, Renderer,
     },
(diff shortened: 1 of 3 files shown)


Drakulix (approved):
> Instead, this makes glow_frame and glow_frame_mut associated functions of the AsGlowRenderer trait. Then it is pretty straightforward to make the RenderElement implementations generic using that and FromGlesError.

Oh neat, that is such a good, although a little clunky, solution to this problem!

PR #685, opened 2024-08-05
Use `EGL_MESA_device_software` as fallback for screen capture

ids1024:
The first commit abstracts things a bit to reduce duplication, with a slight negative diff. So we can add this variant to the kms backend without complicating things much.

This should help for testing in VMs without hardware acceleration, until pixman is fully integrated. `cosmic-workspaces` and `xdg-desktop-portal-cosmic` will need to be updated to not require the dmabuf global.

There are still some cases where panics can occur, but those should be a subset of existing panics. With pixman we should always have a working renderer.

### DIFF
--- a/src/wayland/handlers/screencopy/mod.rs
+++ b/src/wayland/handlers/screencopy/mod.rs
@@ -351,21 +351,10 @@ fn constraints_for_output(output: &Output, backend: &mut BackendData) -> Option<
         }
     };
 
-    let mut _kms_renderer = None;
-    let renderer = match backend {
-        BackendData::Kms(ref mut kms) => {
-            let node = kms
-                .target_node_for_output(&output)
-                .unwrap_or(kms.primary_node.expect("No Software Rendering"));
-            _kms_renderer = Some(kms.api.single_renderer(&node).unwrap());
-            _kms_renderer.as_mut().unwrap().as_mut()
-        }
-        BackendData::Winit(ref mut winit) => winit.backend.renderer(),
-        BackendData::X11(ref mut x11) => &mut x11.renderer,
-        _ => unreachable!(),
-    };
-
-    Some(constraints_for_renderer(mode, renderer))
+    let mut renderer = backend
+        .offscreen_renderer(|kms| kms.target_node_for_output(&output).or(kms.primary_node))
+        .unwrap();
+    Some(constraints_for_renderer(mode, renderer.as_mut()))
 }
 
 fn constraints_for_toplevel(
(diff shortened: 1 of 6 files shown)


src/wayland/handlers/screencopy/mod.rs
```diff
@@ -375,26 +364,20 @@ fn constraints_for_toplevel(
...
 
-    let mut _kms_renderer = None;
-    let renderer = match backend {
-        BackendData::Kms(ref mut kms) => {
+    let mut renderer = backend
+        .offscreen_renderer(|kms| {
             let dma_node = with_renderer_surface_state(&wl_surface, |state| {
                 let buffer = state.buffer()?;
                 let dmabuf = get_dmabuf(&*buffer).ok()?;
                 dmabuf.node()
```
Drakulix, line 372: Hmm, not your fault obviously, but it seems like we have a small inconsistency here.
The built-in screenshot code uses `advertised_node_for_client`, while here we ask the surface buffer for the node. Not sure whats better honestly.

src/wayland/handlers/screencopy/render.rs
```diff
@@ -326,39 +326,66 @@ pub fn render_workspace_to_buffer(
...
-                        .unwrap()
-                }
-                _ => None,
-            };
-            let mut multirenderer = match kms.api.renderer(
-                &render_node,
-                &target_node,
-                buffer_format.unwrap_or(Fourcc::Abgr8888),
+    let renderer = match state.backend.offscreen_renderer(|kms| {
+        let render_node = kms.target_node_for_output(&output).or(kms.primary_node)?;
```
Drakulix, line 330: Looking at this, I wonder if this line is responsible for the flickering we sometimes see in the overview mode for multi-gpu systems. The target node is the native one for direct scanout, but not necessarily a good render_node for offscreen rendering.

(This is more a note, than something actionable.)

Drakulix (approved):
Looks good to me.

src/wayland/handlers/screencopy/mod.rs
ids1024: Yeah. I tried to keep the behavior unchanged here, but it's a little unclear if the nodes it is attempt to use everywhere are ideal.

src/wayland/handlers/screencopy/render.rs
ids1024: It could make sense just to use a `single_renderer` here, with the target node of the buffer.

Or I guess it would make sense if it used the same logic as deciding what GPU to render on screen, but targeting the buffer node instead of the output node?

Though all of this is suboptimal without https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/268, so we can be sure we actually know what node the capture buffer, and the buffers we need to composite, are actually on.

Issue #694, opened 2024-08-07
NVIDIA desktop suspend/resume failure


jacobgkau:
On a Thelio Mira B4 with an RTX 4060 Ti, the system suspends, but hangs at a black screen on resume.

Log file: [out.txt](https://github.com/user-attachments/files/16533501/out.txt)

Outtake of note:

```
Aug 07 19:50:54 pop-os cosmic-comp[2004]: Failed to destroy old mode property blob: No such file or directory (os error 2)
Aug 07 19:50:54 pop-os cosmic-comp[2004]: thread 'surface-HDMI-A-3' panicked at 'internal error: entered unreachable code': src/backend/kms/surface/mod.rs:726
                                             0: <unknown>
                                             1: <unknown>
                                             2: <unknown>
                                             3: <unknown>
                                             4: <unknown>
                                             5: <unknown>
                                             6: <unknown>
                                             7: <unknown>
                                             8: <unknown>
                                             9: <unknown>
                                            10: <unknown>
                                            11: <unknown>
                                            12: <unknown>
                                            13: <unknown>
                                            14: <unknown>
                                            15: <unknown>
```

StayBlue:
I believe this is an issue with Nvidia itself, not Cosmic. You can see the forum post I made on the [Nvidia forum](https://forums.developer.nvidia.com/t/black-screen-after-pc-wakes-up-from-suspend/304023) if you would like, although it doesn't provide any logs. Other people on the forum have reported similar issues with other desktop environments as well.

TigerGorilla2:
I have the same issue (surface-<monitor> threads panicking) with a GTX 1050 Ti on  fedora 42, nvidia driver 580.82.09, and cosmic-comp-1.0.0~beta.1-1.fc42.src.rpm. It happens when the monitor is plugged into the gpu or plugged into the motherboard (using Ryzen 5600G) with the nvidia drivers loaded. It does not happen when i `rmmod` the nvidia driver. Using the patches from PR #699 ([a rebased version onto beta1.1](https://github.com/TigerGorilla2/cosmic-comp-fix-resume/tree/fix-resume)) I dont have resume issues anymore.

jacobgkau:
[...]

PR #699, opened 2024-08-08
surface: Don't crash due to weird state after resume

Drakulix:
@jacobgkau Should fix resume issues from https://github.com/pop-os/cosmic-comp/issues/694 (unable to confirm myself)

### DIFF
--- a/src/backend/kms/surface/mod.rs
+++ b/src/backend/kms/surface/mod.rs
@@ -673,8 +673,27 @@ impl SurfaceThreadState {
         let Some(compositor) = self.compositor.as_mut() else {
             return;
         };
+
+        // handle edge-cases right after resume
+        if !matches!(
+            self.state,
+            QueueState::WaitingForVBlank { .. } | QueueState::Idle
+        ) {
+            match mem::replace(&mut self.state, QueueState::Idle) {
+                QueueState::WaitingForVBlank { .. } | QueueState::Idle => unreachable!(),
+                QueueState::Queued(token) | QueueState::WaitingForEstimatedVBlank(token) => {
+                    self.loop_handle.remove(token);
+                }
+                QueueState::WaitingForEstimatedVBlankAndQueued {
+                    estimated_vblank,
+                    queued_render,
+                } => {
+                    self.loop_handle.remove(estimated_vblank);
+                    self.loop_handle.remove(queued_render);
+                }
+            }
+        }
         if matches!(self.state, QueueState::Idle) {
-            // can happen right after resume
             return;
         }
 


jacobgkau (approved):
With this installed on the Mira B4 w/ 4060 Ti, resume still isn't working correctly.

The first change of behavior is I now get a mouse cursor on top of an otherwise blank screen after resuming; the cursor doesn't move around smoothly, but if I move the mouse, the cursor will pop to a new location a few seconds later.

With this fix, I am able to get to a TTY after the resume (which I wasn't able to do before). If I restart `cosmic-greeter.service`, then I can log back into the GUI successfully.

Here's the system journal after the resume: [after-resume-d580294.txt](https://github.com/user-attachments/files/16566876/after-resume-d580294.txt)

I see these logs after suspend is started but before it's finished suspending:

```
Aug 09 20:54:51 pop-os cosmic-comp[2028]: Failed to submit rendering: Failed to submit result for display
                                          
                                          Caused by:
                                              0: The underlying drm surface encountered an error: DRM access error: Page flip commit failed on device `Some("/dev/dri/card1")` (Permission denied (os error 13))
                                              1: DRM access error: Page flip commit failed on device `Some("/dev/dri/card1")` (Permission denied (os error 13))
                                              2: Permission denied (os error 13)
```

After resume, I see this line repeated a lot in one place:

```
Aug 09 20:55:42 pop-os cosmic-session[1958]: [2024-08-09T20:55:42Z WARN  iced_futures::subscription::tracker] Error sending event to subscription: TrySendError { kind: Full }
```

And then one of these just before I switch to a TTY:

```
Aug 09 20:56:28 pop-os cosmic-comp[2028]: Failed to submit rendering: Failed to submit result for display
                                          
                                          Caused by:
                                              0: The underlying drm surface encountered an error: Device is currently paused, operation rejected
                                              1: Device is currently paused, operation rejected
```

I'm approving this since it's an improvement in behavior, but it seems more work is still required, possibly in cosmic-comp.

Drakulix:
Pushed a commit to address the last log message.

@jacobgkau If you have the time to retest, that would be very appreciated. The system journal would again help with debugging.

jacobgkau:
As stated in https://github.com/pop-os/cosmic-comp/issues/694#issuecomment-3458837981, the problem is no longer reproducible with lab hardware that matches what I reported last year, so we may need to just do regression testing on this.

jacobgkau:
Tried to rebase, referencing the original diff & [a user's updated patchset](https://github.com/pop-os/cosmic-comp/compare/master...TigerGorilla2:cosmic-comp-fix-resume:beta-4-with-fix).

Edit: Squashed the last build fix (cherry-picked from 1fd9544 in that fork) into the commit it was fixing since its commit message wasn't great.

src/backend/kms/surface/mod.rs
```diff
@@ -683,7 +689,8 @@ impl SurfaceThreadState {
             }
         };
 
-        let _ = tx.send(());
+        let _ = tx.send(Ok(()));
```
Drakulix: It seems to me like this can only ever return `Ok(())`. Given we don't handle the result in `pause_session` (which is bad practice, should this ever return an error), I would rather keep the sender/receive, etc sending a `()`, like it did before.

So the only thing that needs to change in this commit imo, is the reordering for `device.drm.pause()`. The rest can be dropped.

jacobgkau: Like this? c6002c050b822d9138647e6f83caad741f23f8b4

Drakulix: yes, perfect.

wash2 (approved):

PR #721, opened 2024-08-13
Add wl-clipboard to tiling exceptions

### LINKED ISSUE #700: `wl-paste` and `wl-copy` cause a tiling animation to happen as if they created a new window even though they don't
With `wl-paste` there  is a full animation:

https://github.com/user-attachments/assets/1ab407e3-adf5-4159-9a97-caa34f77253a

With `wl-copy` there is a very fast flicker:

https://github.com/user-attachments/assets/87577c83-c0dd-4aa2-ba43-eabb6b74f834

Tebro:
This prevents the flickering of windows when using wl-clipboard in certain editors (Helix, neovim)

Fixes #700

### DIFF
--- a/src/shell/layout/mod.rs
+++ b/src/shell/layout/mod.rs
@@ -47,6 +47,7 @@ lazy_static::lazy_static! {
         r"tilda",
         r"zoom",
         r"^.*?action=join.*$",
+        r"",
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
Thanks

PR #726, opened 2024-08-14
Move tiling exceptions to configuration file

Tebro:
This moves out the hard coded tiling exceptions into a new separate config file under CosmicComp

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -926,7 +926,7 @@ dependencies = [
 [[package]]
 name = "cosmic-settings-config"
 version = "0.1.0"
-source = "git+https://github.com/pop-os/cosmic-settings-daemon#362c77f9faaeb7f1b9e4aa79a7d5588001f04874"
+source = "git+https://github.com/pop-os/cosmic-settings-daemon#1ed68808e85ce681da882446ec572d44c68a6866"
 dependencies = [
  "cosmic-config",
  "serde",
--- a/src/shell/layout/mod.rs
+++ b/src/shell/layout/mod.rs
@@ -61,7 +61,10 @@ pub struct TilingExceptions {
 }
 
 impl TilingExceptions {
-    pub fn new(exceptions_config: &Vec<ApplicationException>) -> Self {
+    pub fn new<'a, I>(exceptions_config: I) -> Self
+    where
+      I: Iterator<Item=&'a ApplicationException>
+    {
         let mut app_ids = Vec::new();
         let mut titles = Vec::new();
 
--- a/src/shell/layout/mod.rs
+++ b/src/shell/layout/mod.rs
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: GPL-3.0-only
 
-use cosmic_settings_config::shortcuts::action::Orientation;
+use cosmic_settings_config::{shortcuts::action::Orientation, window_rules::ApplicationException};
 use regex::{Regex, RegexSet};
 use smithay::{
     desktop::WindowSurface,
@@ -8,8 +8,6 @@ use smithay::{
     xwayland::xwm::WmWindowType,
 };
 use tracing::warn;
-
-use crate::config::Config;
 
 use super::CosmicSurface;
 
@@ -63,11 +61,11 @@ pub struct TilingExceptions {
 }
 
 impl TilingExceptions {
-    pub fn new(config: &Config) -> Self {
+    pub fn new(exceptions_config: &Vec<ApplicationException>) -> Self {
         let mut app_ids = Vec::new();
         let mut titles = Vec::new();
 
-        for exception in &config.tiling_exceptions {
+        for exception in exceptions_config {
             if let Err(e) = Regex::new(&exception.appid) {
                 warn!("Invalid regex for appid: {}, {}", exception.appid, e);
                 continue;
--- a/src/shell/layout/mod.rs
+++ b/src/shell/layout/mod.rs
@@ -1,12 +1,13 @@
 // SPDX-License-Identifier: GPL-3.0-only
 
 use cosmic_settings_config::shortcuts::action::Orientation;
-use regex::RegexSet;
+use regex::{Regex, RegexSet};
 use smithay::{
     desktop::WindowSurface,
     wayland::{compositor::with_states, shell::xdg::XdgToplevelSurfaceData},
     xwayland::xwm::WmWindowType,
 };
+use tracing::warn;
 
 use crate::config::Config;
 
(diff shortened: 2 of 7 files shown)


src/shell/layout/mod.rs
```diff
@@ -126,10 +55,35 @@ pub fn is_dialog(window: &CosmicSurface) -> bool {
...
+        for app in &config.cosmic_conf.tiling_exceptions {
+            for title in &app.titles {
+                app_ids.push(app.appid.clone());
+                titles.push(title.clone());
+            }
+        }
+
+        Self {
+            app_ids: RegexSet::new(app_ids).unwrap(),
+            titles: RegexSet::new(titles).unwrap(),
```
Tebro, line 78: Can this fail? Or is it safe to use `unwrap` here?

Tebro: After some more consideration I am pretty sure these would be good to check. The question is how to handle a potential failure.

wiiznokes: Maybe return an error in the TilingExceptions::new function, and log it when you call it.

Tebro: Let's start with that.

wiiznokes: I don't have a full picture of the code, but you should also make sure the log don't spam to much

Tebro: All of this happens in Shell init so it logs early and only once per start.

Drakulix:
Sorry, but we probably won't do this as one single config file.

Reason being that we want to ship the default rules and also be able to update them. So we likely want a new config file for rules added by users **and** default rules to be explicitly disabled.

Externalizing the default rules into a separate config file still makes sense, given we want to allow distros overwriting them.

I don't have the time to do a thorough review on your work now and will try to do so next week. In any case thanks for working on this!

Tebro:
@Drakulix is there something I can do for this at this time?

Drakulix:
> @Drakulix is there something I can do for this at this time?

Yes, so looking at this further, what needs to be done to move this PR along is split the config file into two. One for the defaults and one for user-overrides (so either disabling existing rules from the defaults file or adding additional ones), because we want distros to provide the former and be able to update them without disrupting user configuration.

For this you can take inspiration from our shortcuts-config, which does the same (but is located in the `cosmic-settings-daemon`-crate: https://github.com/pop-os/cosmic-settings-daemon/blob/master/config/src/shortcuts/mod.rs#L53).

That also adds a new config namespace (`com.system76.CosmicSettings.Shortcuts`), we probably want the same for this, something like `com.system76.CosmicSettings.WindowRules` (anticipating, that we might have different rules in the future) and then `tiling-exception-defaults` and `tiling-exception-custom` for the files.

Given these should be configurable from `cosmic-settings` and `cosmic-applet-tiling` in the future, it might also make sense to move the config into the `cosmic-settings-daemon` repo (splitting this feature into two PRs).

The defaults file could still be installed by cosmic-comp, like the shortcuts file.
I don't have a problem with the whole config format (at least for the defaults file, the custom file would need to be slightly different to allow disabling existing rules), but I would prefer, if we didn't skip the whole config, if one regex fails to parse. I would appreciate it, if the deserialization code could skip over broken rules and just log a warning.

I understand, that this is a far bigger tasks, than you probably imagined opening this PR, so please don't see this as a demand. If you don't feel like working on this, we will eventually implement said feature. That said your effort and contribution is very appreciated.

Tebro:
Thanks for the response. Will try to find some time to look into all of this. It would be nice to finish what was started.

Tebro:
This is now updated, needs the PR in cosmic-settings-daemon to be merged first for the checks to pass.

Drakulix (commented):
This looks great!

I think there is just one thing missing, it doesn't appear to reload the config file at runtime. To do that you need to add some code to watch the config file, re-assemble the exceptions again and update the data in `Shell`.

You can see the code doing that for the shortcuts just below your new code in `Config::load` (config.rs).

Tebro:
Got that working now.

src/shell/layout/mod.rs
```diff
@@ -126,10 +54,42 @@ pub fn is_dialog(window: &CosmicSurface) -> bool {
...
 
-pub fn has_floating_exception(window: &CosmicSurface) -> bool {
+#[derive(Debug, Clone, Default)]
+pub struct TilingExceptions {
+    app_ids: RegexSet,
+    titles: RegexSet,
+}
+
+impl TilingExceptions {
+    pub fn new(exceptions_config: &Vec<ApplicationException>) -> Self {
```
Drakulix, line 64: `&Vec<_>` is a weird type for a function argument. It forces the caller to allocate specifically a `Vec`, while you only use it for iterating.

So I suggest something like this instead:
```suggestion
    pub fn new<'a, I>(exceptions_config: I) -> Self
    where
      I: Iterator<Item=&'a ApplicationException>
    {
```

src/shell/mod.rs
```diff
@@ -1218,6 +1220,8 @@ impl Shell {
     pub fn new(config: &Config) -> Self {
         let theme = cosmic::theme::system_preference();
 
+        let tiling_exceptions = layout::TilingExceptions::new(&config.tiling_exceptions);
```
Drakulix, line 1223:
```suggestion
        let tiling_exceptions = layout::TilingExceptions::new(config.tiling_exceptions.iter());
```

src/shell/mod.rs
```diff
@@ -3365,6 +3370,10 @@ impl Shell {
         &self.theme
     }
 
+    pub fn update_tiling_exceptions(&mut self, exceptions: &Vec<ApplicationException>) {
```
Drakulix, line 3373:
```suggestion
    pub fn update_tiling_exceptions<'a, I>(&mut self, exceptions: I)
    where
      I: Iterator<Item=&'a ApplicationException>
    {
```

Drakulix (commented):
Almost, just a small nitpick. Thanks for working on this :)

Drakulix:
https://github.com/pop-os/cosmic-settings-daemon/pull/50 merged, this needs a `cargo update -p cosmic-settings-config`

Cargo.lock
```diff
@@ -926,7 +926,7 @@ dependencies = [
 [[package]]
 name = "cosmic-settings-config"
 version = "0.1.0"
-source = "git+https://github.com/pop-os/cosmic-settings-daemon#362c77f9faaeb7f1b9e4aa79a7d5588001f04874"
+source = "git+https://github.com/pop-os/cosmic-settings-daemon#1ed68808e85ce681da882446ec572d44c68a6866"
```
Drakulix, line 929: > Did the cargo update, some of the version moves in the diff looked confusing, such as libloading 0.8.5 -> 0.7.4. Pushed it up in case it is correct.

Yeah it is weird, I am seeing the same changes locally running that command, but libcosmic definitely doesn't need to be updated for this. (And I very much would prefer it to stay on the current version given I would likely need to re-test a bunch of stuff, if we updated it.)

From my testing locally you can drop every `Cargo.lock` change, except this line and it will work.

Tebro: Alright, that makes sense. forced that up.

PR #737, opened 2024-08-16
Use fallback scale if size is invalid

jackpot51:
This makes it so 200% is used on displays with some axis greater than 1600 pixels and zero size.


Drakulix (approved):
LGTM

PR #741, opened 2024-08-17
DPMS with `wlr-output-power-management-unstable-v1` protocol

ids1024:
This works to set DPMS on or off. It still needs to track the DPMS state and send events to clients when the mode changes. And otherwise be cleaned up a bit.

Components like `cosmic-greeter` may make use of this.

This can be tested with the below code (which also works on Sway):

```rust
use std::{env, process};
use wayland_client::{
    delegate_noop,
    globals::{registry_queue_init, GlobalListContents},
    protocol::{wl_output, wl_registry},
    Connection, Dispatch, Proxy, QueueHandle,
};
use wayland_protocols_wlr::output_power_management::v1::client::{
    zwlr_output_power_manager_v1, zwlr_output_power_v1,
};

struct State;

fn main() {
    let mode = match env::args().skip(1).next().as_ref().map(|x| x.as_str()) {
        Some("on") => zwlr_output_power_v1::Mode::On,
        Some("off") => zwlr_output_power_v1::Mode::Off,
        _ => {
            eprintln!("Usage: dpms on|off");
            process::exit(1);
        }
    };

    let connection = Connection::connect_to_env().unwrap();
    let (globals, mut event_queue) = registry_queue_init::<State>(&connection).unwrap();
    let qh = event_queue.handle();

    let outputs = globals.contents().with_list(|list| {
        list.iter()
            .filter(|global| global.interface == wl_output::WlOutput::interface().name)
            .map(|global| globals.registry().bind(global.name, global.version, &qh, ()))
            .collect::<Vec<wl_output::WlOutput>>()
    });

    let output_power_manager = globals
        .bind::<zwlr_output_power_manager_v1::ZwlrOutputPowerManagerV1, _, _>(&qh, 1..=1, ())
        .unwrap();

    for output in &outputs {
        let output_power = output_power_manager.get_output_power(output, &qh, ());
        output_power.set_mode(mode);
    }
    event_queue.roundtrip(&mut State).unwrap();
}

impl Dispatch<wl_registry::WlRegistry, GlobalListContents> for State {
    fn event(
        _: &mut Self,
        _: &wl_registry::WlRegistry,
        _: wl_registry::Event,
        _: &GlobalListContents,
        _: &Connection,
        _: &QueueHandle<Self>,
    ) {
    }
}

delegate_noop!(State: ignore wl_output::WlOutput);
delegate_noop!(State: zwlr_output_power_manager_v1::ZwlrOutputPowerManagerV1);
delegate_noop!(State: ignore zwlr_output_power_v1::ZwlrOutputPowerV1);

```

### DIFF
--- a/src/backend/kms/surface/mod.rs
+++ b/src/backend/kms/surface/mod.rs
@@ -102,9 +102,9 @@ static NVIDIA_LOGO: &'static [u8] = include_bytes!("../../../../resources/icons/
 
 #[derive(Debug)]
 pub struct Surface {
-    pub(super) connector: connector::Handle,
+    pub(crate) connector: connector::Handle,
     pub(super) crtc: crtc::Handle,
-    pub(super) output: Output,
+    pub(crate) output: Output,
     known_nodes: HashSet<DrmNode>,
 
     active: Arc<AtomicBool>,
@@ -114,6 +114,8 @@ pub struct Surface {
     loop_handle: LoopHandle<'static, State>,
     thread_command: Sender<ThreadCommand>,
     thread_token: RegistrationToken,
+
+    dpms: bool,
 }
 
 pub struct SurfaceThreadState {
@@ -238,6 +240,7 @@ pub enum ThreadCommand {
     ScheduleRender,
     SetMode(Mode, SyncSender<Result<()>>),
     End,
+    DpmsOff,
 }
 
 #[derive(Debug)]
@@ -350,6 +353,7 @@ impl Surface {
             loop_handle: evlh.clone(),
             thread_command: tx,
             thread_token,
+            dpms: true,
         })
     }
 
@@ -380,7 +384,9 @@ impl Surface {
     }
 
     pub fn schedule_render(&self) {
-        let _ = self.thread_command.send(ThreadCommand::ScheduleRender);
+        if self.dpms {
+            let _ = self.thread_command.send(ThreadCommand::ScheduleRender);
+        }
     }
 
     pub fn set_mirroring(&mut self, output: Option<Output>) {
@@ -430,6 +436,21 @@ impl Surface {
         });
 
         rx.recv().context("Surface thread died")?
+    }
+
+    pub fn get_dpms(&mut self) -> bool {
+        self.dpms
+    }
+
+    pub fn set_dpms(&mut self, on: bool) {
+        if self.dpms != on {
+            self.dpms = on;
+            if on {
+                self.schedule_render();
+            } else {
+                let _ = self.thread_command.send(ThreadCommand::DpmsOff);
+            }
+        }
     }
 }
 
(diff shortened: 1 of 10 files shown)


ids1024:
Looking at wlroots/mutter/kwin and DRM documentation, I guess we shouldn't set the `DPMS` property (unless we're using legacy modesetting) and should set `ACTIVE` as part of an atomic modeset.

> DPMS:
>
> Legacy property for setting the power state of the connector. For atomic drivers this is only provided for backwards compatibility with existing drivers, it remaps to controlling the “ACTIVE” property on the CRTC the connector is linked to. Drivers should never set this property directly, it is handled by the DRM core by calling the [drm_connector_funcs.dpms](https://www.kernel.org/doc/html/v4.15/gpu/drm-kms.html#c.drm_connector_funcs) callback. For atomic drivers the remapping to the “ACTIVE” property is implemented in the DRM core. This is the only standard connector property that userspace can change.
>
>  Note that this property cannot be set through the MODE_ATOMIC ioctl, userspace must use “ACTIVE” on the CRTC instead.

src/backend/kms/surface/mod.rs
```diff
@@ -540,6 +561,30 @@ fn surface_thread(
...
+                            state.timings.discard_current_frame()
+                        }
+                        QueueState::WaitingForEstimatedVBlankAndQueued {
+                            estimated_vblank,
+                            queued_render,
+                        } => {
+                            state.loop_handle.remove(estimated_vblank);
+                            state.loop_handle.remove(queued_render);
+                        }
+                    };
```
Drakulix, line 585: hmm, at some point we might want to refactor this into a reset method, but that is fine for now.

Drakulix (commented):
As discussed on mattermost, this needs a way for the compositor to clear the suspension, when the user isn't idling anymore, but otherwise this looks good and quite straight-forward!

ids1024:
Trying to test mirroring here, I ran into a few unrelated issues with cosmic-randr/mirroring (https://github.com/pop-os/cosmic-randr/pull/33, https://github.com/pop-os/cosmic-randr/pull/34).

I think this should be good now, after changing a few things:

- Store the `Mode` that was last sent to each `zwlr_output_power_v1` instance so it can update only when there's a change.
- `set_dpms` now also sets the DPMS state for any outputs mirroring the output
- On any input events (which is when `notify_activity` is called for idle notifier), set DPMS on for all outputs
  * Probably not an issue, but I made sure this is done in an efficient way that avoids multiple loops when there's nothing to change.
- Added an `OutputPowerState:refresh` function, called when it does change the DPMS state on an input event, to send the `mode` to clients.
- Updated  `OutputPowerHandler` to take `Output` instead of `WlOutput`

Drakulix (approved):

PR #748, opened 2024-08-19
Add smart window gaps feature

JCapucho:
This PR implements the smart gaps feature like what is present in i3 and sway https://i3wm.org/docs/userguide.html#gaps

Smart window gaps only add the external and internal gaps when it aids in visually separating the top levels (when using the tiling mode).

This is done by only adding the outer gap when more than one toplevel exists in the output and only adding the inner gap when an adjacent toplevel exists in the given direction.

I don't know if this is a wanted feature but it's absence was really bugging me (I previously used sway), so I implemented it and have been using it so I thought I might as well contribute it upstream :)

## Screenshots

One window (no gaps)
One window

Two windows (outer and inner gaps)
Two windows

## Notes

Requires a change in `libcosmic`, if the feature is wanted I will open the PR (Didn't want to spam two repositories for this feature), the changes are here https://github.com/JCapucho/libcosmic/tree/smart-window-gaps

I also think the inner gap calculations are wrong, currently they are:
```rust
if TilingLayout::has_adjacent_node(tree, &node_id, dir) {
    inner / 2
} else if is_smart_gaps {
    0
} else {
    inner
}
```

I have kept the previous version where if smart gaps isn't enabled and there are no adjacent windows a `inner` gap is added, but this seems wrong because it leads to windows having both an outer and inner gap when they are at the edge of the output. This should instead be:
```rust
if TilingLayout::has_adjacent_node(tree, &node_id, dir) {
    inner / 2
} else {
    0
}
```

I think this hasn't been a problem because by default the outer gap is 0 and there's no way to change it in the settings app, but if it's changed in the file directly it leads to a double border.

## Further work

I have also a similar change to only add the active window hint when there's more than one window in the output, but that one still has some problems.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -885,7 +885,7 @@ dependencies = [
 [[package]]
 name = "cosmic-config"
 version = "0.1.0"
-source = "git+https://github.com/pop-os/libcosmic/#b40839638ab0e1d96de3f817eded647e6952db40"
+source = "git+https://github.com/JCapucho/libcosmic.git?branch=smart-window-gaps#43985bf9eb08d80cdbfe2ec849bc203c98c20875"
 dependencies = [
  "atomicwrites",
  "calloop 0.14.0",
@@ -904,7 +904,7 @@ dependencies = [
 [[package]]
 name = "cosmic-config-derive"
 version = "0.1.0"
-source = "git+https://github.com/pop-os/libcosmic/#b40839638ab0e1d96de3f817eded647e6952db40"
+source = "git+https://github.com/JCapucho/libcosmic.git?branch=smart-window-gaps#43985bf9eb08d80cdbfe2ec849bc203c98c20875"
 dependencies = [
  "quote",
  "syn 1.0.109",
@@ -961,7 +961,7 @@ dependencies = [
 [[package]]
 name = "cosmic-theme"
 version = "0.1.0"
-source = "git+https://github.com/pop-os/libcosmic/#b40839638ab0e1d96de3f817eded647e6952db40"
+source = "git+https://github.com/JCapucho/libcosmic.git?branch=smart-window-gaps#43985bf9eb08d80cdbfe2ec849bc203c98c20875"
 dependencies = [
  "almost",
  "cosmic-config",
@@ -2305,7 +2305,7 @@ dependencies = [
 [[package]]
 name = "iced"
 version = "0.12.0"
-source = "git+https://github.com/pop-os/libcosmic/#b40839638ab0e1d96de3f817eded647e6952db40"
+source = "git+https://github.com/JCapucho/libcosmic.git?branch=smart-window-gaps#43985bf9eb08d80cdbfe2ec849bc203c98c20875"
 dependencies = [
  "dnd",
  "iced_core",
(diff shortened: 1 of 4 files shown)


Drakulix:
First of all, thanks for contribution!

Second, this is a feature purely for cosmic-comp, I don't think this should be stored in the libcosmic-theme. Rather this should be inside cosmic-comp's config, just like the active-border width or gap_size. (Which you can take as inspiration on how to pass this down to the tiling layout code.)

Lastly, I don't know if we want to have that feature. Pinging @pop-os/ux on this one, specifically if we would want to include this inside cosmic-settings. I don't think it would hurt to merge this has a config-file only feature, but that might make it more prone to breakage as I don't know, if we would be testing this frequently. (Then again the tiling code is already complicated enough and rarely touched these days.)

JCapucho:
> Second, this is a feature purely for cosmic-comp, I don't think this should be stored in the libcosmic-theme. Rather this should be inside cosmic-comp's config, just like the active-border width or gap_size. (Which you can take as inspiration on how to pass this down to the tiling layout code.)

That makes sense I looked at the gaps and those were on the theme but I should have looked if there were comp specific settings

maria-komarova:
Thanks for the contribution. 
I am not sure at this point if we want this feature. It seems a little too niche and not necessarily easy to understand unless one is coming from i3 and sway. But if one is new to tiling, then it might be confusing. We show hints for window groups when one is moving a window around because it seems that this is the most useful point when understanding window groups matters. But I'm not sure it is as valuable when one is just working.

JCapucho:
> Thanks for the contribution. I am not sure at this point if we want this feature. It seems a little too niche and not necessarily easy to understand unless one is coming from i3 and sway. But if one is new to tiling, then it might be confusing. We show hints for window groups when one is moving a window around because it seems that this is the most useful point when understanding window groups matters. But I'm not sure it is as valuable when one is just working.

The hints are still there and unchanged I only disabled them by setting the width to 0 in my settings. The difference is in the gaps.

Drakulix:
> I am not sure at this point if we want this feature. It seems a little too niche and not necessarily easy to understand unless one is coming from i3 and sway.

That is certainly the target group for this setting.

> But if one is new to tiling, then it might be confusing. We show hints for window groups when one is moving a window around because it seems that this is the most useful point when understanding window groups matters. But I'm not sure it is as valuable when one is just working.

Adding to the previous comment, this only affects window positions at rest, it should do nothing to the overview when actively moving windows around or swapping them.

As such I think this is a useful feature for people wanting more screen estate and that feel at home with the tiling concepts already, though arguably a bit niche.

maria-komarova:
> Adding to the previous comment, this only affects window positions at rest, it should do nothing to the overview when actively moving windows around or swapping them.

I got it. Just trying to better understand why it is so helpful to see the groups easier when windows are at rest. I can see how it is useful in i3 or sway when you need to move windows and understanding the groups helps one figure out what will happen. But I'm still not sure hoe much value there would be in our case. Muscle memory that people bring with them is understandable but how exactly is it helpful to see the groups of windows at rest in COSMIC is not so clear to me.

Drakulix:
> Muscle memory that people bring with them is understandable but how exactly is it helpful to see the groups of windows at rest in COSMIC is not so clear to me.

Got it. I tend to agree given that our tiling operations aren't based on rotating parts of the tree, which is what makes this information so crucial in i3/sway.

JCapucho:
I rebased everything and moved the configuration to the cosmic-comp's config

JCapucho:
@Drakulix sorry for the ping, but I've rebased again and wanted to know if there has been discussion around accepting this feature upstream.

Drakulix:
> @Drakulix sorry for the ping, but I've rebased again and wanted to know if there has been discussion around accepting this feature upstream.

Sorry no, this simply wasn't very high on our priority list and it likely won't make it for 1.0. I'll try to catch up with community contributions and UX shortly after though.

PR #750, opened 2024-08-19
Fix button events being sent when super is held

### LINKED ISSUE #686: [BUG] Super + Right click passes right through to the application
Holding super to resize, then clicking rightclick passes the rightclick through to the application

JCapucho:
Fixes #686

Previously when doing Super+{Left,Right} Click the compositor would still send the button event to the client, for left click this didn't cause issues most of the time but for right click this caused many issues because it would usually open a context menu which would cause the resize to stop immediately.

This PR fixes the issue by not sending any button events if the logo key is held.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -888,6 +888,7 @@ impl State {
                                                                         &state.common.config,
                                                                         &state.common.event_loop_handle,
                                                                         &state.common.xdg_activation_state,
+                                                                        false
                                                                     );
                                                                     drop(shell);
                                                                     dispatch_grab(res, seat_clone, serial, state);
@@ -910,7 +911,8 @@ impl State {
                                                                         &surface,
                                                                         &seat_clone,
                                                                         serial,
-                                                                        edge
+                                                                        edge,
+                                                                        false
                                                                     );
                                                                     drop(shell);
                                                                     dispatch_grab(res, seat_clone, serial, state);
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -78,6 +78,7 @@ use std::{
     any::Any,
     borrow::Cow,
     cell::RefCell,
+    collections::HashSet,
     os::unix::process::CommandExt,
     thread,
     time::{Duration, Instant},
@@ -87,6 +88,8 @@ pub mod gestures;
 
 #[derive(Default)]
 pub struct SupressedKeys(RefCell<Vec<(Keycode, Option<RegistrationToken>)>>);
+#[derive(Default)]
+pub struct SupressedButtons(RefCell<HashSet<u32>>);
 #[derive(Default, Debug)]
 pub struct ModifiersShortcutQueue(RefCell<Option<shortcuts::Binding>>);
 
@@ -113,6 +116,16 @@ impl SupressedKeys {
                 .flatten()
                 .collect::<Vec<_>>(),
         )
+    }
+}
+
+impl SupressedButtons {
+    fn add(&self, button: u32) {
+        self.0.borrow_mut().insert(button);
+    }
+
+    fn remove(&self, button: u32) -> bool {
+        self.0.borrow_mut().remove(&button)
     }
 }
 
@@ -773,6 +786,7 @@ impl State {
 
                     let serial = SERIAL_COUNTER.next_serial();
                     let button = event.button_code();
+                    let mut pass_event = !seat.supressed_buttons().remove(button);
                     if event.state() == ButtonState::Pressed {
                         // change the keyboard focus unless the pointer is grabbed
                         // We test for any matching surface type here but always use the root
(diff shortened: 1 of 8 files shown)


src/input/mod.rs
```diff
@@ -972,17 +980,19 @@ impl State {
...
-                        self,
-                        &ButtonEvent {
-                            button,
-                            state: event.state(),
-                            serial,
-                            time: event.time_msec(),
-                        },
-                    );
-                    ptr.frame(self);
+                    if pass_event {
```
Drakulix, line 983: This isn't entirely correct, as this would still send the `ButtonReleased`-event without the client seeing a `ButtonPressed`-event previously.

You might want to take inspiration from `surpressed_keys` for keyboard-input. This will need a similar solution for mouse buttons.

Drakulix (changes requested):
Looks mostly good to me, except for mismatched button events.

src/input/mod.rs
JCapucho: I've been trying to fix this for a while and I think I might have taken the wrong approach, contrary to the keyboard `input` where we can intercept the event from the client, we must always send the `button` event because it will be sent to the `PointerGrab`, otherwise it will get stuck.

My idea is that the interception code gets moved to the `MoveGrab` and `ResizeForkGrab` where a new boolean field gets added for whether they are grabs resulting from a Super + {Left, Right} Click or not, And if they are the `button` handler doesn't pass the event to the client.

What do you think @Drakulix ?

JCapucho: I've actually implemented the suppress functionality in the grabs and ended up needing to do both the current approach of suppressing pressed events in `process_input` and release events in the grabs, because the grab is only added in the next idle cycle.

However this still seems to be not enough because in case a grab is not set there exists nothing to suppress the release event. So it seems like something like `suppressed_keys` is also needed.

JCapucho: I've finished adding code to suppress buttons that do not trigger a grab.

Honestly the code is not ideal, there's a lot of argument drilling but I don't know how to architect it better.

src/input/mod.rs
```diff
@@ -849,15 +863,22 @@ impl State {
                                             if let Some(surface) =
                                                 target.toplevel().map(Cow::into_owned)
                                             {
+                                                // If the logo is held then the pointer event is
+                                                // aimed at the compositor and shouldn't be passed
+                                                // to the application.
+                                                pass_event = false;
```
Drakulix, line 869: We should add the button to supressed_buttons right once we decided to not pass the event. Otherwise we might get mismatches, as it isn't deterministic when the idle_callbacks will run. Also having this in the `_`-match clause and `dispatch_grab` seems highly repetitive. We know already that we will skip the event here.

Though I would like to chance even this logic. Currently we discard every mouse button, if super as held, but we only need to discard those, that we will handle ourselves. So we should not prevent middle-clicks and other buttons, just left and right clicks.

Drakulix: Then we should also not need to add this to the grab code, which I really want to avoid as that is leaking abstractions left and right.

Drakulix (commented):
Still not 100% right, but thanks for working on this!

src/input/mod.rs
JCapucho: > Then we should also not need to add this to the grab code, which I really want to avoid as that is leaking abstractions left and right.

The problem is that `ptr.button` needs to be called if there is a grab because otherwise it won't receive the release event and it will be stuck moving/resizing.

And at that point the grab is the one that forwards to the client so it must be the one responsible for suppressing the button.

Drakulix: We could also release the grab inside of the input-handler. Its drop code will be called either way.

JCapucho:
I've switched the code to unset the grab instead of tracking the suppression status on the grab this greatly simplifies the code. I've also change the code so that the suppression is only done in the buttons handled by the compositor by pulling the match out.

I'm sure the code isn't good enough yet but just wanted to know if the overall structure is sound. And thank you for guiding me while implementing this :)

src/input/mod.rs
```diff
@@ -845,40 +859,50 @@ impl State {
                                     // Don't check override redirect windows, because we don't set keyboard focus to them explicitly.
                                     // These cases are handled by the XwaylandKeyboardGrab.
                                     if let Some(target) = shell.element_under(pos, &output) {
-                                        if seat.get_keyboard().unwrap().modifier_state().logo {
+                                        if seat.get_keyboard().unwrap().modifier_state().alt {
```
Drakulix, line 862: Why change the modifier?

src/input/mod.rs
```diff
@@ -969,20 +1001,25 @@ impl State {
...
+                            self,
+                            &ButtonEvent {
+                                button,
+                                state: event.state(),
+                                serial,
+                                time: event.time_msec(),
+                            },
+                        );
+                        ptr.frame(self);
+                    } else {
```
Drakulix, line 1020: Small nitpick here: Given we set the grab with an idle-callback unsetting any grab any time we don't pass events isn't wrong, but technically we only want to unset grabs on a `Released`-event.

```suggestion
                    } else if event.state() == ButtonState::Released {
```

Drakulix (commented):
Looks mostly good to me know! Two small nitpicks, then this is ready for merge imo. Thanks for working on this!

src/input/mod.rs
JCapucho: I changed it for testing because the logo key was being caught by my DE, didn't mean to commit it.

Drakulix (approved):
Thanks!

PR #752, opened 2024-08-20
toplevel-management: Set active output if changed, and move cursor

ids1024:
Previously, alt-tab wouldn't actually focus a window if it was on another output, since the active output/workspace was unchanged.

We need to move the cursor if we activate a window on another output.


Drakulix (approved):
Seems sensible

PR #757, opened 2024-08-21
Update pointer focus before motion occurs

ids1024:
This seems to work, but produces a flood of events. Smithay probably need a method to update focus without producing events unless a surface is entered.

Instead of `time: 0`, this and other synthesized events should probably use `CLOCK_MONOTONIC`. It seems libinput does document that it uses that.

### DIFF
--- a/src/shell/focus/mod.rs
+++ b/src/shell/focus/mod.rs
@@ -7,7 +7,7 @@ use crate::{
 use indexmap::IndexSet;
 use smithay::{
     desktop::{layer_map_for_output, PopupUngrabStrategy},
-    input::Seat,
+    input::{pointer::MotionEvent, Seat},
     output::Output,
     reexports::wayland_server::Resource,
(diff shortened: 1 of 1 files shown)


ids1024:
Actually, we can just test `PointerHandle::current_focus()` to see if we need to change it. So maybe no `smithay` changes are needed.

Drakulix:
> This seems to work, but produces a flood of events.

Why does it flood though. `refresh_focus` is called on an idle-callback, right? So it should only happen once all events are processed and the compositor would go into sleep, no?

ids1024:
I think this should be correct. The use of `current_location().as_global()` seems to match what is used elsewhere, and this now doesn't seem to produce input events unnecessarily.

I think the timestamp here and elsewhere shouldn't be `0`, but if so that can be updated in various places after https://github.com/Smithay/smithay/pull/1509 is merged.

Drakulix (approved):

PR #759, opened 2024-08-22
Add `cosmic_atspi_v1` protocol

ids1024:
This works with the `receive` example of `reis` using `LIBEI_SOCKET=/tmp/atspi-ei-kb.socket`. I've also worked on changes to at-spi2-core, so I'll now be able to try testing it...

If a version of this is ultimately used, it will need a secure way to pass the socket to accessibility tools. Putting it in `/tmp` is a placeholder.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -1464,6 +1464,13 @@ impl State {
                 .unwrap_or(false)
         });
 
+        self.common.atspi_ei.input(
+            modifiers,
+            &handle,
+            event.state(),
+            event.time() as u64 * 1000,
+        );
+
         // Leave move overview mode, if any modifier was released
         if let Some(Trigger::KeyboardMove(action_modifiers)) =
             shell.overview_mode().0.active_trigger()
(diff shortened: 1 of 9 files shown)


ids1024:
It's just a hacky prototype at the moment, but this now partly works with https://github.com/ids1024/at-spi2-core/tree/atspi-ei.

Orca keybindings now work, but the keys are also sent to applications. We need something like X11's `XIGrabKeycode` to register a key combination. But all we need to do is have the compositor track those grabs and not send them to applications. We don't need to feed all input synchronously through the screen reader or anything.

I think that's pretty much the only other thing we need for parity with `AtspiDeviceX11`?

We'll also need a secure mechanism to start the screen reader, and only give it access to this.

ids1024:
I've fixed an `unwrap` panic I noticed sometimes when killing orca, and added a workaround to disable caps lock state. We probably need to add a request to the protocol when a virtual modifier is added.

I think we can merge this now, with the understanding that it will be improved later, and won't cause issues when the protocol isn't in use.

src/input/mod.rs
```diff
@@ -1625,6 +1632,57 @@ impl State {
...
+                .atspi_ei
+                .virtual_mods
+                .contains(&event.key_code())
+        {
+            self.common
+                .atspi_ei
+                .active_virtual_mods
+                .insert(event.key_code());
+
+            tracing::error!(
```
Drakulix, line 1677: can we downgrade this to at least `info` (probably `debug`)?

src/input/mod.rs
```diff
@@ -1649,6 +1707,16 @@ impl State {
             return FilterResult::Intercept(None);
         }
 
+        // TODO modifiers queue
```
Drakulix, line 1710: Is this todo problematic? Can this currently interact weirdly with modifier only keybindings? (mostly for future reference to understand what needs to be done here.)

src/input/mod.rs
ids1024: Oh right, that's a debugging line I added at some point that I didn't mean to commit. (Downgrading to `debug` would be appropriate).

src/input/mod.rs
ids1024: I think that was just a comment to remind myself to look into whether this interacts problematically with that at some point. Not for any known issue.

I'll just remove that comment, but more testing could definitely be done in general.

Drakulix (approved):

Issue #769, opened 2024-08-24
Layer shell focus issue


dcz-self:
Squeekboard is not working reliably when trying to input text to a window.

### Steps to reproduce:

1. Run Squeekboard
2. Run `WAYLAND_DEBUG=1 yad --entry 2&>1 | grep text`
3. Type in any two buttons

Expected: The entry field has text from both input events, and they come from the text-input interface
Actual: the text does not come in as text-input (may come as keyboard events).

### Problem

When clicking on the layer-surface panel of squeekboard, despite KeyboardInteractivity=False, the yad application receives a `leave` event `zwp_text_input_v3@35.leave`, causing it to disable text input. Then the input method has nowhere to send to.

Actual scenario:
1. Click on layer-shell panel
2. yad receives `zwp_text_input_v3@35.leave`
3. yad sends "disable"
4. input method receives "inactive"
5. Input method receives click. *Input method can't send text to inactive interface.*

Expected:
1. Click on layer-shell panel
2. *Yad receives no event directly caused by that*
3. Input method receives click.
4. Input method sends text to the *active* interface
5. Yad receives text

### Diagnosis

This looks like disabling keyboard interactivity on the layer shell doesn't prevent it from taking the input-method focus.

I'll try to figure it out.

dcz-self:
This is no longer a problem when using current smithay.

PR #770, opened 2024-08-24
Layer-shell: don't drop focus when clicking layer-shell

dcz-self:
Fixes https://github.com/pop-os/cosmic-comp/issues/769

A layer shell requesting "KeyboardInteractive=False" would cause the compositor to momentarily drop focus from the original surface, and assume no focus. If the layer shell click produces a text input event, then the original surface can't receive the event, breaking the input method.

* * *

I'm not entirely sure if the change I made makes sense. It's not clear to why anything needs to be done if "done==false".

The overarching question is: what should happen when a layer surface is clicked but doesn't accept focus?

The answer so far was "remove keyboard focus from all applications". I turned it into "keep focus where it was". Considering that the focus jumped immediately back to the same application (but why?), I think this is the right direction.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -793,17 +793,27 @@ impl State {
                         // (in case of a window the toplevel) surface for the focus.
                         // see: https://gitlab.freedesktop.org/wayland/wayland/-/issues/294
                         if !seat.get_pointer().unwrap().is_grabbed() {
+                            enum NewFocusTarget {
+                                Change(KeyboardFocusTarget),
+                                Remove,
+                                KeepOld,
+                            }
+                            use NewFocusTarget::*;
                             let output = seat.active_output();
 
                             let pos = seat.get_pointer().unwrap().current_location().as_global();
                             let relative_pos = pos.to_local(&output);
-                            let mut under: Option<KeyboardFocusTarget> = None;
+                            let under: NewFocusTarget;
 
                             if let Some(session_lock) = shell.session_lock.as_ref() {
-                                under = session_lock
+                                under = match session_lock
                                     .surfaces
                                     .get(&output)
-                                    .map(|lock| lock.clone().into());
+                                    .map(|lock| lock.clone().into())
+                                {
+                                    Some(target) => Change(target),
+                                    None => Remove,
+                                };
                             } else if let Some(window) =
                                 shell.active_space(&output).get_fullscreen()
                             {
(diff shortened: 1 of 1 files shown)


src/input/mod.rs
```diff
@@ -838,24 +851,32 @@ impl State {
...
-                                        {
-                                            under = Some(layer.clone().into());
-                                            true
-                                        } else {
-                                            false
+                                            {
+                                                target = Change(layer.clone().into());
+                                                true
+                                            } else {
+                                                target = Remove;
```
Drakulix, line 868: I think this is somewhat confusing to read. We set the target to `Remove` implying the focus to be taken away, but since we return `false`, we never really evaluate target.

Can we please just make `target` `mut` and default initialize it with `Remove` and then skip these lines? (same goes for line 873)

Drakulix (approved):
This seems like a good change imo, just two small nitpicks on the implementation.

src/input/mod.rs
dcz-self: Tbh I found the entire function confusing. It's not explained what the "done" is and why it's necessary. Without knowing that, I settled on doing a mechanical translation.
I decided to drop the "mut" because it makes the choices implicit when they are not altered, making the purpose of each branch harder to figure out (need to backtrack).
I think the it can be structured better, rather than fal but I'd need a deeper explanation.

Drakulix: Fair.

I don't think `done` is used to great extent any more and is mostly an artifact of previous code. It was introduced as part of [this commit](https://github.com/pop-os/cosmic-comp/commit/85edfc71418ca08bc7a0f00c634990dabb01ac6a) to break apart a never ending cascade of `if-else` statements to figure out the focus target. The previous approach had the problem of never continuing down the tree, if any layer-surface was in the way, no matter if it actually has an input_region there or even took keyboard-focus.

> I decided to drop the "mut" because it makes the choices implicit when they are not altered, making the purpose of each branch harder to figure out (need to backtrack).

Also fair, but in that case I feel like we might want to combine `done` and `target` (possibly also `under`), as the state is far too spread out at the moment imo.

> I think the it can be structured better, rather than fal but I'd need a deeper explanation.

I hope that is helpful, do you have any further questions? Do you feel like tackling this or would you prefer me doing that?

dcz-self: I can do it, but I'm away for a week soon.

Drakulix: > I can do it, but I'm away for a week soon.

No rush. I'll ping you, if I happen to find some time to look into this earlier than you do.

dcz-self:
This works now, thank you!

Issue #586, opened 2024-06-28
do not delete (non-last) workspace after closing its apps


juhp:
I don't think a workspace should disappear in general when it no longer contains apps,
since this effectively causes other workspaces/apps to move up one workspace,
and changes the shortcuts associated with the moved workspaces, which is disruptive.

I actually would also prefer to have a configurable number of workspaces open by default.

Drakulix:
Static workspaces will be added back in a release after alpha 1.

juhp:
(I had meant come back and add that it is slightly better than I thought - the workspace is only closed when one leaves it _or_ if one is on another workspace and an app on the workspace crashes (or closes for some reason by itself) - which is what happened to me)

FriederHannenheim:
Now that Alpha 3 is released, is there any update on when static workspaces are coming?

jacobgkau:
Workspace pinning has now been implemented. You can pin a workspace to prevent it from being closed after closing its apps. The number of pinned workspaces is preserved through restarts, which means this also accomplishes "static" workspaces.

hgaiser:
Thanks, I think that is a good solution to allow both dynamic and static workspaces. Only thing that would make it even better is if we can rename workspaces :).

jacobgkau:
Renaming workspaces is planned for a future update, but isn't in-scope for Epoch 1.

hgaiser:
Awesome! Thanks for the update, appreciate it

PR #774, opened 2024-08-25
Configurable empty workspace removal behaviour

xDarksome:
Related to #586 

Adds a configuration option that allows to choose the behavior of the compositor when an empty workspace gets deactivated.

`RemoveEmpty::All` - same as currently, the empty workspace gets removed and the ones after it get shifted
`RemoveEmpty::Trailing` - removes only "trailing" empty workspaces, the ones after which there are no non-empty workspaces anymore

Been running it for a week, works fine. There was an edge case that caused a crash [here](https://github.com/pop-os/cosmic-comp/compare/master...xDarksome:cosmic-comp:master_jammy?expand=1#diff-4f1e1533acb2e2645a75fc56f2249115a8c1029a541937726ef8e6449b3b0cecR1002), because now previously active workspace may become invalid (out of bounds). Let me now if there's a better way to fix it.

### DIFF
--- a/cosmic-comp-config/src/workspace.rs
+++ b/cosmic-comp-config/src/workspace.rs
@@ -11,6 +11,19 @@ pub struct WorkspaceConfig {
     pub workspace_mode: WorkspaceMode,
     #[serde(default = "default_workspace_layout")]
     pub workspace_layout: WorkspaceLayout,
+    #[serde(default)]
+    pub remove_empty: RemoveEmpty,
+}
+
+/// Setting of which empty workspaces the compositor should automatically remove.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Serialize, Deserialize)]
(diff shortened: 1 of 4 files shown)


Drakulix:
Ping @pop-os/ux, mostly for taking note of this use-case. (Though I would appreciate some input.)

Thank you for the work you put into this, but I honestly don't think we want this contribution - not even as a (likely rarely tested) config option - as this further complicates the workspace logic, which we want to revisit in the future anyway.

Static/Named workspaces are still a feature we want to eventually have and it is unclear how exactly that would interact with these changes. (And we don't want to silently drop features, like this config settings, when we start reworking the logic.)

git-f0x:
This is the planned workspace behavior, which might address your workflow issues: https://github.com/pop-os/cosmic-workspaces-epoch/issues/108

PR #779, opened 2024-08-27
Xwayland de-scaling

Drakulix:
Adds a "X11 windows handle scaling on their own"-option similar to KDE.

Draft until cosmic-session and cosmic-settings PRs are done and the necessary smithay changes are merged.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4703,7 +4703,7 @@ checksum = "3c5e1a9a646d36c3599cd173a41282daf47c44583ad367b8e6837255952e5c67"
 [[package]]
 name = "smithay"
 version = "0.3.0"
-source = "git+https://github.com/smithay//smithay?rev=298bef3#298bef358d4bbffd3a1159dc2ec418b9f81417cc"
+source = "git+https://github.com/smithay//smithay?rev=65c4abf#65c4abf4cbdf122abff96ad17a440c76a8cf7cea"
 dependencies = [
  "appendlist",
  "ash 0.38.0+1.3.281",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -116,4 +116,4 @@ inherits = "release"
 lto = "fat"
 
 [patch."https://github.com/Smithay/smithay.git"]
-smithay = { git = "https://github.com/smithay//smithay", rev = "298bef3" }
+smithay = { git = "https://github.com/smithay//smithay", rev = "65c4abf" }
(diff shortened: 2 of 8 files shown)


ids1024:
With "applications scale themselves", applications that don't do any scaling like `xterm` and `xcalc` appear small on my 1x scale outputs, not only on the 2x scale output.

I assume this is expected, and a limitation of X, so w can't tell applications to use 1x scale or 2x scale depending on what output they're on? (And perhaps it's also desirable if an app has it's own option to change scaling that won't adjust for monitor. Like increasing the font size in xterm.)

ids1024:
Yeah, looking at how XSETTINGS works, I guess it has to work that way. The X server has a single `_XSETTINGS_S0` selection that applies the same settings to all windows. So we can change it at any time, but only for all windows of all clients.

This does presumably also mean that we don't get proper subpixel font rendering on 1x outputs of the mixed DPI system. Since the buffer is downscaled after rendering. And that those windows waste resources rendering at a higher resolution than presently needed. But presumably the KDE implementation has the same limitations.

In theory it would be good if the wording in cosmic-settings to express the implications of downscaling as a tradeoff as well, but that would make the description quite excessively long and technical.

ids1024:
So this looks reasonable and seems to be working as expected.

Drakulix:
> With "applications scale themselves", applications that don't do any scaling like `xterm` and `xcalc` appear small on my 1x scale outputs, not only on the 2x scale output.

This is only if a 2x or more output is also present though, right?

If so: no nothing we can really do, and the factor is global and we have to assume the application respects it for this to be useful.

In the future we could have a second Xwayland instance and some "compatibility mode" settings in the laucher/app-library/dock.

But before doing that we should support lazily starting Xwayland instances.

Drakulix:
> Yeah, looking at how XSETTINGS works, I guess it has to work that way. The X server has a single `_XSETTINGS_S0` selection that applies the same settings to all windows. So we can change it at any time, but only for all windows of all clients.
> 
> 
> 
> This does presumably also mean that we don't get proper subpixel font rendering on 1x outputs of the mixed DPI system. Since the buffer is downscaled after rendering. And that those windows waste resources rendering at a higher resolution than presently needed. But presumably the KDE implementation has the same limitations.

Yes it does, note this was the same with our "HiDPI Daemon" previously and that was also deemed good enough.

It is almost like we need a new windowing system to properly support this. /s

PR #780, opened 2024-08-27
screencopy: Use buffer age of `0` for render to temporary buffer

ids1024:
It seems we allocate a new `GlesRenderbuffer` every time we screencopy to an shm buffer.

We probably should use a more complicated approach to do proper damage tracking without any unnecessary copies, and re-using the GPU buffer, but as long as this allocates a buffer the age of that buffer should be treated as `0`.

Fixes corruption in cosmic-workspaces when shm screencopy is used. (For instance, when Cosmic is run with software rendering.)


Drakulix (approved):
Nice catch!

PR #788, opened 2024-08-28
Add `ext-foreign-toplevel-list-v1` protocol

ids1024:
Seems to be working, testing with https://github.com/Smithay/client-toolkit/pull/465.


Drakulix (approved):
LGTM!

PR #792, opened 2024-08-30
Add `xdg-foreign-unstable-v2` protocol

ids1024:


Drakulix (commented):
Yeah I don't think this is enough. We probably also want to raise xdg-foreign windows in our update_active impl. Or does the smithay code handles this and set parent?

Drakulix (approved):
Looking at smithay's code, it does indeed set `parent` for us. About time we support this protocol then, thanks!!

ids1024:
Yeah.

`xdg-desktop-portal-cosmic` should use this, as well as `xdg-dialog-v1`. But that one will require more work to support.

PR #794, opened 2024-08-30
Natural 4-finger gestures

### LINKED ISSUE #754: Make four-finger workspace swipe respect system setting for natural scrolling
Version: a3c8111

It's jarring to have to keep going between swiping the opposite direction I want to scroll for virtual desktops, and then swiping the direction I want to scroll for scrolling within apps.

In our GNOME session, this gesture respected the natural scrolling setting.

pknepps:
resolves #754 
4-finger gestures should now follow user setting for natural scroll on touchpad

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -1120,6 +1120,12 @@ impl State {
                         );
                         // Decide on action if first update
                         if first_update {
+                            let mut natural_scroll = false;
+                            if let Some(scroll_config) = &self.common.config.cosmic_conf.input_touchpad.scroll_config {
+                                if let Some(natural) = scroll_config.natural_scroll {
+                                    natural_scroll = natural;
+                                }
+                            }
                             activate_action = match gesture_state.fingers {
                                 3 => None, // TODO: 3 finger gestures
                                 4 => {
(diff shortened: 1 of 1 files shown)


src/input/mod.rs
```diff
@@ -1120,6 +1120,12 @@ impl State {
                         );
                         // Decide on action if first update
                         if first_update {
+                            let mut natural_scroll = true;
```
Drakulix, line 1123: So according to the libinput docs, natural_scroll should always be an option for touchpad and thus the scroll_config as well as the natural_scroll property should always be set here: https://github.com/pop-os/cosmic-comp/blob/ed64e26faf4b97221f0883bc1113f997acbadc50/src/config/input_config.rs#L56

**But** safe-guarding is certainly a good idea, though I don't think I agree with the default. libinput seems to generally set this to `false` if no device specific config overrides this (e.g. for Apple Touchpads). As such, I am inclined to say this should be `false` as well for a consistent experience.

```suggestion
                            let mut natural_scroll = false;
```

Drakulix (commented):
Thanks for working on this! Small nitpick, otherwise looks good to me.

pknepps:
Made the change. Thank you!

Drakulix (approved):
Thanks!

PR #806, opened 2024-09-03
Cleanup: remove unnecessary casting

julianbraha:
Removes typecasts where the variable is already that type, e.g.:
```
let offset = data.offset as i32;
let width = data.width as i32;
let height = data.height as i32;
let stride = data.stride as i32;
```
these fields are already `i32`, so `as i32` can be removed.

This resolves 18 instances of this warning.

Found with clippy lint:
https://rust-lang.github.io/rust-clippy/master/index.html#/unnecessary_cast

### DIFF
--- a/src/backend/winit.rs
+++ b/src/backend/winit.rs
@@ -112,9 +112,9 @@ impl WinitState {
             .get::<RefCell<OutputConfig>>()
             .unwrap()
             .borrow_mut();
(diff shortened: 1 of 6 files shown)


Drakulix (approved):
We need to be careful with these lints as these types might differ between architectures, but these changes all seem fine. Thanks!

PR #807, opened 2024-09-03
feat: Add keybinding for `WindowSwitcherReverse`

bbb651:
See https://github.com/pop-os/cosmic-launcher/pull/183

### DIFF
--- a/data/keybindings.ron
+++ b/data/keybindings.ron
@@ -80,6 +80,7 @@
     (modifiers: [Alt], key: "Tab"): System(WindowSwitcher),
     (modifiers: [Alt, Shift], key: "Tab"): System(WindowSwitcherReverse),
     (modifiers: [Super], key: "Tab"): System(WindowSwitcher),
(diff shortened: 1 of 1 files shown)


data/keybindings.ron
```diff
@@ -78,6 +78,7 @@
     (modifiers: [Super], key: "slash"): System(Launcher),
     (modifiers: [Super]): System(Launcher),
     (modifiers: [Alt], key: "Tab"): System(WindowSwitcher),
+    (modifiers: [Alt, Shift], key: "Tab"): System(WindowSwitcherReverse),
     (modifiers: [Super], key: "Tab"): System(WindowSwitcher),
```
skylerberg, line 82: I believe we should also add Super+Shift+Tab here

skylerberg (changes requested):

skylerberg (approved):

git-f0x:
I think this can be closed, since it has been implemented since.

PR #811, opened 2024-09-04
feat(Workspace): add definition for thumbnail placements

acolombier:
Add definition for thumbnail placement in workspace

Used in https://github.com/pop-os/cosmic-settings/pull/557

### DIFF
--- a/cosmic-comp-config/src/workspace.rs
+++ b/cosmic-comp-config/src/workspace.rs
@@ -2,34 +2,35 @@
 
 use serde::{Deserialize, Serialize};
 
-fn default_workspace_layout() -> WorkspaceLayout {
-    WorkspaceLayout::Vertical
-}
-
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Default)]
 pub struct WorkspaceConfig {
     pub workspace_mode: WorkspaceMode,
-    #[serde(default = "default_workspace_layout")]
+    #[serde(default)]
     pub workspace_layout: WorkspaceLayout,
+    pub workspace_thumbnail_placement: WorkspaceThumbnailPlacement,
 }
 
-impl Default for WorkspaceConfig {
-    fn default() -> Self {
-        Self {
-            workspace_mode: WorkspaceMode::OutputBound,
-            workspace_layout: WorkspaceLayout::Vertical,
-        }
-    }
-}
-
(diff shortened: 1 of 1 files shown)


ryanabx:
I feel like workspace cosmetic info that the compositor doesn't need should be in a separate config, perhaps in the cosmic-workspaces-epoch repo. The overall changes are really great though! Thank you for working on this

acolombier:
I'm wondering if it would be even worth merging `WorkspaceLayout` and `WorkspaceThumbnailPlacement` together since the first on is kinda redundant. I feel like having the workspace setting definition in a single place is valuable tho, but agreed it should perhaps be moved out of `cosmic-comp` repo.
I think there is a risk that at some point, cosmic-comp needs information about workspace thumbnail placement to adapt certain behaviour so it would make a point on keeping config in a single crate

ryanabx:
> cosmic-comp needs information about workspace thumbnail placement to adapt certain behaviour

Sorry, I may not be seeing what you're seeing, but why would cosmic-comp need that?

acolombier:
There is no such a behaviour yet! It was more a matter of, this could be needed in the future.

ryanabx:
Yeah, I suppose it's not too harmful to have it in cosmic-comp's config. The reason I contended on that is because cosmic-comp is used in projects that don't utilize the entire cosmic-session, so having a config value that's only used in cosmic-workspaces-epoch is clutter, when the value could be in cosmic-workspaces-epoch instead. The workspace orientation is something comp actually uses, thus it makes sense to have it there, and have cosmic-workspaces-epoch read the value.

ryanabx:
@Drakulix what do you think about the above conversation?

acolombier:
Does it makes sense for the `WorkspaceLayout` to be `Vertical` and `WorkspaceThumbnailPlacement` to be `Left` or `Right`? Looking at the original design in Figma, the answer seems to be now, but appreciate that this could be an oversight during design. If there is a consensus around no tho, I guess it could make sense to merge `WorkspaceLayout` and `WorkspaceThumbnailPlacement` as one enum, rather than two independent, which may be in conflict state (eg `Vertical` vs `Left`/`Right`)

acolombier:
Friendly ping @Drakulix in case you missed the notification :)

Drakulix:
I have not yet reviewed this, as I am still waiting @pop-os/ux to decide, if we want this feature.

acolombier:
This was developed following the Figma design as per mentioned [here](https://github.com/pop-os/cosmic-settings/pull/557). [Feedback](https://github.com/pop-os/cosmic-settings/pull/557#issuecomment-2329571476) about the UI was also provided and addressed.

Issue #734, opened 2024-08-15
Changing language messes up binds


BbIXOD:
If you set your language to something non-Latin (for example Ukrainian) it will prevent de binds from executing. It affects only binds which use letters.

leviport:
Duplicate of https://github.com/pop-os/cosmic-comp/issues/642

PR #813, opened 2024-09-04
fix(input): binds not working on non-Latin layouts

dvtfl:
Workarounds #734.

Replaces [`KeysymHandle::raw_syms()`](https://smithay.github.io/smithay/smithay/input/keyboard/struct.KeysymHandle.html#method.raw_syms) with [`KeysymHandle::raw_latin_sym_or_raw_current_sym()`](https://smithay.github.io/smithay/smithay/input/keyboard/struct.KeysymHandle.html#method.raw_latin_sym_or_raw_current_sym).

`KeysymHandle` returns typed symbols as-is. For example, if "Й" (Cyrillic letter 'Io', corresponding to "Q" on the English layout) is pressed, `raw_syms()` returns "Cyrillic_io". However, keybinds are usually defined as "<something>+Q" (i.e., in Latin), which causes a mismatch.

On the other hand `KeysymHandle::raw_latin_sym_or_raw_current_sym()` tries to return corresponding Latin key symbol.


Drakulix:
I am not a huge fan of this, given I worried what else this might break.

The keybindings shouldn't use keysyms in the long run for anything but describing the keys in the user interface and use keycodes for matching them instead. That would make this change unnecessary without causing a lot of breakage (hopefully).

However I am inclined to merge workarounds in the meantime. Are you aware of any layout/shortcut-combinations this actively breaks?

Drakulix:
> FWIW: IIUC Windows, Gnome, KDE, i3 are using keysyms for keybindings. E.g., pressing a bind that require keys Win+[Latin Q] on English (QWERTY) layout would require pressing Ctrl+[Latin A] on French layout (AZERTY).

Good to know.

> Nope. I’ve updated the commit so that the system first attempts to match the binding with the current layout. If it fails, it falls back to matching the Latin key. This way, binds like Ctrl+é (french layout) shouldn't break (at least i hope, it's getting confusing for me).

This seems like a good change, however this still means we go through all bindings in order and potentially select a weaker `raw_latin_sym_or_raw_current_sym`-bind, although later in our loop a binding could match a `raw_syms`-bind...

Issue #812, opened 2024-09-04
[Feature Request] Allow closing of tabs in a stack with a middle mouse click


w1ll-i-code:
Modern browsers (e.g. firefox and chromium) and text editors (e.g. vscode and intellij) allow one to close open tabs by using a click with the mouse wheel on the respective tab. I think this would also be a really helpful feature for window stacks as it generally is faster and requires less precision and works well with muscle memory from other applications.

PR #814, opened 2024-09-04
Close tab on middle mouse click.

w1ll-i-code:
Implements #812.

Allows to close a tab in a window stack with a middle mouse click. Dispatch a close Message on middle mouse click, the rest was already implemented. 

---

I'm sorry this is my first PR in this repo and I'm not sure about the protocol for this here. If anything is missing, please let me know.


Quackdoc:
Can this be configurable? middle mouse click is something I accidentally hit a lot, accidentally closing the window would be fairly annoying.

w1ll-i-code:
Thanks for the comment. That's a use-case I have not thought or heard about before. But I do agree that we'd want to avoid that.

But I fear making it configurable might not be the right approach here. I think the discoverability of such an option is quite low. We'd definitely have to place that into the cosmic-settings, but even then I fear people struggling with this behaviour might not think that it's configurable. 

If you want, I'd like to first understand your use-case better:

1. In which cases do you usually accidentally hit that button?
    1. When moving the mouse?
    2. When pressing the left or right mouse buttons?
    3. When moving your hand from the keyboard to the mouse?
2. Have you struggled with that also in other applications?
    1. How often do you close tabs in browsers or text editors accidentally?
    3. Have you found any configuration option there to make life easier?
3. Which windows do you usually use in a stack like that? My personal use-case is mostly for terminals. But for other use-cases this might be even more annoying.

Quackdoc:
Due to my RSI, I use an ergonomic mouse, in which the "natural" position has my index finger partially resting on the scroll wheel. With the lack of finger dexterity I have on bad days due to the RSI, I can often accidentally trigger it when try to left click.

It does occasionally happen when using firefox and chromium, if I was working that day, it would probably happen maybe 1 out of 5-20 times I swap a tab. I don't use tabs on editors enough for this to be an issue on them. It's not a major issue in something like a browser since I can just re-open the tab with a hotkey (which I have bound on separate pad), and most of the sites I use load fast on my PC. 

I typically keep my telecoms apps like telegram, matrix, discord, and slack in a stack. I wouldn't loose work or anything, it would be at most a minor inconvenience. 

at the very least, I haven't heard of other people having issues with this like I have anyways. I'm not sure if discoverability would be a major issue. I for sure see the benefit in having the feature.

w1ll-i-code:
Thanks for the detailed response. I totally agree that this is a problem we should address. Accessibility is really important Imo. Sadly I don't have a lot of experience in that field. In this case it really seems like the most pragmatic option. I can definitely implement such a config option.

I'll check later in the code which options are available. I could see either a simple toggle or adjustable key-bindings work well for this. What do you think would be the better option?

w1ll-i-code:
I'm not sure if the maintainers already have a concept for accessibility in mind, where we could hook into. Either for discoverability or documentation.

w1ll-i-code:
@Drakulix Sorry for pinging you. I think for the config options we need some input/feedback from a maintainer... I'm more than happy to implement this, both here and in cosmic-settings, but I wanna make sure first that we are going in the right direction.

Drakulix:
> @Drakulix Sorry for pinging you. I think for the config options we need some input/feedback from a maintainer... I'm more than happy to implement this, both here and in cosmic-settings, but I wanna make sure first that we are going in the right direction.

I think in general close-on-middle-click as a feature is desirable. I agree, that being able to disable this should be an accessibility feature though.

So what we want to have in cosmic-comp is a setting in `src/config.rs` to enable/disable close-on-middle-click.

For the cosmic-settings side, I cannot give you much guidance. We have plans for accessibility settings and some very early design, I reckon, but nothing really where you can just add this.
Pinging @maria-komarova in case I am not up-to-date here.

So I am inclined to say we can leave this a config-file only setting for now, but any implementation that *is* merged should have a way to toggle it.

maria-komarova:
The design is not ready yet for the Accessibility settings so leaving it in the config only for now sounds good to me.

w1ll-i-code:
@Drakulix Sorry it took me a while. Now it is configurable and the configuration will be updated live whenever the file changes.

w1ll-i-code:
I just noticed that I did one change only locally through go-to-definition... iced_runtime would need a change for that to work... I need to investigate how else I can propagate config changes to the windows then.

w1ll-i-code:
Hi @Drakulix. I'm really sorry. This fell completely off my radar, but now its finally working.

marco-zan:
Hi guys, first time commenting here.
Imho adding an option to toggle middle-click as discussed might not address the underlying issue completely. Couldn't an accessibility option to disable middle click entirely or with some filters a better solution?
It's easier to implement and do not rely on implementation on each and every part of the code that implement some action on middle click. 
Also would remove the need of workarounds on the user-side. 

Of course, this is just an idea, and I'd love to hear your thoughts on it. Thanks for all the great work done on cosmic

w1ll-i-code:
That sounds reasonable, but I'm not in a position to make a call on that. Maybe @maria-komarova has an update on the plans for the accessibility features?

Drakulix:
Thanks for working on this!

I can't promise I'll merge this soon though, because of the necessary plumbing of the new config parameter. While the approach seems sound on first glance (and is very similar to what we already do for the theme), I don't feel comfortable merging this as is and would rather refactor how windows receive new configuration. This just seems error prone and smells like a design flaw of cosmic-comp, that I need to address.

However I don't really have time for this refactor atm as other features/fixes have higher priority, sorry.

w1ll-i-code:
No worries. I was already wondering if this kind of config handling would be the final implementation.

PR #818, opened 2024-09-05
Patch for incorrect git url for 'smithay'

Cypher1:


Drakulix:
This is by design (and works) due to https://github.com/rust-lang/cargo/issues/7497 . Otherwise the patch breaks.

Cypher1:
Sorry, the URL has an extra `/` in it causing anyone using
```
[url "git@github.com:"]
	insteadOf = "https://github.com/"
```
To fail to authenticate.

I'd love to understand why it's part of the design of `cosmic-comp`

Drakulix:
It is, a proper fix for cargo is  need before we can fix it though.

PR #844, opened 2024-09-08
implement cursor-shape-v1

m4rch3n1ng:
when a cursor is requested it first renders any Surface cursors, then it renders the named cursors giving priority to the cursors set by the `set_shape` function.

surprising behaviour to me at least is, that it gives priority to Hidden and Surface cursors over the cursors set by `set_shape`, but since that also was the old behaviour i didn't change it without getting feedback first.

i replaced the [manual cursor cache insertion on startup](https://github.com/pop-os/cosmic-comp/blob/79ae56c601176f179870f75719416dbabe3d01d9/src/backend/render/cursor.rs#L259-L310) (because i thought it was silly to repeat this for 34 cursors) and replaced it with a function that loads [the cursors on demand](https://github.com/pop-os/cosmic-comp/blob/dd9adc0616cc692f4707cd92178237f0979196b8/src/backend/render/cursor.rs#L215-L219).

i also changed the behaviour of `Cursor::load`, so that it [first tries to load the default cursor of the theme](https://github.com/pop-os/cosmic-comp/blob/dd9adc0616cc692f4707cd92178237f0979196b8/src/backend/render/cursor.rs#L46) instead of using the fallback cursor immediately.

and last i also changed two calls to `Duration::from(time).as_millis()` with a direct call to `Time::as_millis()` (as of https://github.com/Smithay/smithay/commit/c700970e2cf2fb6698e5fcc47066fc74ce066ce9).

as the actual last point i also didn't know how cursors work in xwayland, so i left that largely unchanged as well.

a lot of the noise in the commit is because i changed the `set_shape()` function to take a `CursorIcon` instead of a `CursorShape` and removed the `CursorShape` enum entirely.

### DIFF
--- a/src/backend/render/cursor.rs
+++ b/src/backend/render/cursor.rs
@@ -275,20 +275,20 @@ where
         })
         .unwrap_or(CursorImageStatus::default_named());
 
-    if let CursorImageStatus::Surface(ref wl_surface) = cursor_status {
-        return draw_surface_cursor(renderer, wl_surface, location.to_i32_round(), scale);
-    } else if let CursorImageStatus::Named(named_cursor) = cursor_status {
-        let integer_scale = scale.x.max(scale.y).ceil() as u32;
+    let seat_userdata = seat.user_data();
+    let mut state_ref = seat_userdata.get::<CursorState>().unwrap().lock().unwrap();
+    let state = &mut *state_ref;
 
-        let seat_userdata = seat.user_data();
-        let mut state_ref = seat_userdata.get::<CursorState>().unwrap().lock().unwrap();
-        let state = &mut *state_ref;
-
-        let current_cursor = state.current_cursor.unwrap_or(named_cursor);
+    let named_cursor = state.current_cursor.or(match cursor_status {
+        CursorImageStatus::Named(named_cursor) => Some(named_cursor),
+        _ => None,
+    });
+    if let Some(current_cursor) = named_cursor {
         if !draw_default && current_cursor == CursorIcon::Default {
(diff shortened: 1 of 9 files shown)


Drakulix:
Looks good on first glance, thanks for working on this!

> when a cursor is requested it first renders any Surface cursors, then it renders the named cursors giving priority to the cursors set by the set_shape function.
> 
> surprising behaviour to me at least is, that it gives priority to Hidden and Surface cursors over the cursors set by set_shape, but since that also was the old behaviour i didn't change it without getting feedback first.

Yeah, I am not sure this is correct, as obviously shapes set by grabs should take precedence. And they might already, because grabs make the surface loose focus, so smithay should unset the `CursorImageState`. But that is just a coincidence, so it still might make sense to prefer explicitly set shapes by the compositor. Could you switch that around?

> as the actual last point i also didn't know how cursors work in xwayland, so i left that largely unchanged as well.

Xwayland also uses `Surface` cursors and renders the cursor for us. We only need to provide it with a rendered default cursor to fallback to, so what you have done it totally fine.

m4rch3n1ng:
done, but i'm not quite happy with the `.or(match`, do you want me to move the stuff to it's seperate function and just call that twice?

> Xwayland also uses Surface cursors and renders the cursor for us.

ah nice, that makes my job significantly easier

Drakulix (approved):
> done, but i'm not super happy with the `.or(match`, do you want me to move the stuff to it's seperate function and just call that twice?

Nah, that is fine.

Thanks! This looks good to go. :)

PR #860, opened 2024-09-13
Allow resize in both direcitons on Meta+Right click

### LINKED ISSUE #858: Meta+Right click resize in two directions at once
In sway, Meta+Right click you can resize in both directions at once, and it's really nice. You don't have to carefully plan which segment of a window to grab to do the resize you want to do. It acts like you're dragging the closest corner of the window. 

For example, in this clip I think it should also resize cosmic files width.

https://github.com/user-attachments/assets/0ae3e056-acdd-4011-aa4e-ef447134f84b

russelltg:
Fixes #858

This is just an initial implementation, it seems to work alright but I'm unfamiliar with this codebase so may have done something in a silly way.


russelltg:
Quick demo:
https://github.com/user-attachments/assets/e60aa4f1-bb76-44b9-ae0c-6c57d77c5ace

russelltg:
Definitely has a some bugs with more windows on screen that I'll have to figure out

src/shell/layout/tiling/grabs/resize.rs
```diff
@@ -94,6 +94,7 @@ impl PointerTarget<State> for ResizeForkTarget {
                         location.as_global(),
                         node,
                         left_up_idx,
+                        None,
```
Drakulix, line 97: Preferably these would also gain the option to do more complex grabs at corners, but I guess we can leave this for now.

src/shell/layout/tiling/grabs/resize.rs
```diff
@@ -184,9 +186,11 @@ pub struct ResizeForkGrab {
     last_loc: Point<f64, Global>,
     old_tree: Option<Tree<Data>>,
     accumulated_delta: f64,
+    accumulated_delta_parent: f64,
     node: NodeId,
     output: WeakOutput,
     left_up_idx: usize,
+    parent_left_up_idx: Option<usize>,
```
Drakulix, line 193: Instead of splitting this in `parent` and `node`/<no suffix>, I would prefer, if we separately determine the `left_idx` and `left_node` and the `up_idx` and the `up_node`. (And have both be an `Option`, though the grab should obviously never be created with both being `None`.)

src/shell/layout/tiling/grabs/resize.rs
```diff
@@ -281,39 +293,42 @@ impl ResizeForkGrab {
...
                     }
                     _ => unreachable!(),
+                };
+
+                if let Some(parent_left_up_idx) = self.parent_left_up_idx {
+                    if let Some(Data::Group {
+                        orientation, sizes, ..
+                    }) = parent.map(|p| tree.get_mut(&p).unwrap().data_mut())
+                    {
+                        if *orientation == child_orientation {
```
Drakulix, line 322: This can happen, we simply have to walk up the tree in a `while`-loop and select the next `parent` to check for it's orientation, until we potentially get `None` (which will happen if there isn't any edge in the other direction as we hit the workspace/output border).

Drakulix (commented):
> This is just an initial implementation, it seems to work alright but I'm unfamiliar with this codebase so may have done something in a silly way.

Seems sensible, I also thought about how to fix this already, so here are some thoughts. Your implementation is quite close to how I envision this being solved, I would appreciate it, should you continue working on this.

src/shell/layout/tiling/grabs/resize.rs
russelltg: The `*_node` values would be instances of something like `enum ResizeDest { Node, NodeParent }`?

russelltg: Okay, looking at your other comment about having to walk up the tree until finding a split of the correct orientation, this needs to be a `ResizeDest(usize)` meaning how many times we need to walk towards the root?

russelltg: Okay, looking at your other comment about having to walk up the tree until finding a split of the correct orientation, this needs to be a `ResizeDest(usize)` meaning how many times we need to walk towards the root?

Drakulix: > 

The *_node values would be instances of something like enum ResizeDest { Node, NodeParent }?


No just `NodeId`s. You don't need to know how far you have walked, if you just mark the nodes we do the resize operations on. (Both of them one for each axis.)

src/shell/layout/tiling/grabs/resize.rs
```diff
@@ -83,6 +83,10 @@ impl PointerTarget<State> for ResizeForkTarget {
             data.common.event_loop_handle.insert_idle(move |state| {
                 let pointer = seat.get_pointer().unwrap();
                 let location = pointer.current_location();
+                let (left_node_idx, right_node_idx) = match orientation {
```
Drakulix, line 86: left and right? Shouldn't this be left and up?

Drakulix (commented):
This seems pretty spot on what I had in mind! Just a couple of nitpicks. :)

Also please remember to remove the draft-status, once you want me to test this and eventually merge.

russelltg:
Just realized I swapped the resize grab directions, will fix tomorrow

russelltg:
Ok, upon further review I don't think I regressed the grab--this snippet is from master (and it's the same on this branch)

I think this branch is ready for testing/review. 

https://github.com/user-attachments/assets/a45e6a32-039a-4914-9a12-18a31a69fbc9

russelltg:
Gentle reminder on this, is there anything left to do to move this forward?

rosalyntg:
Rebased :)

PR #862, opened 2024-09-13
Split off `KeyboardHandle::filter` callback to separate method

ids1024:
Requires https://github.com/Smithay/smithay/pull/1533. If the `input_intercept` changes there are controversial, this could instead use a a few more arguments, or something involving `input_intercept`/`input_forward` instead of `input` (but that's probably not simpler).

https://github.com/pop-os/cosmic-comp/pull/465 will require some additional logic here to filter out the key grabs. I don't really want this to get *more* complicated so a wanted to try splitting things up first. Just moving the callback to a method is a start, though more could be improved.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4481,12 +4481,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "scan_fmt"
-version = "0.2.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


Drakulix (commented):
Seems sensible to me

ids1024:
Updated now that Smithay changes are merged, and added a commit removing the `needs_key_repetition` argument which should be unneeded with https://github.com/Smithay/smithay/pull/1535.

Drakulix (approved):

PR #870, opened 2024-09-18
toplevel-info/mgmt: Update to v2/v3

Drakulix:
Swallows up the foreign-toplevel-list protocol handling as well, to be able to guarantee event-order between the cosmic-toplevel-info protocol and foreign-toplevel-list (e.g. it's `done`-event).

Untested, but this should basically work..

### DIFF
--- a/src/wayland/protocols/toplevel_info.rs
+++ b/src/wayland/protocols/toplevel_info.rs
@@ -10,7 +10,12 @@ use smithay::{
         Client, DataInit, Dispatch, DisplayHandle, GlobalDispatch, New, Resource, Weak,
     },
     utils::{user_data::UserDataMap, IsAlive, Logical, Rectangle},
+    wayland::foreign_toplevel_list::{
+        ForeignToplevelHandle, ForeignToplevelListHandler, ForeignToplevelListState,
+    },
 };
+
+use crate::utils::prelude::{Global, OutputExt, RectGlobalExt};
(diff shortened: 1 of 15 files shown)


Drakulix:
Also adds now an overlap-notify implementation.

src/wayland/protocols/toplevel_info.rs
```diff
@@ -287,22 +348,32 @@ where
...
                     {
                         handle.closed();
                     }
                 }
                 false
             }
         });
+        for instance in &self.instances {
+            if instance.version() >= zcosmic_toplevel_info_v1::EVT_DONE_SINCE {
+                instance.done();
```
ids1024, line 365: Testing this, I'm seeing a flood of `zcosmic_toplevel_info_v1::done` events.

Presumably it shouldn't send `done` unless at least one toplevel has changed (or been added or removed).

ids1024:
Client implementation for toplevel info version 2: https://github.com/pop-os/cosmic-protocols/pull/35. Seems to be working other than the `done` issue. And otherwise this looks reasonable.

src/wayland/protocols/toplevel_info.rs
Drakulix: Should be fixed now. Global `done` also needed some fixes to be only send after we cleaned up all of the state updates successfully.

PR #882, opened 2024-09-25
fix: mouse click on KeyboardSwap crashes

zeroishero:
closes pop-os/cosmic-comp#924 

Added check for Keyboard Swap when Mouse click event happens.


Drakulix:
Thanks, but this was fixed via https://github.com/pop-os/cosmic-comp/commit/b53eb508a8e795bc22f80d85250f73c33195db9d instead.

PR #884, opened 2024-09-26
warp cursor to center of window for `cursor_follows_focus`

m4rch3n1ng:
i'm not sure if you would want to put this behind a config, but i don't know if the cursor warping to the top left of a window is ever the better option over warping to the center, but i can config it if desirable.

i removed the `if is_some()` and then `unwrap()` thing, because `Rectangle` is copy and not borrowed anyway, but i don't exactly know why this was here so i can just revert it if necessary.

### DIFF
--- a/src/shell/focus/mod.rs
+++ b/src/shell/focus/mod.rs
@@ -10,7 +10,7 @@ use smithay::{
     input::{pointer::MotionEvent, Seat},
     output::Output,
     reexports::wayland_server::Resource,
-    utils::{IsAlive, Serial, SERIAL_COUNTER},
+    utils::{IsAlive, Point, Serial, SERIAL_COUNTER},
     wayland::{
         seat::WaylandFocus,
         selection::{data_device::set_data_device_focus, primary_selection::set_primary_focus},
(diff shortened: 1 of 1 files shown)


git-f0x:
This is planned as the default behavior, so probably no config option necessary. (pop-os/cosmic-settings#567) :)

src/shell/focus/mod.rs
```diff
@@ -231,29 +231,29 @@ fn update_focus_state(
...
                 let mut shell = state.common.shell.write().unwrap();
-                // get the top left corner of the target element
+                // get geometry of the target element
                 let geometry = shell.focused_geometry(target.unwrap());
-                //to avoid the nested mutable borrow of state
-                if geometry.is_some() {
-                    let top_left = geometry.unwrap().loc.to_f64();
+                if let Some(geometry) = geometry {
+                    // get the center of the target element
+                    let new_pos = (geometry.loc + geometry.size.downscale(2).to_point()).to_f64();
```
Drakulix, line 238: This is quite the nitpick, but `downscale(2).to_point()` doesn't really communicate the intent of this operation at all. Please do a more obvious devision by 2 or add a comment.

Drakulix (approved):
Thanks!

src/shell/focus/mod.rs
m4rch3n1ng: done, i think

Drakulix (approved):
Thx!

PR #885, opened 2024-09-26
Update `egui`, `smithay-egui`

ids1024:
Requires https://github.com/Smithay/smithay/pull/1195 and https://github.com/Smithay/smithay-egui/pull/16.


Drakulix (approved):

ids1024:
Updated to latest Smithay commit (also updating `drm`, `gbm`.) I've left `cursor_position_hint` doing nothing (which should match previous behavior) since I don't see how to actually reliably get absolute position of a `WlSurface`.

PR #889, opened 2024-09-28
focus: Un-activate non-active workspaces

Drakulix:


wiiznokes:
In my test, windows are not deactivated when switching workspace

wash2 (approved):

PR #890, opened 2024-09-28
Fix context menu styling

git-f0x:
This makes the menu match other context menus in COSMIC, and ensures some spacing between the shortcut and title. Tried to make the menu change with Density (hence the libcosmic update), but it didn't seem to change anything, regardless of the way of incorporating the spacing variables.
screenshot-2024-09-28-11-54-22
(Edit: Not sure why the text color of the title is the same as the shortcut. I've tried switching the title back to how it was, but it didn't change. But this is also the case in menus in apps)

Also tweaked the resize indicators for better styling. The right indicator still gets cut off slightly at smaller window widths, but not sure why that happens. 

The centering of the SSD header title has a minor issue in some apps at minimum widths (only saw it on GitHub desktop so far; other apps have higher minimum widths), but not sure how that can be avoided. Changing the `header_bar` in libcosmic can improve this, but it also messes up COSMIC app headers.
screenshot-2024-10-08-10-37-47

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -869,7 +869,7 @@ dependencies = [
 [[package]]
 name = "cosmic-config"
 version = "0.1.0"
-source = "git+https://github.com/pop-os/libcosmic/#af68a3f660402b850dfd00041372d964d3b098d7"
+source = "git+https://github.com/pop-os/libcosmic/#5306649be1cfb6c384da11e2ab25cafc4be79b14"
 dependencies = [
  "atomicwrites",
  "calloop 0.14.1",
@@ -888,7 +888,7 @@ dependencies = [
 [[package]]
 name = "cosmic-config-derive"
 version = "0.1.0"
-source = "git+https://github.com/pop-os/libcosmic/#af68a3f660402b850dfd00041372d964d3b098d7"
+source = "git+https://github.com/pop-os/libcosmic/#5306649be1cfb6c384da11e2ab25cafc4be79b14"
 dependencies = [
  "quote",
  "syn 1.0.109",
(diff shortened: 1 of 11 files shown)


Drakulix:
> This makes the menu match other context menus in COSMIC, and ensures some spacing between the shortcut and title.

Thanks for working on this!

> Tried to make the menu change with Density (hence the libcosmic update), but it didn't seem to change anything, regardless of the way of incorporating the spacing variables.

Possibly this isn't reload correctly (or even initially), since we can't rely on libcosmic's App implementation to do that for us. E.g.: https://github.com/pop-os/cosmic-comp/blob/master/src/shell/mod.rs#L3700

> Also tweaked the resize indicators for better styling. The right indicator still gets cut off slightly at smaller window widths, but not sure why that happens.

If that is not a new problem we can ignore that for now.

> (hence the libcosmic update)

Libcosmic updates have broken things in cosmic comp in the past given the unusual integration.
It seems like you already tested a lot of the UI elements, but please make sure you tested all of them:
- Window decorations
- Stack decorations
- Context menu
- Resize indicator
- Swap indicator (Triggered in tiling with the swap shortcut)
- Stack indicator (Triggered in tiling when drag & dropping a window on top of another one)

git-f0x:
Everything listed seems to work as expected.
Another issue that was present before (and still is), is that the `Resize` context menu item doesn't extend to the edges and has less height, compared to other items. But that's just a minor visual issue.

Drakulix:
> Everything listed seems to work as expected. :)

Great. Good for merge then imo.

> Another issue that was present before (and still is), is that the `Resize` context menu item doesn't extend to the edges and has less height, compared to other items (as if it has some kind of padding around itself). But that's just a minor visual issue. Edit: Setting [this](https://github.com/pop-os/cosmic-comp/blob/be38da47e9d778e3917b256faf6870750bb7a4e3/src/shell/grabs/menu/item.rs#L88) to 0.0 makes the item look as expected (but looking at the button style in libcosmic, it should already be 0.0?). Should that be changed?

If it works with explicit `0.0` and not otherwise, I would rather figure out, why the theme is giving us wrong values here. But that can be done in another PR.

Drakulix (approved):

iamkartiknayak:
Context menu still doesn't respect the style (`Square`, `Rounded` or `Slightly round`) from Settings

git-f0x:
The selection highlights do change, so this might be a simple fix (other context menus have the same issue). I'll see if I can fix it.

git-f0x:
This should fix it: #935.

PR #901, opened 2024-10-03
Set default output transform based on DRM `panel orientation`

ids1024:
This works, but with multiple outputs results in overlapping outputs in the display configuration (which has weird effects). So I guess that will also need a change go consider the orientation.

### DIFF
--- a/src/backend/kms/device.rs
+++ b/src/backend/kms/device.rs
@@ -645,10 +645,10 @@ fn populate_modes(
         output.add_mode(mode);
     }
     output.set_preferred(output_mode);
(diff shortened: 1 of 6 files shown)


Drakulix:
> This works, but with multiple outputs results in overlapping outputs in the display configuration (which has weird effects). So I guess that will also need a change go consider the orientation.

Should we leave this PR pending then until the algorithm is fixed? Otherwise this might be a regression of functionality.

ids1024:
I added a commit that should fix that, so I think this is an overall improvement now.

I think we've had mention in the past of issues with (manually configured) vertical monitors that maybe this could also help with. Not sure.

PR #908, opened 2024-10-05
use `Grabbing` cursor instead of `Grab` cursor

m4rch3n1ng:
when i implemented the cursor-shape-v1 protocol in #844 i didn't notice that the previous `CursorShape::Grab` actually mapped to `CursorIcon::Grabbing` instead of `CursorIcon::Grab`. this fixes that "regression".


Drakulix (approved):
Thanks!

Issue #698, opened 2024-08-07
Notifications behind applet drop-down prevents drop-down from being interacted with


jacobgkau:
Not sure what component(s) this would involve, but I'm starting it off here since it affects applets.

If I have a notification showing up behind a place I want to click within an applet menu, clicking closes the applet menu instead of doing what it's supposed to.

For example, in the video below, I try to open the "Visible Wireless Networks" list. The first time, with no notifications, it works; the second time, the network applet closes when I click, because a notification was behind where I clicked.

https://github.com/user-attachments/assets/3c3b8927-07a5-4c10-b928-55879bde987f

WatchMkr:
This appears to be fixed. Perhaps it was proper popups in cosmic-applets.

PR #918, opened 2024-10-10
Refactor and unify {render, input}-order

Drakulix:
Draft because this needs more testing.

In theory this should fix:
- https://github.com/pop-os/cosmic-comp/issues/769
- https://github.com/pop-os/cosmic-comp/issues/698
- and probably some more similar issues

Essentially this tries to use common code for iterating through all of the surfaces/clients/whatever that can be rendered and receive input.

Unfortunately this means dropping `SplitRenderElements`, as that was/is a source of disagreement between the input-code (which doesn't have a similar method) and the render-code.

As such this duplicates a bit of code, but usually the popup code path is much shorter anyway, so we introduce separate code paths for popups and toplevels all over.

This also cleans up some of the code introduced as part of the focus-follows-cursor (and vise versa) mode, which duplicated some more of the input logic. Everything should now either use `State::surface_under` or `State::element_under` depending if it would affect pointer or keyboard focus.

Initial tests seem promising and don't show any obvious breakage, however this definitely needs more testing and possibly some in-depth reviews, though I would ideally like to merge this before it starts to seriously bitrot.

### DIFF
--- a/src/shell/focus/order.rs
+++ b/src/shell/focus/order.rs
@@ -1,0 +1,374 @@
+use std::{ops::ControlFlow, time::Instant};
+
+use cosmic_comp_config::workspace::WorkspaceLayout;
+use keyframe::{ease, functions::EaseInOutCubic};
+use smithay::{
+    desktop::{layer_map_for_output, LayerSurface, PopupKind, PopupManager},
+    output::{Output, OutputNoMode},
+    utils::{Logical, Point},
+    wayland::{session_lock::LockSurface, shell::wlr_layer::Layer},
+    xwayland::X11Surface,
+};
+
+use crate::{
+    backend::render::ElementFilter,
+    shell::{
+        layout::{floating::FloatingLayout, tiling::ANIMATION_DURATION},
+        Shell, Workspace, WorkspaceDelta,
+    },
+    utils::{geometry::*, prelude::OutputExt, quirks::WORKSPACE_OVERVIEW_NAMESPACE},
+    wayland::protocols::workspace::WorkspaceHandle,
+};
+
+pub enum Stage<'a> {
+    SessionLock(Option<&'a LockSurface>),
+    LayerPopup {
+        layer: LayerSurface,
+        popup: &'a PopupKind,
+        location: Point<i32, Global>,
+    },
+    LayerSurface {
+        layer: LayerSurface,
+        location: Point<i32, Global>,
+    },
+    OverrideRedirect {
+        surface: &'a X11Surface,
+        location: Point<i32, Global>,
+    },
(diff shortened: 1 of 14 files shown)


git-f0x:
Testing this with floating windows, it seems I can't click on anything outside a rectangle in the top-left corner slightly larger than a quarter of the screen. Though the size of the clickable area seems to be variable (I think it changes when maximizing and unmaximizing an app?).
Tiling and maximized windows seem to behave as expected.

Drakulix:
Thanks for testing! Last commit should fix the bugginess around floating windows.

git-f0x:
The unclickable area issue seems to be fixed! Though there still seem to be some focus issues, where clicking on a window in the background doesn't focus it (though they are interactable), and input sometimes goes to the window below.
screenshot-2024-10-11-16-03-23
Clicking on the `cosmic files` or Nautilus window here doesn't focus them and bring them to the front.
While trying to attach this screenshot, clicking on the **Open** button in the `cosmic-files` dialog, just focused the Firefox window below, until the dialog was moved towards the top-left screen corner.

git-f0x:
I'm not encountering any issues from some initial testing with the latest changes.

ids1024:
I guess `render_input_order` should just return an iterator, but that's maybe awkward until rust has a generator syntax with `yield`. Seems like a good way to do things for now.

src/shell/focus/order.rs
```diff
@@ -0,0 +1,394 @@
...
+    shell::{
+        layout::{floating::FloatingLayout, tiling::ANIMATION_DURATION},
+        Shell, Workspace, WorkspaceDelta,
+    },
+    utils::{geometry::*, prelude::OutputExt, quirks::WORKSPACE_OVERVIEW_NAMESPACE},
+    wayland::protocols::workspace::WorkspaceHandle,
+};
+
+pub enum Stage<'a> {
+    SessionLock(Option<&'a LockSurface>),
```
ids1024, line 24: Would it make sense to not use an `Option` here, and just never invoke the callback if there is no session lock surface for the output?

At least as things work currently, it looks like that would behave the same. `session_lock_elements` produces no elements if its argument is `None`. `surface_under` and `element_under` should return `None`.

Drakulix: I think this is correct, if we have a lock, we are not supposed to render anything else, e.g. in case the client dies.

Drakulix:
@git-f0x Good observation, the new code failed to take fullscreen surfaces correctly into account.

Drakulix:
Seems to work fine after a day of extensive testing / running in production. Lets merge

PR #920, opened 2024-10-11
Fixes for behavior when workspaces are moved back to on output on hotplug

ids1024:
I've been annoyed by how this adds an empty initial workspace which ends up focused instead of one of the non-empty ones. For some reason that's still happening... I also see some other things that could be fixed.

It seems there also is (sometimes?) an issue where a window becomes unresponsive after being moved. Probably a frame callback issue... I need to look into that more.

### DIFF
--- a/src/shell/mod.rs
+++ b/src/shell/mod.rs
@@ -678,6 +678,9 @@ impl Workspaces {
             if set.workspaces.is_empty() {
                 set.add_empty_workspace(workspace_state);
             }
+            for (i, workspace) in set.workspaces.iter_mut().enumerate() {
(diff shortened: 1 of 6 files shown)


hojjatabdollahi:
Also it messes up the workspace numbers when you plug your monitors back in.

image

ids1024:
https://github.com/pop-os/cosmic-comp/pull/920/commits/cc6dace1fb9e3c7424b533f279d4153901f3cefb should fix the workspace numbering.

Perhaps some of this could be manged a bit differently to make sure it stays in sync without keeping track of things in multiple places.

PR #935, opened 2024-10-21
[Menu] Use corner radius from theme

git-f0x:
Just makes the context menu use `@radius_s` for the corner radius, to match the highlight for items inside.
I also tried to make the stack header do the same, but it didn't seem to pick up roundness changes.


Drakulix (approved):
LGTM, thanks!

Drakulix:
What exactly did you try to change the stack? Would like to take a look at that.

git-f0x:
I changed the 8.0 [here](https://github.com/pop-os/cosmic-comp/blob/50d6dc3d21c5c373d81f294d9b869a76205840a2/src/shell/element/stack.rs#L927) to `radius_s` (`theme::active().cosmic().radius_s()[0]`).

PR #948, opened 2024-10-24
set cursor shape for resize grabs

m4rch3n1ng:
similar to the `MoveGrab` <sup>[1]</sup> <sup>[2]</sup>, this sets the `CursorIcon` shape in `ResizeSurfaceGrab::new` and unsets it in the `Drop` impl.

the match over `ResizeEdge` may not be exhaustive, but currently these are the only values being passed into the function.

[1]: https://github.com/pop-os/cosmic-comp/blob/9a9eddb132a308bd4de17a143e1e4919bfc9c7b1/src/shell/grabs/moving.rs#L715-L718
[2]: https://github.com/pop-os/cosmic-comp/blob/9a9eddb132a308bd4de17a143e1e4919bfc9c7b1/src/shell/grabs/moving.rs#L855-L858

### DIFF
--- a/src/shell/layout/floating/grabs/resize.rs
+++ b/src/shell/layout/floating/grabs/resize.rs
@@ -3,6 +3,7 @@
 use std::sync::atomic::{AtomicBool, Ordering};
 
 use crate::{
+    backend::render::cursor::CursorState,
     shell::{
         element::CosmicMapped,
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
LGTM, thanks!

Issue #943, opened 2024-10-23
wl-paste causes a crash


hojjatabdollahi:
Calling wl-paste in the terminal causes a crash.  Here is the error that pops up in the logs:

```
ct 23 15:55:36 cosmic cosmic-comp[5381]: Toplevel for foreign-toplevel-list not registered for cosmic-toplevel-info.
Oct 23 15:55:36 cosmic cosmic-comp[5381]: thread 'main' panicked at 'Callback creating object zcosmic_toplevel_handle_v1@51[2] did not provide any object data.': /build/cosmic-comp-pj8uhi/cosmic-comp-0.1~1729697706~24.04~9>
                                             0: <unknown>
                                             1: <unknown>
                                             2: <unknown>
                                             3: <unknown>
                                             4: <unknown>
                                             5: <unknown>
                                             6: <unknown>
                                             7: <unknown>
                                             8: <unknown>
                                             9: <unknown>
                                            10: <unknown>
                                            11: <unknown>
                                            12: <unknown>
                                            13: <unknown>
                                            14: __libc_start_main
                                            15: <unknown>
Oct 23 15:55:36 cosmic cosmic-comp[5381]: thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: PoisonError { .. }': /build/cosmic-comp-pj8uhi/cosmic-comp-0.1~1729697706~24.04~9a9eddb/vendor/wayland-backe>
                                             0: <unknown>
                                             1: <unknown>
                                             2: <unknown>
                                             3: <unknown>
                                             4: <unknown>
                                             5: <unknown>
                                             6: <unknown>
                                             7: <unknown>
                                             8: <unknown>
                                             9: <unknown>
                                            10: <unknown>
                                            11: <unknown>
                                            12: <unknown>
                                            13: <unknown>
                                            14: <unknown>
                                            15: <unknown>
                                            16: <unknown>
                                            17: <unknown>
                                            18: <unknown>
                                            19: __libc_start_main
                                            20: <unknown>
Oct 23 15:55:36 cosmic cosmic-comp[5381]: thread 'main' panicked at 'panic in a destructor during cleanup': library/core/src/panicking.rs:227
                                             0: <unknown>
                                             1: <unknown>
                                             2: <unknown>
                                             3: <unknown>
                                             4: <unknown>
                                             5: <unknown>
                                             6: <unknown>
                                             7: <unknown>
                                             8: <unknown>
                                             9: <unknown>
[...]

PR #949, opened 2024-10-24
toplevel_info: Fix racy data creation

Drakulix:
Potential fix for #943

### DIFF
--- a/src/wayland/protocols/toplevel_info.rs
+++ b/src/wayland/protocols/toplevel_info.rs
@@ -90,7 +90,7 @@ pub struct ToplevelHandleStateInner<W: Window> {
     title: String,
     app_id: String,
     states: Vec<States>,
-    pub(super) window: W,
+    pub(super) window: Option<W>,
 }
 pub type ToplevelHandleState<W> = Mutex<ToplevelHandleStateInner<W>>;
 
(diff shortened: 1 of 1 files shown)


ids1024:
I no longer see a crash with `wl-paste`, even with `while true; do wl-paste; done`, so it seems like the race condition is gone.

src/wayland/protocols/toplevel_info.rs
```diff
@@ -509,10 +526,14 @@ where
...
-            .iter()
-            .flat_map(|state| (*state as u32).to_ne_bytes())
-            .collect::<Vec<u8>>();
+        let states: Vec<u8> = {
+            let ratio = std::mem::size_of::<States>() / std::mem::size_of::<u8>();
+            let ptr = states.as_mut_ptr() as *mut u8;
+            let len = states.len() * ratio;
+            let cap = states.capacity() * ratio;
+            std::mem::forget(states);
+            unsafe { Vec::from_raw_parts(ptr, len, cap) }
```
ids1024, line 535: Why was this changed back to use `Vec::from_raw_parts`?

It is unsound to use `from_raw_parts` with a type that has a different alignment.

Drakulix: rebase mistake. will remove before merge

PR #969, opened 2024-10-31
Selectively disable direct-scanout to reduce bandwidth

Drakulix:
So this is an attempt for a problem, that has been bugging me for quite some time, but I am not convinced this is a good implementation...

## The problem

So we have run into a few issues with amdgpu cards, where certain modes were not selectable, e.g.:
- https://github.com/pop-os/cosmic-epoch/issues/687
- https://github.com/pop-os/cosmic-epoch/issues/560
- .. (more, but I am unable to quickly find them)

The underlying problem, as I was able to confirm with drm_info logs and talking to some AMD engineers seems to be, that we are using overlay planes. Or rather, that overlay planes contribute to bandwidth usage, making new configurations with higher requirements (faster / more high-res mode) potentially impossible without disabling some planes. Unfortunately the drm-api currently has no way to communicate why a commit failed, so we just reject the new mode.

**Solution 1**: So the initial idea would be to disable planes on a given crtc before attempting to modeset. `compositor.clear()` seems to be an obvious solution here, given we modeset anyway.

Unfortunately that **doesn't** work. Roaming planes (e.g. planes that can be attached to multiple crtcs) exist and given this runs concurrently other `DrmCompositor`s might be able to grab the just released planes before we are done with modesetting. Fun. Also this problem can apply to enabling completely new outputs.

**Solution 2**: So we need to synchronize this for the whole device?

**Almost works**, though `compositor.clear` is pretty ugly here, as we would be blanking all surfaces for but I have also seen another variant of this issue, where drm leasing fails, because we are using too much bandwidth. So now we suddenly need to keep resource usage low, while another process is involved, because we have no (good) way to asses how much bandwidth that adds at all. At this point we just want to disable overlay plane usage (though just disabling direct-scanout for now is roughly equivalent) for both the modeset (to avoid flicker) and while a lease is active.

**Solution 3**: This PR.

## Solution

- b5382896c9bfd37172d4ce88306ff60ca5dc0218 - Add a way to ask the thread to disable direct scanout **and** notify us, once the next (forced) frame returned, so we can be sure the planes are actually not in use. I don't like the complexity this commit introduces.
- 458f668a1b7a2a3c1be1860f4f155a4cade5f4b5 - Disable direct scanout on all surfaces of a device before applying a new output configuration, re-enable afterwards - if applicable
- a46be899f96abc75af38fc834045a097cd0bb7f6 - Disable direct scanout while a lease is active

## Alternatives

Perhaps we should instead synchronize by terminating the surface threads, doing everything on the main thread and restarting the threads. We clearly need more access to the `DrmCompositor`s state and that is tricky. So instead maybe we should just add a way to destruct a `DrmCompositor` into it's `DrmSurface` and make that `Send`. That way we should hopefully be able to still avoid flickering, though we might need to still somehow obtain a fully-composited rendered image to not break the visuals for a frame while disabling planes.

This would also help us solve another issue in the future, that wlroots recently tackled:
- https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4548
- https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4567

Unfortunately it looks like some gpus are so resource constraint, that we need more logic to even enable all possible displays without any planes in use. And we need device-commits to make this work at all, which complicates synchronizing even more and makes shutting down the threads temporarily even more appealing.

That is a much more significant rework of both smithay code and cosmic-comp code though.

cc @ids1024 for general comments on both the problem and code. @cmeissl for ideas on the problem and the smithay side of things.

Draft while design isn't worked out and testing so far as only been done to verify, that this doesn't break anything, not to verify, that this actually fixes these bugs. (Lacking hardware that reproduces this.)

### DIFF
--- a/src/backend/kms/mod.rs
+++ b/src/backend/kms/mod.rs
@@ -35,7 +35,8 @@ use std::{
     borrow::BorrowMut,
     collections::{HashMap, HashSet},
     path::Path,
-    sync::{atomic::AtomicBool, Arc, RwLock},
+    sync::{atomic::AtomicBool, mpsc::sync_channel, Arc, RwLock},
+    time::Duration,
 };
 
 mod device;
@@ -530,6 +531,17 @@ impl KmsState {
 
         let mut all_outputs = Vec::new();
         for device in self.drm_devices.values_mut() {
+            // we want to lower the resource usage to make sure we can modeset and enable new outputs
+            let mut receivers = Vec::new();
+            for surface in device.surfaces.values_mut() {
+                let (tx, rx) = sync_channel(1);
+                surface.set_direct_scanout(false, tx);
+                receivers.push(rx);
+            }
+            for rx in receivers {
+                let _ = rx.recv_timeout(Duration::from_millis(100));
+            }
+
             // we only want outputs exposed to wayland - not leased ones
             // but that is also not all surface, because that doesn't contain all detected, but unmapped outputs
             let outputs = device
@@ -690,6 +702,12 @@ impl KmsState {
                         w += output.config().transformed_size().w as u32;
                     }
                     all_outputs.push(output);
+                }
+            }
+
+            if !test_only && device.active_leases.is_empty() {
+                for surface in device.surfaces.values_mut() {
+                    surface.set_direct_scanout(true, sync_channel(0).0);
                 }
             }
 
--- a/src/backend/kms/surface/mod.rs
+++ b/src/backend/kms/surface/mod.rs
@@ -124,6 +124,7 @@ pub struct SurfaceThreadState {
     primary_node: DrmNode,
     target_node: DrmNode,
     active: Arc<AtomicBool>,
+    allow_direct_scanout: bool,
     compositor: Option<GbmDrmCompositor>,
 
     state: QueueState,
@@ -197,23 +198,67 @@ pub type GbmDrmCompositor = DrmCompositor<
 pub enum QueueState {
     Idle,
     /// A redraw is queued.
-    Queued(RegistrationToken),
+    Queued {
+        queued_render: RegistrationToken,
+        pending_syncs: Vec<SyncSender<()>>,
+    },
     /// We submitted a frame to the KMS and waiting for it to be presented.
     WaitingForVBlank {
         redraw_needed: bool,
+        pending_syncs: Vec<SyncSender<()>>,
     },
     /// We did not submit anything to KMS and made a timer to fire at the estimated VBlank.
-    WaitingForEstimatedVBlank(RegistrationToken),
+    WaitingForEstimatedVBlank {
+        estimated_vblank: RegistrationToken,
+    },
     /// A redraw is queued on top of the above.
     WaitingForEstimatedVBlankAndQueued {
         estimated_vblank: RegistrationToken,
         queued_render: RegistrationToken,
+        pending_syncs: Vec<SyncSender<()>>,
     },
 }
 
 impl Default for QueueState {
     fn default() -> Self {
         QueueState::Idle
+    }
+}
+
+impl QueueState {
+    fn clear(&mut self, evlh: &LoopHandle<SurfaceThreadState>, timings: &mut Timings) {
+        match std::mem::replace(self, QueueState::Idle) {
+            QueueState::Idle => {}
+            QueueState::Queued {
+                queued_render,
+                pending_syncs,
+            } => {
+                evlh.remove(queued_render);
+                for sync in pending_syncs {
+                    let _ = sync.send(()); // compositor.clear satisfies the requirement
+                }
+            }
+            QueueState::WaitingForEstimatedVBlank { estimated_vblank } => {
+                evlh.remove(estimated_vblank);
+            }
+            QueueState::WaitingForVBlank { pending_syncs, .. } => {
+                timings.discard_current_frame();
+                for sync in pending_syncs {
+                    let _ = sync.send(()); // compositor.clear satisfies the requirement
+                }
+            }
+            QueueState::WaitingForEstimatedVBlankAndQueued {
+                estimated_vblank,
+                queued_render,
+                pending_syncs,
+            } => {
+                evlh.remove(estimated_vblank);
+                evlh.remove(queued_render);
+                for sync in pending_syncs {
+                    let _ = sync.send(()); // compositor.clear satisfies the requirement
+                }
+            }
+        };
     }
 }
 
@@ -239,6 +284,7 @@ pub enum ThreadCommand {
     VBlank(Option<DrmEventMetadata>),
     ScheduleRender,
     SetMode(Mode, SyncSender<Result<()>>),
+    SetDirectScanout(bool, SyncSender<()>),
     End,
     DpmsOff,
 }
@@ -401,6 +447,16 @@ impl Surface {
         rx.recv().context("Surface thread died")?
     }
 
+    pub fn set_direct_scanout(&mut self, allow: bool, sync: SyncSender<()>) {
+        if self
+            .thread_command
+            .send(ThreadCommand::SetDirectScanout(allow, sync.clone()))
+            .is_err()
+        {
+            let _ = sync.send(());
+        }
+    }
+
     pub fn suspend(&mut self) {
         let _ = self.thread_command.send(ThreadCommand::Suspend);
     }
@@ -502,6 +558,7 @@ fn surface_thread(
         target_node,
         active,
         compositor: None,
+        allow_direct_scanout: true,
 
         state: QueueState::Idle,
         timings: Timings::new(None, false),
@@ -549,7 +606,7 @@ fn surface_thread(
                     return;
                 }
 
-                state.queue_redraw(false);
+                state.queue_redraw(false, Vec::new());
             }
             Event::Msg(ThreadCommand::UpdateMirroring(mirroring_output)) => {
                 state.update_mirroring(mirroring_output);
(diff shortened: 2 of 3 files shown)


cmeissl:
Okay, I am also procrastinating on exactly this topic for the last few weeks...so this is a quick brain dump:

## General

What I have come up with so far (partially some PoC impls, but mostly just ideas):
- Adding a `mode` parameter to `DrmCompositor::render_frame` allowing to force composition
- Adding a `DrmCompositor::commit_frame` function allowing to make a blocking commit to make sure a specific state is reached.
- Adding a `DrmCompositor::with_format` function in addition to the existing `DrmCompositor::new` allowing to force a format with a specific set of modifiers.
- Adding a `DrmCompositor::set_format` function allowing to change the swapchain format.
- Adding a function on `DrmDevice` to test a combination of CRTC with certain primary plane formats
- Adding a `DrmDeviceManager` which internally manages all `DrmCompositor` with something like a `add_crtc` function.

## DrmDeviceManager

Using an `RwLock` holding a list of current active CRTC/DrmCompositor combinations. `add_crtc` would return a token to the CRTC allowing access to the `DrmCompositor` which would take a read lock allowing concurrent access.
`add_crtc` could evaluate the format(s) with `DrmDevice::test_format` and take a write lock changing formats "atomically" on all crtc.

Before adding a crtc we might need to make sure there is no direct scan-out in place. This could be done by `clear` or by re-rendering all crtc under the write lock + `DrmCompositor::commit_frame`. So changing format might require rendering twice, once with the existing format but with mode Composition and again with the changed format. After that each thread could continue as usual including assigning overlay planes.

## Other

There is still an issue with doing direct scan-out on the primary plane and a second CRTC. It could (actually I have seen this happen...) happen that the direct scan-out on CRTC 1 reduces the overall bandwidth requirement allowing CRTC 2 to use an additional overlay plane. But going back to composition on CRTC 1 might then fail...
I have thought about that extensively and have come up with two possible solutions:
- Always include all other CRTC with their format of the pimary plane during atomic testing
- Allow to limit which format is allowed for direct scan-out on the primary plane (probably just using the same format/modifier as the swapchain has)

The first option might fail in funny ways and the second sounds way easier. The second one could be implemented with something similar to the `Capabilities` we have in the gles renderer.
(Could enable/disable that depending on count of active CRTC in `DrmDeviceManager`)

cmeissl:
Some kind of use_mode wrapper in the device token could use a similar approach. First optimistic by just calling use_mode on the compositor and when that fails take a write lock and retry after making sure all crtc do not use direct scan-out. Issues is more or less the same, we might have to redraw all crtc (or at least commit some buffer)

Drakulix:
> What I have come up with so far (partially some PoC impls, but mostly just ideas):

+1 for extending the `DrmCompositor` interface, making it more suitable for the complex reality of these issues.

>     * Adding a `mode` parameter to `DrmCompositor::render_frame` allowing to force composition
+1
 
>     * Adding a `DrmCompositor::commit_frame` function allowing to make a blocking commit to make sure a specific state is reached.
+1

>     * Adding a `DrmCompositor::with_format` function in addition to the existing `DrmCompositor::new` allowing to force a format with a specific set of modifiers.
So this is to aid a similar approach to wlroots swapchain helper by forcing a specific modifier set?
 
>     * Adding a `DrmCompositor::set_format` function allowing to change the swapchain format.
When would this be used? To fall back to 8-bit for more bandwidth?
 
>     * Adding a function on `DrmDevice` to test a combination of CRTC with certain primary plane formats
Not sure I get the picture here, is this to build the device-wide test commit?

 
> ## DrmDeviceManager
> 
> Using an `RwLock` holding a list of current active CRTC/DrmCompositor combinations. `add_crtc` would return a token to the CRTC allowing access to the `DrmCompositor` which would take a read lock allowing concurrent access. `add_crtc` could evaluate the format(s) with `DrmDevice::test_format` and take a write lock changing formats "atomically" on all crtc.

+1 for the approach of a sharable (assuming `DrmOutputManager` would be `Send+Sync`) manager.
Not sure we actually need an `RwLock` though, we don't need access on more than one thread in parallel, so a simple `Mutex` might be enough?

> Before adding a crtc we might need to make sure there is no direct scan-out in place. This could be done by `clear` or by re-rendering all crtc under the write lock + `DrmCompositor::commit_frame`. So changing format might require rendering twice, once with the existing format but with mode Composition and again with the changed format. After that each thread could continue as usual including assigning overlay planes.

+1

> There is still an issue with doing direct scan-out on the primary plane and a second CRTC. It could (actually I have seen this happen...) happen that the direct scan-out on CRTC 1 reduces the overall bandwidth requirement allowing CRTC 2 to use an additional overlay plane. But going back to composition on CRTC 1 might then fail... I have thought about that extensively and have come up with two possible solutions:

Oh yeah, I forgot about this one...
 
>     * Always include all other CRTC with their format of the pimary plane during atomic testing
This could then fail scanout later, we essentially would need to make two tests, right? I like the idea conceptionally, but I also fear this adds to much complexity.
 
>     * Allow to limit which format is allowed for direct scan-out on the primary plane (probably just using the same format/modifier as the swapchain has)

I feel like this might potentially be too limiting, I don't think there is any guarantee of gbm selecting the same modifier as a client using vulkan/egl might, so we might leave a bunch of performance on the table here...

I guess we could limit the `format`, but not the `modifier` and if fallback fails re-create the swapchain with the same modifier as the previous scan-out modifier? I wonder if that could have  weird performance effects, but it shouldn't be too different from selecting modifiers based on bandwidth for enabling more outputs as wlroots is doing with the helper.

cmeissl:
> > What I have come up with so far (partially some PoC impls, but mostly just ideas):
> 
> +1 for extending the `DrmCompositor` interface, making it more suitable for the complex reality of these issues.
> 
> > ```
> > * Adding a `mode` parameter to `DrmCompositor::render_frame` allowing to force composition
> > ```
> 
> +1
> 
> > ```
> > * Adding a `DrmCompositor::commit_frame` function allowing to make a blocking commit to make sure a specific state is reached.
> > ```
> 
> +1
> 
> > ```
> > * Adding a `DrmCompositor::with_format` function in addition to the existing `DrmCompositor::new` allowing to force a format with a specific set of modifiers.
> > ```
> 
> So this is to aid a similar approach to wlroots swapchain helper by forcing a specific modifier set?

Yeah, in the first version also probably falling back to `Invalid` if we fail to initialize a CRTC after making sure
we do not use overlay planes.

> 
> > ```
> > * Adding a `DrmCompositor::set_format` function allowing to change the swapchain format.
> > ```
> 
> When would this be used? To fall back to 8-bit for more bandwidth?

Basically doing the same as `DrmCompositor::with_format`, but just for already initialized ones. I also thought
about destructing existing ones into their `DrmSurface`. But after thinking about it I do not see any benefit.
The destruction will make it harder to use because it requires introducing some `Options` here and there and
it also will drop the internal fb caches. I also thought about adding a builder which could transfer some of the
state with something like `DrmCompositor::builder()` and `DrmCompositor::into_builder(self)`, but in the end I don't
think it is worth atm.

> 
> > ```
> > * Adding a function on `DrmDevice` to test a combination of CRTC with certain primary plane formats
> > ```
> 
> Not sure I get the picture here, is this to build the device-wide test commit?

Yeah, we need to test all CRTC at once which can be a bit tricky without accessing internal state of `DrmDevice`.
So I thought about adding a function that can exactly do that, issue an atomic test commit enabling all provided
CRTC/Encoders with the specified format/modifier per CRTC and disabling all planes.

> 
> > ## DrmDeviceManager
> > Using an `RwLock` holding a list of current active CRTC/DrmCompositor combinations. `add_crtc` would return a token to the CRTC allowing access to the `DrmCompositor` which would take a read lock allowing concurrent access. `add_crtc` could evaluate the format(s) with `DrmDevice::test_format` and take a write lock changing formats "atomically" on all crtc.
> 
> +1 for the approach of a sharable (assuming `DrmOutputManager` would be `Send+Sync`) manager. Not sure we actually need an `RwLock` though, we don't need access on more than one thread in parallel, so a simple `Mutex` might be enough?
> 

Not exactly, the idea was to have it once and only a single place which is responsible for enabling CRTC. Adding a crtc would return an `DrmOutput` which would be `Send` but not `Sync`. We can use some tricks here to mitigate needed locks, especially for single threaded users. 

In my Poc `DrmOutputManager` internally holds this:

```
type CompositorList<A, F, U, G> = Arc<RwLock<HashMap<crtc::Handle, RefCell<DrmCompositor<A, F, U, G>>>>>;
```

and `DrmOutput` is basically a wrapper around the CRTC and the list.

```
struct DrmOutput<A, F, U, G>
where
    A: Allocator,
    F: ExportFramebuffer<<A as Allocator>::Buffer>,
    <F as ExportFramebuffer<<A as Allocator>::Buffer>>::Framebuffer: std::fmt::Debug + 'static,
    G: AsFd + 'static,
{
    crtc: crtc::Handle,
    compositor: CompositorList<A, F, U, G>,
}
```

> > Before adding a crtc we might need to make sure there is no direct scan-out in place. This could be done by `clear` or by re-rendering all crtc under the write lock + `DrmCompositor::commit_frame`. So changing format might require rendering twice, once with the existing format but with mode Composition and again with the changed format. After that each thread could continue as usual including assigning overlay planes.
> 
> +1
> 
> > There is still an issue with doing direct scan-out on the primary plane and a second CRTC. It could (actually I have seen this happen...) happen that the direct scan-out on CRTC 1 reduces the overall bandwidth requirement allowing CRTC 2 to use an additional overlay plane. But going back to composition on CRTC 1 might then fail... I have thought about that extensively and have come up with two possible solutions:
> 
> Oh yeah, I forgot about this one...
> 
> > ```
> > * Always include all other CRTC with their format of the pimary plane during atomic testing
> > ```
> 
> This could then fail scanout later, we essentially would need to make two tests, right? I like the idea conceptionally, but I also fear this adds to much complexity.
> 

Yeah and I am also not sure this will work on all devices tbh.

> > ```
> > * Allow to limit which format is allowed for direct scan-out on the primary plane (probably just using the same format/modifier as the swapchain has)
> > ```
> 
> I feel like this might potentially be too limiting, I don't think there is any guarantee of gbm selecting the same modifier as a client using vulkan/egl might, so we might leave a bunch of performance on the table here...
> 
> I guess we could limit the `format`, but not the `modifier` and if fallback fails re-create the swapchain with the same modifier as the previous scan-out modifier? I wonder if that could have weird performance effects, but it shouldn't be too different from selecting modifiers based on bandwidth for enabling more outputs as wlroots is doing with the helper.

Not sure this will work, I have seen CRTC fail to enable just because of the modifier. Afaik the intel CCS modifier requires more bandwidth and could again cause this issue when direct scan-out uses something like Y-Tiled. But on the other hand
I also have no good idea here...maybe we can just try to disable the capability in response to failed during render_frame.

Drakulix:
> > I feel like this might potentially be too limiting, I don't think there is any guarantee of gbm selecting the same modifier as a client using vulkan/egl might, so we might leave a bunch of performance on the table here...
> > I guess we could limit the `format`, but not the `modifier` and if fallback fails re-create the swapchain with the same modifier as the previous scan-out modifier? I wonder if that could have weird performance effects, but it shouldn't be too different from selecting modifiers based on bandwidth for enabling more outputs as wlroots is doing with the helper.
> 
> Not sure this will work, I have seen CRTC fail to enable just because of the modifier. Afaik the intel CCS modifier requires more bandwidth and could again cause this issue when direct scan-out uses something like Y-Tiled. But on the other hand I also have no good idea here...maybe we can just try to disable the capability in response to failed during render_frame.

I think I had a typo here in my original response, that made this somewhat hard to pass.

Lets say we let this happen and once we fall-back to composition we fail because of the modifier of our swapchain. Couldn't we re-create the swapchain with the modifier previously in use when we did direct scanout?

PR #978, opened 2024-11-06
design: use active_window_hint as the thickness for floating window snap outline

ryanabx:
This commit changes the snapping indicator's thickness to match the active window hint, per design recommendation by Maria. The thickness for this outline never goes under 1, also per Maria's spec.

screenshot-2024-11-06-02-53-33


Drakulix (approved):
Thanks!

PR #982, opened 2024-11-07
WIP atspi: `add_virtual_modifier`

ids1024:
https://github.com/pop-os/cosmic-protocols/pull/41, https://github.com/pop-os/at-spi2-core/pull/1.

There seems to be a race condition in Orca where it removes all key grabs, then adds a new set of key grabs. As this previously worked, this briefly removed the virtual modifier, which could result in a keypress being sent. (Which could send a caps lock.)

I need to do more testing, and see how other things are working and what other improvements may help, but this seems like a good idea. The virtual modifier is added once at the start.

The fact key input can race with orca/compositor changes to grabs seems bad in general, but I guess that's a limitation of how `AtspiDevice` is currently designed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -899,7 +899,7 @@ dependencies = [
 [[package]]
 name = "cosmic-protocols"
 version = "0.1.0"
-source = "git+https://github.com/pop-os/cosmic-protocols?branch=main#ec1616b90fa6b4568709cfe2c0627b1e8cc887e0"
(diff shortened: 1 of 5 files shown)


Drakulix (commented):
Seems fine to me at first glance

ids1024:
With https://github.com/pop-os/cosmic-comp/pull/1412, we are now using the new upstream DBus protocol, so this is no longer needed.

PR #983, opened 2024-11-09
Unthrottle windows in screencopy sessions

alyssais:
If an individual window was being screencast, and that window was not visible, for example because it was minimized, that window would only be rendered every 995ms, which did not look good on the screencast.

Now, non-visible windows with active screencopy sessions, as well as windows that are mapped on non-visible workspaces with active screencopy sessions, are rendered every 1/60th of a second, which matches the frame rate of the video produced by xdg-desktop-portal-cosmic.  In future, it might be good to let screencopy clients suggest a redraw rate for captured windows.

---

I've tested that minimized windows and hidden windows in a stack can now be screencast smoothly.  I haven't tested screencasts of unfocused workspaces or pointer cursor sessions, because I don't know of clients for those.

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -11,6 +11,7 @@ use crate::{
     input::{gestures::GestureState, PointerFocusState},
     shell::{grabs::SeatMoveGrabState, CosmicSurface, SeatExt, Shell},
     utils::prelude::OutputExt,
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
thx!

PR #986, opened 2024-11-13
fix: apply transform to damage before blitting

wash2:
There seems to be something off about the types of damage being passed around for screencopy. The screencopy damage is clipped because it isn't rotated, so for a rotated screen 1920x1080, it becomes 1080x1080 in `blit_frame_result` when intersected with the dmabuf geometry.  This fix can be tested with https://github.com/pop-os/xdg-desktop-portal-cosmic/pull/105 but this is really just a bandaid. Maybe `blit_frame_result` shouldn't be taking physical damage, because it copies buffer to buffer? I think the dmabuf geometry  used in `blit_frame_result` seems like it should be Buffer too, but I'm not as familiar with smithay.

### DIFF
--- a/src/backend/kms/surface/mod.rs
+++ b/src/backend/kms/surface/mod.rs
@@ -1205,13 +1205,23 @@ impl SurfaceThreadState {
                                     .into_iter()
                                     .flatten();
 
+                                // If the screen is rotated, we must convert damage to match output.
+                                let adjusted = damage.iter().copied().map(|mut d| {
+                                    d.size = d
+                                        .size
+                                        .to_logical(1)
+                                        .to_buffer(1, output_transform)
+                                        .to_logical(1, Transform::Normal)
+                                        .to_physical(1);
+                                    d
+                                });
                                 match frame_result
                                     .blit_frame_result(
                                         output_size,
                                         output_transform,
                                         output_scale,
                                         &mut renderer,
-                                        damage.iter().copied(),
+                                        adjusted,
                                         filter,
                                     )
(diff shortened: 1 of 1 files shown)


src/backend/kms/surface/mod.rs
```diff
@@ -1205,13 +1205,23 @@ impl SurfaceThreadState {
...
                                     .flatten();
 
+                                // If the screen is rotated, we must convert damage to match output.
+                                let adjusted = damage.iter().copied().map(|mut d| {
+                                    d.size = d
+                                        .size
+                                        .to_logical(1)
+                                        .to_buffer(1, output_transform)
+                                        .to_logical(1, Transform::Normal)
+                                        .to_physical(1);
```
Drakulix, line 1215: Ouch, yeah this is ugly.

I think if would possibly a bit cleaner to transform this already in line 1006 https://github.com/pop-os/cosmic-comp/pull/986/files#diff-85c7861d0290a1e31c47e8e1d31b15f37d2bb03a1b4edb046d9173eef8c9f758R1006 .
But given this is quite a hack anyway, we don't have to bother.

Drakulix (approved):
> Maybe blit_frame_result shouldn't be taking physical damage, because it copies buffer to buffer?

Yeah, you are absolutely right. Either it should also take a `Transform`, if it has to work with `Physical` coordinates for some reason and handle the necessary conversion internally or it should just take `Buffer`-coordinates (which I think would be the proper solution here).

Feel free to also make a smithay PR or at least open an issue please to track this and start a discussion.

PR #1004, opened 2024-11-18
Adaptive VRR support

Drakulix:
Draft until cosmic-randr and cosmic-settings are ready

### DIFF
--- a/src/backend/kms/surface/mod.rs
+++ b/src/backend/kms/surface/mod.rs
@@ -5,6 +5,7 @@ use crate::{
         element::{CosmicElement, DamageElement},
         init_shaders, workspace_elements, CursorMode, ElementFilter, GlMultiRenderer, CLEAR_COLOR,
     },
+    config::AdaptiveSync,
     shell::Shell,
     state::SurfaceDmabufFeedback,
     utils::{prelude::*, quirks::workspace_overview_is_open},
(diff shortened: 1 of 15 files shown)


Drakulix:
Now only blocked on smithay.

src/backend/kms/surface/mod.rs
```diff
@@ -401,6 +405,33 @@ impl Surface {
...
+        let (tx, rx) = std::sync::mpsc::sync_channel(1);
+        let _ = self
+            .thread_command
+            .send(ThreadCommand::AdaptiveSyncAvailable(tx));
+        rx.recv().context("Surface thread died")?
+    }
+
+    pub fn use_adaptive_sync(&mut self, vrr: AdaptiveSync) -> Result<bool> {
+        if vrr != AdaptiveSync::Disabled {
+            let (tx, rx) = std::sync::mpsc::sync_channel(1);
```
ids1024, line 418: I wish `std` had a oneshot channel, but I guess it's not worth adding a dependency for it when it's not already in the dependency tree.

src/config/mod.rs
```diff
@@ -98,19 +98,34 @@ pub enum OutputState {
...
 
+#[derive(Debug, Deserialize, Serialize, Clone, Copy, PartialEq)]
+#[serde(rename_all = "lowercase")]
+pub enum AdaptiveSync {
+    #[serde(rename = "true")]
+    Enabled,
+    #[serde(rename = "false")]
+    Disabled,
+    Force,
+}
```
ids1024, line 113: It seems a little odd to use a different terminology (enabled/force vs automatic/always) from the protocol, though it doesn't really matter that much.

ids1024 (approved):
I'm not sure how best to test that VRR is actually working, but the changes here look good.

PR #1006, opened 2024-11-20
Apply output transform to touch/tablet events

ids1024:
Depends on https://github.com/Smithay/smithay/pull/1593.


ids1024:
Rebased now that Smithay has the change needed here.

PR #1022, opened 2024-11-30
screencopy: don't force opaque window backgrounds

alyssais:
This meant that, for windows that were not fully opaque, there was no way to capture their transparency.  No way to put a window with transparency on top of a background in OBS, for example.

(Thanks @tranquillity-codes for teaching me about pre-multiplied alpha when I was going down the wrong track)


Drakulix (approved):
Good catch! Thanks. :)

PR #1024, opened 2024-12-05
Reactivate after output switch

wash2:
Alt tab relies on changes to the activated state, but if there is a single window on each output, they are always activated. This forces activation state changes so that the events are sent to indicate the focus change if an output changes and the newly focused window is already activated. This should work properly with https://github.com/pop-os/cosmic-launcher/pull/225 & https://github.com/pop-os/launcher/pull/249. 

I think there is a cleaner way of doing this if toplevel's also had a focused state, which could only ever set for one window at a time. but this works if we want to avoid adding something like that.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,49 +6,65 @@ name = "cosmic-comp"
 version = "0.1.0"
 
 [workspace]
-members = [
-  "cosmic-comp-config",
-]
+members = ["cosmic-comp-config"]
 
 [dependencies]
-anyhow = {version = "1.0.51", features = ["backtrace"]}
+anyhow = { version = "1.0.51", features = ["backtrace"] }
 bitflags = "2.4"
 bytemuck = "1.12"
-calloop = {version = "0.14.1", features = ["executor"]}
-cosmic-comp-config = {path = "cosmic-comp-config"}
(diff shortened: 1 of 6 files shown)


Drakulix:
Hmm interesting idea. I previously thought about making sure the Activated state only every applies to the last window getting focus, that was this attempt (though I didn't spent time debugging it and making sure it actually works): https://github.com/pop-os/cosmic-comp/pull/889/files

Visually this is of course a bit different to current cosmic behaviour, but given Activated is meant to convey focus, I think that is cleaner? I don't have a strong preference on either though..

wash2:
Does the activated state affect the rate that applications receive frame events? I guess there is at least a visual difference if all monitors have an activated app or just one. @pop-os/ux do you think one way is better?

maria-komarova:
I'm not sure I fully understand the problem. There should only be one focused window at a time.

Drakulix:
> I'm not sure I fully understand the problem. There should only be one focused window at a time.

The problem is, that right now visually (except for the focus indicator) on each monitor one window will appear focused (only the one on the active output will actually get the indicator and be able to receive input).

Since that "appear focus"-state is used by some components to determine the focused state, we have some problems with that behaviour, that could be solved, by simply making all windows on non-focused outputs appear unfocused.

That would however be a slight deviation from the current user-experience.

PR #1033, opened 2024-12-06
Alt shift tab

wash2:
should be merged just after https://github.com/pop-os/cosmic-settings-daemon/pull/58


Drakulix (commented):
Can we drop the Cargo.toml reformatting :sweat_smile: ?

Otherwise LGTM

Drakulix (approved):

PR #1038, opened 2024-12-08
main: Update D-Bus activation environment also on non-systemd systems

### LINKED ISSUE #1037: `DISPLAY` is not provided to the D-Bus activation environment (without systemd)
cosmic-comp provides the activation environment only with systemd:

https://github.com/pop-os/cosmic-comp/blob/7829e7696d276c0b51ea4834986a0a37e40b9ee4/src/main.rs#L48-L51
https://github.com/pop-os/cosmic-comp/blob/7829e7696d276c0b51ea4834986a0a37e40b9ee4/src/systemd.rs#L8-L35

This makes different application broken on non-systemd environments, mainly Steam flatpak, which fails with the error:

```
The unofficial Steam Flatpak app requires a correctly-configured desktop
session, which must provide the DISPLAY environment variable to the
D-Bus session bus activation environment.

On systems that use systemd --user, the DISPLAY environment variable must
also be present in the systemd --user activation environment.

This is usually achieved by running:

    dbus-update-activation-environment DISPLAY

during desktop environment startup.

For more details, please see:
https://github.com/ValveSoftware/steam-for-linux/issues/10554
```

This can be easily fixed by using zbus to provide `DISPLAY`. Pull request incoming. :wink:

vadorovsky:
On systems without systemd, use zbus to update D-Bus activation environment with `WAYLAND_DISPLAY` and `DISPLAY` variables.

Fixes #1037

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -45,9 +45,13 @@ impl State {
         // into systemd and the session?
         self.ready.call_once(|| {
             // potentially tell systemd we are setup now
-            #[cfg(feature = "systemd")]
             if let state::BackendData::Kms(_) = &self.backend {
+                #[cfg(feature = "systemd")]
                 systemd::ready(&self.common);
+                #[cfg(not(feature = "systemd"))]
+                if let Err(err) = dbus::ready(&self.common) {
+                    error!(?err, "Failed to update the D-Bus activation environment");
+                }
             }
 
             // potentially tell the session we are setup now
--- a/src/dbus/mod.rs
+++ b/src/dbus/mod.rs
@@ -1,6 +1,8 @@
-use crate::state::{BackendData, State};
+use crate::state::{BackendData, Common, State};
 use anyhow::{Context, Result};
 use calloop::{InsertError, LoopHandle, RegistrationToken};
+use std::collections::HashMap;
+use zbus::blocking::{fdo::DBusProxy, Connection};
 
 mod power;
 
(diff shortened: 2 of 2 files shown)


src/main.rs
```diff
@@ -45,9 +45,13 @@ impl State {
         // into systemd and the session?
         self.ready.call_once(|| {
             // potentially tell systemd we are setup now
-            #[cfg(feature = "systemd")]
             if let state::BackendData::Kms(_) = &self.backend {
+                #[cfg(feature = "systemd")]
                 systemd::ready(&self.common);
```
vadorovsky, line 50: I was thinking about replacing the `systemd::ready` entirely with `dbus::ready`, I think that should work as well. What's good about the dbus solution is not calling external binaries. For now I didn't remove `systemd::ready` in case you consider that too intrusive. I would love to hear your thoughts here.

What I can confirm is that this patch fixes flatpak Steam on my Gentoo+openrc installation.

Drakulix: The systemd environment is still something separate from the dbus environment, the dbus-broker just happens to import it on systems with systemd.

Just updating the dbus activation environment would this break e.g. systemd user services. I think this is a sensible solution. If calling external binaries would really be a problem, we could also trigger the systemd interface via dbus, but given this is a one time startup thing, I wouldn't bother.

Drakulix (approved):
LGTM. Thanks!

src/main.rs
vadorovsky: Makes sense, thanks!

PR #1091, opened 2024-12-24
Fixes for transformed outputs with mirroring

ids1024:
Also applicable to https://github.com/pop-os/cosmic-comp/pull/1058.

Looks like damage tracking for mirroring with output transforms isn't quite right either (before or after this) so some more fixes are needed.

### DIFF
--- a/src/backend/kms/surface/mod.rs
+++ b/src/backend/kms/surface/mod.rs
@@ -63,7 +63,7 @@ use smithay::{
         },
         wayland_server::protocol::wl_surface::WlSurface,
     },
-    utils::{Buffer as BufferCoords, Clock, Monotonic, Physical, Rectangle, Transform},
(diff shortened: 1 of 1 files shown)


ids1024:
I'm not sure the centering is always right either, in some circumstances, though I'll have to test if that's still an issue after other changes.

ids1024:
Rebased and fixed the centering (it was actually using the right position, but stretching incorrectly with a scale factor).

I'm not really sure about the damage tracking(?) issue I see when the target output of mirroring is rotated. I see one side of the screen update correctly when scrolling (for instance) while the other side is unchanged, and updates after a delay.

ids1024:
It looks like the damage issue is an issue in Smithay with transform handling in general, though it becomes more apparent here than in some other situations: https://github.com/Smithay/smithay/issues/1651

I guess then that doesn't need to be fixed here, and is a separate issue.

Drakulix (approved):

PR #1094, opened 2024-12-25
refactor(shell/element): refactor how decorations height is accessed

yu-re-ka:
This fixes several things:
- The xwayland code previously incorrectly used the SSD_HEIGHT (for Windows) even when the X11 surface was in a stack
- The SSD_HEIGHT was defined in surface.rs, even though rendering serverside decorations is done in the window/stack

Rename (min|max)_size to (min|max)_size_without_ssd in CosmicSurface and make it act accordingly
Add a new (min|max)_size() in CosmicWindow and CosmicStack, which takes the surface's (min|max)_size and adds the decorations.
Change all callers to use (min|max)_size() from the window or stack respectively, except is_dialog() where it does not matter.

### DIFF
--- a/src/shell/element/mod.rs
+++ b/src/shell/element/mod.rs
@@ -468,61 +468,16 @@ impl CosmicMapped {
 
     pub fn min_size(&self) -> Option<Size<i32, Logical>> {
         match &self.element {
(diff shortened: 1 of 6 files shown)


Drakulix (approved):
LGTM. Thanks!

PR #1020, opened 2024-11-29
Use `DrmOutputManager`/`DrmOutput` 

Drakulix:
Draft / test branch for https://github.com/Smithay/smithay/pull/1576.

Supersedes / Attempts to solve issues lied out in #969.

Needs more work to actually solve those issues:
- [x] Currently this screws up the surface threads `QueueState`, fix in cosmic-comp or disabled the event for forced commits in `initialize_output`.
- [x] https://github.com/Smithay/smithay/pull/1576 needs actual format selection (though that code is currently not used?)
- [x] https://github.com/Smithay/smithay/pull/1576 needs better fallbacks (first disable planes, try again, then set the modifier to Invalid, try again, then recover if still not working)
- [x] https://github.com/Smithay/smithay/pull/1576 take care of the case where falling back from primary plane scanout to composition breaks
- [x] https://github.com/Smithay/smithay/pull/1576 add a way to lock all compositors and switch to compositing without initializing a new output
  - [x] Use that for mode-switches
  - [x] Use that when drm-leasing is requested

PR #1097, opened 2024-12-27
Migrate away from `Rectangle::from_loc_and_size`

ids1024:
Depends on https://github.com/Smithay/smithay/pull/1621.

Annoying to update, but overall a little bit better.

Some of these `Rectangle::new` calls should probably be written in a way that doesn't construct new `Point`/`Size` instances from scalars and uses higher-level operations that preserve coordinate space, but this doesn't change that for now. (If we can move Smithay to `euclid` that should have a few more useful methods to clean up things like that too.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4688,7 +4688,7 @@ checksum = "3c5e1a9a646d36c3599cd173a41282daf47c44583ad367b8e6837255952e5c67"
 [[package]]
 name = "smithay"
 version = "0.3.0"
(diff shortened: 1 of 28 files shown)


ids1024:
Updating Smithay now that https://github.com/Smithay/smithay/pull/1621 is merged I see a bunch of build errors related to https://github.com/Smithay/smithay/pull/1576, so I guess I'll update this after https://github.com/pop-os/cosmic-comp/pull/1020 is merged.

Drakulix (approved):

PR #1106, opened 2025-01-01
Update leds when devices are added

### LINKED ISSUE #1104: Keyboard LEDs not being lit when keyboard plugged in.
Steps to reproduce:

1. Press Caps lock, Num lock, or Scroll lock.
2. Verify LED(s) is/are on.
3. Unplug keyboard
4. Plug keyboard back in
5. Test behavior of keys and check LEDs

Expected:
LEDs match state from before unplug, and keyboard behavior functions as expected.

Actual:
LEDs are all off, but the state of the keyboard acts as if they're on. For example, pressing capslock and re/plugging, the led is off, but all text is capitalized when typing.

Cycling the lock keys 3 times brings the leds back into the correct state.

Nashenas88:
This fixes #1104

LEDs on keyboard will now update to match the compositor state when they're plugged in.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -171,7 +171,8 @@ impl State {
             InputEvent::DeviceAdded { device } => {
                 let shell = self.common.shell.read().unwrap();
                 let seat = shell.seats.last_active();
-                seat.devices().add_device(&device);
+                let led_state = seat.get_keyboard().unwrap().led_state();
+                seat.devices().add_device(&device, led_state);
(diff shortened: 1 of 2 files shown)


ids1024:
Hm, perhaps https://docs.rs/input/latest/input/struct.Device.html#method.led_update shouldn't take an `&mut` if it's a clonable ref-counted type. That makes this mildly awkward, though it doesn't really matter.

Drakulix (approved):
Thanks. LGTM!

PR #1107, opened 2025-01-01
Make numlock state on boot configurable

### LINKED ISSUE #369: [Feature request] Remember Num Lock state
If for example Num Lock is enabled in the session, it should stay enabled after a reboot.
If it was disabled, it should stay disabled after a reboot.
Not sure if Caps Lock and Scroll Lock should also be included.
Hopefully this is the right repo for this.

Nashenas88:
This will expose 3 settings for numlock behavior:
1. Numlock is off on boot (this is the current default behavior)
2. Numlock is on on boot
3. Numlock will restore the state from the last boot

Fixes #369

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -2,7 +2,10 @@
 
 use crate::state::State;
 use anyhow::{Context, Result};
+use cosmic_comp_config::NumlockState;
+use smithay::backend::input::{self as smithay_input};
 use smithay::reexports::{calloop::EventLoop, wayland_server::DisplayHandle};
+use smithay::utils::SERIAL_COUNTER;
 use tracing::{info, warn};
 
 pub mod render;
@@ -51,13 +54,14 @@ pub fn init_backend_auto(
             .next()
             .with_context(|| "Backend initialized without output")
             .cloned()?;
-        let initial_seat = crate::shell::create_seat(
+        let (initial_seat, keyboard) = crate::shell::create_seat(
             dh,
             &mut state.common.seat_state,
             &output,
             &state.common.config,
             "seat-0".into(),
         );
+
         state
             .common
             .shell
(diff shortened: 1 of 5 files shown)


git-f0x:
Would it maybe be better to always just use the state from last boot, rather than needing additional settings?
Though I might not be aware of cases where restoring the previous state would be undesirable.

wiiznokes:
> Would it maybe be better to always just use the state from last boot, rather than needing additional settings? Though I might not be aware of cases where restoring the previous state would be undesirable.

Maybe to have the same (clean) state when booting. I don't think having an option for this is a bad thing as long as it configurable via config files

Nashenas88:
@git-f0x I just based it off my own preference and by what's available in other systems.
@wiiznokes it's been designed to be configured through config files. This PR does not touch cosmic settings, but there's no reason GUI controls couldn't be added as well.
@zanaviska I didn't specifically build for that so I'm not sure.

src/shell/seats.rs
```diff
@@ -172,7 +172,7 @@ pub fn create_seat(
     output: &Output,
     config: &Config,
     name: String,
-) -> Seat<State> {
+) -> (Seat<State>, KeyboardHandle<State>) {
```
Drakulix, line 175: Lets not change this api. We can simply call `initial_seat.get_keyboard()` afterwards to get the keyboard.

src/input/mod.rs
```diff
@@ -1449,6 +1449,19 @@ impl State {
             event.time() as u64 * 1000,
         );
 
+        // If the numlock key is pressed ...
+        if event.key_code() == Keycode::new(77) && event.state() == KeyState::Pressed {
```
Drakulix, line 1453: This completely ignores the fact, that we might filter out this event, which would cause the state to mismatch with the actual numlock state.

Instead we should check if the `modifiers_state` has changed after processing keyboard input, e.g. after line 242.

Drakulix: Hint: We also don't want to write the file on every potential keyboard event, so we need to compare the old value with the new one, without `numlock_mut()`.

src/backend/mod.rs
```diff
@@ -66,6 +70,39 @@ pub fn init_backend_auto(
...
+            }
+        };
+
+        // If we're enabling numlock...
+        if toggle_numlock {
+            let mut input = |key_state| {
+                let time = state.common.clock.now().as_millis();
+                let _ = keyboard.input(
+                    state,
+                    smithay_input::Keycode::new(77),
```
Drakulix, line 95: I really dislike magic numbers, can you please add a `const KEY_NUMLOCK = 69` to the beginning of this file and do `KEY_NUMLOCK + 8` here?

Drakulix (commented):
Thanks for working on this. :)

src/backend/mod.rs
zanaviska: For me, `KEY_NUMLOCK + 8` is equally magical as  `77` is

zanaviska:
I am sorry if this is wrong place to ask, but neverthenless
I am trying to implement this my comment myself on top of this branch. But when I try to execute this repo, just app with black screen appears. As far as I understand,  new app is new instance of compositor. How do I know, that observable behaviour of the system(e.g. numlock is on) is the result of new compositor, not the old one? How can I properly test my implementation?

> Could this PR possible fix the bug when num-lock resets to default(or it just turns off?) when switching keyboard languages?

git-f0x:
You can test it by either running the compiled binary from a TTY, or by installing it and logging out and back in.
I don't think the nested compositor instance can affect numlock state.

zanaviska:
@git-f0x When I run binary from TTY, how can I start another app, where I can type(text editor, terminal or anything else)?

git-f0x:
You can open the terminal with Super+T.

Drakulix (approved):
Alright this looks good to me now! Thanks a bunch!

git-f0x:
Tiny question, would LastBoot maybe be a better default?

Drakulix:
> Tiny question, would LastBoot maybe be a better default?

I don't think so, at least I am not aware of other DEs doing that and I think that might be a little unexpected.

git-f0x:
I think GNOME might remember the last state by default, and Windows does as well.

PR #1108, opened 2025-01-02
Add debug session for debugging with existing cosmic install

Nashenas88:
Add a new debug session that piggy-backs on an existing cosmic install so that input interactions can be debugged without replacing the install cosmic-comp. This only works when the configs can be shared between both versions.

I'm not sure if this is wanted, but it made it easier for me to debug some other PRs I was working on.


hojjatabdollahi:
@Nashenas88 what is your workflow for debugging with this PR?

Nashenas88:
@hojjatabdollahi I run `cargo build --release && sudo make install-debug-session`. Then I logout and login to the `Debug Cosmic` session. I'll then launch a terminal and run `journalctl -b -t debug-cosmic-session` to look at logs.

Drakulix:
While I like the idea of a debug session, I certainly do not like how this duplicates a bunch of files. If you want to see this merged, I would suggest to re-model the service and desktop files into template files, that is then generated into the different variants in the install step.

Nashenas88:
The latest PR converts the files to templates. I didn't do this for `debug-data/debug-cosmic-service` as that's really a modified version of https://github.com/pop-os/cosmic-session/blob/master/data/start-cosmic, and it wasn't clear to me how to make it a template given the repos aren't submodules of one or the other.

Issue #315, opened 2024-02-19
Touchpad: Navigation Gestures


ryanabx:
This is a planned feature. See: https://github.com/pop-os/cosmic-settings/issues/35

These are the proposed navigation gestures per the link above:
![image](https://github.com/pop-os/cosmic-comp/assets/56272643/7d687cc5-35fa-4513-9d02-0102f9fd455c)

I didn't however see an issue created yet for the compositor side of things. Are touchpad gestures planned for alpha 1? They are a very useful feature when promoting the use of workspaces. Assuming the compositor side becomes complete, I could work on implementing the cosmic-settings side of things. I'm much less confident on my ability to implement the compositor side of things though :sweat_smile:

ryanabx:
I'm thinking about workspace switching gestures again today, so I'm doing some reading on libinput https://wayland.freedesktop.org/libinput/doc/latest/gestures.html

pknepps:
I would like to work on this. I will open up an issue under cosmic-comp as well.

ryanabx:
> I would like to work on this. I will open up an issue under cosmic-comp as well.

Hi, this issue is under cosmic-comp, right?

Here's an update since the time I made this issue. Workspace switching touchpad gestures are implemented (i.e. to and from workspaces).

However, the overview trigger gesture is not implemented yet, nor is the three finger swipe to switch between windows, so there's work there to be done if you'd like to look into it!

pknepps:
Sorry, I thought this was in cosmic-settings. I've been planning on adding the settings to customize gestures for a while. I will open the issue under cosmic-settings.

lveillard:
Just for future reference, another thing that should be enabled in order to make all navigation gestures work, is the navigation logic itself for backwards/forward navigation. 

It is currently not possible to do it in Cosmic menues

Mmanish7:
When will this be implemented?

taxmeifyoucan:
+1 customizable gestures or just more options for swiping would be greatly appreciated. Ideally on par with touchegg features

cesarviana:
Switching windows using three-finger gesture was very useful for me. It would be nice to have it back (I lost it after moving from Gnome to Cosmic).

an-abnormality:
> Switching windows using three-finger gesture was very useful for me. It would be nice to have it back (I lost it after moving from Gnome to Cosmic).

+1 for three finger gestures preferably. Much more comfortable than four

lveillard:
[...]

PR #1109, opened 2025-01-02
Gesture-based dragging

Nashenas88:
I don't expect this PR to be accepted as-is. I hope it can be used to guide #315 to eventually support 3-finger dragging through configuration. This is a feature I've used with other tooling, e.g. https://github.com/iberianpig/fusuma/issues/173, and you can find others online looking for this same functionality.

The behavior works, but the configuration is not clear / well-defined. I'd also not like to contribute to scope-creep for the cosmic launch.

### DIFF
--- a/cosmic-comp-config/src/input.rs
+++ b/cosmic-comp-config/src/input.rs
@@ -31,6 +31,22 @@ pub struct InputConfig {
     pub tap_config: Option<TapConfig>,
     #[serde(skip_serializing_if = "Option::is_none", default)]
     pub map_to_output: Option<String>,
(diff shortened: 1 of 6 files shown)


hojjatabdollahi:
I used to use 3-finger dragging on Mac all the time to move windows around. I don't necessarily miss it in Cosmic, since I mostly use my keyboard. But it's a lot more ergonomic than tap-tap-drag.

tangowithfoxtrot:
I would love to see this accepted. Three-finger dragging is such a helpful accessibility feature.

PR #1122, opened 2025-01-08
WIP Use `get_unconstrained_geometry` function for popup position

ids1024:
`get_unconstrained_geometry` was added to Smithay in https://github.com/Smithay/smithay/pull/1256 (based on the cosmic-comp implementation) and is used in Anvil and Niri (at least). It would be good to avoid duplicating it.

But I've also noticed popup positioning not working very well. Not sure exactly is wrong with the existing cosmic-comp version, but it seems to not resize popups when it should, causing them to end up off screen.

There are a lot of different cases to consider here, but after some work, and comparison to Anvil, this version seems to be working pretty well, without the issues that I previously saw.

For `unconstrain_xdg_popup_tile`, this required adding an argument to Smithay's `get_unconstrained_geometry` to disable resizing. Maybe it should take a `ConstraintAdjustment` argument, that is combined with the one set by the client?

`get_anchor_point()` could also be moved to Smithay as a method of `PositionerState` since it's just a straightforward implementation of what the protocol specifies.


Drakulix:
> `get_unconstrained_geometry` was added to Smithay in [Smithay/smithay#1256](https://github.com/Smithay/smithay/pull/1256) (based on the cosmic-comp implementation) and is used in Anvil and Niri (at least). It would be good to avoid duplicating it.

This has been on my backlog for quite some time, thanks for taking care of it.

> For `unconstrain_xdg_popup_tile`, this required adding an argument to Smithay's `get_unconstrained_geometry` to disable resizing. Maybe it should take a `ConstraintAdjustment` argument, that is combined with the one set by the client?

That sounds like a good approach to me.

> `get_anchor_point()` could also be moved to Smithay as a method of `PositionerState` since it's just a straightforward implementation of what the protocol specifies.

:+1:

Drakulix (approved):

PR #1125, opened 2025-01-09
Improve window positioning when toggling tiling on

AshtonOhms:
Hi folks! This is a quality of life change I've been working on. I frequently switch back and forth between using tiling and floating windows, and thought it would be nice to be able to position windows in floating mode and preserve those positions when turning on tiling.

This change updates the tiling layout code to attempt to preserve relative window positioning and sizes as much as possible when turning on toggling.

At a high level, the algorithm for this is to iteratively:
* Find candidate positions to "split" the screen based on window geometry
* Determine which candidate positions best minimize overlap with windows
* Assign windows to either side of the split - if a side has one window, insert a `Mapped` leaf into the tree, otherwise create a `Group` and placeholders for the windows to be solved.
* Repeat this process until all placeholder leafs have been solved.

Here's a small demo video: [Demo](https://www.youtube.com/watch?v=QcZlskcIC7Y)

This is the first time I've written a significant amount of Rust, so any feedback on anything non-idiomatic would be appreciated. Cheers!

### DIFF
--- a/src/shell/layout/tiling/mod.rs
+++ b/src/shell/layout/tiling/mod.rs
@@ -341,6 +341,14 @@ pub struct MinimizedTilingState {
     pub sizes: Vec<i32>,
 }
 
+struct OrientationTestResult {
+    start_windows: Vec<CosmicMapped>,
+    end_windows: Vec<CosmicMapped>,
+    separateness: i32,
+    orientation: Orientation,
+    position: i32,
+}
+
 impl TilingLayout {
     pub fn new(theme: cosmic::Theme, output: &Output) -> TilingLayout {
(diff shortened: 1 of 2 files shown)


Drakulix:
Not a bad approach at all and I like the idea, but this likely needs to go through @pop-os/ux before I can consider merging this. Thanks for working on this!

maria-komarova:
Looks like a good improvement to me. Maybe I'm missing some edge cases. I guess depending on the layout one might want the windows to occupy more standard half and quarter tiles but it is difficult to say if this would be the case more frequently. It probably needs testing in usage, at least internally before we merge and release it.

PR #1126, opened 2025-01-10
Don't cancel grabs on Escape.

### LINKED ISSUE #680: Missing first Esc keypress after a window is focused
When I focus a  window, the first time I press Esc I get a key release event but not a key press event. After that, each time I press Esc I get both a press and a release. Until I switch focus away and back again, and then once again the first time I press Esc I only get the release.

- It happens both in tiled mode and floating mode
- It happens whether I use the keyboard or the mouse to change focus
- I've tested several different keys, and only Esc seems to have this behavior
- I *think* this is cosmic-comp's fault because it doesn't happen under sway

jneem:
fcitx5 (and maybe other input methods?) listen for the global input- switching shortcut by maintaining a constant keyboard grab: every time you change focus, they take a grab on the newly focused window. This interacts badly with cosmic-comp's previous behavior, where the first press of Escape after switching windows would always go towards cancelling that grab. This meant that
 - the first Escape on any newly-focused window wouldn't get delivered to the window, and
 - after pressing Escape on any window, fcitx would no longer listen to the input-switching shortcut.

This commit removes the grab-cancelling behavior altogether. Probably it breaks something, because I don't know what the original purpose was. It may be that the correct behavior depends on what kind of grab it is.

Fixes #680


Drakulix:
> This commit removes the grab-cancelling behavior altogether. Probably it breaks something, because I don't know what the original purpose was. It may be that the correct behavior depends on what kind of grab it is.

The original purpose was to provide a global way to cancel grabs mostly for server-side grabs, where the compositor is responsible to allow the user to cancel without taking any action.

So a proper fix likely needs a way to check the type of the grab in smithay.

jneem:
Thanks for the response!

> So a proper fix likely needs a way to check the type of the grab in smithay.

It looks like some version of this is already possible, by calling `Keyboard::with_grab` and downcasting the grab object. I looked for grabs that `cosmic-comp` takes, and they seem to be only `SwapWindowGrab` and `PopupKeyboardGrab`. Would it be reasonable to cancel those two grabs on `Escape`, and ignore all others?

Another approach that would solve my issue is to cancel all grabs except `InputMethodKeyboardGrab`. It bothers me a bit that I don't know the "right" way to distinguish between keyboard grabs that should be cancelled on `Escape` and those that shouldn't...

Drakulix:
> Thanks for the response!
> 
> > So a proper fix likely needs a way to check the type of the grab in smithay.
> 
> It looks like some version of this is already possible, by calling `Keyboard::with_grab` and downcasting the grab object. I looked for grabs that `cosmic-comp` takes, and they seem to be only `SwapWindowGrab` and `PopupKeyboardGrab`. Would it be reasonable to cancel those two grabs on `Escape`, and ignore all others?
> 
> Another approach that would solve my issue is to cancel all grabs except `InputMethodKeyboardGrab`. It bothers me a bit that I don't know the "right" way to distinguish between keyboard grabs that should be cancelled on `Escape` and those that shouldn't...

I think the first approach is reasonable here, we can always extend this list, but given this was initially conceived for compositor-initiated grabs an allow-list seems like the right approach.

Drakulix (approved):
Thanks for looking into this! Seems good to me now :)

Issue #1124, opened 2025-01-09
No signal after `DrmOutputManager`/`DrmOutput` commits


git-f0x:
After `cosmic-comp` was updated from 9b78a2d7807bcb2f4fad02f570f1dc96a71dbbe9 to addcbc5039425b0ffc83c95383366e16cbb835d2, the display no longer received signal when running cosmic-comp, while TTYs still work.
Plugging the HDMI cable into the motherboard, so that the iGPU is used, makes the compositor work as it should, with the monitor receiving a signal.
Happens on both Pop!_OS 24.04 and Fedora 41.
It's not related to any commits after 42af2af37d7443b7e93460d96524ff6b2fa12533, so likely caused by #1020.

### Logs:
Debug build:
[comp-debug.log](https://github.com/user-attachments/files/18367212/comp-debug.log)
Debug build with `RUST_BACKTRACE=1`:
[comp-backtrace.log](https://github.com/user-attachments/files/18367229/comp-backtrace.log)
`dmesg` logs (using [these steps](https://github.com/swaywm/wlroots/wiki/DRM-Debugging), but had to switch to a different TTY to kill `dmesg`):
[dmesg.log](https://github.com/user-attachments/files/18367236/dmesg.log)

### System:
CPU: Intel i5 11400
GPU: AMD Radeon RX 570
Monitor: Asus VG245H

AngryPenguinPL:
Can confirm, same story on Mandriva.

git-f0x:
Since this seems GPU-specific, what GPU do you have?

AngryPenguinPL:
Radeon RX580 8GB, mesa 24.3.3

hayandevTW:
I am using an NVIDIA RTX 2070.

I'm not entirely sure if this is the same issue, but I encounter a "No signal" problem when switching to tty and then returning to the desktop, or after the connection with the monitor is lost and then reconnected, when using only one monitor.

In my case, I found that modifying the code below resolved the issue. If you are facing this problem, you might be able to alleviate it by either rolling back the version or modifying the code as shown below. However, this is not a permanent fix.

src/backend/render/mod.rs: 569
```rust
    let workspace_elements = workspace_elements(
        // ...
    );

    if workspace_elements.is_err() {
        return Ok(Vec::new());
    }
    let workspace_elements = workspace_elements?;
```

[...]

PR #1127, opened 2025-01-10
Better drm logs

Drakulix:
Depends on https://github.com/Smithay/smithay/pull/1635.

Hopefully helps with debugging https://github.com/pop-os/cosmic-comp/issues/1124

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4799,7 +4799,7 @@ checksum = "3c5e1a9a646d36c3599cd173a41282daf47c44583ad367b8e6837255952e5c67"
 [[package]]
 name = "smithay"
 version = "0.3.0"
-source = "git+https://github.com/smithay//smithay?rev=9fbc9fb#9fbc9fbbfe04c94e6716b89080145d061f66031f"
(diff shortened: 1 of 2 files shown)


hayandevTW:
I am using an NVIDIA RTX 2070.

I'm not entirely sure if this is the same issue, but I encounter a "No signal" problem when switching to tty and then returning to the desktop, or after the connection with the monitor is lost and then reconnected, when using only one monitor.

In my case, I found that modifying the code below resolved the issue. If you are facing this problem, you might be able to alleviate it by either rolling back the version or modifying the code as shown below.
However, this is not a permanent fix.

src/backend/render/mod.rs: 569
```rust
    let workspace_elements = workspace_elements(
        // ...
    );

    if workspace_elements.is_err() {
        return Ok(Vec::new());
    }
    let workspace_elements = workspace_elements?;
```

From the `OutputNoMode` error returned at `workspace_elements: 665`, I suspect that the issue arises due to a disconnection and reconnection of the display, during which the workspaces map is not updated.
Since I do not fully understand the intended behavior (can the same output instance be reused when the display is disconnected and reconnected?), I may not be able to investigate this issue further.

PR #1130, opened 2025-01-10
Run `cargo update`

ids1024:
Used `cargo update -p home@0.5.11 --precise 0.5.9` to ensure it still builds on Rust 1.80.0.

This includes https://github.com/pop-os/cosmic-text/commit/166b59f560c551dab391a864f7c1f503c1e18446, which is probably beneficial to cosmic-comp as well. May as well also update other dependencies, and just after an alpha release seems as good a time as any to do so.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 4
 
 [[package]]
 name = "ab_glyph"
-version = "0.2.28"
+version = "0.2.29"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "79faae4620f45232f599d9bc7b290f88247a0834162c4495ab2f02d60004adfb"
+checksum = "ec3672c180e71eeaaac3a541fbbc5f5ad4def8b747c595ad30d674e43049f7b0"
 dependencies = [
  "ab_glyph_rasterizer",
  "owned_ttf_parser",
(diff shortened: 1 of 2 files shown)


ids1024:
Hm, not sure Jenkins is failing.

```
cargo build --release --offline --locked
warning: `/<<PKGBUILDDIR>>/.cargo/config` is deprecated in favor of `config.toml`
note: if you need to support cargo 1.38 or earlier, you can symlink `config` to `config.toml`
error: failed to load source for dependency `smithay`

Caused by:
  Unable to update https://github.com/smithay//smithay?rev=9fbc9fb#9fbc9fbb

Caused by:
  can't checkout from 'https://github.com/smithay//smithay': you are in the offline mode (--offline)
make[1]: *** [Makefile:30: all] Error 101
make[1]: Leaving directory '/<<PKGBUILDDIR>>'
dh_auto_build: error: make -j128 returned exit code 2
make: *** [debian/rules:8: build-arch] Error 25
dpkg-buildpackage: error: debian/rules build-arch subprocess returned exit status 2
```

That shouldn't really have changed. I guess this is an issue where cargo doesn't produce a useful message...

ids1024:
Ah. So `cargo vendor | head -n -1 > .cargo/config` returns `0` even if `cargo vendor` failed, so the source build produced an error, but was treated as success.

Drakulix:
Merged manually

PR #1145, opened 2025-01-14
expose cosmic_comp as a library

yuyuyureka:
Allows other programs to depend on the cosmic_comp library crate and start a cosmic-comp in their process by calling `cosmic_comp::run()`. Customization parameters can be added to `run()`. Note that the interface of the cosmic_comp library crate is not stable in any way and might change anytime.

Also adds a first use of the new hooks system, which allows customizing
cosmic-comp at compile-time.
In this case, the view() function of CosmicWindow / CosmicStack is
hooked and the hook can change what is rendered as the header bar.

foo

### DIFF
--- a/examples/customized-window-decorations.rs
+++ b/examples/customized-window-decorations.rs
@@ -1,0 +1,79 @@
+use cosmic::iced::{Color, Rectangle, Size};
+use cosmic::widget;
+use cosmic::widget::canvas;
+use cosmic_comp::hooks::{Decorations, Hooks};
+use cosmic_comp::shell::element::stack::{
+    DefaultDecorations as DefaultStackDecorations, TAB_HEIGHT,
+};
+use cosmic_comp::shell::element::window::{
+    DefaultDecorations as DefaultWindowDecorations, SSD_HEIGHT,
+};
+use std::sync::Arc;
+
+/// The struct implementing Decorations hook
+#[derive(Debug)]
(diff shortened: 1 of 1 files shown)


ids1024:
I'm not sure exactly why you would want this, instead of just spawning `cosmic-comp` as a subprocess.

yuyuyureka:
I've rebased and switched over to a OnceLock-based global. Please have a look again, thank you!

yuyuyureka:
Hi, did you have a chance to look at this again?

Drakulix (approved):
LGTM, can you do a last (promise :crossed_fingers:) rebase?

yuyuyureka:
Did not notice the build was failing

yuyuyureka:
Checked in the missing hooks.rs and rebased

yuyuyureka:
Tested build locally

yuyuyureka:
for the failing checks [pop-os/staging/jammy/binary-amd64](https://buildchain.system76.com/job/pop-os/job/repos/134856/) I can not see the logs, as the CI page requires a login. Could you please tell me if this needs my attention?

leviport:
Both errors are from Jammy, and I believe we aren't updating cosmic-comp on Jammy anymore. @Drakulix can confirm, but that's probably nothing to worry about.

Drakulix:
No we aren't, so indeed not a concern. :)

PR #1147, opened 2025-01-14
The very basest of base README implementations

awilkins:
It isn't much, but it lets you know which -dev libs you need.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,13 @@
+# COSIC Compositor
+
+This is the Wayland compositor used by COSMIC Epoch.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -0,0 +1,13 @@
+# COSIC Compositor
```
l-const, line 1:
```suggestion
# COSMIC Compositor
```

newsgrep:
libxkbcommon-dev and libpixman-1-dev are missing.
Optional would be to mention cargo.

PR #1160, opened 2025-01-17
chore: log out using cosmic-osd

wash2:


Drakulix:
Oh I though we simply leave the Terminate Action, but change the default keybindings to bind to a new SystemAction calling cosmic-osd. That way both variants exist for either DE-experiments that don't have cosmic-osd or users that rather want to bind Terminate.

wash2:
requires https://github.com/pop-os/cosmic-settings-daemon/pull/66

wash2:
I'll undo the changes to Cargo.toml before merging too.

Issue #1129, opened 2025-01-10
Key input not always working for Firefox in stack


ids1024:
I've been seeing this issue for a while, upon occasion, but I think I can reproduce it consistently.

* Start `cosmic-comp`
* Open a terminal
* Enable stacking
* Open Firefox in the same stack
* Click on URL bar
* Press key on keyboard

Instead of the letter being typed, nothing happens.

Looking at `WAYLAND_DEBUG` logs, I see `wl_keyboard::key` events but no `wl_keyboard::enter`? Which is definitely odd. The issue is probably related to subsurfaces. I also see a log of `discarded` events, including `wl_surface::enter` events; not sure exactly why.

ids1024:
Ah, it's actually not related to subsurfaces, it's related to `xdg_activation`.

`XdgActivationHandler::request_activation` results in a call to `CosmicStack::set_active`. Which then sets `previous_keyboard` to `1`, the same as `active`, so later on key input when `keyboard_leave_if_previous` is run, those are the same so it doesn't call enter.

ids1024:
Is it just wrong to be modifying `previous_keyboard` here? I guess `previous_keyboard` should be "last window `enter` was sent to"?

ids1024:
This appears to be fixed by https://github.com/pop-os/cosmic-comp/pull/1801, with the reproduction steps listed here, and the reproduction I mentioned in https://github.com/pop-os/cosmic-comp/pull/1168 also doesn't seem to happen now. So I'm not sure this is handled perfectly and doesn't have any other issues, but it is working correctly as far as I can tell.

PR #1168, opened 2025-01-22
WIP stack: Update `previous_keyboard` only when `enter()` is called

ids1024:
Attempt at fixing https://github.com/pop-os/cosmic-comp/issues/1129.

This does seem to fix that issue with the reproduction steps listed there, but I still see it if I create a stack with a terminal and firefox, focus the terminal (element index 0 in the stack) and close it, then try to type in firefox (now element index 0).

So managing what child has keyboard focus still needs some work. I'm not sure exactly what conditions apply were I've randomly seen this behavior in actual use (and if this particular change helps) or if this could cause problems.

### DIFF
--- a/src/shell/element/stack.rs
+++ b/src/shell/element/stack.rs
@@ -181,7 +181,6 @@ impl CosmicStack {
                 let old_idx = p.active.swap(idx, Ordering::SeqCst);
                 if old_idx == idx {
                     p.reenter.store(true, Ordering::SeqCst);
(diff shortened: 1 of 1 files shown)


Drakulix:
Yeah, I think this approach is basically right, but I honestly don't remember exactly, why I previously updated the `previous_keyboard` state on these events. I guess the leave/enter sequence was sent in multiple places as well.

Issue #1005, opened 2024-11-19
Changes to the shortcut system for managing windows


maria-komarova:
It has been previously notes that there is a problem that occurs when displays and workspaces are placed on the same axis (displays and workspaces are horizontal, for example). That results in the keyboard shortcuts that have an extended functionality (for example, move a window to a workspace if there is an adjacent one when running out of displays) being confusing to understand. When it comes to those extended shortcuts there is a question of what gets priority - displays or workspaces. 

- [ ] **Stacks**
There is an old suggestion that hasn't yet been implemented but will be relevant for shortcuts down the list. When moving focus outside of a stack (Super+Arrows), show the tab on the stack where the move started when one keeps holding down Super.

- [ ] **Switch focus between windows and move windows: Super+arrows and Shift+Super+arrows**
Workspaces will take priority instead of displays. Move to the next workspace and if there is a display on the same axis, move to that display when out of workpaces. When one holds modifiers without releasing them - same behavior as for stacks. The workspace on the initial display should show the workspace where holding Super started (this avoids the users having the blank workspace left on the initial display)

- [ ] **Moving between workspaces and moving windows to workspaces: Ctrl+Super+arrows and Shift+Ctrl+Super+arrows**
- Remove extended action for the shortcut (don't move to displays)
- Add ability to cycle through workspaces. Cycle through to the first workspace on the display when the end is reached. Do not move to another display. (this ensures the opposite action happens with using the opposite arrow instead of jumping to a different display)

- [ ] **New shortcut - Switch focus to another display with Alt+Super+arrows**
- This shortcut doesn't not have an extended functionality and only applies to displays.

- [ ] **Move window to another display with Shift+Alt+Super+arrows**
- This shortcut doesn't not have an extended functionality and only applies to displays.

**Other changes**
- [ ] Remove shortcut: Move to prev display (Super+Shift+Comma) replaced with directional
- [ ] Remove shortcut:Move to next display (Super+Shift+Period) replaced with directional
This should help avoid confusion as to what display is next or previous.
- [ ] Fix Ctrl+Super+arrows Shortcuts. Currently, the shortcut doesn't work for all direction. For example, when setting workspaces to horizontal from vertical the Ctrl+Super+right doesn't switch to a workspace on the right. We should make sure those shortcuts work by default. 
- [ ] Change the touchpad gesture to match horizontal or vertical workspace orientation.

rxt30:
[...]

PR #1187, opened 2025-01-30
Fix navigation calculations and focus handling

uvera:
I encountered with multi monitor setup.

If I have my laptop display and external display like this
image

focusing to left direction did not work.
This PR also handles the case where monitors are next to each other, where left direction focusing also didn't work.

I used this JS script to test overlap logic

```javascript
function checkYOverlap(monitorA, monitorB) {
    return (
        monitorA.y < monitorB.y + monitorB.height &&
        monitorA.y + monitorA.height > monitorB.y
    );
}

// Test cases
const testCases = [
    {
        name: "Partial Overlap",
        monitorA: { y: 100, height: 100 },
        monitorB: { y: 150, height: 150 },
        expected: true,
    },
    {
        name: "Complete Overlap",
        monitorA: { y: 120, height: 200 },
        monitorB: { y: 100, height: 300 },
        expected: true,
    },
    {
        name: "No Overlap (Above)",
        monitorA: { y: 50, height: 50 },
        monitorB: { y: 150, height: 100 },
        expected: false,
    },
    {
        name: "No Overlap (Below)",
        monitorA: { y: 300, height: 100 },
        monitorB: { y: 150, height: 100 },
        expected: false,
    },
    {
        name: "Edge Touch (No Overlap)",
        monitorA: { y: 100, height: 100 },
        monitorB: { y: 200, height: 100 },
        expected: false,
    },
    {
        name: "Edge Overlapping (Just Touching)",
        monitorA: { y: 100, height: 101 },
        monitorB: { y: 200, height: 100 },
        expected: true,
    },
    {
        name: "Smaller bigger 1",
        monitorA: { y: 100, height: 100 },
        monitorB: { y: 80, height: 150 },
        expected: true,
    },
    {
        name: "Smaller bigger 2",
        monitorA: { y: 80, height: 150 },
        monitorB: { y: 100, height: 100 },
        expected: true,
    }
];

// Run tests
for (const testCase of testCases) {
    const result = checkYOverlap(testCase.monitorA, testCase.monitorB);
    console.log(
        `${testCase.name}: ${result} (Expected: ${testCase.expected})`,
        result === testCase.expected ? "✅" : "❌"
    );
}
```

### DIFF
--- a/src/input/actions.rs
+++ b/src/input/actions.rs
@@ -1133,11 +1133,13 @@ fn to_previous_workspace(
     workspace_state: &mut WorkspaceUpdateGuard<'_, State>,
 ) -> Result<Option<Point<i32, Global>>, InvalidWorkspaceIndex> {
     let current_output = seat.active_output();
-    let workspace = shell
-        .workspaces
-        .active_num(&current_output)
-        .1
-        .saturating_sub(1);
+    let workspace = shell.workspaces.active_num(&current_output).1;
+
+    if workspace == 0 {
+        return Err(InvalidWorkspaceIndex);
+    }
+
+    let workspace = workspace.saturating_sub(1);
 
     shell.activate(
         &current_output,
--- a/src/input/actions.rs
+++ b/src/input/actions.rs
@@ -827,15 +827,41 @@ impl State {
                         };
 
                         if let Some(direction) = dir {
-                            self.handle_shortcut_action(
-                                Action::SwitchOutput(direction),
-                                seat,
-                                serial,
-                                time,
-                                pattern,
-                                Some(direction),
-                                true,
-                            )
+                            match direction {
+                                Direction::Left => {
+                                    self.handle_shortcut_action(
+                                        Action::PreviousWorkspace,
+                                        seat,
+                                        serial,
+                                        time,
+                                        pattern,
+                                        Some(direction),
+                                        true,
+                                    );
+                                }
+                                Direction::Right => {
+                                    self.handle_shortcut_action(
+                                        Action::NextWorkspace,
+                                        seat,
+                                        serial,
+                                        time,
+                                        pattern,
+                                        Some(direction),
+                                        true,
+                                    );
+                                }
+                                _ => {
+                                    self.handle_shortcut_action(
+                                        Action::SwitchOutput(direction),
+                                        seat,
+                                        serial,
+                                        time,
+                                        pattern,
+                                        Some(direction),
+                                        true,
+                                    );
+                                }
+                            }
                         }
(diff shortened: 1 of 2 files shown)


src/input/actions.rs
```diff
@@ -827,15 +827,41 @@ impl State {
...
-                            self.handle_shortcut_action(
-                                Action::SwitchOutput(direction),
-                                seat,
-                                serial,
-                                time,
-                                pattern,
-                                Some(direction),
-                                true,
-                            )
+                            match direction {
```
uvera, line 830: This behaviour makes the most sense to me, we should go to previous or next workspace (depending on direction). Those actions also fall back to SwitchOutput if corresponding methods return `InvalidWorkspaceIndex`

src/input/actions.rs
```diff
@@ -1107,11 +1133,13 @@ fn to_previous_workspace(
...
 ) -> Result<Option<Point<i32, Global>>, InvalidWorkspaceIndex> {
     let current_output = seat.active_output();
-    let workspace = shell
-        .workspaces
-        .active_num(&current_output)
-        .1
-        .saturating_sub(1);
+    let workspace = shell.workspaces.active_num(&current_output).1;
+
+    if workspace == 0 {
```
uvera, line 1138: This is here as workspace is of type `usize` where `0usize.saturating_sub(1) results in 0` which isn't an invalid index. This is here to allow PreviousWorkspace action to fallback to SwitchOutput on error.

src/shell/mod.rs
```diff
@@ -1723,12 +1723,12 @@ impl Shell {
                 let geo = o.geometry();
                 match direction {
                     Direction::Left | Direction::Right => {
-                        !(geo.loc.y + geo.size.h < current_output_geo.loc.y
-                            || geo.loc.y > current_output_geo.loc.y + current_output_geo.size.h)
+                        geo.loc.y < current_output_geo.loc.y + current_output_geo.size.h
```
uvera, line 1726: Flipping the logic around helped avoid the issue where SwitchOutput to left resulted in output switch where it shouldn't as my external monitor is above my laptop regarding display.

Drakulix:
Thanks for looking into this.

I'll take a closer look at the code once working on https://github.com/pop-os/cosmic-comp/issues/1005 (very soon), which also revises shortcuts around output and workspace switching.

Given that will likely conflict a bit, I'll most likely cherry-pick parts of this PR.

Drakulix:
Closing as https://github.com/pop-os/cosmic-comp/pull/1251 is merged now. Thanks again for working on this!

PR #1188, opened 2025-01-30
shell: Remove last workspace if it follows empty active workspace

ids1024:
Partly fixes https://github.com/pop-os/cosmic-workspaces-epoch/issues/83, but it seems like there's at least one other issue with workspaces not being removed when they should be.

The if condition got a bit complicated here, so I've split it up and inverted the condition.

### DIFF
--- a/src/shell/mod.rs
+++ b/src/shell/mod.rs
@@ -553,34 +553,43 @@ impl WorkspaceSet {
 
     fn ensure_last_empty(&mut self, state: &mut WorkspaceUpdateGuard<State>) {
         // add empty at the end, if necessary
-        if self
-            .workspaces
-            .last()
-            .map(|last| !last.is_empty())
-            .unwrap_or(true)
-        {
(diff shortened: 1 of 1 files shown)


ids1024:
Another issue here is `pending_activations`. `retain_tokens` removes tokens with a 5 second delay, so the token added with `create_external_token()` when it's not used make `is_empty()` on the workspace `false` until that delay is over.

Not sure the right change for that one.

Drakulix:
> Another issue here is `pending_activations`. `retain_tokens` removes tokens with a 5 second delay, so the token added with `create_external_token()` when it's not used make `is_empty()` on the workspace `false` until that delay is over.
> 
> Not sure the right change for that one.

The idea here is that windows should spawn on the workspace they were invoked on. So the right fix would probably be to re-run the cleanup logic once the 5 second delay runs out?

Drakulix (approved):

PR #1189, opened 2025-01-31
feat: Snap Window Edges to Close Output Edges

Koranir:
This is a feature that KDE has called "Snap Zones" which I find to be very useful for organizing floating layouts (especially small widgets like the Spotify miniplayer, demonstrated in the video).

When resizing or moving a floating/sticky window, any edge that gets too close to one of the currently active output's edges will be snapped into the output's edge, mitigating any pixel-perfect movements you would have to do otherwise to get window edges to align properly.

https://github.com/user-attachments/assets/9ea74bf9-af80-4e3c-a210-8e610c73a50e

### DIFF
--- a/src/shell/grabs/menu/default.rs
+++ b/src/shell/grabs/menu/default.rs
@@ -288,7 +288,12 @@ pub fn window_items(
                     let _ = handle.insert_idle(move |state| {
                         let mut shell = state.common.shell.write().unwrap();
                         let seat = shell.seats.last_active().clone();
-                        let res = shell.menu_resize_request(&resize_clone, &seat, ResizeEdge::TOP);
+                        let res = shell.menu_resize_request(
+                            &state.common.config,
+                            &resize_clone,
+                            &seat,
+                            ResizeEdge::TOP,
+                        );
 
                         std::mem::drop(shell);
                         if let Some(((target, loc), (grab, focus))) = res {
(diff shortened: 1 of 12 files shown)


Drakulix:
So the verdict is, that we want this behind a cosmic-comp config setting that is off by default (but will be exposed in cosmic-settings).

Could you add this to the PR? If not that is also totally fine, in that case I'll do it, once I have the time.

In any case thanks for working on this :+1:

Koranir:
> So the verdict is, that we want this behind a cosmic-comp config setting that is off by default (but will be exposed in cosmic-settings).

> Could you add this to the PR?

Done. I've also made the actual snapping distance configurable too.

src/shell/grabs/menu/default.rs
```diff
@@ -348,8 +358,12 @@ pub fn window_items(
                     let _ = handle.insert_idle(move |state| {
                         let mut shell = state.common.shell.write().unwrap();
                         let seat = shell.seats.last_active().clone();
-                        let res =
-                            shell.menu_resize_request(&resize_clone, &seat, ResizeEdge::RIGHT);
+                        let res = shell.menu_resize_request(
+                            &state.common.config,
```
Drakulix, line 362: Thanks! I would prefer if you could just pass the threshold value instead of the whole config? (Applies to all the shell and grab methods).

Drakulix (commented):
> Done. I've also made the actual snapping distance configurable too.

Great! I think that makes for a great config option, though we will probably expose this in cosmic-settings purely as a toggle for now.

Drakulix (approved):

Drakulix:
Thanks again!

Koranir:
Aur naur, I forgot to rename the `window_snap_threshold` field/argument names for the `MoveGrab`s. Not a bug or anything but it's inconsistent naming :P

Drakulix:
> Aur naur, I forgot to rename the `window_snap_threshold` field/argument names for the `MoveGrab`s. Not a bug or anything but it's inconsistent naming :P

If you feel like it, feel free to submit a PR correcting that. :)

PR #1191, opened 2025-01-31
Initial support for workspace pinning and moving

ids1024:
Initial compositor-side prototype for https://github.com/pop-os/cosmic-workspaces-epoch/issues/108, using https://github.com/pop-os/cosmic-protocols/pull/48.

This adds a pinned flag for workspaces, which prevents the workspace from being removed automatically when empty, and allows a new workspace after it.

This does not yet persist pinned workspaces. Or allow re-ordering workspaces. We probably want to persist pinned workspaces by monitor and order, with names.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -40,10 +40,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
 dependencies = [
  "cfg-if",
- "getrandom",
+ "getrandom 0.2.15",
  "once_cell",
  "version_check",
- "zerocopy",
+ "zerocopy 0.7.35",
 ]
 
 [[package]]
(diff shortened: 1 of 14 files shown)


ids1024:
Oh, since workspace capabilities are per workspace handle, we could add and remove the `Move` workspace based on the same rules for whether or not we will auto-remove the workspace. Because the empty last workspace shouldn't be movable, but it would be nice to do that without duplicating the logic for that in cosmic-workspaces...

ids1024:
Though another thing cosmic-workspaces is doing is not allowing dragging a workspace after the last workspace, since cosmic-comp currently will auto-remove the empty last workspace if you try to move it somewhere else. Not sure if there's a nice way the protocol could communicate that, if cosmic-comp or a different compositor with the same protocols wanted to offer different behaviors than that.

ids1024:
Updated a couple things. And changed this to use cosmic-protocols main.

There are some TODOs here, and this will need various fixes for workspaces span outputs mode. But we seem to otherwise be fairly happy with how this is working now.

We may also need to change the format used for persisting the config before the final release, but that should be fine.

Drakulix (approved):
Seems good to me!

PR #1198, opened 2025-02-04
shell: Handle maximize requests before commit

Drakulix:
Similar to how we handle fullscreen-requests for unmapped windows.

While I am at it, this also refactors pending windows and layers a bit to make handling nicer. It also allows to unset the fullscreen state on pending windows again, which should be legal according to the protocol, but wasn't handled correctly previously.


Drakulix:
fixed branch name here: https://github.com/pop-os/cosmic-comp/pull/1198

Issue #1071, opened 2024-12-17
Single-click on server-side decorations causes window to enter drag mode


ryanabx:
Probably the worst demonstration of all time, but here:

Let me know if this is a duplicate issue!

https://github.com/user-attachments/assets/f7263bff-0efa-4be4-ab8d-3ef0fe6518ca

ids1024:
I think it detects the start of a drag based on a threshold of a certain number of pixels moved after the click, so maybe that number should be increased. (Or I don't know if there are other checks that other implementations use.)

ryanabx:
I was clicking, then lifting my mouse click without moving my mouse at all, which was what struck out to me. If it's calculating a threshold, yeah it definitely should be increased.

I don't know how other implementations work for sure, but if the mouse is lifted it probably shouldn't keep the drag going

ids1024:
Oh yeah, it shouldn't keep dragging after lifting the mouse. That should cancel the mouse grab and stop the drag.

Probably something https://github.com/pop-os/cosmic-comp/pull/1046 broke.

PR #1201, opened 2025-02-04
Update `libcosmic`

ids1024:
Fixes https://github.com/pop-os/cosmic-comp/issues/1071.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -786,7 +786,7 @@ dependencies = [
 [[package]]
 name = "cosmic-client-toolkit"
 version = "0.1.0"
-source = "git+https://github.com/pop-os/cosmic-protocols?rev=d218c76#d218c76b58c7a3b20dd5e7943f93fc306a1b81b8"
+source = "git+https://github.com/pop-os/cosmic-protocols?rev=29ab323#29ab32305c6457fccf0728caaaf79fcac4cca665"
 dependencies = [
  "cosmic-protocols",
  "libc",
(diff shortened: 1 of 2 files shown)


git-f0x:
While everything seems to work visually, I seem to have very high CPU usage with this (causing choppy cursor movement), especially when an XWayland client is placed in a stack. CPU usage when moving the cursor also seems higher.

git-f0x:
The level of CPU usage and choppiness isn't present without this PR, which is why I added the comment here. Not sure why, maybe `cctk`?

ids1024:
Odd. And that's with a release build?

`cctk` shouldn't really impact performance like that, and moreover isn't even used in cosmic-comp, since it's for Wayland clients.

git-f0x:
Yes, and I'm not seeing this on the `fix-maximize_noble` branch.

ids1024:
I can reproduce this. `IcedElementInternal::update` is taking a lot longer than it should.

It seems the issue isn't iced, libcosmic, or cosmic-protocols, but actually cosmic-freedesktop-icons. Downgraded back to the previous version with `cargo update cosmic-freedesktop-icons --precise bfc22d531596b57646dc47bcd70b2eb6f3152cd4` for now. So this shouldn't regress anything any more.

wash2 (approved):

git-f0x:
Yep, there aren't any perceptible regressions now.

ids1024:
Rebased, since https://github.com/pop-os/cosmic-comp/pull/1206 also updated `Cargo.lock`.

Drakulix (approved):

Issue #979, opened 2024-11-06
Some games(Including warframe) error with "Failed to send Xwayland Mapped-Event", and don't show up.


Dekomoro:
In Warframe's case it appears to be the abnormal resizing of the launcher which breaks.
Error pops up with `journalctl --user -b0 _COMM=cosmic-comp` after running and closing the game.

The workaround I used to fix this is to use `protontricks 230410 winecfg`(for warframe) and disable "Allow the window manager to control the windows"

Dekomoro:
This is exclusive to windowed applications running through proton

Mel34:
Nota bene: Same thing happens with wine{,-staging}

dmac-au:
Adding https://github.com/pop-os/cosmic-comp/issues/1089 as it seems to be related.

Dekomoro:
Closing this, as upstream smithay seems to have fixed this. Eagerly awaiting the fix to land in cosmic-comp as well.

gabriele2000:
> Closing this, as upstream smithay seems to have fixed this. Eagerly awaiting the fix to land in cosmic-comp as well.

Today is indeed a great day to live!

PR #1206, opened 2025-02-07
chore: Update smithay

Drakulix:
Also pulls in some general fixes from https://github.com/pop-os/cosmic-comp/pull/1173


Drakulix:
> Is it going to fix #979? I've read that Smithay upstreamed a fix.

yes the fix is in there

gabriele2000:
> yes the fix is in there

Awesome, thanks!

PR #1213, opened 2025-02-12
Support `ext-workspace-v1`

ids1024:
This should work, though I still need to test with a client.

This doesn't yet add an extension protocol (https://github.com/pop-os/cosmic-protocols/pull/51). We could merge this without that (after some testing) but we'll want to add that soon.


Drakulix:
Looks good on first read apart from the lack of testing obviously. Seems like a good approach to me.

ids1024:
I guess one thing I'm missing here is sending `ext_workspace_group_handle_v1::workspace_enter` and `workspace_leave`. I guess the old revision of the protocol didn't allow moving a workspace to a different group.

We can do the same for now, sending `workspace_enter` on creation, and never sending `workspace_leave`. When we drop the old protocol, we can change from creating new handles on moving a workspace to sending `workspace_leave`/`workspace_enter`.

ids1024:
After updating https://github.com/pop-os/cosmic-protocols/pull/51 with cctk changes, and sending `workspace_enter` here, this seems to be working. (Of course, it's that client doesn't see the v2 extension protocol, which isn't implemented here yet.)

ids1024:
Oh, I guess I still need to make sure it sends `workspace_leave` only for workspaces from the same manager instance.

(Relatedly, this needs a fix like https://github.com/pop-os/cosmic-comp/pull/1061. I guess we don't currently have any clients that end up binding the workspaces protocol in two places.)

Drakulix (approved):

PR #1214, opened 2025-02-12
Reactivated num/caps-lock upon keyboard layour change

zanaviska:
When changing keyboard layout, modifier state of keyboard resets to default. This happens here  `keyboard.set_xkb_config(..)`. In turn this call leads to `KeyboardHandle::update_xkb_state` call, which in turn executes such code
```
let mut state = xkb::State::new(&keymap);
for key in &internal.pressed_keys {
    state.update_key(*key, xkb::KeyDirection::Down);
}
```

To make sure, that numlock/capslock state is not modified, we read old state, and if needed active those modifiers again.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -94,23 +94,7 @@ pub fn init_backend_auto(
         if toggle_numlock {
             /// Linux scancode for numlock key.
             const NUMLOCK_SCANCODE: u32 = 69;
-            /// Offset used to convert Linux scancode to X11 keycode.
-            const X11_KEYCODE_OFFSET: u32 = 8;
-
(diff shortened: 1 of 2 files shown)


zanaviska:
Important things not worth the description:
1) num/caps-lock reset heppens on smithay-crate level. I tried to research how could we prevent this, or some "fine" way to change modifiers state back, but I have found none. So in the end, we just simulate press on respective buttons
2) some of the code(how we press buttons and constants) is actually borrowed from https://github.com/pop-os/cosmic-comp/pull/1107. I'd like to reuse those things, but we need that branch to be in master in order to reuse the code

Drakulix (approved):
Looks good, thanks!

PR #1217, opened 2025-02-14
Make edge snapping variable names more consistent-er

Koranir:
https://github.com/pop-os/cosmic-comp/pull/1189#issuecomment-2659006687


Drakulix (approved):
ty!

PR #1224, opened 2025-02-19
i18n(de) continuing someother's German translation initiation

phoekjelle:


resources/i18n/de/cosmic_comp.ftl
```diff
@@ -1,5 +1,11 @@
 grow-window = Vergrößern
 shrink-window = Verkleinern
-swap-windows = Fenster tauschen
+swap-windows = Fenster wechseln
```
Drakulix, line 3: Why was this changed? The old translation is much more on point?

Drakulix (changes requested):
I am sorry, but this isn't a particularly good translation.

If you aren't particularly versed in a language, please just don't, this is making more work not less.

Also, if you reached out about this before, I could've told you before, that there are more translations in this PR: https://github.com/pop-os/cosmic-comp/pull/1173/files#diff-ff288ccf53a514f093771957e6ed89557cb868253ffda6612c16afc1ffae7621

resources/i18n/de/cosmic_comp.ftl
feikedonia: Sorry!

feikedonia:
I just shouldn't PR German, 'cause I don't speak the language!
It just was a mistake, but, hé, I have enough work maintaining Dutch I think.

PR #1252, opened 2025-02-28
WIP match output by edid, falling back to connector name

ids1024:
Matching by connector name may work well for some uses, but is pretty inconsistent with display controllers that dynamically add DRM connectors, with variable numbering.

My idea is to only match outputs with the same edid vendor/product information (which should be the same physical display), and fallback to connectors if two somehow are identical. I think that would general match expected behavior?

I'm not sure this is working correctly, and I need to update the `output_stack` too. If persisted state ends up with display information in multiple places, it could make sense to track in just one place somehow...

This should also be used for persisted pinned workspaces (not sure if we want to persist a whole `output_stack`, or only the last output it was explicitly placed on?)

### DIFF
--- a/src/backend/kms/device.rs
+++ b/src/backend/kms/device.rs
@@ -2,7 +2,7 @@
 
 use crate::{
     backend::render::{output_elements, CursorMode, GlMultiRenderer, CLEAR_COLOR},
-    config::{AdaptiveSync, OutputConfig, OutputState},
+    config::{AdaptiveSync, EdidProduct, OutputConfig, OutputState},
     shell::Shell,
     utils::prelude::*,
     wayland::protocols::screencopy::Frame as ScreencopyFrame,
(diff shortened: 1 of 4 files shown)


Drakulix:
For reference, we definitely need a fallback: https://invent.kde.org/plasma/kscreen/-/merge_requests/187

We might in general just copy what KWin/KDE does here, given the amount of cursed edge-cases I have heard from zamundaaa on this.

ids1024:
Yeah, I noticed the two monitors I have have the same serial but different manufacture weeks, which still makes the possible to distinguish by EDID. But clearly indicates not all displays would be.

I guess the KDE version is similar to what I'm thinking of, though hashing the EDID info may not be a bad idea. As long as readability of the display config isn't important.

And perhaps instead of just the connector name, we should be storing the GPU as well, though `/dev/dri` numbering can also be inconsistent... I guess there are also PCIe IDs and driver names...

With connector hotplug from DisplayPort MST on this Intel iGPU I'm running on, for some reason disconnecting monitors then reconnecting them doesn't always use the same numbering. Maybe it would work best if it compared configuration by the number of DP devices and their relative numbers, but not their absolute value...

There may be no perfect way to handle every edge case here.

Issue #883, opened 2024-09-25
High CPU usage when resizing apps


git-f0x:
Most apparent when resizing in 2 dimensions, with `cosmic-comp` and `cosmic-workspaces` processes reaching around 100% CPU usage reported by `htop`, when holding and moving the mouse in a circle somewhat quickly.
Both <kbd>Super</kbd> + <kbd>Right Click</kbd> resize and regular resize have the same behavior.

Running on hardware (i5 11400, RX 570).
Pop!_OS 24.04,
cosmic-comp: `0.1~1727082801~24.04~afdb656`

Drakulix:
Yeah we need to throttle xdg_shell configure events.

git-f0x:
This has been massively improved (the workspaces no longer spike), but it's still higher than e.g. kwin (70-95%, as opposed to kwin at ~40%). So not sure if this should be closed.

peterkrull:
In a similar vein [to this issue](https://github.com/pop-os/cosmic-workspaces-epoch/issues/116), though not quite as bad. The heavy hitter when resizing seems to be [this section of code](https://github.com/pop-os/cosmic-comp/blob/master/src/wayland/protocols/toplevel_info.rs#L371-L408), more specifically the calls to `send_toplevel_to_client` and `surface.upgrade()`.

This is the flamegraph of upstream:
![Image](https://github.com/user-attachments/assets/cb868c43-c5c2-4447-b4dc-fbe9b9fcd291)

And this is with that entire section throttled to only run once every second.
![Image](https://github.com/user-attachments/assets/fe0c93bc-5146-4328-96f0-88c61e342541)

I am not sure if there are any side effects to this, but so far I am not seeing anything weird. Like with the mentioned issue, there is likely a more intelligent way to solve this, but this seems to work on my machine

Is this section of code just meant to clean up information from closed top-levels? If so, shouldn't it be fine to just do so less frequently?

Issue #711, opened 2024-08-11
Rearranging tiled windows with mouse is awkward due to timing dependence


peterkrull:
When initiating a tile-rearrange with the mouse using the Super + Left-click drag, there seems to be a time-delay before windows start to rearrange. This means it is impossible to quickly rearrange windows just by a quick flick of the mouse. Another issue is, even after the timeout, the mouse must move a tiny amount, before the rearrange triggers. If the mouse is moved and held still, it will just return to its original position.

https://github.com/user-attachments/assets/e54d1f05-ac75-465f-93fc-c4f0c8ae0b5d

Is this delay necessary? If it is, could it be made configurable/disable-able?

peterkrull:
I have found that removing the overdue check makes this problem significantly better:
https://github.com/pop-os/cosmic-comp/blob/master/src/shell/layout/tiling/mod.rs#L3548-L3564

And it makes the animations more responsive. Perhaps the animations can be visually a bit more jarring, but I think it is a good tradeoff for the sake of responsiveness and usability.

https://github.com/user-attachments/assets/db4cfb72-7d67-4a47-8a02-cfaeb19bb572

romanstingler:
yes moving / rearranging windows via the mouse should be significantly faster.
Any tiling manager like hyprland,sway it is butter smooth.

peterkrull:
To add some more context. Not only is the long amount of time an issue to me, but the fact that I have to move the cursor _after_ the delay really just feels broken. Perhaps at the end of the relevant delay, an event is emitted to check the cursor position, even if it has not been updated after the ended delay.

l-const:
Could this be addressed somehow?

PR #1279, opened 2025-03-04
perf(shell): Throttle toplevel refresh

peterkrull:
Addresses my own comment:
https://github.com/pop-os/cosmic-comp/issues/883#issuecomment-2605535768

This PR may not be the optimal solution, but I have not had any issues running it on my end.

edit: Also, the 100 ms is just a number i pulled from thin air.

### DIFF
--- a/src/wayland/protocols/toplevel_info.rs
+++ b/src/wayland/protocols/toplevel_info.rs
@@ -1,6 +1,10 @@
 // SPDX-License-Identifier: GPL-3.0-only
 
-use std::{collections::HashSet, sync::Mutex};
+use std::{
+    collections::HashSet,
+    sync::Mutex,
+    time::{Duration, Instant},
+};
 
 use smithay::{
     output::Output,
@@ -46,6 +50,7 @@ pub struct ToplevelInfoState<D, W: Window> {
     dirty: bool,
     last_dirty: bool,
     pub(in crate::wayland) foreign_toplevel_list: ForeignToplevelListState,
+    refresh_throttle: Instant,
     global: GlobalId,
     _dispatch_data: std::marker::PhantomData<D>,
 }
@@ -324,6 +329,7 @@ where
             last_dirty: false,
             foreign_toplevel_list,
             global,
+            refresh_throttle: std::time::Instant::now(),
             _dispatch_data: std::marker::PhantomData,
         }
     }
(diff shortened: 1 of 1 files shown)


Drakulix:
Yeah, unfortunately this isn't a good solution as it could keep the latest state from being send out.

The really issue here is just how frequent the [idle-callback from running the event-loop](https://github.com/pop-os/cosmic-comp/blob/master/src/main.rs#L121-L175) is called, which will call into the toplevel refresh at some point.

My guess is, that ever since cosmic-comp switched to using threads for rendering, the event-loop of the main thread is turning idle much more frequently. Which is theoretically a good thing, but means we run certain house-keeping tasks way to frequently. (Note: This is educated guess work, best would be to profile how many times we actually enter this and as the result of which events, but given the linked thread it is relatively safe to say we do this "too much".)

It is difficult to say how frequently we *should* be running these tasks, the nice thing about the idle callback was, that it is supposed to be called once the event-loop goes idle, so just after a bunch of events have likely changed the state we care about.

Which is why this patch can miss the last state update, if it happens before the 100ms have passed and then we go idle for longer. In turn causing bugs like having to move the cursor before state is represented correctly.

So imo what we should do is replacing the idle-callback with something else:
- Insert a calloop::Timer triggering 100ms from now into the event loop, that calls the old idle-callback code.
- If we already have a timer going, remove it and replace it with a new 100ms timer.

That however would cause the loop to trigger on it's own indefinitely, as the triggering timer will reset the idle-callback, scheduling itself again...

We can't really exclude a certain event source (the timer) from triggering the idle-callback, so we have to refactor again:
- Either we need to make this conditional by tracking, if we actually updated anything, before adding the timer. toplevel_refresh already has a `changed`-state somewhere inside, but many other refresh-methods don't and we would still need to bubble this up and collect it.
- Or we move the the timer to another calloop event-loop on a secondary thread. Kinda ugly, but fixes the self-triggering nature.

These are just the thoughts I have had on this problem so far, really profiling and actually addressing this is somewhere on my todo-list. I am writing this down, mostly in case you or anybody else want to tackle this problem or just understand why this isn't as trivial.

peterkrull:
Thanks for the details. I didn't expect my PR to go through anyway, but I just wanted to get some discussion going on the issue, so changing to draft. As i mentioned in the other issue, the main problem is really the `toplevel_info::refresh` so if there is a quick solution for it specifically, that would be a big improvement.

Screenshot_2025-03-04_15-24-00
Screenshot_2025-03-04_15-24-11

peterkrull:
Sort of unrelated, but to tie in " bugs like having to move the cursor before state is represented correctly" with another issue: https://github.com/pop-os/cosmic-comp/issues/711#issuecomment-2532145734

Being able to schedule a callback that keeps being "pushed forward" as long as there is activity is also useful in other contexts.

PR #1283, opened 2025-03-05
i18n(nl) update Dutch translation

feikedonia:
Makes Dutch translation for the zoomed view consistent with Cosmic-Settings


Drakulix (approved):
Thanks!

PR #1285, opened 2025-03-06
Throttle refresh routines

Drakulix:
(Re-opened due to branch naming CI stuff.)

Fixes https://github.com/pop-os/cosmic-comp/issues/883.
Supersedes https://github.com/pop-os/cosmic-comp/pull/1279.

Profiling with [samply](https://github.com/mstange/samply) shows a very significant drop in `refresh` calls as well in the overall cpu usage.

Notably this only fixes the cosmic-comp side to some degree. Some clients consuming toplevel-info like `cosmic-files-applet` are still using a lot of cpu, as is `cosmic-term`, during resizing for testing this. But that warrants a separate issue in the respective projects.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: GPL-3.0-only
 
+use calloop::timer::{TimeoutAction, Timer};
 use smithay::{
     reexports::{
         calloop::{generic::Generic, EventLoop, Interest, Mode, PostAction},
@@ -9,8 +10,15 @@ use smithay::{
 };
 
 use anyhow::{Context, Result};
-use state::State;
-use std::{env, ffi::OsString, os::unix::process::CommandExt, process, sync::Arc};
+use state::{LastRefresh, State};
+use std::{
+    env,
+    ffi::OsString,
+    os::unix::process::CommandExt,
+    process,
+    sync::Arc,
+    time::{Duration, Instant},
+};
 use tracing::{error, info, warn};
 use wayland::protocols::overlap_notify::OverlapNotifyState;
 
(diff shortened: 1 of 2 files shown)


git-f0x:
Oh, wow, this is an absolutely massive reduction, including when just moving the cursor. I think it's one of the lowest I've seen in any compositor!

git-f0x:
I've edited to comment to highlight that minimizing apps is only sluggish when clicking their icon in the app tray. Minimizing by clicking the minimize button is instant, as before. So maybe something could be tweaked with just the panel?

Drakulix:
Okay, I can reproduce. Given we re-schedule the refresh until after everything is resolved, you can essentially stall the minimized preview appearing by just frantically moving the cursor.

Maybe we shouldn't re-schedule, but just limit to at most every 100ms. That would likely increase cpu usage again a bit, but not cause this broken behaviour. More profiling is needed it seems.

Drakulix:
Alright, the new commit doesn't allow the refresh to be delayed until infinity any more. Unfortunately this means the cpu usage during e.g. moving the cursor is a bit higher again. But an easier way to drop that might simply be to profile and optimize the refresh routines.

Looking at the profiler the biggest contributor to `ToplevelInfoState::refresh` being so expensive seems to be smithay's `Output::client_outputs`-call and specifically the vector allocation inside of it. Fixing this isn't super trivial, as there is a `Mutex` involved, but should still be easily fixable with a slightly more complicated API taking a closure for example (or `MappedMutexGuard` once that becomes stable to return an iterator).

In any case, this should still be a significant improvement and further improvements will come as I find time for more optimization. (Though further external contributions on the matter are still very welcome, there are a lot of low-hanging fruits here.)

git-f0x:
I'm not seeing a notable increase in CPU usage while moving the cursor or resizing, and the issue is fixed! The largest CPU usage is by `cosmic-files-applet` on the desktop, but that is unrelated, and will likely be fixed with reactive rendering after another iced rebase.

PR #1293, opened 2025-03-08
fix: Remove unnecessary scaling of borders

peterkrull:
I noticed that the active window hint scales more than the rest of the UI, so I tried removing the seemingly special treatment given to border scaling:
```rust
let thickness = (thickness as f64 * scale).round() as u8;
let radius = (radius as f64 * scale).round() as u8;
```
And now it scales in line with everything else.


Drakulix (approved):
thx

PR #1298, opened 2025-03-11
chore: smithay update

Drakulix:
This is the big framebuffer-api-update.

Overall seems to work just as well, but touches a bunch of the screencopy logic, that previously implicitly passed the framebuffer around as state of the renderer.

So it is likely better to leave this for a couple of days and test with cosmic-workspaces and xdg-desktop-portal-cosmic before merging.


Drakulix:
In tests both screen sharing outputs and windows, as well as cosmic-workspaces and minimized window previews were all working fine. Calling this good to merge.

PR #1299, opened 2025-03-11
Workspace related fixes/improvement

ids1024:
A couple changes to how `output_add` and `output_remove` handle workspace moves. (Plus a minor improvement to the workspace protocol.)

I think this fixes the issues where output hotplug can results in 0 or 2 workspaces being shown as active instead of 1, and also addresses another case where hotplug results in an empty focused workspace followed by other workspaces (that one's not technically a bug, but I think this is better).


Drakulix (approved):
These changes seem all good to me. Thanks!

Issue #1099, opened 2024-12-28
Clicking empty workspace in workspace overview spawns new workspaces


BrodieRobertson:
If you open up the the Workspace overview, click on an empty workspace, close the overview and then open it again you can continue spawning new empty workspaces. As this is not intended in COSMIC if you then try to swap to any of them they will begin to be deleted.

[workspace-spawning.webm](https://github.com/user-attachments/assets/a85d3534-7837-4dad-adef-f92e5a61d11c)

feikedonia:
Which version do you run? 'Cause it might be fixed in the new version...  

I'm unable to replicate on
cosmic-session-git 1.0.0.alpha.4.r6.g998cf3f-1.

(might be dough, that I'm not following the staps properly)

(typing is hard, btw)

BrodieRobertson:
Just tested in COSMIC Alpha 5 and the bug is still present

feikedonia:
my workspace applet button doesnt work at all
![afbeelding](https://github.com/user-attachments/assets/e92d3105-e877-4038-a547-09e63c436c50)

feikedonia:
I shall report it against the cosmic-applets thingy

ids1024:
This seems to be another issue related to `pending_activations` in the compositor (https://github.com/pop-os/cosmic-comp/pull/1188#issuecomment-2625729307). Removing the test for `pending_activations` in `Workspaces::is_empty()` prevents it. I guess the way that is handled needs some change.

PR #1307, opened 2025-03-13
Improve handling of XDG activation tokens in shell

ids1024:
Requires https://github.com/Smithay/smithay/pull/1676.

This changes two things:
* `Workspace::is_empty` no longer checks if there are activation tokens,
  but a separate `Workspace::can_auto_remove` checks if the workspace is
  empty and has no activation tokens.
  - When we add workspace pinning, that can also be checked there.
* `Workspace` no longer contains a `pending_tokens` list that is updated
  on `refresh`. Instead, `can_auto_remove` takes the xdg activation
  state as an argument.

Since `Workspace::refresh` normally is run for focused workspaces, this
fixes allowing non-focused workspaces to be removed when an activation
token expires. It seems generally good to avoid tracking the activation
tokens in two places, and this is probably more efficient than needing
to refresh in more places.

By splitting this, we still don't remove an empty workspace if it has a
pending activation token, but we also don't add an empty workspace for
an activation token.

This mitigates the confusing behavior with activation tokens that aren't
used, but having to wait a few seconds in some cases before a workspace
is removed is still a little confusing. (We probably want `cosmic-term`
and `cosmic-workspace` to either consume the activation tokens they are
passed, or not be passed tokens when started by keybinding?)

Fixes https://github.com/pop-os/cosmic-comp/issues/1099.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4672,7 +4672,7 @@ checksum = "3c5e1a9a646d36c3599cd173a41282daf47c44583ad367b8e6837255952e5c67"
 [[package]]
 name = "smithay"
 version = "0.5.0"
-source = "git+https://github.com/smithay/smithay?rev=a503d98#a503d981d1be9a54b286ab5e160e4b9edddb500f"
+source = "git+https://github.com/smithay/smithay//?rev=796c41c#796c41c0294ccc195b0f59228d6467cc6c8f5de8"
 dependencies = [
  "aliasable",
  "appendlist",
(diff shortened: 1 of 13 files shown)


Drakulix:
> We probably want cosmic-term
and cosmic-workspace to either consume the activation tokens they are
passed, or not be passed tokens when started by keybinding?

Any of our applications should read the env-variable and call activate on startup. Maybe we can model this in libinput's Application abstraction?

ids1024:
That seems to be how this already works, except it doesn't work in cosmic-workspaces because there's no `main_window_id()` (it dynamically creates a window per output). And I guess cosmic-term doesn't enable the `wayland` feature (does that feature now work without breaking support for non-wayland backends?).

Drakulix (approved):
Yeah this seems like a better way of handling things. LGTM, thx!

PR #1321, opened 2025-03-25
config: Refactor WorkspaceLayout to use the Default macro

it-a-me:
cosmic-comp-config uses an extra function and #[serde(default = ...)] macro rather than simply applying the default at the enum level. This aligns better with the code style of TileBehavior in lib.rs


Drakulix (approved):
thx

PR #1323, opened 2025-03-25
Argument support with clap_lex

LinuxBoy-96:
Support for --help and --version argument in a terminal.

Very useful for when COSMIC will be officially released and someone want more info.

Sorry to force clap_lex on you, but since this it was decided that I have to use clap_lex and to standardize the method across all apps, I had no choices.

Other App, already merged or soon to be, include:
cosmic-greeter
cosmic-files
cosmic-player
cosmic-edit
cosmic-term


build.rs
```diff
@@ -1,13 +1,7 @@
...
 fn main() {
-    if let Some(output) = Command::new("git")
-        .args(&["rev-parse", "HEAD"])
-        .output()
-        .ok()
-    {
-        let git_hash = String::from_utf8(output.stdout).unwrap();
-        println!("cargo:rustc-env=GIT_HASH={}", git_hash);
-    }
+    EmitBuilder::builder().git_sha(true).emit().unwrap();
```
Drakulix, line 5: Why do we have to introduce a new dependency for this? The old code worked just fine and is easy to follow.

LinuxBoy-96: @Drakulix I just made build.rs the same way I had to for cosmic-greeter

I have asked to use vergen too. :(

mmstick: If the build.rs already had a means to provide the hash then there isn't a need to introduce a dependency.

src/main.rs
```diff
@@ -95,7 +100,29 @@ impl State {
...
+    let mut cursor = raw_args.cursor();
+
+    // Parse the arguments
+    while let Some(arg) = raw_args.next_os(&mut cursor) {
+        match arg.to_str() {
+            Some("--help") | Some("-h") => {
+                print_help(env!("CARGO_PKG_VERSION"), env!("VERGEN_GIT_SHA"));
+		return Ok(());
+            }
+            Some("--version") | Some("-v") => {
```
mmstick, line 114: Usually `-V` would be used for version in case `-v` is introduced later for `verbose`.

build.rs
Drakulix: Agreed. There is not necessarily value in doing it exactly the same in every crate, this just adds another dependency to the supply chain and actually breaks code, as some of the debug code uses the `GIT_HASH` env variable as well.

Please just drop the `vergen` stuff, the `clap_lex` additions look fine to me.

Drakulix: > Could we, alternatively, use vergen for that as well?

In theory, yes.

> it make thing so much easier to read.

To me the opposite is true. There is no way for me to decipher, what exactly is happening inside the call without looking at vergen's source code. I also don't know how the string is formatted nor if the format is stable (that is if vergen considers the format of the string as part of their API or not).

> Yes, it's a other dependency [...]

Which means more code, I need to trust or review on every update.

> Is using using vergen really that bad if it has no adverse side effect on binary? (cause it only insert a string in the code)
> 
> I don't think Vegen will diseapear anytime soon

These are not things I am concerned on.

My concern is with relying on a bunch of crates I don't have the time to review frequently for mundane tasks. This is what might as well be three lines of code (if not for rustfmt) replaced by one introducing a library that seems to have many feature we don't make any use of.

That is a future maintenance burden for something I previously never touched after writing it once.

LinuxBoy-96:
@Drakulix Everything is like you wanted : Clap_lex but without Vergen

Drakulix (approved):

PR #1325, opened 2025-03-26
EDID matching for output stack

ids1024:
EDID matching in https://github.com/pop-os/cosmic-comp/pull/1252, but for output stacks.

It seems better to add here first (where it isn't persisted) and overhaul the on-disk format for output configuration along with pinned workspace persisting, which will also need this.

For persistent workspaces, I think it makes sense not to persist the whole output stack, but only the last output a workspace was explicitly placed on (where it was created, or moved to by the user). Though there may be edge cases where this is undesirable and confusing. (Which is also a potential problem for any other alternative.)

So when a workspace is moved explictly by the user, the output stack is cleared. And the earliest output in the stack is where it was placed explictly.

Getting the behaviors right when multiple displays have the same edid information is harder.
- We want to ignore connectors, and match by edid information when possible
- But if two monitors with same edid information are connected *at the same time*, they should be distinguished

I had the idea that `Workspaces::add_output` should check if the edid matches an existing output on a different connector, then try to move workspaces that were explictly assigned to that output with that connector, though the way `.matches()` is used for truncating the output stack will need to change too...

### DIFF
--- a/src/backend/kms/device.rs
+++ b/src/backend/kms/device.rs
@@ -2,7 +2,7 @@
 
 use crate::{
     backend::render::{output_elements, CursorMode, GlMultiRenderer, CLEAR_COLOR},
-    config::{AdaptiveSync, OutputConfig, OutputState, ScreenFilter},
+    config::{AdaptiveSync, EdidProduct, OutputConfig, OutputState, ScreenFilter},
     shell::Shell,
     utils::prelude::*,
     wayland::protocols::screencopy::Frame as ScreencopyFrame,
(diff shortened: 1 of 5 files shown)


ids1024:
I guess in truncation it can truncate entries until the one that matches the edid, but still add a new entry if the connector is different.

The `PATH` property should help with consistent connector names for mst displays, though I'm not sure exactly how consistent it is in practice. (I wonder if we could just change connector names to use the parent connector, then an additional dash and number...)

Drakulix (approved):
Seems good to me.

ids1024:
This should now disambiguate properly if two connected outputs have the same EDID. The last commit should be a fairly simple way to do that.

Drakulix (approved):

PR #1328, opened 2025-03-28
Allow Xwayland to eavesdrop on certain key/mouse events to allow global shortcuts

Drakulix:
.. and a couple of xwayland related fixes

~~Blocked on https://github.com/Smithay/smithay/pull/1694~~

### DIFF
--- a/src/shell/focus/target.rs
+++ b/src/shell/focus/target.rs
@@ -28,9 +28,12 @@ use smithay::{
         },
         Seat,
     },
-    reexports::wayland_server::{backend::ObjectId, protocol::wl_surface::WlSurface, Resource},
+    reexports::wayland_server::{
+        backend::ObjectId, protocol::wl_surface::WlSurface, Client, Resource,
+    },
     utils::{IsAlive, Logical, Point, Serial, Transform},
     wayland::{seat::WaylandFocus, session_lock::LockSurface},
+    xwayland::xwm::XwmId,
 };
 
 #[derive(Debug, Clone, PartialEq)]
@@ -155,6 +158,15 @@ impl PointerFocusTarget {
             _ => None,
         }
     }
+
+    pub fn is_client(&self, client: &Client) -> bool {
+        match self {
+            PointerFocusTarget::WlSurface { surface, .. } => {
+                surface.client().is_some_and(|c| c == *client)
+            }
+            _ => false,
+        }
+    }
 }
 
 impl KeyboardFocusTarget {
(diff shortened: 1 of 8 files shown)


ids1024:
The implementation here generally looks good.

But testing with `xinput test-xi2 --root`, it seems it sometimes gets stuck with a flood of `KeyPress` events with `flags: repeat`? Not sure exactly when that's reproducible, but I guess this is missing release events somewhere.

src/shell/focus/target.rs
```diff
@@ -155,6 +158,15 @@ impl PointerFocusTarget {
             _ => None,
         }
     }
+
+    pub fn is_xwm(&self, client: &Client) -> bool {
```
ids1024, line 162: Calling this function `is_xwm` is a bit confusing when it just compares the client id of the wl_surface to the `client` argument. At least `client` should be named to specify what client it should be?

Though also, `client` is the client id of the `Xwayland` instance, and not really related to the "xwm" specifically, right? (This is different from `KeyboardFocusTarget::is_xwm` here.)

Drakulix: Yeah, this was supposed to work like the `KeyboardFocusTarget`, but then I realized I couldn't get the `XwmId` from the `WlSurface`. Renamed to `is_client`.

> Though also, client is the client id of the Xwayland instance, and not really related to the "xwm" specifically, right?

Technically yes, practically the function in `xwayland.rs` using this matches the client using the `XwmId` as you only have one WM per Xwayland instance. So with a less confusing name, this should be good.

Drakulix:
> But testing with xinput test-xi2 --root, it seems it sometimes gets stuck with a flood of KeyPress events with flags: repeat? Not sure exactly when that's reproducible, but I guess this is missing release events somewhere.

Hmm, I tried to replicate this by swapping focus, hitting a bunch of keys, mixed wayland and xwayland windows, but I wasn't able to. Anything specific you can tell me about when you hit this error? So far I am not running into this...

ids1024:
I'm not sure. It seems to be an issue somewhat frequently, but it's not clear exactly under what circumstances...

Drakulix:
> I'm not sure. It seems to be an issue somewhat frequently, but it's not clear exactly under what circumstances...

- Fixed a bug, where a modifier-only shortcut could keep the modifier pressed from xwayland's point of view
- Fixed a bug, where spawning a lock-screen via a shortcut would not send the appropriate release events

I couldn't find/reproduce anything else. If you have the time, I would appreciate another round of testing.

Drakulix:
Fixed

PR #1335, opened 2025-04-01
backend/kms: Use mst `PATH` for output naming

ids1024:
This may provide more consistent connector names, though that could use more testing.

On my system, I get `eDP-1`, `DP-2-1`, `DP-2-2`. I don't know how to test situations where child connector uses multiple mst "ports" (probably with a monitor I don't have).

This doesn't seem to match other compositors. But there's no expectation that it does.

If we prefer to expose the existing output names to the user, this could be stored separately and used internally for output matching. (Though with https://github.com/pop-os/cosmic-comp/pull/1325, etc, that would only be needed for outputs with the same edid). I'm not sure what naming is most intuitive. The arbitrary dynamic connector numbering MST outputs already have may be confusing.

### DIFF
--- a/src/backend/kms/device.rs
+++ b/src/backend/kms/device.rs
@@ -44,7 +44,7 @@ use tracing::{error, info, warn};
 use std::{
     cell::RefCell,
     collections::{HashMap, HashSet},
-    fmt,
+    fmt::{self, Write},
     path::{Path, PathBuf},
     sync::{atomic::AtomicBool, mpsc::Receiver, Arc, RwLock},
 };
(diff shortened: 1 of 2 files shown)


hojjatabdollahi:
Can't wait until the day that I wake my laptop up or when I connect it to my dock, for the screens to show up how I set them up last! 

I can test anything you want on my setup. It happens multiple times a day for me.

Drakulix:
> On my system, I get `eDP-1`, `DP-2-1`, `DP-2-2`. I don't know how to test situations where child connector uses multiple mst "ports" (probably with a monitor I don't have).

Yeah I think this is a good idea. I am not sure how consistent these are, but if we want user-readable output for e.g. `cosmic-randr` this is probably the way to go.

Though I would still prefer to use EDID-hashes (+ this as a fallback) for storing state.

src/backend/kms/device.rs
```diff
@@ -737,7 +737,25 @@ fn create_output_for_conn(drm: &mut DrmDevice, conn: connector::Handle) -> Resul
...
+        let mut interface = parent_interface;
+        interface.push('-');
+        let mut first = true;
+        for port in ports {
+            if !first {
+                interface.push(',');
+            }
+            write!(&mut interface, "{}", port).unwrap();
+            first = false;
+        }
```
Drakulix, line 754: Can we make this into another helper function, please?

PR #1337, opened 2025-04-02
Fractional Xwayland scaling

Drakulix:
Seems to work surprisingly well, but some apps don't match the fractional scale factor (how would they, given this was never supported by any toolkit).

Many will increase their font size with the dpi value (Qt apps) or be slightly off (in the worse case up to 50% until the nearest integer value), appearing larger/smaller accordingly (e.g. GTK3 apps).

Others like Steam can scale to whatever and still don't seem to get it completely right (window is slightly too small by default), but given the custom toolkit doesn't fit into the desktop, it looks correct.

It certainly works much much better with games, which now see the correct resolution on the monitor with the highest scaling value (no matter with fractional or not), while only the others are slightly off and will be scaled accordingly. (If we would want to make this truly configurable, we should allow user to set the "primary" display as exposed to Xwayland and take that output's scale for the client-scale.)

I am a bit weary about all the potential for rounding errors this introduces, but I haven't found anything obviously broken related to input or rendering.

Not sure, if we want this and if so, if this should replace our current implementation (which works more consistently with apps, but not games) or be a separate option (how do we explain this in the UI?).

Anyway, draft for testing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -331,6 +331,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
 
 [[package]]
+name = "atomic_float"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "628d228f918ac3b82fe590352cc719d30664a0c13ca3a60266fe02c7132d480a"
+
+[[package]]
 name = "atomicwrites"
 version = "0.4.2"
 source = "git+https://github.com/jackpot51/rust-atomicwrites#043ab4859d53ffd3d55334685303d8df39c9f768"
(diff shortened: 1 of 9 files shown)


Adam-Cosner:
Looks like cursor size gets bigger than normal on this PR when hovering over XWayland apps

Adam-Cosner:
Is there a particular reason that the scale is rounded before DPI multiplication?

Mithras:
Can you please explain what these options actually do?

Drakulix:
> Can you please explain what these options actually do?

The "Optimize for applications" setting equals what used to be "let apps scale themselves" in COSMIC. This rounds the scaling factor to the next integer scale, as that works best and most consistently for x11 applications, which generally don't really support fractional scaling.

"Optimize for fullscreen / gaming" does the same, but allows fractional scaling values to be represented and doesn't round. As a result you get matching screen resolutions (at least for the primary output, if your outputs don't have the same scaling value) in X11 applications, but given many x11 don't support fractional scaling well, they might appear a little smaller or larger as their wayland counterparts (as they will effectively round to the nearest integer themselves, which cosmic can't be aware of). This is how "let apps scale themselves" works on KDE.

We consider both approaches valuable for different users, which is why we decided to split this into two options.

PR #1340, opened 2025-04-04
Change default X11 eavesdropping to None

jacobgkau:
Requested by @WatchMkr.

May be less confusing for users than having some functionality work but not others by default. This way, something is clearly not working and they'll hopefully know to look the setting up so they can learn all the available options and their security implications.

### DIFF
--- a/cosmic-comp-config/src/lib.rs
+++ b/cosmic-comp-config/src/lib.rs
@@ -166,9 +166,9 @@ pub struct XwaylandEavesdropping {
 
 #[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Deserialize, Serialize)]
 pub enum EavesdroppingKeyboardMode {
+    #[default]
     None,
     Modifiers,
-    #[default]
     Combinations,
     All,
 }


Drakulix (approved):

gabriele2000:
> they'll hopefully know to look the setting up so they can learn all the available options and their security implications

I too like to dream sometimes.
A lot of people are installing a linux-based OS to run away from Windows 11... I doubt a lot of users have the ability to read, let alone "look the setting up".

jacobgkau:
@gabriele2000 There's no need to insult peoples' literacy in a GitHub comment.

My point with that statement was that there's a greater chance someone will realize something's broken and at least do a web search (or make a forum complaint) if the feature is very obviously not working, than if the feature's partially working (i.e. only for certain keys) and it's not clear whether the application itself or the desktop environment only supports certain keys. It's just a secondary upside to this decision, not the primary motivator (which is security).

I agree it would be nice for people who require the feature if it was turned on by default. However, for people who don't require the feature, it's better to have it turned off by default to prevent trivial keylogging from being possible. (Can you imagine the backlash if a keylogger utilizing this feature was discovered, and people who don't use any related applications were exposed because we catered our defaults to X11-only voice chat app users?) GNOME doesn't support this feature at all, so we're providing more compatibility than they are, if slightly more manually than KDE is with their current default.

This decision can always be revised later if we see a lot of people confused about it. The setting will also hopefully get less relevant over time as Wayland protocols continue to develop and affected applications continue to be moved over to Wayland.

PR #1357, opened 2025-04-17
Remove cosmic-workspace-v1, and improve workspace protocol implementation

ids1024:
I want to test this a bit more (not that it should cause issues) and add a change to use the same workspace handle when a workspace is moved to another output.


ids1024:
Ah, looks like cosmic-panel still needs to be updated for ext-workspace.

ids1024:
https://github.com/pop-os/cosmic-panel/pull/375

ids1024:
I think with https://github.com/pop-os/cosmic-panel/pull/375 merged everything should use the new protocols. So that should be fine. And the updated `move_workspace_to_group` appears to be working as expected. So I think this should be good to merge now.

Drakulix (approved):

PR #1369, opened 2025-04-23
shell/grabs: Adjust reactive popups when window is moved

dcz-self:
XDG popups with positioner::reactive should be reconstrained if the window is moved:

> When set reactive, the surface is reconstrained if the conditions used for constraining changed, e.g. the parent window moved.

* * *

This can be tested with the example from https://codeberg.org/dcz/stiwri :

```
cargo run --bin popup
```

This is what it looks like. Laggy because of my computer, but also I think I'm redrawing the surface one frame late.

The popup is allowed to slide to the right if the space to the left is not big enough. Its allowed to flip down if the space above is not big enough.

[Screencast_20250423_145018.webm](https://github.com/user-attachments/assets/81e305a5-9b20-40ea-a900-77e56483634b)

Before this change, you needed to drag the edge and resize for the popup to move.

This will be very useful for the input method protocol update I'm working on.

### DIFF
--- a/src/shell/grabs/moving.rs
+++ b/src/shell/grabs/moving.rs
@@ -11,7 +11,7 @@ use crate::{
         },
         focus::target::{KeyboardFocusTarget, PointerFocusTarget},
         layout::floating::TiledCorners,
-        CosmicMapped, CosmicSurface, Direction, ManagedLayer,
+        CosmicMapped, CosmicSurface, Direction, ManagedLayer, WindowSurface,
     },
     utils::prelude::*,
     wayland::protocols::toplevel_info::{toplevel_enter_output, toplevel_enter_workspace},
@@ -490,7 +490,32 @@ impl PointerGrab<State> for MoveGrab {
         event: &MotionEvent,
     ) {
         self.update_location(state, event.location);
-
+        {
+            let mut borrow = self
+                .seat
+                .user_data()
+                .get::<SeatMoveGrabState>()
+                .map(|s| s.lock().unwrap());
+            if let Some(grab_state) = borrow.as_mut().and_then(|s| s.as_mut()) {
+                // this is the global location
+                let location =
+                    grab_state.window_offset.as_global().to_f64() + grab_state.location.as_global();
+                for (window, offset) in self.window.windows() {
+                    match window.0.underlying_surface() {
+                        WindowSurface::Wayland(toplevel) => {
+                            update_reactive_popups(
+                                toplevel,
+                                // round() is used for consistency with fractional scaling
+                                (location + offset.as_global().to_f64()).to_i32_round(),
+                                self.window_outputs.iter(),
+                            );
+                        }
+                        // X11 surfaces have access to the global coordinate space unlike wayland clients, so they do the re-adjustment themselves.
+                        _ => {}
+                    }
+                }
+            }
+        }
         // While the grab is active, no client has pointer focus
         handle.motion(state, None, event);
         if !self.window.alive() {


src/shell/grabs/moving.rs
```diff
@@ -490,7 +490,32 @@ impl PointerGrab<State> for MoveGrab {
...
+                let location =
+                    grab_state.window_offset.as_global().to_f64() + grab_state.location.as_global();
+                for (window, offset) in self.window.windows() {
+                    match window.0.underlying_surface() {
+                        WindowSurface::Wayland(toplevel) => {
+                            update_reactive_popups(
+                                toplevel,
+                                // TODO: give some thought to round vs floor
+                                (location + offset.as_global().to_f64()).to_i32_round(),
+                                self.window_outputs.iter(),
```
dcz-self, line 510: The outputs are for finding the anchor point and the anchor point must be within the surface, so it will be within the same outputs as the window, right?

src/shell/grabs/moving.rs
```diff
@@ -490,7 +490,32 @@ impl PointerGrab<State> for MoveGrab {
...
         self.update_location(state, event.location);
-
+        {
+            let mut borrow = self
+                .seat
+                .user_data()
+                .get::<SeatMoveGrabState>()
+                .map(|s| s.lock().unwrap());
+            if let Some(grab_state) = borrow.as_mut().and_then(|s| s.as_mut()) {
+                // is this workspace location or output location?
```
Drakulix: Looking at the code, it should be `Global` (I wonder why this piece of code isn't storing it that way), which means global coordinate space across outputs. `update_reactive_popups` wants global, so this should be correct.

src/shell/grabs/moving.rs
```diff
@@ -490,7 +490,32 @@ impl PointerGrab<State> for MoveGrab {
...
+                let location =
+                    grab_state.window_offset.as_global().to_f64() + grab_state.location.as_global();
+                for (window, offset) in self.window.windows() {
+                    match window.0.underlying_surface() {
+                        WindowSurface::Wayland(toplevel) => {
+                            update_reactive_popups(
+                                toplevel,
+                                // TODO: give some thought to round vs floor
```
Drakulix: Round is fine here. We use that in most of the code-base, so this would be consistent when fractional scaling is applied.

src/shell/grabs/moving.rs
Drakulix: The outputs are for figuring out, which output contains the toplevel/popup to constrain that accordingly (and not let them go off-screen). `window_outputs` should be fine here, as they contain every output the window currently overlaps with. :+1:

src/shell/grabs/moving.rs
```diff
@@ -490,7 +490,32 @@ impl PointerGrab<State> for MoveGrab {
...
+                                (location + offset.as_global().to_f64()).to_i32_round(),
+                                self.window_outputs.iter(),
+                            );
+                        }
+                        // Not sure what to do wth X11 surfaces. Perhaps their popups don't need adjustments.
```
Drakulix: They do, but since they have access to the global coordinate space unlike wayland clients, they do the re-adjustment themselves.

Drakulix (approved):
Thanks and sorry for looking at this so late.

Code-wise this seems all fine, can you update the code-comments I commented on to reflect the state of things better for merging? (Or remove them, where they were just questions.)

dcz-self:
Thanks for taking a look, I pushed a version with improved comments (though my explanations are surface-deep).

If you feel like more popup-related contributions, I can do them as soon as smithay gets https://github.com/Smithay/smithay/pull/1745

PR #1371, opened 2025-04-24
Profiling / Performance fixes

Drakulix:
Draft as this needs more testing on different systems.

This touches the frame scheduling logic again, but this time all profiled with tracy, so that we actually have some data:

1. According to kwin, mesa drivers return from `page_flip`/`commit` once the commit is staged for the next vblank. So we can use the time until submission (not presentation) to optimize our latency by moving closer, if we finish submission early. Waiting for presentation was somewhat borked anyway, as that depends on the refresh rate, so most likely we were rendering asap anyway.
2. This also introduces a workaround for nvidia cards, where this doesn't seem to help at all. Especially on multi-gpu setups submission might take ages, while we wait for the render operation to complete and we don't have a feedback channel. (I guess we might start putting the composited dmabuf into our event-loop to figure out the render time at least, but that is difficult to setup with our `DrmCompositor`.) So lets bail on the whole scheduling logic and just render asap
  Note: This should check for the driver rather than the vendor, given how good nouveau has gotten.
3. At the same time we also increase the safety margins a bit as we know are much closer to the deadline potentially running into kernel scheduling weirdness.
4. We also fix the vrr-deadline calculation for the min-refresh-rate. I assume this is what people described as stuttering, as we were basically guaranteed to miss the deadline for cursor updates with the old code. Now the cursor operates at *at least* 30fps, while still allowing content to drive the FPS in fullscreen. Feels very nice in game.

Other than that the tracing showed `refresh` on the main thread of frequently block the render-threads from acquiring the shell-lock, due to `IcedElement::refresh` taking ages. We can't easily fix this as most of the time in spent in `UserInterface::build` (specifically `diff_children`) and `layout` (not `draw` as one would expect, so finally switching to the wgpu renderer wouldn't help much), but we can reduce how often we need to call this as it turns out.

Doesn't fully eliminate the problem, but we are able to hit high refresh rates (>=120) much more reliably now.

This is all tested on an `oryp7` with integrated Intel and dedicated nvidia 3070 gpu with an external 120hz, 4k, VRR display. So in some ways the most problematic device (limited multi-gpu bandwidth, 4k, high refresh rate, vrr and nvidia...), but I would still want to give this some tests on other systems with tracy attached - mainly an AMD system - before merging.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -859,6 +859,7 @@ dependencies = [
  "tracing",
  "tracing-journald",
  "tracing-subscriber",
+ "tracy-client 0.18.0",
  "wayland-backend",
  "wayland-scanner",
  "xcursor",
@@ -1481,7 +1482,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "33d852cb9b869c2a9b3df2f71a3074817f01e1844f839a144f5fcef059a4eb5d"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -2926,7 +2927,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fc2f4eb4bc735547cfed7c0a4922cbd04a4655978c09b54f1f7b228750664c34"
 dependencies = [
  "cfg-if",
- "windows-targets 0.48.5",
+ "windows-targets 0.52.6",
 ]
 
 [[package]]
@@ -4081,7 +4082,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "afbdc74edc00b6f6a218ca6a5364d6226a259d4b8ea1af4a0ea063f27e179f4d"
 dependencies = [
  "profiling-procmacros",
- "tracy-client",
+ "tracy-client 0.17.6",
 ]
 
 [[package]]
@@ -4400,7 +4401,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.4.15",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


kode54:
How do I use it?

Drakulix:
> How do I use it?

Wait for it to be released.

kode54:
You said you wanted it to be tested on AMD hardware. Would you prefer if I mail you my hardware?

Drakulix:
> You said you wanted it to be tested on AMD hardware. Would you prefer if I mail you my hardware?

I said I want to test it on some more of hardware. I do have enough systems to confirm improvements on a variety of gpu vendors.

> Edit: Full disclosure, I already built the branch against alpha 7, and have Tracy installed. Clearly I need to do something more if I am to collect traces from my hardware under test scenarios.

If you are tech-savy and want to give this a try, you can build cosmic-comp with `cargo build --profile fastdebug --features profile-with-tracy` to get a build with enabled tracing, then `tracy` should automatically find it.

Try to trigger some usually problematic workflows/applications while recoding the trace. It would be nice to know, if they behave better than before, if you see obvious frame pacing issue in the trace or alternatively save and upload the trace file.
 
> Or I suppose I could wait for the beta or whenever this is folded into the release, and go back to using a different DE that has working VRR support that doesn't need further testing.

Testing is very welcome, but especially with complex changes like these, I don't necessarily have capacity to instruct every potential contributor on how to test changes. Many also not interested in providing valuable feedback, but just fixing the bug they experience on their system without realizing the headaches they might cause later on driving untested code.

So my usual response without the context I got from you last message would indeed be: Please wait.

Drakulix:
Alright, this seems to an improvement on all my machines, if not in performance than at least in consistency. Merging.

kode54:
Sorry for being impatient. I didn't have any bottlenecks on the branch when I tested, so I didn't know if there was anything specific that needed profiling. I'll stick to the alphas or future beta for now, though I'm holding off for the time being. Sorry if I wasted your time.

PR #810, opened 2024-09-04
fix: flake.nix under NixOS

dvtfl:
There are multiple issues with `flake.nix`.

1. `rust-src` is missing in `rust-toolchain.toml`.
Because of that, Rust Analyzer fails to resolve `std`, rendering RA pretty much unusable. 

(Zed log)
```
[WARN  project] Language server rust-analyzer-2024-09-02 (id 1) status update: Workspace `~/cosmic/cosmic-comp/Cargo.toml` has sysroot errors: can't load standard library from sysroot
{sysroot_path}
(discovered via `rustc --print sysroot`)
try installing the Rust source the same way you installed rustc
```
image
`flake.nix` imports rust toolchain from it:
https://github.com/pop-os/cosmic-comp/blob/52280e9823506a480248e84caaf7545426a871c8/flake.nix#L26 

2. When running in Wayland session in NixOS, `cosmic-comp` fails since `libwayland-egl.so` is missing in runtime.
```
2024-09-04T12:46:55.406051Z ERROR backend_winit: panic: thread 'main' panicked at 'Library libwayland-egl.so could not be loaded.': /home/dvtfl/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wayland-sys-0.31.4/src/egl.rs:46
<...>
  26: smithay::backend::winit::init
             at /home/dvtfl/.cargo/git/checkouts/smithay-0141b1fcb9e16729/65c4abf/src/backend/winit/mod.rs:71:5
  27: cosmic_comp::backend::winit::init_backend
             at /home/dvtfl/projects/rust/cosmic/cosmic-comp/src/backend/winit.rs:132:9
  28: cosmic_comp::backend::init_backend_auto
             at /home/dvtfl/projects/rust/cosmic/cosmic-comp/src/backend/mod.rs:35:25
  29: cosmic_comp::main
             at /home/dvtfl/projects/rust/cosmic/cosmic-comp/src/main.rs:108:
<...>
```
[full backtrace](https://yaso.su/o6UzZvyb)
3. When running in X11 session in NixOS, `cosmic-comp` fails since it fails to instantiate Vulkan and GBM.
```
INFO cosmic_comp::backend::x11: EGL hardware-acceleration enabled.
WARN cosmic_comp::backend::x11: Failed to instanciate vulkan, falling back to gbm allocator. err=Load(LoadError)
INFO smithay::backend::drm::device::fd: Dropping device: Some("/dev/dri/renderD128")
[...]

PR #1374, opened 2025-04-28
flake.nix: refactor

faukah:
This PR:
- cleans up the flake 
- removes unneeded inputs and replaces them with a few lines of simle nix code
- adds a nix-systems input, which allows for local overriding of the systems which cosmic-comp supports.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -2,50 +2,45 @@
   description = "Compositor for the COSMIC desktop environment";
 
   inputs = {
-    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
+    # <https://github.com/nix-systems/default-linux>
+    systems.url = "github:nix-systems/default-linux";
 
-    parts.url = "github:hercules-ci/flake-parts";
-    parts.inputs.nixpkgs-lib.follows = "nixpkgs";
+    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
 
     crane.url = "github:ipetkov/crane";
 
-    rust.url = "github:oxalica/rust-overlay";
-    rust.inputs.nixpkgs.follows = "nixpkgs";
-
-    nix-filter.url = "github:numtide/nix-filter";
+    rust-overlay.url = "github:oxalica/rust-overlay";
+    rust-overlay.inputs.nixpkgs.follows = "nixpkgs";
   };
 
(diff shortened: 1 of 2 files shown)


Drakulix:
@dvtfl Mentioning you because of [this comment](https://github.com/pop-os/cosmic-comp/pull/810#issuecomment-2360677110) and me having no clue, if these changes make sense.

flake.nix
```diff
@@ -2,115 +2,120 @@
...
-
-            nativeBuildInputs = with pkgs; [
-              pkg-config
-              autoPatchelfHook
-              cmake
-            ];
+  outputs = inputs @ {
+    self,
+    crane,
+    ...
```
spikespaz, line 19: It would now be sane to take all inputs instead of using `inputs.systems` and `inputs.rust-overlay`

spikespaz (approved):
Banger

flake.nix
```diff
@@ -2,115 +2,123 @@
...
-              stdenv.cc.cc.lib
-              pixman
-              libdisplay-info
-            ];
+    fs = lib.fileset;
+    eachSystem = lib.genAttrs (import systems);
+    pkgsFor = eachSystem (system:
+      import nixpkgs {
+        localSystem.system = system;
+        overlays = [(import rust-overlay)];
```
spikespaz, line 31: Nit: 
`rust-overlay` works this way because it contains the `default.nix` at the repo root. If you want to keep importing it this way, mark the flake input `flake = false`.

My preference would be to use `rust-overlay.overlays.default` in this position, and keep the input the way it is.

flake.nix
```diff
@@ -2,115 +2,123 @@
...
+      });
+  in {
+    packages =
+      lib.mapAttrs (system: pkgs: let
```
spikespaz, line 35: Optional, can be changed later: 

If you wish to provide an overlay, for users that want to introduce `cosmic-comp` into `pkgs`, and build it with *their* nixpkgs instead of this flake's transitive input, move the package to a file (`nix/package.nix`) and make a flake output `overlays.default = final: prev: { cosmic-comp = final.callPackage "${self}/nix/package.nix" {}; };`, then `inherit (pkgsFor.${system}) cosmic-comp; default = self.packages.${system}.cosmic-comp;` in flake output `packages`.

Typing on a phone, sorry for errors.

Issue #1305, opened 2025-03-13
Change to display config causes screencopy frame to never receive `success` or `failed`


ids1024:
This is part of the cause behind https://github.com/pop-os/xdg-desktop-portal-cosmic/issues/130, though I'm confused about how blocking in the `process` callback of the pipewire stream is preventing a different stream on a different thread from working. (I think blocking there is correct? And it creates the other stream, but pipewire never changes it from paused to playing state.)

It appears this gets to the ` let _ = tx.send((frame, damage))` call in `src/backend/kms/surface/mod.rs`. But I guess the receiver is user data associated with the queued/pending frame in the `DrmCompositor`, so it's never received if we destroy the `DrmCompositor` on display config state changes?

Drakulix:
This indicates we are leaking `Output`s somewhere on display configuration. Because both frames and session are wrapped to make sure they `fail` or `stop` on `Drop`.

ids1024:
Ah yeah, a leak makes sense.

PR #1383, opened 2025-04-30
Use `DropableFrame` in kms frame udata to make sure `failed` is sent

ids1024:
This seems to fix https://github.com/pop-os/cosmic-comp/issues/1305. xdg-desktop-portal-cosmic prints an error, buy retries (as it should for an `Unknown` error; though maybe there should be a retry limit) and the session continues working.

Not sure if it should be sending `failed`, or queing it with the next frame so it can send `success` to the client, but this works and is desirable as a failsafe anyway.

### DIFF
--- a/src/backend/kms/device.rs
+++ b/src/backend/kms/device.rs
@@ -5,7 +5,7 @@ use crate::{
     config::{AdaptiveSync, EdidProduct, OutputConfig, OutputState, ScreenFilter},
     shell::Shell,
     utils::prelude::*,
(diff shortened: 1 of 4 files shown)


ids1024:
It should probably pass `DropableFrame` in various places instead of `Frame`, so I'm leaving this as a draft before trying to refactor that a bit. But functionally, it seems like an improvement.

ids1024:
Not sure why I pushed this to my fork of cosmic-comp; so https://github.com/pop-os/cosmic-comp/pull/1395 updates the source branch.

PR #1385, opened 2025-05-01
add custom ssd height

### LINKED ISSUE #275: Make windows titles optional
I use some apps which are meant to not have windows titles, but the Cosmic compositor seems to force them.

One of these is Emacs, which should basically look like it's running in a terminal, but looks like this:  

screenshot-2024-04-24-10-28-42

I am running the pgtk version of Emacs 29.3, so it should be Wayland only. But it also is the same issue on the X-version.

Will this be a feature in the final version (disable windows titles)?

cramt:
fixes pop-os/cosmic-epoch#275

depends on pop-os/libcosmic#865 being merged first

### DIFF
--- a/src/shell/element/mod.rs
+++ b/src/shell/element/mod.rs
@@ -878,8 +878,9 @@ impl CosmicMapped {
 
     pub fn ssd_height(&self, pending: bool) -> Option<i32> {
         match &self.element {
-            CosmicMappedInternal::Window(w) => (!w.surface().is_decorated(pending))
-                .then(|| crate::shell::element::window::SSD_HEIGHT),
+            CosmicMappedInternal::Window(w) => {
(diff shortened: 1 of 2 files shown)


Ochiruko:
Hello cosmic maintainers! Thank you for all the amazing work you have done, we're all super excited for the beta! I'm excited to explore the DE later today on asahi linux, but have been waiting until SSDs can be hidden to make the leap. This appears to be a cornerstone commit for that issue. What is its current status? Is there any plan to merge this, or do you plan to hold off on pulling it until other decisions are made?

cramt:
Superseded by #2666, which does this the way you asked in pop-os/libcosmic#865: a single `show_titlebars` boolean in the comp config rather than a configurable SSD height, and no libcosmic change at all. Closing this one.

PR #1395, opened 2025-05-03
protocols/screencopy: Make frame/session send stopped/fail on drop

ids1024:
This seems to fix https://github.com/pop-os/cosmic-comp/issues/1305. xdg-desktop-portal-cosmic prints an error, buy retries (as it should for an `Unknown` error; though maybe there should be a retry limit) and the session continues working.

Not sure if it should be sending `failed`, or queuing it with the next frame so it can send `success` to the client, but this works and is desirable as a failsafe anyway.

### DIFF
--- a/src/backend/kms/device.rs
+++ b/src/backend/kms/device.rs
@@ -5,7 +5,7 @@ use crate::{
     config::{AdaptiveSync, EdidProduct, OutputConfig, OutputState, ScreenFilter},
     shell::Shell,
     utils::prelude::*,
-    wayland::protocols::screencopy::Frame as ScreencopyFrame,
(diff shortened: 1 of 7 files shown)


ids1024:
I think it's best if `Frame` itself implemented drop, `ScreencopyHandler::frame` and `ScreencopyHandler::cursor_frame` take `Frame`, but `frame_aborted` just gets a handle to match against frames. And sessions can do the same thing.

Drakulix (approved):

ids1024:
This seems to be working well for making sure frames and sessions send `failed`/`stopped` when they should.

PR #1398, opened 2025-05-06
Improved handling of XWayland grabs

ids1024:
A solution for https://github.com/Smithay/smithay/issues/1714.

With this, the lock screen is able to get keyboard focus normally, but focus then reverts to the XWayland grab surface. This can be tested with the example client from the issue.

Adding a new variant of `KeyboardFocusTarget` is annoying. Maybe it could map to a different variant of the enum. But it presumably needs to handle any `wl_surface` XWayland uses. (Override redirect surfaces? Subsurfaces?)

### DIFF
--- a/src/shell/mod.rs
+++ b/src/shell/mod.rs
@@ -53,6 +53,7 @@ use smithay::{
         session_lock::LockSurface,
         shell::wlr_layer::{KeyboardInteractivity, Layer, LayerSurfaceCachedState},
         xdg_activation::XdgActivationState,
+        xwayland_keyboard_grab::XWaylandKeyboardGrab,
     },
     xwayland::X11Surface,
 };
@@ -266,6 +267,7 @@ pub struct Shell {
     pub session_lock: Option<SessionLock>,
     pub seats: Seats,
     pub previous_workspace_idx: Option<(Serial, WeakOutput, usize)>,
+    pub xwayland_keyboard_grab: Option<XWaylandKeyboardGrab<State>>,
 
     theme: cosmic::Theme,
     pub active_hint: bool,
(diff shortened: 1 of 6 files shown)


Drakulix:
Yeah, I think given the circumstances this is a fine solution.

ids1024:
The `KeyboardFocusTarget` bloat will be reduced if we merge https://github.com/Smithay/smithay/pull/1334 and https://github.com/pop-os/cosmic-comp/pull/1431 anyway.

ids1024:
Gnome appears to disable XWayland grabs by default, without gconf settings `xwayland-allow-grabs` and `xwayland-grab-access-rules`. And I guess those settings aren't exposed in the UI either?

I think we probably do want to support it (by default?), since it's important for some types of applications, but may want a setting in our "X11 Applications Compatibility" section. Which would be easy enough.

ids1024:
Updated, with `get_output_for_focus` looking up the element for the `WlSurface`.

This should be a good improvement over what we have anyway.

ids1024:
... and fixed building with the `debug` feature.

ids1024:
Added a commit to also inhibit shortcuts when the grab is active, which matches X sessions.

src/shell/mod.rs
```diff
@@ -1661,7 +1664,10 @@ impl Shell {
             }?;
 
             focus_target = KeyboardFocusTarget::Element(new_target);
-        };
+        } else if let KeyboardFocusTarget::XWaylandGrab(surface) = &focus_target {
+            let new_target = self.element_for_surface(surface)?.clone();
+            focus_target = KeyboardFocusTarget::Element(new_target);
```
Drakulix: This doesn't work for override redirect surfaces, no?

(Also won't work with my upcoming fullscreen changes, which just wrap a `CosmicSurface`, but that is unrelated to this PR, more a note to myself, when rebasing.)

ids1024:
Updated `get_output_for_focus` to handle override-redirect windows by choosing the output the window's geometry overlaps the most with. I think that's the best way to handle that.

PR #1404, opened 2025-05-09
i18n(es): Add spanish translation

enriquezaf:
Not completely happy with the translation of unstack, but "dispersar" is the best word that I have found so far, it means "To separate and scatter what was or used to be gathered together", the  thing is, stack translates to "apilar", in spanish adding "des-" to a word usually means inversion of the meaning, but the word "desaplilar" doesn't exist, I think people would understand "desaplilar", I asked a few people (native speakers like me) and some could tell me what it means but they had to think about it for a bit, the official dictionary not having the word doesn't literally means that the word is wrong either, so I'm not sure if it should be "dispersar", "desapilar" or other word.

dispersar: [https://dle.rae.es/dispersar?m=form](https://dle.rae.es/dispersar?m=form)
des-: [https://dle.rae.es/des-?m=form](https://dle.rae.es/des-?m=form)
apilar: [https://dle.rae.es/desapilar?m=form](https://dle.rae.es/desapilar?m=form)


Drakulix (approved):
thanks!

PR #1408, opened 2025-05-12
kms/surface: Fix thread crash on error return of `redraw`

ids1024:
If `redraw()` returned early, before updating `self.state`, but after calling `queue_frame`, `on_vblank` would later be called, and reach `unreachable` since state isn't set to `WaitinfForVBlank`.

In particular, this was happening when the dmabuf from the image copy frame failed to `bind`.

To avoid this, make sure to update `self.state` immediately after calling `queue_frame`, before any early return from an error.

This fixes a crash I see with Intel+Nvidia screen capture and https://github.com/pop-os/xdg-desktop-portal-cosmic/pull/76. I need to try another fix to see if I can get capture to actually work.

### DIFF
--- a/src/backend/kms/surface/mod.rs
+++ b/src/backend/kms/surface/mod.rs
@@ -1431,6 +1431,25 @@ impl SurfaceThreadState {
                     x @ Ok(()) | x @ Err(FrameError::EmptyFrame) => {
                         self.timings.submitted_for_presentation(&self.clock);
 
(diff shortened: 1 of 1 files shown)


ids1024:
There doesn't seem to be a simple way to make capture work properly. Though it seems like this should at least be safe to merge. It at least makes it possible to test client changes without crashing the DRM thread.

Drakulix (approved):
LGTM

PR #1412, opened 2025-05-14
WIP `org.freedesktop.a11y.Manager` dbus protocol

ids1024:
With the rough incomplete version here, upstream Orca and at-spi2-core are able get key input from the server. More needs to be done to handle grabs correctly, restrict access to the protocol, and make sure all the details work correctly.


ids1024:
To restrict interfaces to only certain well-known names, I've added `NameOwners`, similar to something Mutter and Gnome Shell have. (Still need to actually use it properly, and test.) A little arcane async stuff, but not too complicated, and I think should work well.

I guess we also need to remove grabs for clients that are disconnected...

Conan-Kudo:
Any chance this will get completed in time for 1.0? Or can we get the existing at-spi2-core patch upstreamed so that non-pop distros can have working screen reader support?

WatchMkr:
@Conan-Kudo yes, we intend to get this in for that very reason.

Conan-Kudo:
Excellent, thank you!

ids1024:
Okay, this should now be correctly restricting the protocol to only Orca (as Gnome and KDE do), and it detects client disconnects to remove grabs. I've also fixed a mistake I made that was causing some issues in my testing, and cleaned this up a bit.

I think there still be issues in edge cases with Orca's global structural navigation in recent releases? On Gnome too? May be worth looking into, but this should already be a good improvement to the existing implementation of our COSMIC-specific protocol (which has a couple issues this should fix), in addition to supporting all distros with a recent enough Orca and at-spi2-core.

I could also do a bit more refactoring of the cosmic-comp DBus code, but that can be done later.

This should be tested along with the `orca-update` staging repo, which has packaging for `at-spi2-core`, `orca`, and `libwnck3`. Those will need to be merged into `master_noble` branches along with this.

leviport (approved):
I'm not great at using Orca, but it still seems to be working as well as it did before with this change and the `orca-update` branch added.

ids1024:
Merged. I've updated `orca`, `libwnck3`, and `at-spi2-core` to use `master_noble` branches.

We probably won't need, but for archival purposes the previous patched `at-spi2-core` is preserved in a tag: https://github.com/pop-os/at-spi2-core/releases/tag/cosmic-AtspiDevice

PR #1414, opened 2025-05-15
Update Rust version, and dependencies

ids1024:
`libcosmic` has a `rust-version` of 1.85, so update toolchain to that, and update libcosmic and other dependencies.


git-f0x:
Oh, I broke SSDs a bit, since a change I made in libcosmic to match designs doesn't leave enough room for window control buttons in SSDs.
I might need to undo that, or potentially add an `is_ssd` field to the `header_bar` widget, and handle that case differently.

ids1024:
I ran `cargo update -p libcosmic`, then squashed that into the existing `cargo update` commit. So the libcosmic fix should be here now.

git-f0x:
It doesn't do anything unless `.is_ssd(true)` is added for the widget linked above.

ids1024:
Ah, right. Added a commit adding `.is_ssd(true)` , and removing `.start`.

SSDs in simple X11 clients look right, anyway.

Drakulix (approved):

PR #1422, opened 2025-05-19
screencopy: Don't advertise implicit modifiers for dmabuf capture

Drakulix:
Too many clients just try to negotiate any shared modifier and import any dmabuf into whatever gpu they might using. Worse pipewire has no way of communicating a dmabuf node.

Given that cosmic-comp might advertise different nodes for different capture sources on the same system, we shouldn't advertise implicit modifiers as supported as cross-gpu might happen.

### DIFF
--- a/src/wayland/handlers/screencopy/mod.rs
+++ b/src/wayland/handlers/screencopy/mod.rs
@@ -418,7 +418,9 @@ fn constraints_for_renderer(
                 .fold(
                     HashMap::<Fourcc, Vec<Modifier>>::new(),
                     |mut map, format| {
-                        map.entry(format.code).or_default().push(format.modifier);
+                        if format.modifier != Modifier::Invalid {
+                            map.entry(format.code).or_default().push(format.modifier);
+                        }
                         map
                     },
                 )


ids1024:
I was thinking it might be better for xdg-desktop-portal-cosmic to filter out the implicit modifier from what it advertises to pipewire. I don't think this is really an issue in `ext-image-copy-capture-v1`, since it does have a `dmabuf_device` event that tells the client which GPU the implicit or explicit modifier is valid for.

I would have opened a PR on the portal for that yesterday, but I'm having a bit of trouble getting it to fallback to shm instead (but changing in the compositor instead won't help anything). The pipewire documentation definitely suggests that should be possible and correct though:

> The producer is also responsible to check if all clients announce sufficient capabilities or fallback to shared memory buffers when possible.

https://docs.pipewire.org/page_dma_buf.html

Drakulix:
> I was thinking it might be better for xdg-desktop-portal-cosmic to filter out the implicit modifier from what it advertises to pipewire.

Yeah that seems sane as well.

PR #1425, opened 2025-05-20
Do not focus layer surfaces with KeyboardInteractivity::None

jackpot51:
This will allow for an on-screen keyboard to provide input to other applications.

Draft until I have tested it.

### DIFF
--- a/src/shell/focus/mod.rs
+++ b/src/shell/focus/mod.rs
@@ -115,6 +115,13 @@ impl Shell {
                 .unwrap()
                 .element_for_surface(window)
                 .cloned(),
+            Some(KeyboardFocusTarget::LayerSurface(layer)) => {
+                // Do not focus layer surfaces with KeyboardInteractivity::None
+                if !layer.can_receive_keyboard_focus() {
+                    return;
+                }
+                None
+            }
             _ => None,
         };
 


src/shell/focus/mod.rs
```diff
@@ -115,6 +115,13 @@ impl Shell {
                 .unwrap()
                 .element_for_surface(window)
                 .cloned(),
+            Some(KeyboardFocusTarget::LayerSurface(layer)) => {
+                // Do not focus layer surfaces with KeyboardInteractivity::None
+                if !layer.can_receive_keyboard_focus() {
+                    return;
+                }
+                None
+            }
```
Drakulix, line 124: Yeah this isn't completely right. We want to avoid setting it as the focused surface for smithay's keyboard abstraction instead. Which means not creating a `KeyboardFocusTarget` in the first place from [`element_under`](https://github.com/pop-os/cosmic-comp/blob/master/src/input/mod.rs#L1982).

ids1024:
`State::element_under` already has a check for this, so I think the change here is unneeded.

https://github.com/pop-os/cosmic-comp/pull/1427 appears to fix the issue with squeekboard.

jackpot51:
Fixed by #1427

Issue #913, opened 2024-10-08
Incorrect GPU selection for session


flukejones:
I have a new laptop with AMD and NVIDIA GPU. The AMD gpu is the iGPU. However there are a few issues:

1. The Nvidia GPU is before the AMD on the PCI bus
```
65:00.0 VGA compatible controller: NVIDIA Corporation AD106M [GeForce RTX 4070 Max-Q / Mobile] (rev a1)
66:00.0 Display controller: Advanced Micro Devices, Inc. [AMD/ATI] Strix [Radeon 880M / 890M] (rev c1)
```

2. The Nvidia GPU is marked as boot_vga

3. This causes the dGPU to be in consistent use and never suspend. It will suspend if switched to a TTY and desktops stopped.

The issue isn't solely cosmic-comp. Gnome is affected also. However if I work around the issue with `nvidia_drm.modeset=0` this prevents gnome grabbing the card, but makes cosmic crash as it tries to use `card0` with now doesn't have DRM ability.

Related issue on [`drm/amd`](https://gitlab.freedesktop.org/drm/amd/-/issues/3673)

Quackdoc:
temporary fix should be `COSMIC_RENDER_DEVICE=/dev/dri/renderD1##` env var? I believe `COSMIC_RENDER_AUTO_ASSIGN=y` may also be warranted.

flukejones:
Unfortunately that does not help. Still selects incorrect gpu

superm1:
Since the same thing came up on mutter I did have a [proposal](https://gitlab.gnome.org/GNOME/mutter/-/issues/3746) there for them to better pick things.  Maybe cosmic can do something similar.

flukejones:
I can confirm that using the udev rule changes as @superm1 suggests in the link above does indeed fix the issues I've experienced in Gnome.

I also discovered that Nvidia 560 driver was causing further issues that maybe masked some things adversely, so I will retest cosmic with 550 driver.

flukejones:
Umm, yeah.. Have to start cosmic with `COSMIC_RENDER_DEVICE=/dev/dri/renderD128 cosmic-session` and with nvidia driver 550 the dgpu suspends. But there are too many things going on for me to adequately track while also trying to work. Some notes:

- Nvidia 550
- kernel 6.12-rc3
- Applied https://lore.kernel.org/linux-pci/20241014152502.1477809-1-superm1@kernel.org/T/#u
- zed editor fails to load (I will get logs another time)
[...]

PR #1429, opened 2025-05-22
DRM debugging / profling

Drakulix:
Series of commits currently blocked on https://github.com/Smithay/smithay/pull/1739.

- Adds new env-variables (and common helpers) to parse various device identifiers to make testing with multi-gpu setups easier
- Updates the code to select the primary-gpu to prefer devices with built-in connectors (eDP, LVDS). (Should address https://github.com/pop-os/cosmic-comp/issues/913)
- Makes the primary node properly dynamic, allowing runtime switches as good as possible, when the device disappears as well as keeping the same device as long as possible. Can still be over-written with an env var and potentially switched at runtime for debugging purposes in the future.
- More profiling, which revealed:
  - Some threads can hog the shell-lock, breaking our timing for the KMS threads. Lets use parking-lot, which has a concept of eventual fairness. In practice this means much more consistent timings
  - Profiling revealed that while we do render at e.g. 120Hz without issues, programs can end up having their refresh rate halved. Turns out this was an oversight from before we limited the frame-callbacks of each surface to one per vblank. Uppps.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -843,6 +843,7 @@ dependencies = [
  "log-panics",
  "once_cell",
  "ordered-float",
+ "parking_lot 0.12.3",
  "png",
  "profiling",
  "rand 0.9.1",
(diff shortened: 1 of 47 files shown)


ids1024:
I've been wanting to have a way to filter what DRM devices are used, without modifying the code, so that's definitely a nice addition.

PR #1441, opened 2025-05-28
Add security context indicator

nesteroff:
This adds the ability to enable a security context indicator around floating windows similar to the focus indicator.

The security context protocol is commonly used when running applications in sandboxed environments. Being able to enable an indicator helps clearly distinguish between sandboxed and regular application windows. This is especially useful when running multiple apps with different sandbox settings. For example, a user might have a "trusted" browser running in a strict secure sandbox with a green window border and a regular one with red.

The indicator can be enabled by defining rules based on the sandbox engine and appid from the security context protocol and by specifying an indicator color in the `security_context` configuration file.

This might not be the most widely requested feature but it’s quite useful in certain scenarios. I’m sharing it here to see if there’s any interest in merging something like this or if others have thoughts or suggestions on how it could be improved.

### DIFF
--- a/cosmic-comp-config/src/lib.rs
+++ b/cosmic-comp-config/src/lib.rs
@@ -99,6 +99,7 @@ pub struct CosmicCompConfig {
     pub edge_snap_threshold: u32,
     pub accessibility_zoom: ZoomConfig,
     pub appearance_settings: AppearanceConfig,
+    pub security_context: SecurityContextConfig,
 }
 
 impl Default for CosmicCompConfig {
@@ -135,6 +136,7 @@ impl Default for CosmicCompConfig {
             edge_snap_threshold: 0,
             accessibility_zoom: ZoomConfig::default(),
             appearance_settings: AppearanceConfig::default(),
+            security_context: SecurityContextConfig::default(),
         }
     }
 }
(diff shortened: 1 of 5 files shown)


ids1024:
One limitation of this is that it won't display anything for X11 windows in a sandbox. It may be good to fallback to a different method of detection for that.

(I intend to use https://github.com/pop-os/cosmic-protocols/pull/59 to allow clients to do this; `/proc/$PID/root/.flatpak-info` is apparently what other things use to determine if a process is running in  a Flatpak.)

nesteroff:
> One limitation of this is that it won't display anything for X11 windows in a sandbox. It may be good to fallback to a different method of detection for that.
> 

That’s a good point. The idea behind this is to have an indicator specifically for the security context protocol. It shows that the app is running in a sandbox (though this can mean different things) and that it has restrictions on which Wayland protocols it can access. For X11 windows, that probably isn't the case so it’s somewhat logical that there wouldn’t be a security context indicator visible.

It would definitely be nice to have a more general "sandbox indicator" mechanism but that really depends on sandboxing engine implementation. I'm not sure we’d want to hardcode logic into the compositor to parse the internals of specific sandbox structures or something like that.

In my specific setup, for example, applications are sandboxed inside virtual machines so there isn't really much information available on the compositor side.

PR #1456, opened 2025-06-05
kms/surface: Split off a couple parts of `redraw()` into different functions, and change how screencopy errors are handled

ids1024:
This follows up https://github.com/pop-os/cosmic-comp/pull/1408 by making sure errors in screencopy don't result in `redraw()` exiting early with an error, which is easiest done by splitting that part of the function off into a different function. It also splits off a couple more functions from `redraw`.

This is still a bit ugly, but it seems nice to structure things a bit more, and not keep growing `redraw()` with more functionality. (It would help if `take_screencopy_frames()` could work without mutating `elements`...)

### DIFF
--- a/src/backend/kms/surface/mod.rs
+++ b/src/backend/kms/surface/mod.rs
@@ -22,10 +22,18 @@ use anyhow::{Context, Result};
 use calloop::channel::Channel;
 use smithay::{
     backend::{
-        allocator::{format::FormatSet, gbm::GbmAllocator, Fourcc},
+        allocator::{
+            format::FormatSet,
+            gbm::{GbmAllocator, GbmBuffer},
+            Fourcc,
+        },
         drm::{
-            compositor::{BlitFrameResultError, FrameError, FrameFlags, PrimaryPlaneElement},
+            compositor::{
+                BlitFrameResultError, FrameError, FrameFlags, PrimaryPlaneElement,
+                RenderFrameResult,
+            },
             exporter::gbm::GbmFramebufferExporter,
+            gbm::GbmFramebuffer,
             output::DrmOutput,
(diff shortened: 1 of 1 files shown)


ids1024:
It's also nice this is a slight negative diff, despite adding functions. Due to de-duplicating some error handling, and just reducing the indent level.

src/backend/kms/surface/mod.rs
```diff
@@ -1454,256 +1308,26 @@ impl SurfaceThreadState {
...
-                                        .reset();
-                                    tracing::warn!(?err, "Failed to screencopy");
-                                }
+                                session
+                                    .user_data()
+                                    .get::<SessionData>()
+                                    .unwrap()
+                                    .lock()
+                                    .unwrap()
+                                    .reset();
```
Drakulix, line 1329: So we are not doing `frame.fail` here anymore. Is there a reason for that? Do we just re-try the frame next time?

Drakulix (approved):
Overall looks like a nice improvement!

src/backend/kms/surface/mod.rs
ids1024: `send_screencopy_result()` consumes the `Frame`,  and since https://github.com/pop-os/cosmic-comp/pull/1395 dropping that will call `fail` with `FailureReason::Unknown`. So the behavior shouldn't be changed here.

We can't call `fail()` explicitly in `redraw()` since the `Frame` has been consumed, unless we do some awkward thing passing `&mut Option<Frame>` or returning `Frame` along with the errors. We can call it in `send_screencopy_result()`, but then don't get nice `?` error handling. So just relying on `Drop` seems best.

Drakulix: Agreed

PR #1457, opened 2025-06-06
Fixes for `frame` callbacks, and surface redraw queing, for dnd and cursor surfaces

ids1024:
It would be nice if some of this code iterating over all surfaces could be shared (it's partly similar to `render_input_order`; but not exactly). But that isn't changed here.

(It would probably also be nice to avoid having to iterate through everything so often in so many places, if possible. No specific suggestions for that though.)

Perhaps `visible_output_for_surface` should also return multiple outputs, when a surfaces overlaps multiple. To queue redraws on every output that's overlapped.

### DIFF
--- a/src/backend/render/cursor.rs
+++ b/src/backend/render/cursor.rs
@@ -20,8 +20,7 @@ use smithay::{
     reexports::wayland_server::protocol::wl_surface,
     render_elements,
     utils::{
-        Buffer as BufferCoords, IsAlive, Logical, Monotonic, Physical, Point, Scale, Size, Time,
-        Transform,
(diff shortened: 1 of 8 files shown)


ids1024:
Oh yeah, I forgot there was an earlier PR for this.

Presumably sending dmabuf-feedback should be done too. This seems to work with the clients tested (I haven't tried anything using animated dmabuf cursors), but I guess it would just be using default feedback. Or the previous feedback for windows being dragged.

Definitely would be nice to somehow reduce the number of places we have code iterating over every type of surface...

ids1024:
Added a commit updating `send_dmabuf_feedback()`.

PR #1465, opened 2025-06-13
feat: power button handling

wash2:
requires:
https://github.com/pop-os/cosmic-settings-daemon/pull/85
https://github.com/pop-os/cosmic-osd/pull/110

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -7,7 +7,7 @@ use crate::{
             cosmic_keystate_from_smithay, cosmic_modifiers_eq_smithay,
             cosmic_modifiers_from_smithay,
         },
-        Action, Config, PrivateAction,
+        Action, Config, PrivateAction, X11_KEYCODE_OFFSET,
     },
     input::gestures::{GestureState, SwipeAction},
     shell::{
@@ -35,8 +35,8 @@ use calloop::{
     RegistrationToken,
 };
 use cosmic_comp_config::{workspace::WorkspaceLayout, NumlockState};
-use cosmic_settings_config::shortcuts;
 use cosmic_settings_config::shortcuts::action::{Direction, ResizeDirection};
+use cosmic_settings_config::{shortcuts, Binding};
 use smithay::{
     backend::input::{
         AbsolutePositionEvent, Axis, AxisSource, Device, DeviceCapability, GestureBeginEvent,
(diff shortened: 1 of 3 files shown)


src/input/mod.rs
```diff
@@ -213,6 +213,33 @@ impl State {
...
+
+                        self.handle_action(
+                            Action::Shortcut(action.clone()),
+                            &seat,
+                            SERIAL_COUNTER.next_serial(),
+                            Event::time_msec(&event),
+                            binding.clone(),
+                            None,
+                        );
+                    }
```
Drakulix, line 240: Imo we shouldn't go through `handle_active` and try to map `XF86_PowerOff` at all.

Just call directly into `self.handle_shortcut_action` with `shortcuts::Action::System(System::PowerOff)` completely bypassing the key bindings and remove the entry from `keybindings.ron`.

src/input/mod.rs
```diff
@@ -213,6 +213,33 @@ impl State {
                     self.common.idle_notifier_state.notify_activity(&seat);
 
                     let keycode = event.key_code();
+
+                    // if power key is pressed, just use the system action for power off
+                    // this is a workaround for the fact that the power key is not
+                    // available in the keymap, so we cannot use the keymap to determine
+                    // if the key is pressed
+                    if keycode.raw() == 116 + 8 {
```
Drakulix: Is there really no const name for this keycode in `xkbcommon`, urgh...

Please introduce two consts with names at least, like e.g. here: https://github.com/pop-os/cosmic-comp/blob/master/src/config/mod.rs#L780

src/input/mod.rs
```diff
@@ -213,6 +213,25 @@ impl State {
                     self.common.idle_notifier_state.notify_activity(&seat);
 
                     let keycode = event.key_code();
+
+                    const POWER_OFF: u32 = 116;
+                    // if power key is pressed, just use the system action for power off
+                    // this is a workaround for the fact that the power key is not
+                    // available in the keymap, so we cannot use the keymap to determine
+                    // if the key is pressed
+                    if keycode.raw() == POWER_OFF + X11_KEYCODE_OFFSET {
```
Drakulix: We probably also want to check `event.state()` to be `Pressed` to not trigger this twice.

src/input/mod.rs
```diff
@@ -213,6 +213,25 @@ impl State {
...
+                        self.handle_action(
+                            Action::Shortcut(shortcuts::Action::System(
+                                shortcuts::action::System::PowerOff,
+                            )),
+                            &seat,
+                            SERIAL_COUNTER.next_serial(),
```
Drakulix: Also given that we continue with the rest of the function (which is fine to make sure the keyboard state is correct), we don't want to increase the serial twice.

So probably just move this block after the variable setup and also re-use `state` and `time`?

Drakulix (commented):
Sorry, I missed some minor things in the first review.

Drakulix (approved):
Thanks!

PR #1471, opened 2025-06-18
WIP Inihibit logind power button behavior

ids1024:
Gnome sets inhibitors in https://gitlab.gnome.org/GNOME/gnome-settings-daemon/-/blob/main/plugins/media-keys/gsd-media-keys-manager.c?ref_type=heads. But as long as we handle the bindings entirely in cosmic-comp, it makes sense to register inhibitors in the same place. This has some quick hacky code that works to inhibit just the power button.

I want to do a bit more to clean up how we handle zbus and futures in cosmic-comp. Which is also needed by https://github.com/pop-os/cosmic-comp/pull/1412. Using the calloop executor is probably ideal, but I've had issues with that in cosmic-workspaces (maybe that's all working after https://github.com/Smithay/calloop/pull/227?).

Since https://github.com/pop-os/cosmic-comp/pull/1465 and https://github.com/pop-os/cosmic-comp/pull/1466, I also see the shutdown dialog when waking from suspend with the power button. We need to do something to prevent that (I don't know if the button press is waking the system, but button release is then seen by the OS?)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -830,6 +830,7 @@ dependencies = [
  "egui",
  "egui_plot",
  "futures-executor",
+ "futures-util",
  "i18n-embed",
  "i18n-embed-fl",
  "iced_tiny_skia",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -829,6 +829,7 @@ dependencies = [
  "cosmic-text",
  "egui",
  "egui_plot",
+ "futures-executor",
  "i18n-embed",
  "i18n-embed-fl",
  "iced_tiny_skia",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,6 +67,7 @@ reis = { version = "0.5", features = ["calloop"] }
 clap_lex = "0.7"
 parking_lot = "0.12.3"
 futures-executor = { version = "0.3.31", features = ["thread-pool"] }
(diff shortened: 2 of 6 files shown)


Drakulix:
We already do this in cosmic-session: https://github.com/pop-os/cosmic-session/commit/b2f42771222b1d0acd267355a83776abd465eff7

Which I also feel like is the better fit, should someone want to mess with the compositor in a different session or vise-versa.

ids1024:
Ah, I didn't check cosmic-session, and I've been testing by manually starting components.

Looks like the issue with the shutdown dialog showing up on wake from suspend doesn't happen with cosmic-session when the greeter is running and shows a lock screen on wake. But does if cosmic-greeter isn't started by cosmic-session.

For a cosmic session using all the normal components, it doesn't really matter which component takes the inhibitor lock. I suppose it depends on how we want to make it configurable (but we don't really need a way to disable the inhibitor lock; just a way to configure what cosmic-comp does in response to the button). Not sure what fits best when using cosmic-comp in some other sort of session, or a different compositor on cosmic.

ids1024:
Async executor changes moved to https://github.com/pop-os/cosmic-comp/pull/1412. Best to not have *too* many extra threads and executors.

Issue #1463, opened 2025-06-11
[BUG] Cosmic removing global wayland objects and crashing apps


thatdevsherry:
# Summary

After screen timeout is reached, it fades out and the display turns off, but is soon again turned on and lock screen is shown indefinitely. During this process apps like qbittorrent and kitty crash. Seems like a cosmic issue.

I noticed if I set timeout to say 15s for testing, after the display goes to sleep, it will always turn back on and show lock screen. If i move mouse a bit and leave it again for 15s on lock screen, display shuts off but again turns on to lock screen. The two errors then reappear (failed to destroy old mode & failed to set xwayland primary output).

# Steps to reproduce
1. Open qbittorrent or kitty
2. Set screen timeout to 2mins for testing
3. Wait for timeout
4. Screen fades out
5. Display goes to sleep
6. Display turns back on quickly
7. Lock screen shown indefinitely
8. Enter password and the apps will have crashed

# Logs

Using locally built cosmic-comp 5eeff37. Asked AI about this and it pointed that the udev rule triggering when display goes to sleep is being interpreted by cosmic-comp as a new display connection, so it creates a new output but apps are still referring to old output

```text
❯ journalctl -f | grep -E "(cosmic-comp)"
Jun 12 15:28:32 pop-os cosmic-comp[43688]: DEBUG: set_dpms called - output: "HDMI-A-1", on: false
Jun 12 15:28:32 pop-os cosmic-comp[43688]: DEBUG: BACKTRACE:
Jun 12 15:28:32 pop-os cosmic-comp[43688]: DEBUG: set_dpms completed for: HDMI-A-1
Jun 12 15:28:32 pop-os cosmic-comp[43688]: Restoring focus to None
Jun 12 15:28:33 pop-os cosmic-comp[43688]: Restoring focus to Some(LockSurface(LockSurface { shell_surface: ExtSessionLockSurfaceV1 { id: ObjectId(ext_session_lock_surface_v1@108[8], 321), version: 1, data: Some(Any { .. }), handle: WeakHandle { handle: WeakInnerHandle[rs] { .. } } }, surface: WlSurface { id: ObjectId(wl_surface@99[8], 320), version: 5, data: Some(Any { .. }), handle: WeakHandle { handle: WeakInnerHandle[rs] { .. } } } }))
Jun 12 15:28:47 pop-os cosmic-comp[43688]: Udev event: type=change, devnum=Some(57857) devnode=Some("/dev/dri/card1")
[...]

PR #1477, opened 2025-06-24
Send `wl_surface::leave` before destroying output global

ids1024:
Fix for https://github.com/pop-os/cosmic-comp/issues/1463. Requires https://github.com/Smithay/smithay/pull/1763.


ids1024:
The `smithay` update here also brings in https://github.com/Smithay/smithay/pull/1764.

Drakulix (approved):

PR #1478, opened 2025-06-24
protocols/workspace: Set ext workspace `id` for pinned workspace

ids1024:
The `id` is defined to be sent only once, on creation of the handle or later. And only for workspaces that are "likely to be stable across multiple sessions".

Set we add an `id` initially for pinned workspaces, and add one when the workspace is pinned.

The `id` is not supposed to be human readable, so we just use a random value.

I don't think we currently have any plans that would need this feature, but this will allow scripts to recognize that a pinned workspace is the same as a workspace from a previous session, and do things like move a certain application to that workspace.

We'll need to update clients with https://github.com/pop-os/cosmic-protocols/pull/62 before merging this.

### DIFF
--- a/cosmic-comp-config/src/workspace.rs
+++ b/cosmic-comp-config/src/workspace.rs
@@ -43,5 +43,6 @@ pub struct OutputMatch {
 pub struct PinnedWorkspace {
     pub output: OutputMatch,
     pub tiling_enabled: bool,
-    // TODO: name, id
+    pub id: Option<String>,
+    // TODO: name
 }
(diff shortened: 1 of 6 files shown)


ids1024:
https://github.com/pop-os/xdg-desktop-portal-cosmic/pull/179, https://github.com/pop-os/cosmic-workspaces-epoch/pull/199, https://github.com/pop-os/cosmic-applets/pull/982, and https://github.com/pop-os/cosmic-panel/pull/411 should be all the necessary client fixes to avoid crashes with this.

(Annoying I missed this event that was added since our earlier version of the protocol was forked; I wish there was a way to make non-exhaustive matching a warning....)

ids1024:
Oh yeah. Now those have been merged for a while, and everything here seems to be working.

PR #1492, opened 2025-06-30
Update Cargo.toml to strip the release binary

LinuxBoy-96:
Strip the binary for the release build.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -118,6 +118,7 @@ inherits = "release"
 
 [profile.release]
 lto = "fat"
+strip = true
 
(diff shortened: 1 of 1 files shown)


ids1024:
I think Linux package managers generally have a mechanism to strip/split debug info when packaging executables, so it's probably not necessary or desirable to configure cargo to strip by default.

LinuxBoy-96:
> I think Linux package managers generally have a mechanism to strip/split debug info when packaging executables, so it's probably not necessary or desirable to configure cargo to strip by default.

You are right, @mmstick told me stripping is not an acceptable way to do that. :(

PR #1493, opened 2025-06-30
kms: Do not attempt to import dmabuf on device not supporting format

ids1024:
Previously, if `expected_node` couldn't import a buffer, it would print an error, then try the next node. There shouldn't really be a reason to attempt import if the format/modifier isn't in `dmabuf_texture_formats`.

It seems the issue I've been seeing with cosmic-workspaces crashing the Nvidia driver is fixed by removing this, and disabling `dma_shadow_copy` (which was producing the same error). Importing Nvidia buffers on the AMD GPU seems to be causing issues.

Not sure how `dma_shadow_copy` should be fixed, but a test for supported formats is easy to add here anyway.

### DIFF
--- a/src/backend/kms/mod.rs
+++ b/src/backend/kms/mod.rs
@@ -16,6 +16,7 @@ use smithay::{
         allocator::{
             dmabuf::Dmabuf,
             gbm::{GbmAllocator, GbmBufferFlags},
+            Buffer,
         },
         drm::{output::DrmOutputRenderElements, DrmDeviceFd, DrmNode, NodeType},
         egl::{context::ContextPriority, EGLContext, EGLDevice, EGLDisplay},
@@ -455,6 +456,17 @@ impl KmsState {
                 _egl = Some(init_egl(&device.gbm).context("Failed to initialize egl context")?);
                 &_egl.as_ref().unwrap().display
             };
+
+            if !egl_display
+                .dmabuf_texture_formats()
+                .contains(&dmabuf.format())
+            {
+                trace!(
+                    "Skipping import of dmabuf on {:?}: unsupported format",
+                    device.render_node
+                );
+                continue;
+            }
 
             let result = egl_display
                 .create_image_from_dmabuf(&dmabuf)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4732,7 +4732,7 @@ checksum = "8917285742e9f3e1683f0a9c4e6b57960b7314d0b08d30d1ecd426713ee2eee9"
 [[package]]
 name = "smithay"
 version = "0.7.0"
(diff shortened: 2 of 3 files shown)


src/backend/kms/mod.rs
```diff
@@ -456,6 +457,17 @@ impl KmsState {
                 &_egl.as_ref().unwrap().display
             };
 
+            if !egl_display
+                .dmabuf_texture_formats()
+                .contains(&dmabuf.format())
+            {
+                trace!(
+                    "Skiopping import of dmabuf on {:?}: unsupported format",
```
Drakulix:
```suggestion
                    "Skipping import of dmabuf on {:?}: unsupported format",
```

Drakulix (approved):
Nitpicks, but seems reasonable

Drakulix:
> It seems the issue I've been seeing with cosmic-workspaces crashing the Nvidia driver is fixed by removing this, and disabling `dma_shadow_copy` (which was producing the same error). Importing Nvidia buffers on the AMD GPU seems to be causing issues.
> 
> Not sure how `dma_shadow_copy` should be fixed, but a test for supported formats is easy to add here anyway.

So this sounds like this is just a workaround as most nvidia buffers will likely use a modifier not supported by amdgpu anyway. When we explicitly select a working format in `dma_shadow_copy` this breaks however?

How exactly does it break? Any dmesg output? Because `dma_shadow_copy` deliberately tries to allocate a buffer readable/writable by two different gpus and that is definitely supposed to work.

ids1024:
Testing with `dma_shadow_copy` disabled but without this PR, it looks like the driver crash is specifically related to the shadow copy, and this change isn't needed for that.

Both issues do cause `EGL] 0x3009 (BAD_MATCH) eglCreateImageKHR: createImageFromDmaBufs failed` errors to be logged though, so this change should be good if only to avoid a flood of error messages.

And yeah, the shadow copy seems like it should work. (And not crash the driver even if it doesn't.) So at least part of this is a driver bug. I need to do a bit more testing to figure out exactly why this is failing. (While `VK_ICD_FILENAMES=/usr/share/vulkan/icd.d/nvidia_icd.json` seems to result in an Nvidia linear buffer being copied to the AMD GPU without issue.)

ids1024:
Ah, it looks like it's trying to use implicit modifiers for shadow copy. In which case, it makes sense that it's causing issues (albeit not driver crashes) and it should be simple enough to address.

ids1024:
https://github.com/Smithay/smithay/pull/1767 seems to work. It no longer produces errors, or driver crashes, and successfully uses shadow copy instead of mem copy.

ids1024:
Added a commit to also bump smithay with that fix.

PR #1498, opened 2025-07-03
screencopy: Fix damage tracking bug with shm screencopy of output

ids1024:
Similar to the change in https://github.com/pop-os/cosmic-comp/pull/780, but also updates it to be a little clearer than just an uncommented `age = 0` line.

Ideally we want some robust system to re-use the offscreen buffer (but not allocate more buffers indefinitely if the client doesn't capture with the same `wl_buffer`).

### DIFF
--- a/src/backend/kms/surface/mod.rs
+++ b/src/backend/kms/surface/mod.rs
@@ -39,7 +39,7 @@ use smithay::{
         },
         egl::EGLContext,
         renderer::{
-            buffer_dimensions,
+            buffer_dimensions, buffer_type,
             damage::Error as RenderError,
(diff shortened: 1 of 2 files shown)


Drakulix:
I wonder if we could instead simply not copy everything from offscreen-buffer to the shm-buffer, but instead the same damaged regions we re-rendered. Then it wouldn't really matter except for the overhead of allocating the buffer.

But yeah, in the meantime this seems fine.

ids1024:
True, that sounds like it should work too. Though we probably also want to avoid reallocating buffers each time as well.

Issue #1474, opened 2025-06-23
Refine focus indication


maria-komarova:
When moving focus with keyboard shortcuts Super+arrows or Super+Alt+arrows from display 1 with open windows to an empty display remove active hint from the previously focused window on display 1.

PR #1507, opened 2025-07-09
Fix workspace key bindings in workspace span outputs mode; activate workspace when it is migrated to an output

ids1024:
The first change here seems to fix the issue described in https://github.com/pop-os/cosmic-workspaces-epoch/issues/95. Not sure if there are more keybinding-related issues with workspaces span outputs mode (there are some other sorts of issues that still need to be addressed).

### DIFF
--- a/src/input/actions.rs
+++ b/src/input/actions.rs
@@ -539,7 +539,7 @@ impl State {
                         res
                     };
 
-                    if let Ok(Some(new_pos)) = res {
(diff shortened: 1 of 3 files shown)


ids1024:
Added a fix for https://github.com/pop-os/cosmic-comp/issues/1474, and a related issue when "focus follows cursor" was enabled, but "cursor follows focus" wasn't.

PR #1518, opened 2025-07-15
kms: Don't use more then 10-bit by default

Drakulix:
Draft because I am not sure this actually fixes anything and will be more annoying to deal with, once we might actually want to scanout XRGB16161616 with HDR content.

But there isn't really any reason this is necessary currently, we never use a format with more than 10-bits, and this might run into bandwidth issues otherwise, which currently require manual edits of `outputs.ron`.

We possibly want to tie this to the selected format in the future and downgrade step-by-step both the format and requested max_bpc, if we get atomic-commit errors.

### DIFF
--- a/src/backend/kms/device.rs
+++ b/src/backend/kms/device.rs
@@ -839,7 +839,7 @@ fn populate_modes(
     position: (u32, u32),
 ) -> Result<()> {
     let conn_info = drm.get_connector(conn, false)?;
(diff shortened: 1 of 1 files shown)


seidler2547:
Could we please make this configurable? There are quite some reports indicating that, for example, using USB-C for high-resolution monitors requires max bpc to be set to 8, and there is currently no way to specify the maximum bit depth. Worse yet, if I limit the bit depth on Intel graphics to 8 via debugfs, cosmic-comp fails to initialise the display because it read the theoretical bpc as 10 and tries to apply it, but fails because the driver doesn't let it.

Issue #144, opened 2023-08-02
Crash with Displaylink dock attached / upon attaching


MHoogesteger:
I've just installed cosmic and tried running it with a displaylink dock connected on a fresh Pop install, but get a crash to gray screen and eventually I see the laptop manufacturers BIOS/UEFI startup logo. Attaching the dock after login causes the same thing.

It runs fine without the displaylink dock. Pop on Wayland runs fine with it.

Installed through apt and not too savy with linux in general, but cat /var/log/syslog | grep cosmic gives me the attached log file. It contains both sessions.
[cosmic_log.txt](https://github.com/pop-os/cosmic-comp/files/12242397/cosmic_log.txt)

Issue #136 seems to be crashing on the same unwrap function?

I've also attached the drm_info output.
[drm_info.txt](https://github.com/pop-os/cosmic-comp/files/12242452/drm_info.txt)

Drakulix:
Right, DisplayLink devices are drm-devices without a render-node, which is essentially the same problem, that some virtual GPUs have (like #136).

We need to offload the rendering another GPU (or for the virtual case to some software renderer like llvmpipe) in this case, which cosmic-comp doesn't currently support.

juhp:
Cosmic also crashes for me when I disconnect my Lenovo USB-C Dock.
I am assuming this is the same problem or related - otherwise I can open a new ticket.

Drakulix:
> Cosmic also crashes for me when I disconnect my Lenovo USB-C Dock.

Does it work before you disconnect it? If so, no this isn't related. Please open a new issue and please also post the output of `drm_info`, so we can check how this dock connects internally to our system and what might cause the crash.

zsilet:
I just came across this same issue on the pop_os 24.04 Alpha1 install. 
My system started crashing immediately after installation of the latest displaylink 6.0 driver package from Synaptic. 
Will monitor this issue and will be happy to test once SW rendering becomes available.

squalou:
For the records, tested on different flavor of dell displaylink docks, same thing (as expected). Plug, crash, relaunch, crash immediately. Keeping an eye on this issue as well :)

ChadAPSheridan:
I have to say, I've fallen in love with CosmicDE, but I got a new work laptop, and it uses a displaylink dock, and now I have to use a different DE until this is sorted.  Sadly no other DE's have the feature set I've grown to love in cosmic thus far.  I'll be watching this like a hawk.

[...]

PR #1521, opened 2025-07-16
Lock during screen configuration changes.

Drakulix:
Draft as it depends on https://github.com/Smithay/smithay/pull/1778 being merged first.

This re-orders a bunch of screen-conf operations, which could cause commits changing the mode being issued with wrong `CRTC_W` and `CRTC_H` values. Additionally this completely locks the kms backend during screen changes to avoid any threads to fail due to temporary state changes.

Should fix https://github.com/pop-os/cosmic-comp/issues/144 (more testing pending).

### DIFF
--- a/src/backend/kms/device.rs
+++ b/src/backend/kms/device.rs
@@ -1,7 +1,10 @@
 // SPDX-License-Identifier: GPL-3.0-only
 
 use crate::{
-    backend::render::{output_elements, CursorMode, GlMultiRenderer, CLEAR_COLOR},
+    backend::{
+        kms::render::gles::GbmGlowBackend,
+        render::{init_shaders, output_elements, CursorMode, GlMultiRenderer, CLEAR_COLOR},
+    },
     config::{AdaptiveSync, EdidProduct, OutputConfig, OutputState, ScreenFilter},
     shell::Shell,
     utils::{env::dev_list_var, prelude::*},
@@ -20,10 +23,11 @@ use smithay::{
         drm::{
             compositor::{FrameError, FrameFlags},
             exporter::gbm::GbmFramebufferExporter,
-            output::DrmOutputManager,
+            output::{DrmOutputManager, LockedDrmOutputManager},
             DrmDevice, DrmDeviceFd, DrmEvent, DrmNode, NodeType,
         },
         egl::{context::ContextPriority, EGLContext, EGLDevice, EGLDisplay},
+        renderer::glow::GlowRenderer,
         session::Session,
     },
     desktop::utils::OutputPresentationFeedback,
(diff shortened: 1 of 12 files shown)


src/backend/kms/device.rs
```diff
@@ -111,6 +128,22 @@ impl fmt::Debug for Device {
...
+    pub outputs: &'a HashMap<connector::Handle, Output>,
+    pub surfaces: &'a mut HashMap<crtc::Handle, Surface>,
+    pub drm: LockedGbmDrmOutputManager<'a>,
+    pub gbm: &'a GbmDevice<DrmDeviceFd>,
+
+    pub leased_connectors: &'a mut Vec<(connector::Handle, crtc::Handle)>,
+    pub leasing_global: &'a mut Option<DrmLeaseState>,
+    pub active_leases: &'a mut Vec<DrmLease>,
+    pub active_buffers: &'a HashSet<Weak<WlBuffer>>,
+}
```
ids1024: I would probably do this by moving the fields of `Device`, other than `GbmDrmOutputManager`, into a `inner: DeviceInner` field so it could have just one reference in `LockedDevice` instead of a bunch. But I guess that just means a different kind of bloat added to the code.

ids1024:
It's hard to review every  line that's changed when a lot of things are being moved around.

Anyway, the obvious risk with adding locks (that are potentially held for a prolonged period) is that we could have a deadlock if the kms thread and main thread acquire the lock on the `shell` and on the `DrmCompositor` in different orders, and hold them at the same time. I'll have to read a bit more to see if this is possible currently, but it is a potential risk if we aren't careful.

Other than that, this seems pretty reasonable.

src/backend/kms/device.rs
Drakulix: I changed the code to use an `InnerDevice` and it actually cleaned up the changes quite a bit getting rid of the weird `MaybeLockedDevice` trait and more duplicated code.

Drakulix:
> Anyway, the obvious risk with adding locks (that are potentially held for a prolonged period) is that we could have a deadlock if the kms thread and main thread acquire the lock on the `shell` and on the `DrmCompositor` in different orders, and hold them at the same time. I'll have to read a bit more to see if this is possible currently, but it is a potential risk if we aren't careful.

While this is a very legitimate concern, the chances this particular PR would introduce a deadlock as especially slim in my opinion, because:
- This introduces basically no changes to the surface thread, which is already architected very 
carefully to never lock both the compositor and the shell at the same time.
- The changes to the main thread are mostly wrapped by `LockedDevice` and `LockedBackend`, which both include a mutable borrow, which means we cannot deadlock purely inside the main-thread as this is statically verified by the compiler.
- The only other case, where we directly lock the `DrmOutputManager` is `resume_session`.

I'll run this branch for a couple of days now just to be sure, but I haven't seen any deadlock in all my (pretty extensive) testing so far.

man0lis:
Just tested and this PR does also fix https://github.com/pop-os/cosmic-epoch/issues/1468 for me.
Thank you <3

Issue #1475, opened 2025-06-23
[BUG] cosmic-comp crashes when attempting to stack DBeaver while it's launching


zsilet:
I can get cosmic-comp to crash almost 100% of the time if I attempt to stack DBeaver Community Edition with any other application, while DBeaver is launching. 

This seems to have something to do with the DBeaver launch sequence, as it does not happen once the app has launched/stabilzed. The DBeaver app takes a solid  4-5 seconds to launch though, during which time its window goes away and comes back at least once, making the issue a bit easier to repeat. This MIGHT be a general vulnearbility/issue, that I am just lucky enough to have come upon. 

I will attach a video. 

Journalctl logs are attached. Look for 'Jun 23 16:37:29 pop-os cosmic-comp[2616]: thread 'main' panicked at 'Placeholder group without real window?': src/shell/layout/tiling/mod.rs:3559'

[log.txt](https://github.com/user-attachments/files/20872139/log.txt)

[sys_info_sw_2025-Jun-23-1628.txt](https://github.com/user-attachments/files/20872141/sys_info_sw_2025-Jun-23-1628.txt)

ids1024:
This is easy to reproduce, at least.

SteamRex25:
I had a similar issue: https://pastebin.com/2SZQxRfi 

I'm on latest Cosmic Arch build. 


Sys Info (Just in-case):
Kernel: 6.18.5-arch1-1
CPU: Ryzen 7 9700x
GPU: rtx 3070
Motherboard: MSI tomahawk b850 max wifi
RAM: 32GB DDR5 6000 MT/s

PR #1528, opened 2025-07-17
grabs/moving: cleanup drag in tiling layer if window doesn't exist

ids1024:
It looks like https://github.com/pop-os/cosmic-comp/issues/1475 isn't happening after this, but I need to test a bit more since it's a little hard to reproduce consistently.

This is also meant to fix an issue that can be seen by closing a window (for instance `exit` in cosmic-term, or super+q) while dragging it still leaving a placeholder in the tiling layout. I'm still not entirely sure why that wasn't causing a crash, but dbeaver was.

This may not be exactly the right way to fix this. I've also noticed an issue where if an output doesn't have any windows, a placeholder isn't created when dragging a window to that output.

### DIFF
--- a/src/shell/grabs/moving.rs
+++ b/src/shell/grabs/moving.rs
@@ -783,6 +783,7 @@ impl Drop for MoveGrab {
         let previous = self.previous.clone();
         let window = self.window.clone();
         let is_touch_grab = matches!(self.start_data, GrabStartData::Touch(_));
+        let cursor_output = self.cursor_output.clone();
 
         let _ = self.evlh.0.insert_idle(move |state| {
             let position: Option<(CosmicMapped, Point<i32, Global>)> = if let Some(grab_state) =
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
Seems fine to me

ids1024:
I guess I didn't get around to looking for better solutions here or adding more fixes to the PR.

But this should be working fine anyway. I guess if @Drakulix thinks it looks fine, we can merge it, and improve the tiling shell code more in later PRs.

* The first commit here fixes the crash with DBeaver, demonstrated in the video on https://github.com/pop-os/cosmic-comp/issues/1475. It seems pretty easy to reproduce with DBeaver if you follow the right steps.
* The second commit fixes a a visual issue with dragging tiling windows.
   - Previously, on dragging to an output with no windows on it, no placeholder would be created to indicate the drag. While a placeholder would be crated to show where the window would be dragged to on the output it was dragged from, but would disappear but not reappear and moving to another output and back.
   - Now it adds placeholders as expected for workspaces with no windows on them.

leviport (approved):
Both fixes are working and I'm not seeing any regressions.

PR #1529, opened 2025-07-17
fix: Make window snapping apply properly on multiple-output configurations

### LINKED ISSUE #1253: Window snapping only works correctly on 1 of my monitors
I run a triple monitor setup and on my left hand monitor the window snapping works exactly as I would expect, but on my other 2 monitors window snapping only works when I resize the window to the edge, when I move the window to the edge no snapping takes place.

Koranir:
Fixes #1253

~~As a bonus, now it snaps to edges across outputs. There might be issues if there are multiple outputs with edges very close to each other, as there is no priority measures in place to, for example, snap to the one closest or has a cursor on, but in practice nobody should have that kind of setup.~~

### DIFF
--- a/src/shell/grabs/moving.rs
+++ b/src/shell/grabs/moving.rs
@@ -383,37 +383,6 @@ impl MoveGrab {
             let mut window_geo = self.window.geometry();
             window_geo.loc += location.to_i32_round() + grab_state.window_offset;
 
-            if matches!(self.previous, ManagedLayer::Floating | ManagedLayer::Sticky) {
-                let loc = (grab_state.window_offset.to_f64() + grab_state.location).as_local();
-                let size = window_geo.size.to_f64().as_local();
-                let output_geom = self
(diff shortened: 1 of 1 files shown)


Drakulix:
> There might be issues if there are multiple outputs with edges very close to each other, as there is no priority measures in place to, for example, snap to the one closest or has a cursor on, but in practice nobody should have that kind of setup.

If developing cosmic-comp has thought me anything, that it is that no matter how unusual a setup enough users will have it and have (from their perspectives) good reasons for it.

So I don't think a solution to this issue should move the code into the output-loop, but still rely on the `cursor_output`. It seems to me, that the fix for the linked issue is simply not converting the output-geometry into `local`-coordinates (which makes a lot of sense and is probably an oversight from when this feature was initially introduced).

Could you please change the PR to simply do that without moving the code around?

Drakulix (approved):
Thanks!

Issue #1526, opened 2025-07-17
[BUG] cosmic-comp crash when attempting to resize Steam Client window.


Hyperchaotic:
1. Open up the Steam Client
2. Grab left or top edge and drag to expand/resize.

- Instead of the grabbed edge the opposite side of the window will resize with the mouse motion. 
- When releasing the mouse button cosmic-comp crashes.

Using Pop!_OS 24.04 and latest updates as of submitting this.

```
Jul 16 17:27:08 pop-os cosmic-comp[2047]: thread 'main' panicked at 'invalid resize state: None': src/shell/layout/floating/grabs/resize.rs:585
                                             0: <unknown>
                                             1: <unknown>
                                             2: <unknown>
                                             3: <unknown>
                                             4: <unknown>
                                             5: <unknown>
                                             6: <unknown>
                                             7: <unknown>
                                             8: <unknown>
                                             9: <unknown>
                                            10: <unknown>
                                            11: <unknown>
                                            12: <unknown>
                                            13: <unknown>
                                            14: <unknown>
                                            15: <unknown>
                                            16: <unknown>
                                            17: <unknown>
                                            18: __libc_start_main
                                            19: <unknown>
                                            
```

Hyperchaotic:
The crash is fixed by pr  #1531 . 

Resize is still the opposite edge when I try to resize using a laptop touchpad.

Hyperchaotic:
https://github.com/user-attachments/assets/54203bfb-d0da-483d-b190-2d305dc59db7

EDIT: It only happens when grabbing+dragging top and left edge, grabbing+dragging bottom or right edge works as expected.

grellanl:
I'm seeing the same with Steam with cosmic-comp ~f2d6f70 - the Steam window resizes normally using mouse, but as soon as I release the mouse button it panics and restarts the session.

cosmic-comp[22168]: thread 'main' panicked at 'invalid resize state: None': src/shell/layout/floating/grabs/resize.rs:585

Hyperchaotic:
Yes, it does seem like the crash and my peculiar "resizing moves opposite side"-thing are different issues, seeing as PR #1531 fixes my crash when I build cosmic-comp with it.

grellanl:
Oh actually, if I resize from the left I do seem to have the "opposite movement" thing before it panics, so maybe they're related (resizing works normally from the right though, until the panic when releasing the mouse button).

AAnzel:
[...]

Issue #1468, opened 2025-06-17
`MoveGrab` started from touch event in SSD is treated as a mouse drag


ids1024:
In `src/shell/element/window.rs`, `Shell::move_request` is called with the `Serial` from `last_seat`. I believe this used to work, but now is getting a higher serial than the one for the touch down event, so `check_grab_preconditions` doesn't find a touch grab for that serial, and adds a mouse grab instead.

Either it needs a better way to get the serial, or for SSD we shouldn't use the serial to determine this.

PR #1531, opened 2025-07-17
floating/resize: Don't panic if something else cancelled the resize

Drakulix:
Fixes crashes like https://github.com/pop-os/cosmic-comp/issues/1526, though I am not sure, what the result of this case will be as Steam resizes perfectly fine on my system with current master.

However I was able to trigger this by starting a resize grab and then also resizing via shortcuts without releasing the pointer. Then on ungrab the `ResizeState` would already have been cleaned up by the shortcut resize.

This is a very constructed scenario where I am not sure what the user would expect anyhow, but we shouldn't crash.

### DIFF
--- a/src/shell/layout/floating/grabs/resize.rs
+++ b/src/shell/layout/floating/grabs/resize.rs
@@ -31,6 +31,7 @@ use smithay::{
     output::Output,
     utils::{IsAlive, Logical, Point, Rectangle, Serial, Size},
 };
+use tracing::debug;
 
 /// Information about the resize operation.
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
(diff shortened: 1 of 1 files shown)


Drakulix:
Ping @ids1024 in case you have a better idea?

ids1024:
I suppose in principle this sort of issue could also occur with different pointers in multi-seat, if we ever did properly support that.

Anyway, I guess it would be better if starting enough resize cancelled the existing one (removing the grab)? A bit more of a hassle to do though.

Drakulix:
> Anyway, I guess it would be better if starting enough resize cancelled the existing one (removing the grab)? A bit more of a hassle to do though.

Yeah and given we a shortcut resize isn't really "starting" a resize (if you don't actually hit any arrow-key, it won't touch the resize-state) I don't know how practical this is. I guess you might also run into this when resizing with the pointer and a touchscreen, I can't test that one though. In that case cancelling might definitely be better.

ids1024:
Good point that this might apply to touch too. Though at the moment I don't seem to be able to start a resize with touch. (I think that used to work? Touch moves currently run into https://github.com/pop-os/cosmic-comp/issues/1468.)

Issue #1522, opened 2025-07-16
Can't alt tab in Celeste


therealmate:
I can't alt tab out of Celeste, it doesn't let me switch windows as seen in the video.
cosmic-comp is git (post-fullscreen-rework)

https://github.com/user-attachments/assets/ab43a7bc-08ce-43ce-8fa2-c883dd747f47

therealmate:
The issue seems to be fixed with #1532 on vanilla Celeste, but still persist on Everest (the Celeste mod loader), however, I can now alt tab, but I have to press it two times very fast

https://github.com/user-attachments/assets/376d4248-60c7-478c-8484-e06726edd49d

PR #1532, opened 2025-07-17
shell: Don't unconditionally focus on unfullscreen

### LINKED ISSUE #1520: Celeste fullscreen can grab focus in questionable places
Open celeste in fullscreen

Observed behavior:
Clicking on another monitor immediately refocuses celeste.
Opening the launcher via super immediately refocuses celeste.
Typing in the lockscreen sometimes doesn't register keys. It seems like the lockscreen and celeste are fighting over the keyboard.
(Enable sound to hear key presses)

https://github.com/user-attachments/assets/cef766d2-7c0d-4801-a7d8-327e9b2fcb76

Moving from the celeste workspace to another one causes celeste to fullscreen on the new workspace.
Clicking on another monitor after that now also moves the fullscreened app to that monitor instead of just grabbing focus.

https://github.com/user-attachments/assets/9fa67fb6-5174-4698-8534-64985f44ba0f

I'm not sure how much of this is allowed by the wayland spec and just a bad implementation by celeste but keyboard grabbing on the lock screen is definitely a cosmic bug. [this issue](https://github.com/pop-os/cosmic-greeter/issues/131) seems related

Drakulix:
Fixes #1520 and #1522.

@ids1024 can you test that it works with the wayland backend on the itch.io variant as well?

### DIFF
--- a/src/wayland/handlers/toplevel_management.rs
+++ b/src/wayland/handlers/toplevel_management.rs
@@ -181,11 +181,9 @@ impl ToplevelManagementHandler for State {
         window: &<Self as ToplevelInfoHandler>::Window,
     ) {
         let mut shell = self.common.shell.write();
-        if let Some(target) = shell.unfullscreen_request(window, &self.common.event_loop_handle) {
-            let seat = shell.seats.last_active().clone();
-            std::mem::drop(shell);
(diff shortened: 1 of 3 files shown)


ids1024 (approved):
Changes look reasonable, and seems to fix behavior (including with alt-tab) in native Wayland client.

PR #1539, opened 2025-07-21
Update dependencies and migrate to Rust 2024

git-f0x:
Updates all dependencies, except `egui` and `egui-plot` (CI reports errors, but builds fine locally).
Removes `once_cell` and `lazy_static` in favor of using `std::sync::LazyLock`.
The `xdg` crate had breaking changes ("changed the return of most constructors from Result<BaseDirectories, Error> to just BaseDirectories"), so there were some changes needed around that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -825,7 +825,7 @@ dependencies = [
  "cosmic-protocols",
  "cosmic-settings-config",
  "cosmic-text",
- "egui 0.32.0",
+ "egui",
  "egui_plot",
  "i18n-embed",
  "i18n-embed-fl",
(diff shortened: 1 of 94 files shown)


Drakulix:
Any idea why changing the edition causes such a big re-format of everything? All the code is already rustfmt-ed and this will cause a bunch of conflicts with on-going pull requests.

Thanks anyway, I have been meaning to do this, I just wonder if we can do this in a less intrusive way.

git-f0x:
The default formatting has changed with Rust 2024, with types, etc. being put at the beginning of imports, while modules, etc. are now placed after.
I just saw that there's an option to specify with which edition's style to format with, so can maybe set it to 2021, and update it at a later point. Though it will format to 2024 for others if they don't configure it on their systems as well.

git-f0x:
When you don't have a lot of pending work, let me know to rebase this, since it shouldn't be too hard with imports rearranging being most of the formatting changes.

git-f0x:
I've set the style to 2021, so formatting should be unchanged. Is the CI error something that I can fix?

git-f0x:
Oh, it appears only `sudo ldconfig` was needed for `libseat` to be found. :)
I think this should be relatively conflict free when merging.

PR #1541, opened 2025-07-21
fix: only activate the active window of the stack on activation

wash2:
should fix https://github.com/pop-os/cosmic-applets/issues/449


hojjatabdollahi:
I actively :heart: the title.

PR #1543, opened 2025-07-24
Set gamma for night light

shocklateboy92:
Submitting early draft for feedback in terms of approach/direction:

I'm currently storing the setting in `ColorFilter` for the moment because it was already passed to apply_config_for_outputs(), but I wonder where the proper place to store it is. How would I even get a `CosmicCompConfig` instance in there?


Drakulix:
I am sorry, but I am not going to merge this at this point.

Mostly because this feature will be part of our color rendering-pipeline for HDR and other color-graded content after the 1.0 release, so I don't want to commit to a particular algorithm (especially with best-guess constants like the BRIGHTNESS_MULTIPLIER), as that is a sure way to get some users complaining that their night light values suddenly looks a bit different, when we inevitably switch to something else.

Additionally this requires some changes in smithay, as we don't want to use the legacy `set_gamma` drm-ioctl, but make the color ramps part of the atomic commits, so that we can smoothly ramp them in lock step with the content being shown.

As a result this code just adds a maintenance burden for planned future changes, that I am not going to commit to at this stage of the project.

shocklateboy92:
Fair enough. Thanks for the quick review!

I can just use this branch on my local machine until your full color rendering pipeline is ready 😸 .
I would offer to help with that, but color management is really out of my depth 😅 

My original version tried to do an atomic commit with a property like so:
```
        // Apply gamma LUT
        let gamma_lut_prop = crate::backend::kms::drm_helpers::get_prop(
            device, crtc, "GAMMA_LUT"
        )?;
        
        let mut req = AtomicModeReq::new();
        req.add_property(crtc, gamma_lut_prop, property::Value::Blob(blob_id));
        device.atomic_commit(AtomicCommitFlags::ALLOW_MODESET, req)?;
```

but that just returned `EINVAL` which it's not even supposed to return _at all_ [according to the kernel docs](https://www.kernel.org/doc/html/v4.15/gpu/drm-kms.html#kms-core-structures-and-functions). I'm beginning to suspect developing on low level graphics stack might require special kind of madness 😶‍🌫️

Drakulix:
Yeah no, that commit would never work, it has to be part of the full commit request assembled in smithay. (And yes it does require a certain kind of madness...)

PR #1545, opened 2025-07-25
kms: Fallback to first available DRM device when no primary GPU is found

ZlordHUN:
Should fix https://github.com/pop-os/cosmic-epoch/issues/1468

Specifically the issue I mentioned in my last comment.

Cosmic-comp is looking for card0 which doesn't exist. Added fallback for such cases.

Edit: Should probably add more context. So comp seems to look for card0 on boot which in my case is the integrated GPU. I disabled my integrated GPU and only use my dedicated RTX 4080. Disabling the integrated GPU didnt change the fact that its still designated as card0 from the looks of it and my dedicated GPU is card1. While Im still learning Rust this fix seems to work fine.

### DIFF
--- a/src/backend/kms/mod.rs
+++ b/src/backend/kms/mod.rs
@@ -235,8 +235,18 @@ fn determine_primary_gpu(
         }
     }
 
-    // else just take the first
-    Ok(drm_devices.values().next().map(|dev| dev.inner.render_node))
+    // else use first available render node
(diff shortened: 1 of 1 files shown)


Drakulix:
This code is doing exactly the same as the old one.

As such it isn't fixing anything, the issue seems to be a race condition.

ZlordHUN:
Crud, back to the drawing board. Thank you for the reply! Helps a lot!

PR #1546, opened 2025-07-28
Fix panic on startup with multiple outputs/gpus after #1521

Drakulix:


ids1024 (approved):
Yep, this seems to fix it.

I had noticed that collecting `others` (which I did to try to print) at the start of the function fixed it, but I hadn't noticed `others` was being used inside a couple loops. So not the issue makes sense, and this seems correct.

PR #1549, opened 2025-07-31
WIP fifo and commit timing

### LINKED ISSUE #2775: FIFO and commit timing

ids1024:
This needs testing, and still need to signal the commit timing.

I'd like to de-duplicate the different places that iterate over all different types of surface (instead of adding more code for this for fifo and commit timing, along with the existing `send_frames`, `update_primary_output`, `send_dmabuf_feedback`, `visisible_output_for_surface`, as well as the similar but slightly different code already shared in `render_input_order()`). But these seem to all be slightly different, so I'm not sure how best to unify them...

### DIFF
--- a/src/backend/kms/surface/mod.rs
+++ b/src/backend/kms/surface/mod.rs
@@ -1061,6 +1061,13 @@ impl SurfaceThreadState {
             vrr = has_active_fullscreen;
         }
 
+        // TODO commit timing `signal_until`
+        {
+            let shell = self.shell.read();
+            // XXX correct way to set time?
+            shell.signal_commit_timing(&self.output, self.clock.now() + estimated_presentation);
+        }
+
         let mut elements = output_elements(
             Some(&render_node),
             &mut renderer,
@@ -1375,6 +1382,9 @@ impl SurfaceThreadState {
                             // If postprocessing, use states from first render
                             let states = pre_postprocess_data.states.unwrap_or(frame_result.states);
                             self.send_dmabuf_feedback(states);
+
+                            let shell = self.shell.read();
+                            shell.signal_fifos(&self.output);
                         }
 
                         if x.is_ok() {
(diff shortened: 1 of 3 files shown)


Drakulix:
> But these seem to all be slightly different, so I'm not sure how best to unify them...

That is the reason they are separate... I didn't want to sacrifice readability for some complex code-deduplication that can handle all the little differences.

ids1024:
Yeah. If it were just as simple as needing an iterator over all types of windows (and other top level surfaces that may have their own popups and subsurfaces), that wouldn't be too bad. And all the uses could exhaustively match against any surface types. But then something like `visisible_output_for_surface` isn't just iterating over ever surface on an output, but trying to do the reverse.

There may be some way to structure all of this better, without a lot of duplication, of fairly opaque use of some very complicated surface iteration API, but I'm not sure the best way to do it.

skygrango:
After two weeks of testing, the current result is already sufficiently usable for me. At this point, I feel that I have reached a meaningful milestone, and I do not currently see a clear path toward further short-term improvements. Because of that, I wanted to share these findings with you in case they are useful

This is my result :https://github.com/skygrango/cosmic-comp/commit/a562e8981d28c0a58c8e167cce013241a0f5ab9e

my current conclusion from the investigation is that calling `blocker_cleared` requires access to the `CompositorHandler` trait. As a result, barrier release handling is currently constrained to the main thread, and the ownership model makes this difficult to manage cleanly.

the same limitation also exists in the `winit` backend. Additionally, since the winit backend's render loop cannot be interfered with, this creates a bottleneck for latency management.

By the way, DRM on_vblank event handling also does not currently run on a dedicated thread, so there is no reliable mechanism to guarantee low-latency timing there either.

Because of these constraints, only the KMS thread (VRR with Scanout) is able to achieve relatively stable frame pacing.

I moved the commit-timing check and barrier release logic into queue_redraw. One reason is that this allows redraw scheduling to be adjusted dynamically in order to satisfy the client's timing requirements. The other reason is to give the main thread enough opportunity to process and release the barrier in time.

You can test this using `present-timing`, a utility from https://github.com/themaister/granite

This is a very useful tool for testing the support and effectiveness of fifo-v1 and commit-timing-v1

for reference (with VRR):
Screenshot_2026-05-24_13-13-07

PR #1553, opened 2025-07-31
fix: check layer map for root popup surface

wash2:
There are some additional panel and applet changes for grabs, but this should enable grabs for popups on layer shell surfaces.

### DIFF
--- a/src/wayland/handlers/xdg_shell/mod.rs
+++ b/src/wayland/handlers/xdg_shell/mod.rs
@@ -4,11 +4,12 @@ use crate::{
     shell::{focus::target::KeyboardFocusTarget, grabs::ReleaseMode, CosmicSurface, PendingWindow},
     utils::prelude::*,
 };
+use smithay::desktop::layer_map_for_output;
 use smithay::{
     delegate_xdg_shell,
     desktop::{
         find_popup_root_surface, PopupGrab, PopupKeyboardGrab, PopupKind, PopupPointerGrab,
-        PopupUngrabStrategy,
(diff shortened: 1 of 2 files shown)


src/wayland/handlers/xdg_shell/mod.rs
```diff
@@ -75,52 +78,75 @@ impl XdgShellHandler for State {
...
+            return;
+        }
+        let maybe_root: Option<KeyboardFocusTarget> = maybe_root.and_then(|root| {
+            let shell = self.common.shell.read();
+            shell
+                .element_for_surface(&root)
+                .map(|r| r.clone().into())
+                .or_else(|| {
+                    shell.outputs().find_map(|o| {
+                        layer_map_for_output(o).layers().find_map(|layer| {
```
Drakulix: We do have `layer_for_surface`, which is probably more appropriate here.

So something like: `layer_map_for_output(o).layer_for_surface(root, WindowSurfaceType::ALL).cloned().map(KeyboardFocusTarget::LayerSurface)`, which would then also be able to find subsurfaces.

Drakulix (commented):
Looks good to me apart from some minor nitpick. I like the additional log messages, that should make debugging unexpected behaviour easier in the future.

src/wayland/handlers/xdg_shell/mod.rs
wash2: Ok, I'll try that, thanks!

Drakulix (approved):
Thanks!

PR #1555, opened 2025-07-31
Activation token

wash2:
This should resolve https://github.com/pop-os/cosmic-applets/issues/855

### DIFF
--- a/src/wayland/handlers/xdg_activation.rs
+++ b/src/wayland/handlers/xdg_activation.rs
@@ -42,23 +42,20 @@ impl XdgActivationHandler for State {
             })
             .unwrap_or(false)
         {
-            if let Some(seat) = data.serial.and_then(|(_, seat)| Seat::from_resource(&seat)) {
-                let output = seat.active_output();
-                let mut shell = self.common.shell.write();
-                let workspace = shell.active_space_mut(&output).unwrap();
-                let handle = workspace.handle;
-                data.user_data
-                    .insert_if_missing(move || ActivationContext::Workspace(handle));
-                debug!(?token, "created workspace token for privileged client");
-            } else {
-                data.user_data
-                    .insert_if_missing(|| ActivationContext::UrgentOnly);
-                debug!(
-                    ?token,
-                    "created urgent-only token for privileged client without seat"
-                );
-            }
-
+            let seat = data
+                .serial
+                .and_then(|(_, seat)| Seat::from_resource(&seat))
+                .unwrap_or_else(|| {
+                    let shell = self.common.shell.read();
+                    shell.seats.last_active().clone()
+                });
+            let output = seat.active_output();
+            let mut shell = self.common.shell.write();
+            let workspace = shell.active_space_mut(&output).unwrap();
+            let handle = workspace.handle;
+            data.user_data
+                .insert_if_missing(move || ActivationContext::Workspace(handle));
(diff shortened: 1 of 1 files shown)


src/wayland/handlers/xdg_activation.rs
```diff
@@ -134,13 +132,34 @@ impl XdgActivationHandler for State {
                         .map(|w| *w == current_workspace.handle)
                         .unwrap_or(false);
```
Drakulix: While we are at it:

```suggestion
                        .is_some_and(|w| *w == current_workspace.handle);
```

src/wayland/handlers/xdg_activation.rs
```diff
@@ -134,13 +132,34 @@ impl XdgActivationHandler for State {
...
+                                .position(|w| {
+                                    element_workspace
+                                        .as_ref()
+                                        .map(|e| *e == w.handle)
+                                        .unwrap_or(false)
+                                })
+                                .map(|i| (o.clone(), i))
+                        });
+
+                        if let Some((output, idx)) = maybe {
```
Drakulix:
```suggestion
                        if let Some(idx) = shell.workspaces.idx_for_handle(&element_output, &element_workspace);
```

src/wayland/handlers/xdg_activation.rs
```diff
@@ -134,13 +132,34 @@ impl XdgActivationHandler for State {
                         .map(|w| *w == current_workspace.handle)
```
Drakulix: I can't put this comment five lines higher (thanks github):

`let Some((element_output, element_workspace)) = shell.space_for(&element).map(|w| (w.output.clone(), w.handle.clone())) else { return };`

This gives us the option to use `workspaces.idx_for_handle` later (see other comments) and gets rid of the whole dance around the optional `element_workspace`. If we can't determine the workspace, we can't really activate it anyway.

src/wayland/handlers/xdg_activation.rs
```diff
@@ -134,13 +132,34 @@ impl XdgActivationHandler for State {
...
                         .unwrap_or(false);
 
-                    if in_current_workspace {
-                        current_workspace
-                            .floating_layer
-                            .space
-                            .raise_element(&element, true);
+                    if !in_current_workspace {
```
Drakulix, line 136: Ah, so this will partially address https://github.com/pop-os/cosmic-comp/issues/192 as well. Nice, I was planning on making this change anyway.

src/wayland/handlers/xdg_activation.rs
```diff
@@ -134,13 +132,34 @@ impl XdgActivationHandler for State {
...
+                                .map(|i| (o.clone(), i))
+                        });
+
+                        if let Some((output, idx)) = maybe {
+                            let _ = shell.activate(
+                                &output,
+                                idx,
+                                WorkspaceDelta::new_shortcut(),
+                                &mut self.common.workspace_state.update(),
+                            );
```
Drakulix: We probably want to abort and return here, if we don't find an `idx` or for some reason `activate` fails (that should be the same error condition, so maybe a warning is also good enough). We don't want to call `raise_element` on some unrelated workspace.

Drakulix (commented):
Looks good! I think we can clean up the `request_activation` logic a bit though.

Drakulix (approved):
thx!

Issue #350, opened 2024-03-10
"stuck" keys on xwayland 


c0h2r:
## description

So, I have 2 xwayland windows (one is a terminal, and another one doesn't really matter). Then I exit the shell with `<C-d>` which closes the terminal, and 'd' gets kinda "stuck" (for the lack of a better word in my vocab)  for all other xwayland windows I focus after that and until I press something else.

## demo

https://github.com/pop-os/cosmic-comp/assets/59478667/525c1c14-49ea-400c-bb6d-25e4439419cc

Here I show the behavior on 2 instances of  wezterm with `enable_wayland = false` specified in the config, but the same also goes for 2 `st`s or wezterm and brave (without the env vars that tell chromium stuff to run natively)

## details

OS: NixOS
Cosmic-comp commit: probably 9cb1adb37cc7e0dab9b5e7076aee292e244c421b (i.e. the one in https://github.com/lilyinstarlight/nixos-cosmic/blob/d9f0cdbc53b3a602e9faea5b2efbe7804639cb3c/pkgs/cosmic-comp/package.nix)

Drakulix:
Oh that is a fun bug... thanks for reporting.

Initial guess what is causing this issue (more a note to myself): Application closes, cosmic-comp revokes focus from the window, we don't have a valid target to send keyboard events to, Xwayland never gets `wl_keyboard::key`-event for the released key. (It should however get the `keyboard::leave` event, but most likely doesn't do anything with it.)

Needs some more in-depth debugging and probably some workaround for release-events with long-running clients.

WatchMkr:
I'm unable to reproduce this bug. Can it be closed?

hgaiser:
I still have this issue on alpha 5. If I run `glxgears` and exit by pressing Escape, then run `glxgears` again it closes immediately because it receives an Escape key event.

The same thing seems to happen in any XWayland application that can be closed by a key event. Ie. if I write an OpenCV image viewing application and add an option to exit through some key event, the same thing happens.

Small disclaimer: it seems to be XWayland applications only, but I'm not 100% sure. For example if I try to fix it by going to a Qt application, hit the stuck key, then `glxgears` still exits immediately.

Let me know if you have some pointers in the code where I can put some debugging code to help resolve this issue.

leviport:
It can also be recreated with Chromium. If I close Chromium with Super + Q, relaunching it causes it to close immediately. 

[...]

PR #1567, opened 2025-08-06
xwayland: Don't race/drop focus events

Drakulix:
Turns out the bug behind https://github.com/pop-os/cosmic-comp/issues/350 wasn't in smithay, which correctly sends `wl_keyboard::leave`-events even to dead surfaces (resulting in the null-surface on the client side, which has already destroyed the handle, which is still correctly handled by Xwayland), but in cosmic-comps handling, as we send both `enter` **and** `leave` events before/after a wayland surface was associated with the window.

This is an alternative this approach of fixing this upstream in case that doesn't work out: https://github.com/Smithay/smithay/pull/1793

(Draft until a decision has been reached.)

### DIFF
--- a/src/wayland/handlers/xwayland_shell.rs
+++ b/src/wayland/handlers/xwayland_shell.rs
@@ -1,14 +1,61 @@
 // SPDX-License-Identifier: GPL-3.0-only
 
-use crate::state::State;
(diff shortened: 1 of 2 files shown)


Drakulix:
https://github.com/Smithay/smithay/pull/1793 was merged

PR #1568, opened 2025-08-06
improv(stack): optimize, improve styling, and fix keyboard focus

mmstick:
- Fixed keyboard focus sometimes being lost (sent to wrong window in stack) when focusing windows in a stack
- Fix tab overflow overlay gradient colors to use the recommended gradients in Figma
- Fix padding of elements in the tab, which will now match the Figma design
- Track Tab hover state to apply hover background tab colors
- Cache tab model data in CosmicStackInternal to borrow and reuse between view method calls
- Refactored properties in CosmicStackInternal to not require Arc
- Refactored CosmicStack to queue window add/remove/swap messages for CosmicStackInternal's update method
    - Eliminates need for a Mutex when accessing windows attached to CosmicStackInternal
    - Lets CosmicStackInternal cache tab model data, and keep it in sync with window order changes
    - Allows CosmicStackInternal to redraw itself without needing to force a redraw
    - Fixes keyboard focus sometimes being sent to the wrong window in the stack on focus changes

### DIFF
--- a/src/shell/element/stack/tab_text.rs
+++ b/src/shell/element/stack/tab_text.rs
@@ -15,6 +15,34 @@ use cosmic::{
 
 use super::tab::Model;
 
+const OVERLAY_DARK: Color = Color {
+    a: 1.0,
+    r: 0.149,
+    g: 0.149,
+    b: 0.149,
+};
+
+const OVERLAY_DARK_SELECTED: Color = Color {
+    a: 1.0,
+    r: 0.196,
+    g: 0.196,
+    b: 0.196,
+};
+
+const OVERLAY_LIGHT: Color = Color {
+    a: 1.0,
+    r: 0.894,
+    g: 0.894,
+    b: 0.894,
+};
+
+const OVERLAY_LIGHT_SELECTED: Color = Color {
+    a: 1.0,
+    r: 0.831,
+    g: 0.831,
+    b: 0.831,
+};
+
 /// Text in a stack tab with an overflow gradient.
 pub fn tab_text(model: &Model, selected: bool) -> TabText {
     TabText::new(model, selected)
@@ -152,30 +180,10 @@ impl<'a, Message> Widget<Message, cosmic::Theme, cosmic::Renderer> for TabText<'
             renderer.with_layer(bounds, |renderer| {
                 if state.overflowed {
                     let overlay = match (theme.cosmic().is_dark, self.selected) {
-                        (true, false) => Color {
-                            a: 1.0,
-                            r: 0.149,
-                            g: 0.149,
-                            b: 0.149,
-                        },
-                        (true, true) => Color {
-                            a: 1.0,
-                            r: 0.196,
-                            g: 0.196,
-                            b: 0.196,
-                        },
-                        (false, false) => Color {
-                            a: 1.0,
-                            r: 0.894,
-                            g: 0.894,
-                            b: 0.894,
-                        },
-                        (false, true) => Color {
-                            a: 1.0,
-                            r: 0.831,
-                            g: 0.831,
-                            b: 0.831,
-                        },
+                        (true, false) => OVERLAY_DARK,
+                        (true, true) => OVERLAY_DARK_SELECTED,
+                        (false, false) => OVERLAY_LIGHT,
+                        (false, true) => OVERLAY_LIGHT_SELECTED,
                     };
 
                     let transparent = Color { a: 0.0, ..overlay };


src/shell/element/stack.rs
```diff
@@ -77,34 +78,38 @@ use self::{
...
 
-static SCROLLABLE_ID: Lazy<Id> = Lazy::new(|| Id::new("scrollable"));
+fn scrollable_id() -> Id {
+    static SCROLLABLE_ID: OnceLock<Id> = OnceLock::new();
+    SCROLLABLE_ID.get_or_init(|| Id::new("scrollable")).clone()
+}
 
 #[derive(Clone, PartialEq, Eq, Hash)]
-pub struct CosmicStack(pub(super) IcedElement<CosmicStackInternal>);
+pub struct CosmicStack(pub(super) Arc<IcedElement<CosmicStackInternal>>);
```
Drakulix: Lets please not do that. The whole story around how the iced types can and should be cloned has been an issue in the past, especially around stacks. Lets just put the `CosmicStackInternal` into an `Arc`.

src/shell/element/stack.rs
```diff
@@ -838,13 +825,22 @@ impl Program for CosmicStackInternal {
...
+                self.windows.iter().enumerate().for_each(|(i, w)| {
+                    if i == self.active.load(Ordering::SeqCst) {
+                        w.set_suspended(false);
+                    } else {
+                        w.set_suspended(true);
+                    }
+                    w.send_configure();
+
+                    SpaceElement::refresh(w)
+                });
```
Drakulix: We can't do any of this in `update` as update might be called when rendering, which happens on a different thread. All operations on the windows should be done in the main thread, which is why the other message dispatch these onto the `loop_handle`.

For these reasons I am also not 100% sure why we move a lot of the login into `update` in the first place instead of handling them, where they happen and we already are in the main thread?

Drakulix:
Sorry if that came over as a bit harsh.

> Fix tab overflow overlay gradient colors to use the recommended gradients in Figma
> Fix padding of elements in the tab, which will now match the Figma design
> Track Tab hover state to apply hover background tab colors
> Cache tab model data in CosmicStackInternal to borrow and reuse between view method calls

These are all very welcome and I appreciate the design around the tab-model a lot. Given this is all bundled into a single commit, I just felt like I had to address the issues first, sorry!

src/shell/element/stack.rs
mmstick: This would require to `impl Program on Arc<CosmicStackInternal>`, and thus make all properties immutable. Wrapping the tab models in a mutex would make it impossible to borrow in the view method.
                                                                                                                                                                                                                                                                                                                          
Seems the only option then is to wrap the tab models into an immutable Arc<[T]>, and then regenerating the models when a Refresh message is received.

src/shell/element/stack/tab_text.rs
```diff
@@ -142,49 +142,65 @@ impl<Message> Widget<Message, cosmic::Theme, cosmic::Renderer> for TabText {
...
-                    gradient::Linear::new(Degrees(90.))
-                        .add_stop(0.0, transparent)
-                        .add_stop(1.0, background),
-                )),
-            );
-        }
+            renderer.with_layer(bounds, |renderer| {
+                if state.overflowed {
+                    let overlay = match (theme.cosmic().is_dark, self.selected) {
+                        (true, false) => Color {
```
Drakulix: I guess it would be nice to move these into constants with proper names, but given they are only used once, this isn't important.

src/shell/element/stack.rs
```diff
@@ -340,9 +335,8 @@ impl CosmicStack {
                             p.scroll_to_focus.store(true, Ordering::SeqCst);
                             (true, true)
                         } else {
-                            let new = prev_idx.unwrap().1;
-                            let old = p.active.swap(new, Ordering::SeqCst);
-                            if old != new {
+                            let old = p.active.swap(0, Ordering::SeqCst);
+                            if old != 0 {
```
Drakulix: I don't think this is correct. We want to restore the previously focused tab, when we move through a whole stack with one modifier-press. See https://github.com/pop-os/cosmic-comp/issues/1005

src/shell/element/stack.rs
```diff
@@ -371,9 +365,8 @@ impl CosmicStack {
                             p.scroll_to_focus.store(true, Ordering::SeqCst);
                             (true, true)
                         } else {
-                            let new = prev_idx.unwrap().1;
-                            let old = p.active.swap(new, Ordering::SeqCst);
-                            if old != new {
+                            let old = p.active.swap(max - 1, Ordering::SeqCst);
+                            if old != max - 1 {
```
Drakulix: Same as above, see https://github.com/pop-os/cosmic-comp/issues/1005

src/shell/element/stack.rs
```diff
@@ -415,12 +408,9 @@ impl CosmicStack {
...
                     if !p.group_focused.load(Ordering::SeqCst) {
-                        let new = prev_idx.unwrap().1;
-                        let old = p.active.swap(new, Ordering::SeqCst);
-                        if old != new {
-                            p.previous_keyboard.store(old, Ordering::SeqCst);
-                            p.scroll_to_focus.store(true, Ordering::SeqCst);
-                        }
+                        p.previous_keyboard
+                            .store(prev_idx.unwrap(), Ordering::SeqCst);
+                        p.scroll_to_focus.store(true, Ordering::SeqCst);
```
Drakulix: Same as above, this doesn't update `active`: https://github.com/pop-os/cosmic-comp/issues/1005

src/shell/element/stack.rs
```diff
@@ -466,20 +454,20 @@ impl CosmicStack {
...
-                } else if active == windows.len() {
-                    p.active.store(active - 1, Ordering::SeqCst);
-                    p.scroll_to_focus.store(true, Ordering::SeqCst);
-                }
                 window.try_force_undecorated(false);
                 window.set_tiled(false);
 
+                let active = moved_horizontally
+                    .then(|| if active == 0 { 0 } else { windows.len() - 1 })
+                    .unwrap_or_else(|| active.checked_sub(1).unwrap_or(0));
```
Drakulix: Again see https://github.com/pop-os/cosmic-comp/issues/1005, this should use the prev_idx if set.

mmstick:
Rebased and removed the last commit. Still testing though

PR #1575, opened 2025-08-10
Set wlr_output_head serial number

duvholt:
This is a follow up on https://github.com/Smithay/smithay/pull/1795 where I added output serial number support in Smithay. 

As mentioned in that PR my main use case for adding serial numbers is for Shikane/Kanshi to correctly identify my monitor setup since I have a dual monitor setup with two DELL U2715H.

`wlr-randr` output after this change:

```
[..]
DP-5 "Dell Inc. - DELL U2715H - DP-5"
  Make: Dell Inc.
  Model: DELL U2715H
  Serial: GH85D5AL0DWL
[..]
```

I also plan to extend cosmic-randr to show serial number as it is currently only shown when using `cosmic-randr list --kdl` right now.

### DIFF
--- a/src/backend/kms/device.rs
+++ b/src/backend/kms/device.rs
@@ -949,6 +949,10 @@ fn create_output_for_conn(drm: &mut DrmDevice, conn: connector::Handle) -> Resul
                 .as_ref()
                 .and_then(|info| info.model())
                 .unwrap_or_else(|| String::from("Unknown")),
+            serial_number: edid_info
+                .as_ref()
+                .and_then(|info| info.serial())
+                .unwrap_or_else(|| String::from("Unknown")),
         },
     );
     if let Some(edid) = edid_info.as_ref().and_then(|x| x.edid()) {
--- a/src/backend/winit.rs
+++ b/src/backend/winit.rs
@@ -148,6 +148,7 @@ pub fn init_backend(
         subpixel: Subpixel::Unknown,
         make: "COSMIC".to_string(),
         model: name.clone(),
+        serial_number: "Unknown".to_string(),
     };
     let mode = Mode {
         size: (size.w, size.h).into(),
--- a/src/backend/x11.rs
+++ b/src/backend/x11.rs
@@ -94,6 +94,7 @@ impl X11State {
             subpixel: Subpixel::Unknown,
             make: "COSMIC".to_string(),
             model: name.clone(),
+            serial_number: "Unknown".to_string(),
         };
         let mode = Mode {
             size: (size.w as i32, size.h as i32).into(),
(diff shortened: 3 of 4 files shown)


hojjatabdollahi:
@duvholt are you telling me I don't have to open cosmic settings and re-arrange my monitors every time I connect my dock at work? :face_holding_back_tears:

Drakulix:
> @duvholt are you telling me I don't have to open cosmic settings and re-arrange my monitors every time I connect my dock at work? 🥹

Tbf, serial numbers aren't the perfect solution to these issues (though I am glad to provide some infrastructure to support scripting things like these via kanshi/shikane given we support the relevant protocol anyway). There are (especially no-name) monitors reporting the same serial number for all monitors of the same model.

The correct solution here is unfortunately hashing/comparing the whole EDID (something which really isn't very user-friendly). We already started using that for pinned-workspaces to experiment more with this approach. The goal is to eventually extend this to the stored output config, so you'll get stable positions/layouts in these circumstances as well.

hojjatabdollahi:
Ah! Thanks for the explanation. 
I just tested it by pinning a workspace per monitor, unplugged my dock and plugged it back in and even though the displays were rearranged (You can tell since the cursor jumps from the left monitor to the right and then the middle) but the workspaces stayed on the correct monitor. That's awesome.

Can't wait until that lands for the display config too. Thank you.

hojjatabdollahi:
@Drakulix Just FYI, after this PR if I connect my external monitors cosmic-settings wouldn't show anything in the display page. This is being printed out open cosmic-settings:
```
  ERROR  error fetching displays, why: "`cosmic-randr` KDL format error"
    at cosmic-settings/src/pages/display/mod.rs:623 on main
```
The only difference between the kdl output of my internal display and the external one is the serial_number field, so I tried disabling it.

Commenting out the bit in cosmic-randr that prints out the serial_number fixed the issue for me.
```
diff --git a/cli/src/main.rs b/cli/src/main.rs
index 6865a36..81b0ec1 100644
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -683,9 +683,9 @@ fn list_kdl(context: &Context) {
                 })
                 "\n"
             }
-            if !head.serial_number.is_empty() {
-                "  serial_number=\"" (head.serial_number) "\"\n"
-            }
+            // if !head.serial_number.is_empty() {
+            //     "  serial_number=\"" (head.serial_number) "\"\n"
+            // }
             "  modes {"
         );
```

It looks like cosmic-settings does not have a field for "serial number" and it wasn't a problem before since cosmic-comp wasn't passing that to cosmic-randr.

PR #1578, opened 2025-08-12
smithay update for Xwayland focus fixes

Drakulix:
Smithay update bringing in https://github.com/Smithay/smithay/pull/1793 to fix https://github.com/pop-os/cosmic-comp/issues/350 and possibly a lot of xwayland app startup focus issues (mostly affecting games).

Branch to test more than the bare minimum before merging given this affects focus handling.


Drakulix:
Seems to work fine and improves some games as well (though some still end up minimized on startup..).

PR #1580, opened 2025-08-12
WIP Use `glFinish` to synchronize dmabuf image-copy-capture

ids1024:
This seems to fix the visual corruption issues, that appear at least on certain Intel hardware with multiple outputs. And also fixes the issue with https://github.com/pop-os/cosmic-protocols/blob/main/client-toolkit/examples/screenshot-screencopy-dma.rs capturing an entirely empty image.

Using `sync.wait()` *should* have essentially the same effect, but doesn't seem to work. Looking at https://github.com/Smithay/smithay/blob/1c962f97818922ab2ac4db48409318c1def9c5b3/src/backend/drm/compositor/frame_result.rs#L349-L373, perhaps the issue is that it's returning a sync point before it blits the damage regions (if `elements_to_render.is_empty()`)?

In any case, ideally we don't want to be blocking here. So if implicit sync won't work (and it's unclear how we might make it work if it doesn't just magically work as it should?) and the image-copy-capture protocol doesn't have explicit sync, I suppose we want to get a sync point that does indicate all the blits are done, and then call `.success` only after that sync point is signaled...

### DIFF
--- a/src/wayland/handlers/screencopy/render.rs
+++ b/src/wayland/handlers/screencopy/render.rs
@@ -18,7 +18,10 @@ use smithay::{
     desktop::space::SpaceElement,
     input::Seat,
     output::{Output, OutputNoMode},
-    reexports::wayland_server::protocol::{wl_buffer::WlBuffer, wl_shm::Format as ShmFormat},
+    reexports::{
+        glow::HasContext,
+        wayland_server::protocol::{wl_buffer::WlBuffer, wl_shm::Format as ShmFormat},
+    },
     utils::{
         Buffer as BufferCoords, IsAlive, Logical, Physical, Point, Rectangle, Scale, Size,
         Transform,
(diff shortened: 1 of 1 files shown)


ids1024:
https://github.com/Smithay/smithay/pull/1805 has some changes to produce a `SyncPoint` we can use here. Which would be simple to wait on, but we want to use it instead as a calloop source that doesn't block the thread. Which also shouldn't be too hard.

Hard to tell if synchronization is correct. Simply creating the `EGLFence`, or doing a `glFlush`, seems to make the artifacts I can reproduce go away, even without waiting on the fence. But that isn't a correct fix, unless `glFlush` is really all we need for implict sync to end up working here. (And even then, it may not be good enough for cross-device uses, or on Nvidia.)

I guess we do need a proposal for explicit sync in the image-copy protocol...

ids1024:
Closing in favor of https://github.com/pop-os/cosmic-comp/pull/1595, which should address the same cases but without blocking on a `glFinish` call.

PR #1585, opened 2025-08-15
focus: Don't consider xwayland grabs to be exclusive

### LINKED ISSUE #449: Input handling issues with VirtualBox
When using VirtualBox in COSMIC Epoch, I sometimes find my keyboard input still going to VirtualBox and not to other apps after escaping from VirtualBox and switching to another workspace.

This has occurred both with mouse pointer integration enabled and disabled (the latter meaning I have to click inside of the VM/use the right-ctrl key to escape in order to capture/release input). However, when it happens, switching from one to the other seems to work around it for a short time.

Drakulix:
Should fix #449

### DIFF
--- a/src/debug.rs
+++ b/src/debug.rs
@@ -344,7 +344,6 @@ fn format_keyboard_focus(focus: Option<KeyboardFocusTarget>) -> String {
         Some(Popup(x)) => format!("Popup {}", x.wl_surface().id().protocol_id()),
         Some(Group(_)) => format!("Window Group"),
         Some(LockSurface(x)) => format!("LockSurface {}", x.wl_surface().id().protocol_id()),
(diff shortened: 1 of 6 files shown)


Drakulix:
Okay, new commit, new approach.

If we want this to behave like Gnome or KDE, then we basically only want the grab to inhibit shortcuts, but neither have it take over focus nor force the focus on the window.

Aaaannnd looking at the KWin source code, that is exactly what other compositors seem to do anyway.

This appears to be working well now for me, can you re-test @jacobgkau?

jacobgkau (approved):
This is working as expected now! The keyboard (including both regular keys and modifiers) goes to the VM when it's focused, and to the host when it's not focused.

PR #1595, opened 2025-08-19
kms/surface: Don't send screencopy`success()` until sync point is reached

ids1024:
Uses https://github.com/Smithay/smithay/pull/1805 to get a sync point from the blit, and then waits on it in a calloop source without blocking the thread.

Like https://github.com/pop-os/cosmic-comp/pull/1580, this should fix some screen capture synchronization issues, but without the overhead of a `glFinish`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4792,7 +4792,7 @@ checksum = "67b1b7a3b5fe4f1376887184045fcf45c69e92af734b7aaddc05fb777b6fbd03"
 [[package]]
 name = "smithay"
 version = "0.7.0"
-source = "git+https://github.com/smithay/smithay.git?rev=ae1faae#ae1faaeb71e2664f4ddb3db7423a5ec746a6a51d"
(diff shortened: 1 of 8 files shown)


ids1024:
Actually, `PendingImageCopyData` can also be used in the return value of `render_session`, then we can make sure this synchronization is handled the same way for every type of capture (without adding duplication).

ids1024:
I think this should be good to merge pending https://github.com/Smithay/smithay/pull/1805.

Drakulix (approved):
LGTM.

There is always optimizations to be had here regarding deferring the copy, but those are difficult, when you have to deal with GLs global state.

ids1024:
> There is always optimizations to be had here regarding deferring the copy, but those are difficult, when you have to deal with GLs global state.

Yeah, that part could be easier with Vulkan, though synchronization is known to be one of the more challenging aspects of Vulkan, so it has it's own difficulties. Probably should be possible to improve this on GL too, but that can be done later.

Rebased and updated for merged smithay changes , so this should be good now.

Issue #732, opened 2024-08-15
touchpad toggle function key not working on oryp4


qrkourier:
This System76 laptop has a function alt key for enable/disable of the touchpad. This is useful to avoid palm checking when an external mouse is used. It's on the same key as F1, and has no apparent effect in Cosmic on PopOS 24.

Please guide me to debug?

Possibly related: https://github.com/pop-os/cosmic-comp/issues/315

leviport:
It's not just oryp4 either. We'll have to fix that for all of our laptops.

qrkourier:
I'll take a quick inventory of the function keys on this oryp4. I didn't find any open issues for sys76 laptops, so I edited the title to be more broad.

1. [ ] touchpad toggle - no apparent effect 
2. [ ] lcd toggle - built-in goes dark on press and returns to normal
3. [x] mute
4. [x] NA
5. [x] vol down
6. [x] vol up
7. [ ] external display - not tested
8. [x] display dim
9. [x] display brighten
10. [x] screenshot
11. [ ] airplane mode - bluetooth and wifi radios are power cycled individually, but not together, and repeated pressing never re-enables both
12. [x] suspend
13. [x] printscrn (launches cosmic screenshot)
14. [ ] insert - not tested

WatchMkr:
Airplane mode appears to be working as expected now.

We need to fix touchpad disable.

leviport:
External display toggle (F7) also doesn't seem to do anything. 

We'll also want to check mic mute, since some of our machines have that one now.

PR #1597, opened 2025-08-20
config: Allow overriding touchpad state

Drakulix:
Touchpad toggle functionality for https://github.com/pop-os/cosmic-comp/issues/732 via cosmic-config.

@ids1024 This needs obviously a new system-action as well, but otherwise is this enough for cosmic-osd?

### DIFF
--- a/cosmic-comp-config/src/input.rs
+++ b/cosmic-comp-config/src/input.rs
@@ -56,6 +56,13 @@ pub enum DeviceState {
     DisabledOnExternalMouse,
 }
 
+#[derive(Debug, Default, Deserialize, Serialize, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum TouchpadOverride {
+    #[default]
+    None,
+    ForceDisable,
+}
+
 impl Default for DeviceState {
     fn default() -> Self {
         Self::Enabled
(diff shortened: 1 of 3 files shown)


ids1024:
On Gnome, the fn key seems to toggle the dconf key `/org/gnome/desktop/peripherals/touchpad/send-events`, which is also toggled by the `Touchpad` toggle at the top of touchpad settings in gnome-control-center (which makes the other controls on that page insensitive).

If we want something similar to Gnome, it would probably make sense just to use `state` in the `input_touchpad` config of cosmic-comp? `cosmic-settings` could also get a toggle for this on the touchpad settings (which we don't currently seem to have).

Drakulix:
> If we want something similar to Gnome, it would probably make sense just to use state in the input_touchpad config of cosmic-comp? 

Then a device-specific config could override it. That is why I went with a separate config entry.

ids1024:
Yeah, I suppose if we ultimately want to expose per-device settings in Cosmic Settings, we'd expect the touchpad toggle key to work even when the touchpad has its own configuration.

ids1024 (approved):
I would have hoped our existing input settings could cover this, but I guess they don't, so this works. Should be easy to create an OSD that monitors for changes to this setting.

Issue #1592, opened 2025-08-18
Overlapping displays after closing and opening the lid w/ external displays connected


jacobgkau:
Version: b28f92a
Hardware: bonw16 (Core Ultra 9 275HX/RTX 5080 hybrid)

Steps to recreate:
1. Plugged in HDMI and USB-C DisplayPort.
2. Closed lid of laptop-- observed no visible change, but assumed the laptop's workspace would be available for the other displays if I'd plugged in an external mouse.
3. Opened lid of laptop.

I initially had two mouse cursors-- one on the internal display and the other on the USB-C display. Attempting to click showed the one on the USB-C display was working to click (and the one on the internal display wasn't):

https://github.com/user-attachments/assets/05492e0e-5a63-4c62-acf8-d16b18f1d230

After unplugging the external displays, the internal display had a hard lock-up initially. Eventually, the screen changed and showed the panel and wallpaper not extending across the entire display anymore:

![Image](https://github.com/user-attachments/assets/335a53b6-8951-4ab0-b767-a1f30b597a0c)

Over 2 minutes later, the system recovered and was working normally with the internal display again.

Logs from this boot: [display-problems.txt](https://github.com/user-attachments/files/21840862/display-problems.txt)

jacobgkau:
On the serw14 (Core Ultra 9 275HX/RTX 5070 hybrid), I just had the internal display visibly overlapping one of the external displays in Settings after re-opening the lid. When I tried to recreate it, the next time I opened the lid, there was a gap between an external display and the internal display in Settings, and my mouse cursor appeared to be stuck on the external display/couldn't traverse the gap back to the internal display until I thought to use the Super + Arrow shortcut to move focus to a window on the internal display. These seem related to this issue.

Drakulix:
So two things stand out to me here:
1. The overlapping displays. This really shouldn't happen since the configurations between lid-open and lid-closed are entirely separate, but I could imagine we can somehow reach this state, after some recovery of the display configuration went wrong. Could you tell me, what the layout of the displays was before closing the lid?
2. Recovery of the display configuration is very likely looking at the dmesg log, which clearly shows that the nvidia-driver panic'd. Which also explains why it took the system some time to reach a working state again (likely some internal timeout). That one I can't fix or debug, given the driver is a black-box and the logs don't indicate that cosmic-comp configured anything wrong, but rather that the driver somehow chocked on standby...

jacobgkau:
[...]

PR #1598, opened 2025-08-20
kms/device: Don't reconfigure existing outputs on `connector_added`

Drakulix:
This should fix https://github.com/pop-os/cosmic-comp/issues/1592 by making sure we don't accidentally re-configure outputs, that already existed, but were disabled previously.

Ping @jacobgkau

### DIFF
--- a/src/backend/kms/device.rs
+++ b/src/backend/kms/device.rs
@@ -762,11 +762,11 @@ impl InnerDevice {
 
             Ok((output, false))
         } else {
-            output
+            let new_config = output
                 .user_data()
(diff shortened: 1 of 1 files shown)


jacobgkau (approved):
Confirmed on serw14 this fixes https://github.com/pop-os/cosmic-comp/issues/1592. The invalid, overlapping configuration from before installing this was still present when first connecting the external display (not sure if that should be checked for in general somewhere?), but after dragging the displays off of each other, I can no longer re-create the overlap by closing and re-opening the lid (tested with a variety of display position configs).

PR #1603, opened 2025-08-22
Install a default `tiling_exceptions_custom` config to `/share`

ids1024:
Fixes "tiling exceptions custom config error" error that cosmic-comp was printing.

Alternately we could modify `cosmic-settings-dameon-config` to ignore the error for this particular setting. But if we generally expect global config files to be present for Cosmic settings, it seems best to just be consistent and add a simple file here.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,7 @@ TARGET_BIN="$(DESTDIR)$(bindir)/$(BINARY)"
 
 KEYBINDINGS_CONF="$(DESTDIR)$(sharedir)/cosmic/com.system76.CosmicSettings.Shortcuts/v1/defaults"
 TILING_EXCEPTIONS_CONF="$(DESTDIR)$(sharedir)/cosmic/com.system76.CosmicSettings.WindowRules/v1/tiling_exception_defaults"
+TILING_EXCEPTIONS_CUST_CONF="$(DESTDIR)$(sharedir)/cosmic/com.system76.CosmicSettings.WindowRules/v1/tiling_exception_custom"
 
 all: extract-vendor
 	cargo build $(ARGS)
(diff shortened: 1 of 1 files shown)


Drakulix:
Hmm, "custom" sounds like user-provided for me, so I don't really see the point of providing an empty file... Maybe we should downgrade that error to a debug-statement.

hepp3n:
> I don't really see the point of providing an empty file

Hey, if I can add something to it :)

From my perspective as an user. I was needed to edit this file. I did not know the name of it so I had to copy file from /usr/share (the original one). If it's helps, the empty file would be good idea at least I think like that :D

ids1024:
Yeah, we could ignore that error, or have it as a debug print, while still having an error if `tiling_exceptions_defaults` is missing.

> From my perspective as an user. I was needed to edit this file. I did not know the name of it so I had to copy file from /usr/share (the original one). If it's helps, the empty file would be good idea at least I think like that :D

Yeah, this could make things clearer if someone wants to manually edit the configuration. This shows what the file should be named, and then `tiling_exceptions_defaults` shows what format its contents would have.

But I don't have any strong opinion on this.

Drakulix:
> From my perspective as an user. I was needed to edit this file. I did not know the name of it so I had to copy file from /usr/share (the original one). If it's helps, the empty file would be good idea at least I think like that :D

Ideally you don't have to do that though, as the tiling applet should provide a window chooser to add new exceptions. Sure it is nice, if there is a file you can edit manually, but it doesn't even have the same format as `tiling_exceptions_defaults`, so in the end you need to take a look at the source code anyway to figure that out.

(https://github.com/pop-os/cosmic-settings-daemon/blob/master/config/src/window_rules/mod.rs#L41-L44)

Or you read about it in some issue, which likely points you to the right file anyway. Which brings me to my second point, we want to have and should have documentation on the config files at some point.

So with a GUI-way to add exceptions *and* some docs, the empty file would again make no sense really.

> Yeah, we could ignore that error, or have it as a debug print, while still having an error if tiling_exceptions_defaults is missing.

Yeah, we definitely want an error if defaults is missing though. If a distribution chooses to ship an empty array, that is fine, but I want the error to be a hint, that the file should be there for packaging.

hepp3n:
Wow did know about it: (https://github.com/pop-os/cosmic-settings-daemon/blob/master/config/src/window_rules/mod.rs#L41-L44)

Yeah I can agree with you @Drakulix  :) 

Keep going what you are all guys doing because it's an incredible work :)

ids1024:
We could also document this by adding a default file with a commented example. But there are perhaps better ways to document configuration (when we get around to that), and in any case it's reasonable to just not error here: https://github.com/pop-os/cosmic-settings-daemon/pull/100.

PR #1612, opened 2025-08-26
refactor(config): fix no features enabled

wash2:


wash2:
> Out of curiosity, where was this a problem?

it became a problem in the cosmic-settings update

PR #1620, opened 2025-09-03
cosmic-comp-config: Fix building without `output` feature

ids1024:
`EdidProduct` is needed in `workspace`.

Only `load_outputs` really needs to be gated by the `output` feature. Just the type definitions don't require any extra dependencies.

### DIFF
--- a/cosmic-comp-config/src/lib.rs
+++ b/cosmic-comp-config/src/lib.rs
@@ -7,7 +7,6 @@ use std::collections::HashMap;
 use crate::input::TouchpadOverride;
 
 pub mod input;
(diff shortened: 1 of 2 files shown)


wash2:
oh i think https://github.com/pop-os/cosmic-comp/pull/1612 did a similar fix but wasn't merged. i think either works though.

Drakulix:
I don't mind merging either, I just forgot to do it after approving. Feel free to select one.

ids1024:
Closing, since https://github.com/pop-os/cosmic-comp/pull/1612 was merged.

PR #1624, opened 2025-09-05
Software rendering support

Drakulix:
Fixes https://github.com/pop-os/cosmic-comp/issues/144

Turns out the stupidest solution is the fastest...

The biggest changes are in smithay to fix llvmpipe (https://github.com/Smithay/smithay/pull/1811), but a few additional commits here clean up support for display link and other software devices. In general we don't want to expose hardware acceleration related features:
- Never consider a software based gpu a primary node (we have fallbacks for when no primary node is set)
- Don't advertise dmabuf-support for software gpus
- Don't advertise their outputs for leasing
- Don't test dmabuf imports on them
- Work around cursor bugs on evdi specifcially

I have another commit that uses what mutter describes as "zero-copy" (https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/810) or how anvil supports display link, but...
- Some render gpus don't support that anyway (e.g. nvidia can't render into a linear buffer), so we need the fallback path.
- Some gpu-less display drivers also don't support it.
- It isn't truly zero-copy, as of course the driver will still read from the dmabuf, which causes it to be migrated into memory, which ruins performance on all of my test machines. So I am getting similar latency and better performance across the board by just letting our existing multi-gpu infrastructure download and copy the new contents with damage.
- On Meteor/Arrow Lake we get interesting artifacts with this approach, just like when using screencopy.

I also spent a lot of time trying to get the PixmanRenderer and a second backend working instead of relying on llvmpipe. Except for the annoyance of implementing our shaders in software, it turns out that large effort did not only lead no or only marginal benefits, but instead didn't work at all on evdi.

The driver doesn't allow us to map dmabufs created from it's GbmBuffers or DumbBuffers, which we need to render with both the `PixmanRenderer` and more annoyingly `DrmCompositor`. Looking at the source code of the module, mapping the DumbBuffer directly with the corresponding ioctls would probably work, but that would need more architectural changes in smithay, at which point I just gave up.

So in the end we just use llvmpipe and mostly treat the drm device like any other.

### DIFF
--- a/src/backend/kms/surface/mod.rs
+++ b/src/backend/kms/surface/mod.rs
@@ -87,7 +87,7 @@ use smithay::{
         shm::{shm_format_to_fourcc, with_buffer_contents},
     },
 };
-use tracing::{error, trace, warn};
+use tracing::{error, info, trace, warn};
 
 use std::{
     borrow::{Borrow, BorrowMut},
@@ -98,6 +98,7 @@ use std::{
         mpsc::{Receiver, SyncSender},
         Arc, RwLock,
     },
+    thread::JoinHandle,
     time::Duration,
 };
 
@@ -124,6 +125,7 @@ pub struct Surface {
     loop_handle: LoopHandle<'static, State>,
     thread_command: Sender<ThreadCommand>,
     thread_token: RegistrationToken,
+    thread: Option<JoinHandle<()>>,
 
     dpms: bool,
 }
@@ -213,9 +215,11 @@ pub enum ThreadCommand {
         node: DrmNode,
         gbm: GbmAllocator<DrmDeviceFd>,
         egl: EGLContext, // TODO: Option for software rendering
+        sync: SyncSender<()>,
     },
     NodeRemoved {
         node: DrmNode,
+        sync: SyncSender<()>,
     },
     UpdateMirroring(Option<Output>),
     UpdateScreenFilter(ScreenFilter),
@@ -262,7 +266,7 @@ impl Surface {
         let active_clone = active.clone();
         let output_clone = output.clone();
 
-        std::thread::Builder::new()
+        let thread = std::thread::Builder::new()
             .name(format!("surface-{}", output.name()))
             .spawn(move || {
                 if let Err(err) = surface_thread(
@@ -350,6 +354,7 @@ impl Surface {
             loop_handle: evlh.clone(),
             thread_command: tx,
             thread_token,
+            thread: Some(thread),
             dpms: true,
         })
     }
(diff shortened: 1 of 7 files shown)


src/backend/kms/surface/mod.rs
```diff
@@ -699,24 +717,19 @@ impl SurfaceThreadState {
...
-        */
 
         Ok(())
     }
 
     fn node_removed(&mut self, node: DrmNode) {
         self.api.as_mut().remove_node(&node);
-        //self.software_api.as_mut().remove_node(node);
+        // force enumeration
+        let _ = self.api.single_renderer(&self.target_node);
```
ids1024: That's ugly, but I guess it works...

If `needs_enumeration` is true, calling `api.devices()` should also force an `enumerate` call. Which I guess would be mildly better here than using `single_renderer`.

src/backend/kms/surface/mod.rs
```diff
@@ -364,17 +369,24 @@ impl Surface {
...
-            .thread_command
-            .send(ThreadCommand::NodeAdded { node, gbm, egl });
+        let (tx, rx) = std::sync::mpsc::sync_channel(1);
+        let _ = self.thread_command.send(ThreadCommand::NodeAdded {
+            node,
+            gbm,
+            egl,
+            sync: tx,
+        });
+        let _ = rx.recv();
```
ids1024, line 379: I guess blocking here isn't really an issue since it should be uncommon to remove a device node (though not that uncommon with DisplayLink) and shouldn't block for very long. But it should probably at least have a comment explaining why we're waiting here?

ids1024 (commented):
Other than those minor details, this looks good. Though I haven't tested on my DisplayLink device yet. This is simpler than I expected it to be to get EVDI working.

Proper pixman renderering support would be nice, but is definitely going to complicate things. LLVMPipe should be good enough in most cases (after all, I think that's the only software renderer Gnome Shell has. The rendering requirements were controversial when Gnome 3.0 released in 2011). Maybe more noticeable on systems with fairly slow CPUs, as well as a lack of a suitable graphics driver. But that isn't a huge priority.

src/backend/kms/surface/mod.rs
Drakulix: Yes it is, because we otherwise can't unwrap the `DeviceFd` `Arc` later when we want to call `session.close_fd`. And we need to do that to be able to open the device later again with `session.open_fd`. Supposedly the service manager will figure out, that the file descriptor is gone at some point as well, but I had races and thus denied open-calls otherwise.

Drakulix:
> Other than those minor details, this looks good.

Fixed.

PR #1633, opened 2025-09-10
config: Make read_outputs failable

Drakulix:
Based on https://github.com/pop-os/cosmic-comp/pull/1632 as some scenarios here aren't really testable with the deadlock.

Previously we ignored when we had no output configuration **and** failed to apply the automatically created config (which is just every output enabled with it's preferred mode).

This leads to two problems:
- If this happens on startup, we end up with no outputs being added to the shell and we quit.
- If this happens later, we might end up in an inconsistent state, where the shell thinks we have an output, when it didn't light up, for similar reasons.

Thus `read_outputs` is failable and handling that very much depends on the where is was called from, because `read_outputs` doesn't know what configuration was active before.

Thus make it failable and provide useful mitigations everywhere
possible:
- Try to enable just one output in case we fail on start-up.
- Don't enable any additional outputs, when we fail on hot-plug.
- Log the error like previously in any other case (and come up with more
  mitigations, once we understand these cases better).
  
This is another output-configuration change close to the beta release, which might sound scary, but this addresses three real issues with the serw14 and it's 300Hz mode, which is currently the preferred mode. The reason being that 300hz takes so much bandwidth, that no additional output can be enabled.

Scenarios previously broken and fixed now:
- If you have another output plugged in on startup in a configuration the system hasn't seen before, we end up looping in cosmic-session, due to cosmic-comp quitting as it isn't able to initialize all outputs at once. With this fix it falls back to enabling only the internal display.
- If you plug in the output later, while being on 300hz, it would still show up as enabled under cosmic-settings and the shell state. With this fix the additional monitor initializes as disabled and can be enabled once the mode has been lowered.
- Convoluted setup, but if you use an output on the HDMI port (which is connected to the nvidia GPU this doesn't influence bandwidth), close the lid, then attach a DisplayPort output and then re-open the lid, you get a black internal screen with cosmic-settings showing the internal one as enabled similarly to the previous scenario. With this fix it shows correctly as disabled. You can't enable it without disabling the DisplayPort output first, because cosmic-settings won't give you a refresh rate selection on a disabled output. But that is purely an UI issue, as it works correctly with cosmic-randr.

### DIFF
--- a/src/backend/kms/device.rs
+++ b/src/backend/kms/device.rs
@@ -171,9 +171,14 @@ pub fn init_egl(gbm: &GbmDevice<DrmDeviceFd>) -> Result<EGLInternals> {
 }
 
 impl State {
-    pub fn device_added(&mut self, dev: dev_t, path: &Path, dh: &DisplayHandle) -> Result<()> {
+    pub fn device_added(
+        &mut self,
+        dev: dev_t,
+        path: &Path,
+        dh: &DisplayHandle,
+    ) -> Result<Vec<Output>> {
         if !self.backend.kms().session.is_active() {
-            return Ok(());
+            return Ok(Vec::new());
         }
 
         if let Some(allowlist) = dev_list_var("COSMIC_DRM_ALLOW_DEVICES") {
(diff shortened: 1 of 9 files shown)


ids1024 (approved):
The change to `read_outputs` seems straightforward, while the different cases here where we try to enable an output to get a working configuration all look correct and reasonable enough.

There may be more edge cases we want to handle, but hopefully this should be an improvement.

PR #1651, opened 2025-09-17
Handle damage in image-copy-capture protocol correctly

ids1024:
Based on https://github.com/pop-os/cosmic-comp/pull/1638.

It seems `age_for_buffer` shouldn't really be necessary; the client should accumulate damage, and send it with `buffer_damage`. Though trying to change just that revealed that `additional_damage` wasn't being handled properly; it was placed incorrectly in the z-order of the render elements, so it didn't work properly where it was occluded by render elements with opaque regions.

https://github.com/pop-os/cosmic-workspaces-epoch/pull/217 updates cosmic-workspaces to accumulate damage, and send it in `damage_buffer`. `xdg-desktop-portal-cosmic` already does.

This appears to work, but may be worth testing a bit more.

I guess we also don't have anything to delay toplevel and workspace capture of a new frame until there is new damage? We probably want that, though hopefully if we track damage properly it won't add too much overhead to capture extra frames that are identical.

### DIFF
--- a/src/backend/kms/surface/mod.rs
+++ b/src/backend/kms/surface/mod.rs
@@ -1345,13 +1345,6 @@ impl SurfaceThreadState {
                                 (&session, frame, res),
                                 now.into(),
                             ) {
-                                session
-                                    .user_data()
-                                    .get::<SessionData>()
-                                    .unwrap()
-                                    .lock()
-                                    .unwrap()
-                                    .reset();
                                 tracing::warn!(?err, "Failed to screencopy");
                             }
                         }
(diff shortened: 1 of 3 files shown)


Drakulix:
> I guess we also don't have anything to delay toplevel and workspace capture of a new frame until there is new damage? We probably want that, though hopefully if we track damage properly it won't add too much overhead to capture extra frames that are identical.

No we don't, but I guess it would be trivial to lock that to commits of any toplevel's surface?

ids1024:
Merged as part of https://github.com/pop-os/cosmic-comp/pull/2058.

PR #1652, opened 2025-09-18
floating: Keep elements positioned relatively on recalculate

Drakulix:
Keeps windows positioned relatively to their size and location on scale and transform changes of the same output.

This also will keep cosmic-initial-setup centered, when adjusting scale.


jacobgkau (approved):
This gets cosmic-initial-setup to stay centered while changing the "Interface size" and "Additional scale options" settings. It also works for additional windows.

I tried to break this by having a window hanging off near the edge of the screen before changing the scale, but the window pops back to the center of the screen instead of going completely out-of-view, so that edge case is fine.

I'm seeing a visual issue with the dock that I'll open an issue for in https://github.com/pop-os/cosmic-panel, but it was already happening when adjusting scale before this cosmic-comp change, so it's not a regression.

Issue #487, opened 2024-05-17
Fullscreen window letterboxing


git-f0x:
When a window is full-screened, using the keyboard shortcut for maximize creates grey bars. The placement of the bars is correlated to the placement of non-autohiding panels. This also happens when a maximized window is full-screened.
Opened here since it looks like some interaction between the panels and cosmic-comp, but feel free to move this to cosmic-panel if it's more relevant.

git-f0x:
Might be related: similar behavior is also present when making a video fullscreen in maximized browsers (when browsers aren't maximized, videos full-screen properly). The letterboxing is present and it makes it hard to click the exit fullscreen button for the video.

git-f0x:
This might need to be an **alpha 1** issue @WatchMkr, since it interferes with a common use case of using a maximized browser and full-screening e.g. a YouTube video (since it also makes the video settings hard to interact with if there's a dock on the bottom edge).

louie-github:
This issue still appears when full-screening videos from a window that was opened in maximized mode from the beginning. If you restore then maximize the window again, the letterboxes disappear.

louie-github:
This issue also occurs when you show an OSD, for example, volume adjustment, when in full-screen playing videos.

ellieplayswow:
this is still happening, and easily replicable:
- pop 24.04, alpha 5, nvidia GPU (unsure if gpu matters)
- open firefox, navigate to a youtube video
- fullscreen it
- press the super key

![Image](https://github.com/user-attachments/assets/28ec1164-a3f4-4e2f-b5c7-82173cbcd1f6)

OTOH, if you use the firefox picture in picture first, this behaviour doesn't occur

marcerhans:
I have also encountered this issue in alpha 6. Steps to reproduce:
- Start firefox and navigate to youtube.
- Fullscreen firefox (hotkey: Super+M)
- Start a video and make that fullscreen (hotkey: f).
- Adjust volume via laptop function keys (or similar).

Then grey bar on bottom shows up until i exit fullscreen.

Pandapip1:
Performing a screen capture is also sufficient to reproduce.

git-f0x:
Fixed by #1481.

peterkrull:
I am not sure this is fully fixed. Or at least there is another related issue.

https://github.com/user-attachments/assets/03910166-78de-42ad-a8a6-e84b88f0112d

[...]

PR #1658, opened 2025-09-20
[fix] fullsreen: remove grey bars, reposition popups

peterkrull:
Attempt to address #487

Scaling during animation is calculated as a factor between geometries, rather than geometry/bbox. Also scaling is conditional on an animation being in progress. Also popups were seemingly not handled in Shell::unconstrain_popup for fullscreen parent surfaces.


Drakulix (approved):
Looks fine to me. Thanks!

leviport (approved):
I tried a bunch of fullscreening, both with the F key in a youtube video, as well as F11 in Firefox, and I didn't see the gray bar bug once. Before this patch, I could recreate it both ways pretty quickly.

PR #1665, opened 2025-09-24
feat: corner-radius protocol support

wash2:
This seems to work well with the client implementation that I've been working on. It is missing support for per-corner radius values in the shader, and doesn't force a redraw of the focus element yet though.

### DIFF
--- a/src/shell/grabs/moving.rs
+++ b/src/shell/grabs/moving.rs
@@ -14,11 +14,19 @@ use crate::{
         CosmicMapped, CosmicSurface, Direction, ManagedLayer,
     },
     utils::prelude::*,
-    wayland::protocols::toplevel_info::{toplevel_enter_output, toplevel_enter_workspace},
+    wayland::protocols::{
+        corner_radius::CornerRadiusData,
+        toplevel_info::{toplevel_enter_output, toplevel_enter_workspace},
+    },
+};
+use smithay::{
+    reexports::wayland_server::{Resource, Weak},
+    wayland::{compositor::with_states, seat::WaylandFocus},
 };
 
 use calloop::LoopHandle;
 use cosmic::theme::CosmicTheme;
+use cosmic_protocols::corner_radius::v1::server::cosmic_corner_radius_toplevel_v1::CosmicCornerRadiusToplevelV1;
 use smithay::{
     backend::{
         input::ButtonState,
@@ -109,6 +117,29 @@ impl MoveGrabState {
             - scaling_offset;
 
         let active_window_hint = crate::theme::active_window_hint(theme);
+        let radius = self
+            .window()
+            .wl_surface()
+            .and_then(|surface| {
+                with_states(&surface, |s| {
+                    let d = s
+                        .data_map
+                        .get::<Mutex<Weak<CosmicCornerRadiusToplevelV1>>>()?;
+                    let guard = d.lock().unwrap();
+
+                    let weak_data = guard.upgrade().ok()?;
+
+                    let corners = weak_data.data::<CornerRadiusData>()?;
+
+                    let guard = corners.lock().unwrap();
+
+                    // TODO support multiple radius values
+                    Some(guard.top_left)
+                })
+            })
+            .unwrap_or(self.indicator_thickness);
+
+        tracing::error!("{radius}");
         let focus_element = if self.indicator_thickness > 0 {
             Some(
                 CosmicMappedRenderElement::from(IndicatorShader::focus_element(
@@ -125,6 +156,7 @@ impl MoveGrabState {
                     )
                     .as_local(),
                     self.indicator_thickness,
+                    radius,
                     alpha,
                     [
                         active_window_hint.red,
--- a/src/shell/layout/tiling/mod.rs
+++ b/src/shell/layout/tiling/mod.rs
@@ -29,6 +29,7 @@ use crate::{
     wayland::{
         handlers::xdg_shell::popup::get_popup_toplevel,
         protocols::{
+            corner_radius::CornerRadiusData,
             toplevel_info::{
                 toplevel_enter_output, toplevel_enter_workspace, toplevel_leave_output,
                 toplevel_leave_workspace,
@@ -37,7 +38,9 @@ use crate::{
         },
     },
 };
+use smithay::{reexports::wayland_server::Resource, wayland::compositor::with_states};
 
+use cosmic_protocols::corner_radius::v1::server::cosmic_corner_radius_toplevel_v1::CosmicCornerRadiusToplevelV1;
 use cosmic_settings_config::shortcuts::action::{FocusDirection, ResizeDirection};
 use id_tree::{InsertBehavior, MoveBehavior, Node, NodeId, NodeIdError, RemoveBehavior, Tree};
 use keyframe::{
@@ -59,13 +62,13 @@ use smithay::{
     desktop::{layer_map_for_output, space::SpaceElement, PopupKind, WindowSurfaceType},
     input::Seat,
     output::Output,
-    reexports::wayland_server::Client,
+    reexports::wayland_server::{Client, Weak as WsWeak},
     utils::{IsAlive, Logical, Physical, Point, Rectangle, Scale, Size},
     wayland::{compositor::add_blocker, seat::WaylandFocus},
 };
 use std::{
     collections::{HashMap, VecDeque},
-    sync::{Arc, Weak},
+    sync::{Arc, Mutex, Weak},
     time::{Duration, Instant},
 };
 use tracing::trace;
@@ -5015,6 +5018,27 @@ where
                     x => Some(x),
                 }
             }));
+            let radius = mapped
+                .active_window()
+                .wl_surface()
+                .and_then(|surface| {
+                    with_states(&surface, |s| {
+                        let d = s
+                            .data_map
+                            .get::<Mutex<WsWeak<CosmicCornerRadiusToplevelV1>>>()?;
+                        let guard = d.lock().unwrap();
+
+                        let weak_data = guard.upgrade().ok()?;
+
+                        let corners = weak_data.data::<CornerRadiusData>()?;
+
+                        let guard = corners.lock().unwrap();
+
+                        // TODO support multiple radius values
+                        Some(guard.top_left)
+                    })
+                })
+                .unwrap_or(indicator_thickness);
             if is_minimizing && indicator_thickness > 0 {
                 elements.push(CosmicMappedRenderElement::FocusIndicator(
                     IndicatorShader::focus_element(
@@ -5022,6 +5046,7 @@ where
                         Key::Window(Usage::FocusIndicator, mapped.clone().key()),
                         geo,
                         indicator_thickness,
+                        radius,
                         alpha,
                         [window_hint.red, window_hint.green, window_hint.blue],
                     ),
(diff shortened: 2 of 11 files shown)


src/shell/layout/tiling/mod.rs
```diff
@@ -5035,10 +5035,15 @@ where
                         let guard = corners.lock().unwrap();
 
                         // TODO support multiple radius values
```
Drakulix: I guess that todo can be dropped as well. :tada:

src/backend/render/shaders/rounded_outline.frag
```diff
@@ -8,10 +8,14 @@ varying vec2 v_coords;
...
 
-float rounded_box(vec2 center, vec2 size, float radius) {
-    return length(max(abs(center) - size + radius, 0.0)) - radius;
+float rounded_box(in vec2 p, in vec2 b, in vec4 r)
+{
+    r.xy = (p.x > 0.0) ? r.xy : r.zw;
+    r.x = (p.y > 0.0) ? r.x : r.y;
+    vec2 q = abs(p) - b + r.x;
+    return min(max(q.x, q.y), 0.0) + length(max(q, 0.0)) - r.x;
 }
```
Drakulix, line 19: Thanks for cleaning this up!

src/shell/layout/tiling/mod.rs
```diff
@@ -5423,11 +5432,16 @@ where
                                     let guard = corners.lock().unwrap();
 
                                     // TODO support multiple radius values
```
Drakulix: see above

src/shell/grabs/moving.rs
```diff
@@ -109,6 +117,29 @@ impl MoveGrabState {
...
+
+                    let corners = weak_data.data::<CornerRadiusData>()?;
+
+                    let guard = corners.lock().unwrap();
+
+                    // TODO support multiple radius values
+                    Some(guard.top_left)
+                })
+            })
+            .unwrap_or(self.indicator_thickness);
```
Drakulix: It would be nice, if we could have a method on `CosmicSurface`, which does this (and then possibly proxies on `CosmicWindow` and `CosmicMapped`. (Not sure what to do with stacks here, stacked clients shouldn't have rounded corners anyway, so probably just return that.)

src/shell/grabs/moving.rs
```diff
@@ -109,6 +117,29 @@ impl MoveGrabState {
...
+
+        tracing::error!("{radius}");
```
Drakulix: Just noting here before we forget, that we need to drop this.

src/wayland/handlers/corner_radius.rs
```diff
@@ -0,0 +1,44 @@
...
+use crate::wayland::protocols::corner_radius::{
+    delegate_corner_radius, CornerRadiusData, CornerRadiusHandler,
+};
+
+use crate::state::State;
+
+impl CornerRadiusHandler for State {
+    fn corner_radius_state(
+        &mut self,
+    ) -> &mut crate::wayland::protocols::corner_radius::CornerRadiusState {
```
Drakulix: please add a `use` for these longer types.

src/wayland/handlers/corner_radius.rs
```diff
@@ -0,0 +1,44 @@
...
+    fn set_corner_radius(
+        &mut self,
+        _toplevel: &cosmic_protocols::corner_radius::v1::server::cosmic_corner_radius_toplevel_v1::CosmicCornerRadiusToplevelV1,
+        _data: &CornerRadiusData,
+        _top_left: u32,
+        _top_right: u32,
+        _bottom_right: u32,
+        _bottom_left: u32,
+    ) {
+        // TODO force redraw? of focus element?
```
Drakulix: Yeah, we probably want to get the surface from the toplevel and then get it's output from the shell and do `backend.schedule_render(output)`.

src/wayland/handlers/corner_radius.rs
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    fn unset_corner_radius(
+        &mut self,
+        _toplevel: &cosmic_protocols::corner_radius::v1::server::cosmic_corner_radius_toplevel_v1::CosmicCornerRadiusToplevelV1,
+        _data: &CornerRadiusData,
+    ) {
+        // TODO force redraw?
```
Drakulix: Same force redraw logic.

src/backend/render/shaders/rounded_outline.frag
wash2: Ya, this is copied from https://iquilezles.org/articles/distfunctions2d/  and it works pretty well. If any corner radius is set to be larger than half the smallest dimension of the window though, it can end up with a disconnected outline if the radii do not match. If they do match, there will be sharp points on the edges, which I guess is ok, but I'm not really sure if it is something people would really miss. I left it out but I think maybe the radius (pre adjustment by half the thickness) should be set to the min of the radius and half the smallest dimenstion to guard against these weird cases.

Drakulix: Yes, we should sanitize these values.

I think part of that is also covered by protocol errors, which this implementation currently doesn't enforce.

wash2:
> This looks really good already, do you have an easy test client somewhere to verify? Preferably one where you can tweak each corner separately?

Ya, I've been testing with the multi-window example of https://github.com/pop-os/libcosmic/pull/980 but any example enabling the wayland feature should work by default. I think adjusting the theme radius_s would allow us to test each corner, but I didn't try that yet for some reason.

src/backend/render/shaders/rounded_outline.frag
wash2: Yeah, though a valid corner radius for a large window could suddenly become too large when the window is shrunk, so I'm a bit hesitant to emit errors for large values. If there is a good way to handle that in the client, I have no problem with it though. I've added a fix for it in the most recent commit.

src/shell/grabs/moving.rs
wash2: I added a method to `CosmicSurface` that reduces duplication a ton :)

src/backend/render/shaders/rounded_outline.frag
Drakulix: Well the corner radius is double buffered and part of the commit state (another thing which this PR needs to do before merging, this needs to be part of the [cached_state](https://smithay.github.io/smithay/smithay/wayland/compositor/struct.SurfaceData.html#structfield.cached_state)).

So any actual geometry changes can and have to be acked by the client, meaning accompanying radius changes can easily be submitted with the corresponding commit to prevent any races.

src/shell/element/surface.rs
```diff
@@ -124,6 +127,34 @@ impl CosmicSurface {
...
+                    .get::<Mutex<Weak<CosmicCornerRadiusToplevelV1>>>()?;
+                let guard = d.lock().unwrap();
+
+                let weak_data = guard.upgrade().ok()?;
+
+                let corners = weak_data.data::<CornerRadiusData>()?;
+
+                let guard = corners.lock().unwrap();
+
+                let size = self.0.bbox().size;
```
Drakulix: This is not the correct size, the protocol mentions the xdg-toplevel geometry. So this should be `<CosmicSurface as SpaceElement>::geometry()` instead, I think.

src/wayland/protocols/corner_radius.rs
```diff
@@ -0,0 +1,232 @@
...
+                let corner_radius_state = state.corner_radius_state();
+                corner_radius_state.instances.retain(|i| i != _resource);
+            }
+            cosmic_corner_radius_manager_v1::Request::GetCornerRadius { id, toplevel } => {
+                let data = Mutex::new(CornerRadiusInternal {
+                    toplevel: toplevel.downgrade(),
+                    top_left: 0,
+                    top_right: 0,
+                    bottom_right: 0,
+                    bottom_left: 0,
```
Drakulix: So this means any toplevel, that has a corner-radius object gets no rounded outline, if the client never sets radius values?

That would be incorrect, we want to apply the same behaviour here, as if the client didn't use the protocol. Same after `unset_radius` btw until it sets a new one.

Which means this needs to be an `Option` probably.

src/backend/render/shaders/rounded_outline.frag
Drakulix: Cached state usage looks good now! I think the only thing the protocol needs now is a pre-commit hook that checks for the dimensions protocol error.

src/wayland/handlers/corner_radius.rs
```diff
@@ -39,6 +42,45 @@ impl CornerRadiusHandler for State {
             tracing::warn!("Failed to force redraw for corner radius reset.");
         }
     }
+
+    fn add_corners(
+        &mut self,
+        toplevel: &xdg_toplevel::XdgToplevel,
+        toplevel_obj: cosmic_corner_radius_toplevel_v1::CosmicCornerRadiusToplevelV1,
+    ) {
```
Drakulix: This shouldn't be part of the handler, but the protocol impl and be done transparently to the rest of the code.

(The protocol-part would be what could be moved into smithay once we have an upstream protocol, while the handler-part is compositor-specific. Errors aren't specific.)

Note that we need to make sure to only add the hook once per surface. Even if a surface creates the RadiusToplevel a second time after destroying the first. So we might need to store the returned hookid inside the datamap.

src/wayland/protocols/corner_radius.rs
```diff
@@ -45,6 +45,11 @@ impl CornerRadiusState {
 }
 
 pub trait CornerRadiusHandler {
+    fn add_corners(
+        &mut self,
+        toplevel: &XdgToplevel,
+        toplevel_obj: cosmic_corner_radius_toplevel_v1::CosmicCornerRadiusToplevelV1,
+    );
     fn corner_radius_state(&mut self) -> &mut CornerRadiusState;
     fn toplevel_from_resource(&mut self, toplevel: &XdgToplevel) -> Option<ToplevelSurface>;
```
Drakulix: I think we can also get rid of this function, if we just make the trait depend on the `XdgShellHandler`-trait from smithay?

src/wayland/protocols/corner_radius.rs
```diff
@@ -111,7 +116,8 @@ where
                     toplevel: toplevel.downgrade(),
                     corners: None,
                 });
```
Drakulix: We also need to check here for the `corner_radius_exists` protocol error.

src/wayland/handlers/corner_radius.rs
wash2: Ok, thanks, sorry if this would've just been easier for you to implement :sweat_smile:

Drakulix: no worries, I think it is good if more people have some understanding how cosmic-comp is structured and what the design decisions are, I certainly haven't done a good job documenting them.

and no, it is a lot of boilerplate and moving stuff around is always easier than writing it in the first place. thanks for stepping up to it so quickly!

src/wayland/protocols/corner_radius.rs
```diff
@@ -0,0 +1,378 @@
...
+
+                        with_states(surface.wl_surface(), |surface_data| {
+                            let hook_ids = surface_data.data_map.get_or_insert_threadsafe(|| {
+                                Mutex::new(HashMap::<
+                                    WlSurface,
+                                    (HookId, Weak<CosmicCornerRadiusToplevelV1>),
+                                >::new())
+                            });
+                            let mut guard = hook_ids.lock().unwrap();
+                            guard.insert(surface.wl_surface().clone(), (hook_id, obj_downgrade));
```
Drakulix: This unfortunately creates a circular reference to the surface, given we store a strong-reference to the `WlSurface` inside the user-data of the `WlSurface`. This will cause some cleanup to not be handled correctly.

I am not sure, why we store a `HashMap` in the first place, it looks like we only store a single surface inside it or am I mistaken?

If we need the `HashMap`, we need to use a weak-reference for the key as well.

I would also prefer any `data_map` types to have a type-alias, so it is less likely to accidentally get the type wrong and make it easier to update the type in the future, if necessary.

src/wayland/protocols/corner_radius.rs
```diff
@@ -0,0 +1,378 @@
...
+        match request {
+            cosmic_corner_radius_toplevel_v1::Request::Destroy => {
+                let mut guard = data.lock().unwrap();
+                guard.corners = None;
+
+                let Ok(toplevel) = guard.toplevel.upgrade() else {
+                    resource.post_error(
+                        cosmic_corner_radius_toplevel_v1::Error::ToplevelDestroyed as u32,
+                        format!("{:?} No toplevel found", resource),
+                    );
```
Drakulix: The protocol states:
>  If the xdg_toplevel associated with the cosmic_corner_radius_toplevel_v1 object has been destroyed, this object becomes inert. Any further requests other than destroy will raise the toplevel_destroyed protocol error.

So this specific case shouldn't error, but simply return.

Drakulix (approved):
🎉

Issue #1625, opened 2025-09-06
[REQ]: Add alt+f4 as a default keybindings


Quackdoc:
It's fairly ubiquitous and it feels odd to not have support for it

PR #1672, opened 2025-09-26
chore: add Alt+F4 shortcut to close windows

therealmate:
Fixes https://github.com/pop-os/cosmic-comp/issues/1625


maria-komarova (approved):
Good on UX side

jacobgkau (approved):
Successfully adds the default Alt+F4 binding to the "Close window" action. Super+Q still works. Alt+F4 can be removed manually in Settings.

Issue #453, opened 2024-04-30
Previously focused app takes focus while the currently focused app is grabbed


git-f0x:
The active hint doesn't move, but the previous app becomes visibly focused when the current app is grabbed.
When e.g. the search bar in Firefox is selected, and then switching to another app and grabbing it, typing while the window is grabbed will input text to the Firefox search bar.

Drakulix:
fixed by https://github.com/pop-os/cosmic-comp/commit/d4926c3a9e66a2b699fbe626b2609ed5d373cce4

PoisonFrog:
I'm still having this issue on master.

Drakulix:
> I'm still having this issue on master.

Which distro? Which commit exactly?

PoisonFrog:
> > I'm still having this issue on master.
> 
> Which distro? Which commit exactly?

pop_OS!
I added popdev:master to the sources and the system is fully up to date

PoisonFrog:
I just realized that it doesn't happen when you drag the title bar, but it still happens when you drag the window while holding the Super key.

git-f0x:
I can kind of reproduce it if the previously focused app is a GTK app (that app visually changes when another is dragged), but doesn't have any usability issues (e.g. text input). It's just a visual issue.

PoisonFrog:
> I can kind of reproduce it if the previously focused app is a GTK app (that app visually changes when another is dragged), but doesn't have any usability issues (e.g. text input). It's just a visual issue.

I still have this issue with all kinds of apps, including libcosmic apps and I can still input to the background window if holding the Super key to drag. If dragging the title bar I only have this visual issue with GTK apps.

git-f0x:
I can reproduce the visual changes in most apps with Super dragging, but text input doesn't go through.

Drakulix:
> I just realized that it doesn't happen when you drag the title bar, but it still happens when you drag the window while holding the Super key.

Oh thanks, that is what I was missing here. Reopening.

Drakulix:
Fixed by https://github.com/pop-os/cosmic-comp/commit/6c517bd8de9e4de74a24866c274b2e1c8ac6d7e1

PoisonFrog:
It's mostly fixed now, but there are still a few quirks. For example, if you drag a window (using either the title bar or the Super key) and press Super + Q, it closes the window in the background instead of the one you're dragging. This also happens with other window shortcuts like maximize.

[...]

PR #1687, opened 2025-09-30
WIP Don't add focus indicator to another floating window when a floating window is dragged

ids1024:
This has been annoying me when using floating windows. Colored indicators appearing where they shouldn't (and don't with tiling drags) is a bit distracting.

This fix seems to work on a single output, but with multiple outputs messes up the focus indicators on the other output. Since the floating layout (but not the tiling layout) can show focus indicators on multiple outputs, and we only want to omit it on the output we drag a window from.

I thought modifying `refresh_focus_stack` not to remove a window from the stack if it in the `SeatMoveGrabState` might work, but it seems `append_focus_stack` is being called when the move grab starts? (In `set_focus`, called by `process_input_event`).

### DIFF
--- a/src/shell/workspace.rs
+++ b/src/shell/workspace.rs
@@ -8,7 +8,7 @@ use crate::{
     },
     shell::{
         layout::{floating::FloatingLayout, tiling::TilingLayout},
-        OverviewMode, ANIMATION_DURATION,
+        OverviewMode, SeatMoveGrabState, ANIMATION_DURATION,
     },
     state::State,
     utils::{prelude::*, tween::EaseRectangle},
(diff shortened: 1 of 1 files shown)


git-f0x:
Could this potentially fix #453?

Drakulix:
Yeah I am not sure, if simply not drawing the focus indicator is the right approach here? I definitely would like to make sure, we actually keep the focus on the dragged window and don't send any unnecessary enter/leave events (especially given Xwayland misbehaving if we do).

(But maybe this is already the case the issue is just, that we use the focus stack for drawing and updating the Activated state? I haven't looked into this for quite some time.)

git-f0x:
This PR did sound like just a visual fix, but wasn't sure. The dragged window definitely still stops being treated as focused (or even existing) since, as mentioned in the issue, the trailing dynamic workspace disappears if that window is the only one on the next-to-last dynamic workspace.

ids1024:
Testing it, the window being dragged has keyboard focus (I can type into a terminal while dragging it around), so as far as I could tell only a visual change is needed; but there may be other issues I haven't noticed.

git-f0x:
The linked issue contains some examples. E.g. the trailing dynamic workspace disappearing, or previously focused windows visually changing to focused (e.g. Firefox titlebar buttons becoming brighter) when the currently focused window is dragged.

Issue #1510, opened 2025-07-13
tf2 (native linux build) always starts windowed, even when set to fullscreen


ellieplayswow:
this seems to be a regression with the recent fullscreen changes

```
$ apt policy cosmic-comp
cosmic-comp:
  Installed: 0.1~1751912143~24.04~ef5a1a3
  Candidate: 0.1~1751912143~24.04~ef5a1a3
  Version table:
 *** 0.1~1751912143~24.04~ef5a1a3 1001
       1001 http://apt.pop-os.org/release noble/main amd64 Packages
        100 /var/lib/dpkg/status
```

when loading tf2, it starts fullscreen & correct res, but when the actual main menu loads in it forces itself to become a minimised window, despite still recognising it as fullscreen in the settings. to fix this, i need to go from fullscreen -> windowed -> back to fullscreen again

(i apologise for the funny image i used to block out my friends list, it was the only remotely appropriate image i could find when making this demo vid)

https://github.com/user-attachments/assets/188e2667-a56a-4524-ad33-b2d7966ca46b

Drakulix:
Hmm, that looks like two issues.

- Firstly a weird focus issue, as some apps will minimize themselves, when they loose focus, but I am not sure why it suddenly would loose focus when switching to the main menu?
- Secondly not restoring to fullscreen correctly.

Thankfully this is a free game, so I'll investigate once I have the time.

WatchMkr:
Goat Simulator also starts off screen to the right and the menus can't be clicked. Super+F11 to full screen is a work-around.

ids1024:
I believe this is SDL's "minimize on focus loss" behavior. In which case it can be worked around with the `SDL_VIDEO_MINIMIZE_ON_FOCUS_LOSS=0` env var.

As I understand, the problem is that when cosmic-comp fullscreens a window, it changes the focus to `KeyboardFocusTarget::Fullscreen`, which causes a `wl_keyboard::leave` , before later sending a `wl_keyboard::enter` for the same surface.

ids1024:
[...]

PR #1691, opened 2025-09-30
focus: Check if X11 surface matches in `KeyboardFocus::replace()`

ids1024:
Tested on `SDL_VIDEO_MINIMIZE_ON_FOCUS_LOSS=1 supertux2`, using XWayland. If the game starts with fullscreen configured, it minimizes immediately on start. But after this change, this no longer happens.

This hopefully will improve behaviors with various X11/wine fullscreen clients. It may fix https://github.com/pop-os/cosmic-comp/issues/1510, though looking at the video the behavior there may be a bit different.

### DIFF
--- a/src/shell/focus/target.rs
+++ b/src/shell/focus/target.rs
@@ -33,7 +33,7 @@ use smithay::{
     },
     utils::{IsAlive, Logical, Point, Serial, Transform},
     wayland::{seat::WaylandFocus, session_lock::LockSurface},
-    xwayland::xwm::XwmId,
+    xwayland::{xwm::XwmId, X11Surface},
 };
 
 #[derive(Debug, Clone, PartialEq)]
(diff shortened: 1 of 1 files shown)


ids1024:
Testing TF2, I suppose https://github.com/pop-os/cosmic-comp/issues/1510 can be considered *partly* fixed.

* Before the change, I see it minimize on start, but after the change I end up with a fullscreen window as expected. So the issue with TF2 minimizing on start is fixed.
* As shown in the video, there's a weird thing where the window shrinks and grows a couple times before starting? Not sure what's going on
* When I focus another window, it minimized (presumably intended; I think it's using SDL, so this is the same "minimize on focus loss feature"). But focusing the app again makes the window end up non-fullscreened.

So we want another fix so the client can restore fullscreen properly when unminimized. At least.

ids1024:
Failing to restore fullscreen on unminimize also seems to be an XWayland thing, testing supertux2 on both wayland and x11 backends.

ids1024:
> Urgh, so somewhere during these events Xwayland is creating a new wl_surface for the same window, making our check fail?

Yeah, something like that. I didn't dive into the specific X and Wayland logs, but guessed this might help and it did. And it seems safe enough to add this.

Drakulix:
> > Urgh, so somewhere during these events Xwayland is creating a new wl_surface for the same window, making our check fail?
> 
> 
> 
> Yeah, something like that. I didn't dive into the specific X and Wayland logs, but guessed this might help and it did. And it seems safe enough to add this.

Yes, Xwayland definitely is handling this just fine.

Normal wayland clients would not be fine without the corresponding enter/leave events.

PR #1693, opened 2025-10-01
chore: binding for display toggle

wash2:


Drakulix (commented):
Why does this drop the touchpad toggle?

PR #1726, opened 2025-10-10
Update smithay with layer shell updates

ids1024:
Update to smithay from https://github.com/Smithay/smithay/pull/1819.

Some changes are needed to adapt to https://github.com/Smithay/smithay/pull/1817. Even after looking at the changes there, I'm not entirely sure how the replace `configure_serial` and `current_serial`...

### DIFF
--- a/src/shell/element/surface.rs
+++ b/src/shell/element/surface.rs
@@ -43,7 +43,9 @@ use smithay::{
     wayland::{
         compositor::{SurfaceData, TraversalAction, with_states, with_surface_tree_downward},
         seat::WaylandFocus,
-        shell::xdg::{SurfaceCachedState, ToplevelSurface, XdgToplevelSurfaceData},
+        shell::xdg::{
+            SurfaceCachedState, ToplevelCachedState, ToplevelSurface, XdgToplevelSurfaceData,
+        },
     },
     xwayland::{X11Surface, xwm::X11Relatable},
 };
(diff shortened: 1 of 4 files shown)


ids1024:
Reading though the relevant `smithay` code again, and the uses of this in cosmic-comp, I think I've fixed how `last_acked` is used now. Using the `.current()` cached state where applicable.

ids1024:
This should fix https://github.com/pop-os/cosmic-epoch/issues/2199, though that appears to be an  inconsistent race condition that may not be easily reproducible.

The layer `arrange` fix works as expected when starting the panel with `sleep 5 && cosmic-panel` then showing the workspaces overview before that timer runs out. After the change, the workspaces view is resized to make room for the panel exclusive zone, which did not happen before.

src/shell/element/surface.rs
```diff
@@ -583,12 +531,18 @@ impl CosmicSurface {
...
 
                     let current_server = attributes.current_server_state();
-                    if attributes.current.size == current_server.size {
+                    let mut guard = states.cached_state.get::<ToplevelCachedState>();
+                    if guard
+                        .current()
+                        .last_acked
+                        .as_ref()
+                        .is_some_and(|configure| configure.state.size == current_server.size)
+                    {
```
Drakulix, line 540: Can't we simplify this whole thing with the new `with_committed_state` helper or am I misunderstanding how that works internally? Seems to me, that we don't need to manually check the ack'd serial.

Drakulix (commented):
Looks mostly good to me

src/shell/element/surface.rs
ids1024: Yeah, I guess `with_committed_state()` would work here; though we also use `states.data_map`, in addition to the `states.cached_state`, to get the `current_server_state()`.

I guess it wouldn't really be a problem to take the `with_states` lock twice (but not at the same time) here? Though it might not simplify things much.

ids1024:
Rebased and updated to smithay commit including fix in https://github.com/Smithay/smithay/pull/1853.

PR #1746, opened 2025-10-16
Update dependencies

git-f0x:
Mainly for updating libdisplay-info, but I wanted to address clippy as well, and there's a very big diff because of that. There's also a potential bug fix for iterating over crts, where clippy noted that the loop doesn't actually loop. The `'outer` loop never iterated over `conn_info`.
Also updated the CI.

From some testing, there don't seem to be regressions (in libcosmic parts or otherwise).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 4
 
 [[package]]
 name = "ab_glyph"
-version = "0.2.31"
+version = "0.2.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


src/backend/kms/mod.rs
```diff
@@ -797,7 +797,9 @@ impl KmsGuard<'_> {
                             break 'outer;
                         }
                     }
+                }
 
+                if !new_pairings.contains_key(&conn) {
```
Drakulix, line 801: Hmm, I guess this is indeed better for the very rare case of a connector with multiple encoders.

PR #1748, opened 2025-10-16
Call `dbus::ready` even when systemd is enabled

ids1024:
The Gentoo overlap packaged needed this to get socket activation working properly (https://github.com/fsvm88/cosmic-overlay/pull/88).

For some reason I didn't seem to need this testing socket activation on Pop!_OS? Unless someone knows exactly why or has a better idea, for now apply this change along with the socket activation change in https://github.com/pop-os/xdg-desktop-portal-cosmic/pull/176.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -60,7 +60,6 @@ impl State {
             if let state::BackendData::Kms(_) = &self.backend {
                 #[cfg(feature = "systemd")]
                 systemd::ready(&self.common);
-                #[cfg(not(feature = "systemd"))]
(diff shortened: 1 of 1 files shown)


Drakulix (approved):

ids1024:
https://github.com/pop-os/xdg-desktop-portal-cosmic/pull/176 should be tested with this, so I used the same branch name. But this change should be fine to merge without that too.

jacobgkau (approved):
Tested with https://github.com/pop-os/xdg-desktop-portal-cosmic/pull/176.

PR #1766, opened 2025-10-28
Remove concept of "privileged" clients; but instead recognize clients as "not sandboxed"

ids1024:
Compositor changes corresponding to https://github.com/pop-os/cosmic-session/pull/154.

Simply removing `privileged` doesn't quite work with how we still need to distinguish sandboxed clients, not counting the `com.system76.CosmicPanel` sandbox engine.

So instead of just a commit removing `privileged`, I've ended up with several commits that remove what we no longer need, and add a new `not_sandboxed()` check to use instead.

This should be good, and I don't anticipate issue, but I still need to do a bit of testing before marking it not a draft.

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -88,8 +88,10 @@ use smithay::{
         seat::WaylandFocus,
         security_context::{SecurityContext, SecurityContextState},
         selection::{
-            data_device::DataDeviceState, primary_selection::PrimarySelectionState,
-            wlr_data_control::DataControlState,
+            data_device::DataDeviceState,
+            ext_data_control::DataControlState as ExtDataControlState,
+            primary_selection::PrimarySelectionState,
+            wlr_data_control::DataControlState as WlrDataControlState,
         },
         session_lock::SessionLockManagerState,
         shell::{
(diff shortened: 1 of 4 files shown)


ids1024:
Added a commit to also make the wlr data control protocol available to non-sandboxed clients, without an env var.

This affects the backend `wl-copy` and `wl-paste` use, and should allow some other clipboard manager clients to work.

ids1024:
Oh, I guess we also don't currently have ext-data-control. Added that as well. Basically just a renaming of the wlr protocol.

Drakulix (approved):
LGTM.

We might also be able to get rid of the session socket on systemd systems, once we switched over to systemd user-services and can simply import the environment variables there.

But I guess it makes sense to keep around cosmic-session and thus this mechanism for non-systemd systems.

ids1024:
Yeah, I'd like to maintain support for non-systemd distros (and BSDs), even if we can't guarantee the same level of support and testing as on systemd. So some mechanism for starting everything other than systemd user services should be kept around.

jacobgkau (approved):
This gets `wl-paste` working and able to store clipboard history in e.g. `cliphist`. Not seeing any other functional changes.

PR #1772, opened 2025-10-30
shell: Updates to focus stack handling

ids1024:
Replaces https://github.com/pop-os/cosmic-comp/pull/1687, and includes https://github.com/pop-os/cosmic-comp/pull/1698.

Unlike the previous solution, this doesn't have issues on multiple outputs, but I'm having trouble getting the behavior right when moving a window out of a stack...

### DIFF
--- a/src/shell/mod.rs
+++ b/src/shell/mod.rs
@@ -3415,15 +3415,6 @@ impl Shell {
             return None;
         }
 
-        if !move_out_of_stack {
-            for workspace in self.workspaces.spaces_mut() {
-                for seat in self.seats.iter() {
-                    let mut stack = workspace.focus_stack.get_mut(seat);
-                    stack.remove(&old_mapped);
-                }
-            }
-        }
-
         let (window, _) = old_mapped
             .windows()
             .find(|(w, _)| w.wl_surface().as_deref() == Some(surface))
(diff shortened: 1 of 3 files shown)


src/shell/mod.rs
```diff
@@ -3441,6 +3432,20 @@ impl Shell {
...
         };
 
+        if move_out_of_stack {
+            // Update focus stack to set focus to the window being dragged out of
+            // the stack.
+            for workspace in self.workspaces.spaces_mut() {
+                for seat in self.seats.iter() {
+                    let mut stack = workspace.focus_stack.get_mut(seat);
+                    // XXX Not finding CosmicStack in focus stack?
+                    if stack.remove(&old_mapped) {
```
Drakulix: I don't think we want to remove the stack. The new window should take precedence, but we still want focus to fall back to the stack, if the new window gets destroyed for example. So the old code was likely wrong in the first place.

Drakulix (approved):
looks good so far

src/shell/mod.rs
ids1024: Yeah, good point. It may be best to leave it in the stack as well.

> So the old code was likely wrong in the first place.

I actually inverted the if condition too, so that's probably the reason it previously removed only for `!move_out_of_stack`.

ids1024: I'm not sure why this isn't finding the `old_mapped` `CosmicStack` in the focus stack.

But if we just want to append to the stack instead of removing, it seems we just use `space_for_mut` to get the workspace, and just the the `seat` passed to the function. Instead of iterating and searching through focus stacks.

ids1024:
I think the focus indicators are working well now, in all cases.

git-f0x:
This seems to fix almost everything. The only thing I noticed that still occurs is the trailing dynamic workspace disappearing if the only window on the next-to-last workspace is grabbed, which doesn't seem like it should be intended behavior (should only potentially change on drop).

Drakulix (approved):
LGTM!

ids1024:
> The only thing I noticed that still occurs is the trailing dynamic workspace disappearing if the only window on the next-to-last workspace is grabbed, which doesn't seem like it should be intended behavior (should only potentially change on drop).

Yeah, that might make more sense. Though it's also not useful to move a window from a (non-pinned) workspace to the dynamic workspace after it, since that will just auto-remove one of the workspaces regardless.

PR #1793, opened 2025-11-07
WIP `ext_workspace_image_capture_source_manager_v1`

ids1024:
https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/463


Drakulix:
Mind mentioning that as an implementation on the linked wayland-protocols issue?

ids1024:
I was going to add a client implementation too, and link both.

PR #1797, opened 2025-11-10
Xwayland DnD support

Drakulix:
Draft on top of #1726 to play around with https://github.com/Smithay/smithay/pull/1848

### DIFF
--- a/src/wayland/handlers/data_device.rs
+++ b/src/wayland/handlers/data_device.rs
@@ -5,15 +5,14 @@ use smithay::{
     delegate_data_device,
     input::{
         Seat,
-        pointer::{CursorImageStatus, CursorImageSurfaceData},
+        dnd::{DnDGrab, DndGrabHandler, DndTarget, GrabType},
+        pointer::{CursorImageStatus, CursorImageSurfaceData, Focus},
     },
-    reexports::wayland_server::protocol::{wl_data_source::WlDataSource, wl_surface::WlSurface},
+    reexports::wayland_server::protocol::wl_surface::WlSurface,
     utils::{IsAlive, Logical, Point},
     wayland::{
         compositor::{self, SurfaceAttributes},
-        selection::data_device::{
-            ClientDndGrabHandler, DataDeviceHandler, DataDeviceState, ServerDndGrabHandler,
-        },
+        selection::data_device::{DataDeviceHandler, DataDeviceState, WaylandDndGrabHandler},
     },
 };
 use std::sync::Mutex;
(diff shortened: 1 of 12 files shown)


src/wayland/handlers/data_device.rs
```diff
@@ -89,10 +92,39 @@ impl ClientDndGrabHandler for State {
...
+impl DndGrabHandler for State {
+    fn dropped<T: DndFocus<Self>>(
+        &mut self,
+        _target: Option<&T>,
+        _validated: bool,
+        seat: Seat<Self>,
+    ) {
         seat.user_data()
             .get::<Mutex<Option<DnDIcon>>>()
             .unwrap()
```
ids1024: I'm seeing a panic on this `.unwrap()` with a drag from an X11 window. I guess there isn't a `DnDIcon` when the drag icon in an override redirect X11 window?

Drakulix: Fixed. Yes, X11 grabs handle the icon internally, so the user-data never gets set, if no wayland dnd grab occurred previously.

Drakulix:
Rebased and ready for initial Q/A

Drakulix:
This is finally ready for review!

Drakulix:
> This gets drag-and-drop working to Xwayland apps and between Xwayland apps.

(Note that drag&drop between Xwayland apps was likely working fine before and doesn't need any special support from us. This purely concerns DnD between a Wayland and an Xwayland application. Though both DnD between Xwayland and between Wayland apps should be regression tested.) 

> It does seem like drag-and-drop from Xwayland to Wayland isn't working, or at least I haven't found any scenarios where it works yet. When I launch Chrome under Wayland (default), I can drag an image from it into LibreOffice Writer, but when I launch Chrome with its Ozone platform forced to X11, the same image doesn't get put into LibreOffice Writer when I try to drop it there. Are things like that supposed to work?

Yes they are. I tested previously and this morning again, that I can drag&drop files both ways between cosmic-files and nautilus launched with `GDK_BACKEND=x11`. I can reproduce your issue though. I seems to be specific to chrome though, as the same thing works with firefox though (started with `WAYLAND_DISPLAY` being unset).

I am investigating the chrome bug, as that likely also affects a bunch of electron apps.

Drakulix:
Found the issue with chromium-based apps: https://github.com/Smithay/smithay/pull/1864

Waiting for smithay review and merge

Drakulix:
Rebased with new smithay changes to fix chromium-based apps.

jacobgkau:
Looks like the issue with Chrome no longer responding to mouse input might be triggered specifically by dropping the image onto an existing image in LibreOffice Writer (rather than into an empty space in the document).

Drakulix:
> Looks like the issue with Chrome no longer responding to mouse input might be triggered specifically by dropping the image onto an existing image in LibreOffice Writer (rather than into an empty space in the document).

Fixes for this and the stuck dnd preview are send to upstream here: https://github.com/Smithay/smithay/pull/1867

Will update the PR as soon as it is merged.

Drakulix:
Pushed fix

Drakulix:
I am slowly starting to loose my mind over this PR, but I found this bug as well: https://github.com/Smithay/smithay/pull/1868

PR update pending as usual.

Drakulix:
Fixed. Tested various dnd interactions between cosmic-files (wayland), nautilus (X11), google-chrome (ozone X11) and libreoffice writer (wayland) and could not find any more crashes or regressions.

ids1024 (approved):

Drakulix:
> At this point, the functionality outweighs the tiny visual glitch, so I'm going ahead and approving. I'll plan to open an issue to track this if you don't address it here.

Please do, since it might be a bit until I find the time to figure this one out. The preview image from X11-clients is done and controlled by the X11 clients themselves. So potentially there is nothing to be done here by us, but if this happens less frequently on other desktops, I would want to at least look into it.

Issue #1660, opened 2025-09-21
[bug/regression] Crash when swapping group of tiled windows


peterkrull:
Similar to #695 , but crashes on a [different line of code](https://github.com/pop-os/cosmic-comp/blob/b19f66702fcdc9c4a7ab066e8dbca4e94bb5063e/src/shell/layout/tiling/mod.rs#L5566). And it only seems to crash the thread associated with the monitor being swapped to.  Might be a regression since I can reproduce it with the same description given then:

-  In workspace 1, open one terminal instance
-  In workspace 2, open two terminal instances
-  In workspace 2, do `Super + U` to highlight both terminals
-  Use `Super + X` to swap, and swap with the terminal in workspace 1 (`Super + X + 1`)

Or if using multiple monitors, swap the group on monitor 1 with a window on monitor 2.

```
Sep 21 21:44:20 fedora cosmic-comp[797415]: thread 'surface-DP-3' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidNodeIdForTree': src/shell/layout/tiling/mod.rs:5566
Sep 21 21:44:24 fedora cosmic-comp[797415]: thread 'surface-HDMI-A-1' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidNodeIdForTree': src/shell/layout/tiling/mod.rs:5566
```

BrodieRobertson:
I've just witnessed what appears to be the same issue when using swap mode and moving in the direction of another monitor

PR #1800, opened 2025-11-11
layout/tiling: Fix crash swapping across workspaces

Drakulix:
Fixes https://github.com/pop-os/cosmic-comp/issues/1660


leviport (approved):
Confirmed to fix https://github.com/pop-os/cosmic-comp/issues/1660, and I didn't notice any regressions during my testing.

jackpot51 (approved):

Issue #1674, opened 2025-09-26
focus error in stack window using mouse


dp-fun:
Repeated Issue - it is still in beta Version:

When using more than 1 tab in stack-view, there is a focus error when using the mouse clicking on editable fields (changing tabs in stackview with mouse)

To Reproduce ( works with any browser):
1.) open firefox
2.) press 'super' + 'y', then press 'super' + 's' (stack-view)
3.) open in firefox a new private window (will be added to stack)
4.) click with the mouse on the new added firefox stack-window
5.) click now with the mouse on an editable field and press some buttons (abc) -> no text will be shown
6.) in beta-Version of pop os 24.04 the focus stayed on 1st window in stack-view -> all text input went to wrong window

Expected:
Pressed buttons will be shown in text field
(when using only the keyboard to change the stackview windows everything works fine)
Once the 1st window is used, everything works also on the other stack-windows

When using cosmic-terminal in stack-view, I can change the terminals (with mouse) but the 1st pressed key on keyboard did not show up but changes the focus. The 2nd pressed key on keyboard is shown in cosmic-terminal...

OS: Pop 0S 24.04 - Beta (Cosmic)

AnomalRoil:
I can confirm this is impacting me as well and is easy to reproduce using Firefox:
- Open Firefox in stacked mode
- Focus on a text field, e.g. ctrl-L to focus on URL bar, type something in there: "test"
- Open a new private window using the keybinding Ctrl-Maj-P
- New window has focus in URL bar in theory, and yet keyboard inputs keep going to the text field in the previous windows. 

Actually this also works with this text field in Github issues: if I open a new window with Ctrl-N it correctly switches focus to it and allow me to type in the URL bar, but if I use Ctrl-Shift-N or Ctrl-Shift-P the focus doesn't follow.
Looks like it might be related to using the Shift key in the process?

Using Super-Shift-Arrows after having created that new window with Ctrl-Shift-P properly moves the new window around and not the old one, and then the focus properly sticks to that new window.

kalingibbons:
I can't consistently reproduce, but I have this happen in vscode fairly often (electron app). Usually I can get around it by cycling through windows using the keyboard. I feel like I've also had it happen in cosmic files, but I might be misremembering; I'm not worried about losing work if I need to kill the window of my file manager and start over.

kalingibbons:
Happened when I switched to Telegram. Couldn't click into the message field until I cycled through the windows using keyboard.

[...]

PR #1801, opened 2025-11-11
shell/stack: Don't reset `previous_keyboard` on noop `set_active` call

Drakulix:
Fixes https://github.com/pop-os/cosmic-comp/issues/1674, or at least what has been reported by OP specifically with firefox (didn't work with chrome).

Key to this bug is, that the two windows need to be from the same client, which then gets confused due to a missing enter event as to which window should receive the key-events sent by us.


ids1024:
I guess this is for the same issue I was trying to fix with https://github.com/pop-os/cosmic-comp/pull/1168 (also with Firefox in a stack)? With that change, one reproducible issue seemed to go away, but another set of reproduction steps seemed to still have issues.

Issue #1714, opened 2025-10-06
When moving a window from a floating workspace to tiled, tile the window


WatchMkr:

maria-komarova:
Just a few aspects of the behavior to specify:
- Moving a fullscreen window - keep the window fullscreen, as now. 
- Moving a maximized window - restore the window and tile it. 
- We won't keep track of Super+G

Drakulix:
Fixed via https://github.com/pop-os/cosmic-comp/pull/1802

PR #1802, opened 2025-11-11
shell: Tile windows moved to a tiled workspace

Drakulix:
Addresses https://github.com/pop-os/cosmic-comp/issues/1714

### DIFF
--- a/src/shell/mod.rs
+++ b/src/shell/mod.rs
@@ -3017,6 +3017,8 @@ impl Shell {
     ) -> Option<(KeyboardFocusTarget, Point<i32, Global>)> {
         let from_output = self.workspaces.space_for_handle(from)?.output.clone();
         let to_output = self.workspaces.space_for_handle(to)?.output.clone();
+        let to_is_tiling = self.workspaces.space_for_handle(to).unwrap().tiling_enabled;
+
         let from_workspace = self.workspaces.space_for_handle_mut(from).unwrap(); // checked above
 
         let is_minimized = window.is_minimized();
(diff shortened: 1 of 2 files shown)


leviport:
All three points in https://github.com/pop-os/cosmic-comp/issues/1714#issuecomment-3373827917 are working as outlined, but I think I found a corner-case that might need to be considered. If I fullscreen an app on a floating workspace with Super + F11 and move it to a tiling workspace, it stays fullscreen. However, un-fullscreening it on that tiled workspace leaves it floating instead of tiling it:

https://github.com/user-attachments/assets/ec28127c-0761-4d02-ab52-77e717b76132

If that's out of scope of this PR, I'm happy to approve it.

Drakulix:
> All three points in [#1714 (comment)](https://github.com/pop-os/cosmic-comp/issues/1714#issuecomment-3373827917) are working as outlined, but I think I found a corner-case that might need to be considered. If I fullscreen an app on a floating workspace with Super + F11 and move it to a tiling workspace, it stays fullscreen. However, un-fullscreening it on that tiled workspace leaves it floating instead of tiling it

Ping @pop-os/ux about how we want to address this corner-case.

That is fixable, albeit slightly annoying because I'd have to update the restore-data of the fullscreen window on move, but definitely doable.

maria-komarova:
From the user perspective it would be good to tile the window, because it might seem like something is not working as it should otherwise. If it's too much annoyance we can come back to it later but if it's not too much trouble better make the changes now?

Drakulix:
@leviport Pushed a commit, that should address the corner-case.

leviport (approved):
That did the trick!

leviport:
I did notice a small bug with transparency in cosmic-term, but I wasn't sure that it was related to this PR so I just opened an issue here: https://github.com/pop-os/cosmic-term/issues/622

Issue #1657, opened 2025-09-20
Fullscreen video freezes on 16:10 screen


DeLachs:
When I put a video (in Firefox, MPV, VLC) in fullscreen on my laptop display (2880*1800, 16:10) the screen freezes while the audio continues to play. Everything returns to normal after I exit the fullscreen video. Everything works normal if I use an external 16:9 display via e.g. USB-C.

Hardware is a ThinkPad with a Core Ultra 225U with the iGPU and no extra GPU.

The Cosmic version is the current git version installed via the AUR package and all other packages are up to date.

BjornTheProgrammer:
I also have a 16:10 aspect ratio (2560x1600), and I get the same issue, but I haven't tested with an external display. I tried disabling variable refresh rate, and that didn't help. Also using an Intel Core Ultra 9 185H cpu with Meteor Lake-P internal graphics, the computer also has a Nvidia GeForce RTX 4080 Max-Q / Mobile gpu, although it isn't active. 

Using the most up-to-date POP!_OS 24.04 LTS.

filmor:
Hah, I have this problem as well and I never noticed that it only happens on the laptop screen and not on the external one. This is the log that I see when it happens:

```
cosmic-comp[395127]: Failed to submit rendering: Failed to submit result for display
                     Caused by:
                         0: The underlying drm surface encountered an error: DRM access error: Page flip commit failed on device `Some("/dev/dri/card1")` (Invalid argument (os error 22))
                         1: DRM access error: Page flip commit failed on device `Some("/dev/dri/card1")` (Invalid argument (os error 22))
                         2: Invalid argument (os error 22)
```

The log is completely spammed with this for about 3 seconds.

adettori:
Hi, I have the same issue with a 16:10 Thinkpad T14s with the same logs as above.
The problem seems to disappear by setting the VRR in the settings to Disabled and restarting the DE.
This also fixed issues with other applications/games freezing when toggling fullscreen.

ghost:
I have discovered that this problem isn't there with the i915-sriov-dkms package from the AUR while having the xe driver on the blacklist. It may be something with the xe driver that isn't a problem with the i915 driver.

filmor:
I can confirm what @adettori reports, disabling VRR and restarting the DE (that last part is required) fixes the issue for me as well.

ghost:
I can also confirm, that the workaround from @adettori works.

Drakulix:
[...]

Issue #1372, opened 2025-04-24
Page flip errors on Surface Pro 9 when fullscreening window (vrr)


ids1024:
I've noticed that on the builtin monitor of my Surface Pro 9 (with an i7-1255U), fullscreening a Youtube video doesn't work and freezes the graphics. `journalctl` shows the page flip failed:

```
Apr 24 13:08:51 Superficies cosmic-comp[1004744]: Failed to submit rendering: Failed to submit result for display

                                                  Caused by:
                                                      0: The underlying drm surface encountered an error: DRM access error: Page flip commit failed on device `Some("/dev/dri/card1")` (Invalid argument (os error 22))
                                                      1: DRM access error: Page flip commit failed on device `Some("/dev/dri/card1")` (Invalid argument (os error 22))
                                                      2: Invalid argument (os error 22)
```

I assumed it was direct scanout related, but setting `COSMIC_DISABLE_DIRECT_SCANOUT` and `COSMIC_DISABLE_OVERLAY_SCANOUT` don't change anything.

No such issue occurs fullscreening a window on an external monitor. This also happens with the `simple_window` cctk example if `set_fullscreen` is added, so it isn't related to drm `wl_buffer`s or anything special the client does.

Looking at how `src/backend/kms` handles fullscreen windows, it seems this is caused by the `vrr = has_active_fullscreen;` line. Setting `vrr` to `true` there is causing page flips to fail on this system.

chrismazanec:
I'm experiencing exactly the same issue and behaviour on my Fujitsu Lifebook FMV Zero Note U (WU1-K1), with Intel Core Ultra 7 258V (Lunar Lake)

Tested Chrome stable/dev and Firefox, all the same.

Fullscreen works on external display only. Using latest COSMIC Alpha 7.

kode54:
I see nobody has mentioned which kernel version they're using. Kernel version is important when it comes to async flips. You likely need something 6.14 or newer.

chrismazanec:
just tested `6.14.4` with the same result/error
```
Linux black 6.14.4-061404-generic #202504251003 SMP PREEMPT_DYNAMIC Fri Apr 25 12:15:31 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
```

```
Apr 28 07:42:51 black cosmic-comp[1711]: Failed to submit rendering: Failed to submit result for display

                                         Caused by:
[...]

PR #1810, opened 2025-11-13
chore: smithay update

Drakulix:
Updates smithay, notably including [this fix](https://github.com/Smithay/smithay/pull/1856).

Which might not fully fix the adaptive sync logic, but it should avoid bugs like:
- https://github.com/pop-os/cosmic-comp/issues/1657
- https://github.com/pop-os/cosmic-comp/issues/1372

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4961,7 +4961,7 @@ checksum = "67b1b7a3b5fe4f1376887184045fcf45c69e92af734b7aaddc05fb777b6fbd03"
 [[package]]
 name = "smithay"
 version = "0.7.0"
-source = "git+https://github.com/smithay/smithay.git?rev=d743e1a#d743e1a317fa0f01d1c4cadd96d277a1ec7b59d9"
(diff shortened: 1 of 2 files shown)


jacobgkau:
This is working alright in my testing so far, but I haven't been able to recreate the VRR fullscreen video issues without the fix yet. @ids1024 Does this fix the YouTube fullscreen issue on your Surface Pro 9?

ids1024:
This seems to improve behavior on the Surface Pro 9. It ends up showing the video fullscreened instead of freezing.

But fullscreening and unfullscreening the Youtube video does result in the screen turning black for about a second (which doesn't happen on another system).

Drakulix:
> This seems to improve behavior on the Surface Pro 9. It ends up showing the video fullscreened instead of freezing.
> 
> 
> 
> But fullscreening and unfullscreening the Youtube video does result in the screen turning black for about a second (which doesn't happen on another system).

This is somewhat expected, as smithay falls back to a modeset. 

Further changes to smithay are needed to determine earlier, if a modeset would be necessary and thus update the settings from "Automatic" to "Disabled" by default.

This PR is only meant to avoid freezing and crashing.

ids1024 (approved):

Issue #725, opened 2024-08-13
Quickly repeating 4-finger swipe when at first or last workspace also repeats animation


peterkrull:
Example: If I am at the 2nd workspace and i swipe towards the 1st workspace multiple times in quick succession, the animation will sort of cancel, snap me back to the 2nd workspace and redo the animation. This can be done repeatedly.

In the first half of the view I am only swiping left-to-right. In the second half I swipe in the opposite direction.

https://github.com/user-attachments/assets/e8bfb4a2-1508-4894-8758-fb2bf990ee7a

I have stumbled upon this when quickly swiping a few times left-to-right to get to my first workspace.

BrunoWallner:
This also happens on the vertical layout and with the keyboard when pressing `Super` + `Arrow`. 
But using the keyboard it only occurs on the first and not last workspace.

https://github.com/user-attachments/assets/6846fe3a-8684-4738-aaaf-00ebfbfe860d

jacobgkau:
@peterkrull https://github.com/pop-os/cosmic-comp/pull/761 only fixed this for the first workspace, and it looks like it's still happening for the last workspace, correct?

peterkrull:
Ah you are right, I missed that!

WatchMkr:
Worskpaces also don't cycle back to the first or last workspace with gestures. Cycling through workspaces was recently added with a shortcuts refactor.

jacobgkau:
This was rendered obsolete when wrapping/cycling was implemented (without an option to disable it), but it seems the underlying cause is still present, as the bug's manifesting after disabling wrapping with https://github.com/pop-os/cosmic-comp/pull/2215.

PR #1825, opened 2025-11-18
gestures: Cycle through workspaces

Drakulix:
Fix for https://github.com/pop-os/cosmic-comp/issues/725 / workspaces not cycling through when using the swipe touchpad gesture.

This required a few more small changes through out the code base to additionally track the direction of the swipe, since it cannot be inferred any more from the indicies of the workspaces alone.


jacobgkau (approved):
Working as expected. Tested with both vertical and horizontal workspaces. Also tested with an external display with both separate and combined workspaces; I discovered https://github.com/pop-os/cosmic-comp/issues/1832, but that's not a regression.

PR #1844, opened 2025-11-25
Fix focus stack crash

hammerlink:
Fixes pop-os/cosmic-epoch#2548

# Issue
The focus stack was not properly being cleaned up after moving the element to another workspace.
This was not properly working for both tiling and floating layout. 

`shift_remove` does not remove  the occurrences within the focus_stack. This was caused because the elements were not properly hashed, which is required for `indexmap::Equivalent`.

### DIFF
--- a/src/shell/focus/mod.rs
+++ b/src/shell/focus/mod.rs
@@ -29,7 +29,7 @@ use super::{SeatExt, grabs::SeatMoveGrabState, layout::floating::FloatingLayout}
 mod order;
 pub mod target;
 
-#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub enum FocusTarget {
     Window(CosmicMapped),
     Fullscreen(CosmicSurface),
@@ -56,6 +56,15 @@ impl indexmap::Equivalent<FocusTarget> for CosmicMapped {
 impl indexmap::Equivalent<FocusTarget> for CosmicSurface {
     fn equivalent(&self, key: &FocusTarget) -> bool {
         key == self
+    }
+}
+
+impl Hash for FocusTarget {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        match self {
+            FocusTarget::Window(window) => window.hash(state),
+            FocusTarget::Fullscreen(surface) => surface.hash(state),
+        }
     }
 }
 


src/shell/workspace.rs
```diff
@@ -609,9 +609,10 @@ impl Workspace {
             None
         };
 
-        self.focus_stack.0.values_mut().for_each(|set| {
-            set.shift_remove(mapped);
-        });
+        self.focus_stack
+            .0
+            .values_mut()
+            .for_each(|set| set.retain(|m| m != mapped));
```
Drakulix: Hmm, these lines should really be equivalent, given this is a `Set`, there can also not be multiple instances of the same window inside the stack.

Can you add a debug print and compare the differences the between focus stacks you see here? Because this seems to be an upstream bug, if true, which we need to report.

src/shell/layout/tiling/mod.rs
```diff
@@ -1303,9 +1303,14 @@ impl TilingLayout {
...
+
+        let tree = &self.queue.trees.back().unwrap().0;
+        let tree_node = tree
+            .get(&node_id)
+            .map_err(|_| NodeIdError::NodeIdNoLongerValid)?;
+
         let state = {
-            let tree = &self.queue.trees.back().unwrap().0;
-            tree.get(&node_id).unwrap().parent().and_then(|parent_id| {
+            tree_node.parent().and_then(|parent_id| {
```
Drakulix: This is not a correct fix, the `unwrap` is very much there on purpose to catch these logic errors, we should never try to unmap an element, that is not part of the layout.

Do you have a backtrace of this crash? We would want to figure out instead, where the `unmap` call comes from and why it ends up being called wrongly.

hammerlink: This is the backtrace I have. You can get it by following the reproduction steps of the issue. The second time you execute `Super + shift + 0` this will occur as the element remained in the focus_stack of the wrong workspace
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidNodeIdForTree': src/shell/layout/tiling/mod.rs:1308
   0: log_panics::Config::install_panic_hook::{{closure}}
   1: std::panicking::rust_panic_with_hook
   2: std::panicking::begin_panic_handler::{{closure}}
   3: std::sys::backtrace::__rust_end_short_backtrace
   4: rust_begin_unwind
   5: core::panicking::panic_fmt
   6: core::result::unwrap_failed
   7: cosmic_comp::shell::layout::tiling::TilingLayout::unmap
   8: cosmic_comp::shell::workspace::Workspace::unmap_element
   9: cosmic_comp::shell::Shell::move_element
  10: cosmic_comp::shell::Shell::move_current
  11: cosmic_comp::input::actions::<impl cosmic_comp::state::State>::handle_shortcut_action
  12: cosmic_comp::input::actions::<impl cosmic_comp::state::State>::handle_action
  13: cosmic_comp::input::<impl cosmic_comp::state::State>::process_input_event
  14: <core::cell::RefCell<calloop::sources::DispatcherInner<S,F>> as calloop::sources::EventDispatcher<Data>>::process_events::{{closure}}
  15: <smithay::backend::libinput::LibinputInputBackend as calloop::sources::EventSource>::process_events
  16: <core::cell::RefCell<calloop::sources::DispatcherInner<S,F>> as calloop::sources::EventDispatcher<Data>>::process_events
  17: calloop::loop_logic::EventLoop<Data>::run
  18: cosmic_comp::run
  19: cosmic_comp::main
  20: std::sys::backtrace::__rust_begin_short_backtrace
  21: main
  22: __libc_start_call_main
             at ./csu/../sysdeps/nptl/libc_start_call_main.h:58:16
  23: __libc_start_main_impl
             at ./csu/../csu/libc-start.c:360:3
  24: _start
```

Drakulix: Hmm, I was under the assumption, that since https://github.com/pop-os/cosmic-comp/pull/1772 we are using the `KeyboardFocus` and not the `FocusStack` for all key-bindings. In any case this seems to be a symptom then of the broken FocusStack cleanup. Thanks for investigating!

@ids1024 Can you take a look at this?

src/shell/workspace.rs
hammerlink: I added some debugging logs. There are no multiple instances of the same window. But the hashes in the set do not match. That's probably why the shift_remove does not remove it.

```
pop-os cosmic-comp[15550]: Found 1 occurrences but shift_remove failed to remove any!
pop-os cosmic-comp[15550]: Hash comparison - found: 15618893515951602619, mapped: 744033539690621968
```

Drakulix: Oh, `indexmap::Equivalent` requires the hashes to be equivalent as well, I guess. Well at least that is an easy fix, we just need a custom `Hash` implementation on `FocusTarget`, that simply forwards to the internal `CosmicSurface` and `CosmicMapped` struct respectively.

Drakulix: Can you check if https://gist.github.com/Drakulix/121677c51697789abf7a272de2138b07 fixes the issue as well?

hammerlink: Yeah that fixes it!

ids1024:
Oops. I probably should have considered the fact that `Hash` was also a trait bound... and I actually wasn't aware that `Hash` producing the same value for both types is a requirement for `Borrow`.

A custom `Hash` impl seems like a good solution, anyway. :+1:

PR #1845, opened 2025-11-25
Support `_NET_ACTIVE_WINDOW` to allow X11 windows to set focus

ids1024:
Requires https://github.com/Smithay/smithay/pull/1863. This is based on https://github.com/pop-os/cosmic-comp/pull/1797 to compile with an updated Smithay.

This draft version duplicates logic from `XdgActivationHandler`. Though we probably don't want to set `WState::Urgent` like that (and should instead use the `UrgencyHint` from `WM_HINTS`).

We may want to restrict when a client can set this, but it's not clear exactly how. Comparing other implementations may help.

This seems for fix https://github.com/pop-os/cosmic-applets/issues/550, testing with the Slack app indicator. https://github.com/pop-os/cosmic-notifications/issues/35, with notifications, may the same problem.

### DIFF
--- a/src/wayland/handlers/xdg_activation.rs
+++ b/src/wayland/handlers/xdg_activation.rs
@@ -106,131 +106,144 @@ impl XdgActivationHandler for State {
         let Some(context) = token_data.user_data.get::<ActivationContext>() else {
             return;
         };
-        let mut shell = self.common.shell.write();
 
         match context {
             ActivationContext::UrgentOnly => {
+                let shell = self.common.shell.write();
                 if let Some((workspace, _output)) = shell.workspace_for_surface(&surface) {
                     let mut workspace_guard = self.common.workspace_state.update();
                     workspace_guard.add_workspace_state(&workspace, WState::Urgent);
                 }
             }
             ActivationContext::Workspace(_) => {
-                let seat = shell.seats.last_active().clone();
-                let current_output = seat.active_output();
-
-                if let Some(element) = shell.element_for_surface(&surface).cloned() {
-                    if element.is_minimized() {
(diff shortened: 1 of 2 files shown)


ids1024:
Okay, I've updated it to share code with xdg-activation instead of duplicating, so we're not adding duplicated code that can get out of sync. That should be good enough to merge it now, though we can always improve things more later.

I did end up having it set the urgent state in the same way as with xdg-activation. Though in the future, we could natively handle `UrgencyHint`. Not a huge priority though.

This does not attempt to add the activation to `pending_activations`. `ActivationContext::Workspace` doesn't really make sense since there's no specific workspace associated with the X11 event (well, if `currently_active_window` is has a value, we could try to find what workspace that is on, or try to use the timestamp in some way, but for the app indicators or notifications, that doesn't apply anyway).

It also seems odd to me that the workspace in `ActivationContext::Workspace` isn't actually used except when it's added to `pending_activations`? In any case, this doesn't change that.

The `pending_activation: Option<(ActivationKey, ActivationContext)>` argument is a little awkward. I thought about instead returning a `bool` and having the caller handle `pending_activations`, but then there are also some early `return` statements, so it would probably have to return three states in one way or another...

Anyway; if no one has a better idea, this seems like an adequate way to do things without code duplication, and we can improve things with more major changes later if needed. This should do what we want regardless, and allow X11 clients to active windows in the same way Wayland clients can.

Drakulix:
> It also seems odd to me that the workspace in `ActivationContext::Workspace` isn't actually used except when it's added to `pending_activations`? In any case, this doesn't change that.

Yeah that is a bit of bad naming. This is simply done, so that we record the workspace, when the token is requested by some launcher for mapping later. But obviously we don't want to move windows on activate.

Issue #1732, opened 2025-10-12
[Usability] VS Code Window resizing inconsistency/odd behavior


rw-van:
I have the the app tray merged into the bottom bar on the window.

Using Visual Studio Code, I have the window maximized.

- I click once on the top bar of the VS Code  window, just to the left of the "Customize layout button". It displays the "grab" cursor, and it immediately becomes over-sized, taking up more than the screen height and width (the bottom bar is covering the bottom of the VS Code display).
- If I click again, it switches to upper half of the screen.
- If I hold the mouse cursor down while full-screen, it immediately moves the window. Normally there should be some resistance to moving a full-size window, like the cursor should be dragged a bit before the window starts to move.
- Clicking that same spot multiple times causes the window size to jump between over-sized and half-screen somewhat randomly.
- Double-click returns the window to maximized.

WatchMkr:
Clicking the header is resulting in a grab. Other apps do this as well.

* Chrome
* Kdenlive
* Mattermost

Are a few examples.

AndrewLuecke:
I've replicated the grab behavior on fedora too in a few apps (but I'll try to work out which ones). In my case, it seems intermittent though

PR #1847, opened 2025-11-26
shell: Add `DelayGrab` for client initiated move requests

Drakulix:
Addresses https://github.com/pop-os/cosmic-comp/issues/1732 and adds the same one logical pixel move requirement libcosmic apps already employ internally for every client-grab.

(Looking at this we possibly want to change smithay to include default implementations for `PointerGrab` and `TouchGrab`, that just forward to `Handle`. And we possibly want a macro in cosmic-comp to generate implementations for enums holding multiple possible grabs...)

### DIFF
--- a/src/shell/grabs/delay.rs
+++ b/src/shell/grabs/delay.rs
@@ -1,0 +1,285 @@
+use smithay::{
+    input::{
+        Seat, SeatHandler,
+        pointer::{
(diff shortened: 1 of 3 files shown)


jacobgkau (approved):
Confirmed this fixes https://github.com/pop-os/cosmic-comp/issues/1732. Intentional grab operation seem to still work as expected.

wash2 (approved):

PR #1854, opened 2025-12-02
Handle `xdg-system-bell-v1`, marking workspace as "urgent"

ids1024:
This can be tested with terminals like `foot`, though not with the version packaged by 24.04, using commands like `tput bel`. Or `sleep 2 && tput bel` then switching to a different workspace.

It seems reasonable that a "bell" will cause a window to be urgent to get the user's attention. If we want a more complete concept of urgency, we should probably have it as as part of the toplevels state, not just the workspace state, and can highlight urgent windows with window borders, and in `cosmic-workspaces`.


ids1024:
If we want this functionality, we'll also want to add it to `cosmic-term`.

PR #1856, opened 2025-12-03
fix: Add second touchpad hotkey config for Pangolin keyboards

### LINKED ISSUE #1855: [24.04] pang12-15 touchpad hotkeys not working
I attempted to add support for the pang15's touchpad enable/disable hotkey here: https://github.com/pop-os/systemd/pull/60 

But the pang15's hotkey still doesn't work on 24.04 with that patch. The pang13, which already had its patch present in our 24.04 branch, also doesn't work; nor does the pang12, which is [already present in the code](https://github.com/pop-os/systemd/blob/1ecac500c6a3e7d01a4c23b831d541b5f258bf8f/hwdb.d/60-keyboard.hwdb#L1872) and shouldn't need a patch.

I'm not sure if the problem's with systemd or not, but I'm starting the issue here since it's where the only Pangolin-specific touchpad hotkey code I'm aware of is.

jacobgkau:
Fixes #1855.

Certain Pangolin models send Ctrl+Super+XF86TouchpadToggle instead of just XF86TouchpadToggle. I guess GNOME somehow didn't care about this?

### DIFF
--- a/data/keybindings.ron
+++ b/data/keybindings.ron
@@ -122,4 +122,5 @@
     (modifiers: [], key: "XF86AudioNext"): System(PlayNext),
     (modifiers: [], key: "XF86PowerOff"): System(PowerOff),
     (modifiers: [], key: "XF86TouchpadToggle"): System(TouchpadToggle),
+    (modifiers: [Super, Ctrl], key: "XF86TouchpadToggle"): System(TouchpadToggle),
 }


leviport (approved):

jacobgkau:
@Drakulix Let us know if you think this is correct, or if cosmic-comp should handle XF86 toggles differently so modifiers don't interfere with them. I tried looking for documentation on the topic (the closest I found was the [libxkbcommon source code comments](https://github.com/xkbcommon/libxkbcommon/blob/master/include/xkbcommon/xkbcommon-keysyms.h)), but not sure if it's actually standardized.

PR #1859, opened 2025-12-04
input: Make scroll zoom respect the zoom increment setting, center the zoom view when zooming

RooXChicken:
Fixes the issue where the Super+Scroll to zoom doesn't respect the Zoom Increment setting. Also fixes the issue where if you move the cursor then zoom in, it will be off center until you move the cursor.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4952,7 +4952,7 @@ checksum = "67b1b7a3b5fe4f1376887184045fcf45c69e92af734b7aaddc05fb777b6fbd03"
 [[package]]
 name = "smithay"
 version = "0.7.0"
-source = "git+https://github.com/smithay/smithay.git?rev=593b2d5#593b2d5fb32a24652d8cde12b53d6871d1bd2275"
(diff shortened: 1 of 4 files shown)


maria-komarova:
> The fix to center the zoom view when zooming is good.

It behaves a bit jumpy on my system actually. It either starts zooming where I zoomed before and then jumps to the place where the cursor is or it doesn't jump. In any case, the experience feels somewhat disorienting to me but maybe this is just my system?

> I'm finding the increased scroll-zoom speed at the default 50% scroll interval setting to be too high on a touchpad with this change. Even at a 25% scroll interval setting (the minimum currently configurable in the COSMIC Settings GUI), the touchpad zoom is faster than it was before this change and feels too fast.

I agree, the touchpad zoom feels too sensitive right now which might be a challenge for a lot of people.

JPDoesDev:
Hi all,
Is this issue still being worked?
I am a visually impaired user and this functionality (at it's current state) is a roadblock from me being able to use Cosmic.
Is there anything I can do to help, testing, documenting, feedback etc?

Thanks.

PR #1861, opened 2025-12-05
shell: Implement cursor position hints

RooXChicken:
Implements the `cursor_position_hint` from the Pointer Constraints protocol.
Allows windows to set the cursor position. Supports decoration offsets, window position, and fullscreen windows. Works across workspaces. (AKA it just works as expected)

Tested with Blender and Minecraft, where the issues outlined in https://github.com/pop-os/cosmic-epoch/issues/808 were fixed.

TODO:
[X] Create a small program to help with test cases
[] Implement the opposite of `surface_under` in `input`
[] Provide a wrapper for `pointer.set_location` that updates the entire state so there isn't duplicate code
[] Make old cases of pointer set logic use new `pointer.set_location`
[] Scan entire surface tree for surfaces
[X] Support sticky windows
[X] Ensure windows can't move the pointer beyond their geometry


src/wayland/handlers/pointer_constraints.rs
```diff
@@ -27,11 +27,78 @@ impl PointerConstraintsHandler for State {
...
+        surface: &WlSurface,
+        pointer: &PointerHandle<Self>,
+        location: Point<f64, Logical>,
     ) {
-        // TODO
+        if with_pointer_constraint(surface, pointer, |constraint| {
+            constraint.is_some_and(|c| c.is_active())
+        }) {
+            for (out, set) in &self.common.shell.read().workspaces.sets {
+                for workspace in &set.workspaces {
```
Drakulix: Given you don't seem to be using the output, you can just use `.spaces()` on `Shell`:

```suggestion
            for workspace in self.common.shell.read().spaces() {
```

src/wayland/handlers/pointer_constraints.rs
```diff
@@ -27,11 +27,78 @@ impl PointerConstraintsHandler for State {
...
+                    // get the right window position based on if the window is fullscreen
+                    // this isn't the global position, rather the position inside the workspace
+                    let origin = if let Some(fullscreen) = workspace.get_fullscreen() {
+                        if fullscreen.wl_surface().as_deref() == Some(surface) {
```
Drakulix: It's sadly not this easy. Any subsurface is also allowed to take a pointer lock, so we have to check the full surface tree and also take the subsurface offset into account.

src/wayland/handlers/pointer_constraints.rs
```diff
@@ -27,11 +27,78 @@ impl PointerConstraintsHandler for State {
...
+
+                                    Some((pos.x, pos.y))
+                                })
+                            }
+                            else {
+                                None
+                            }
+                        }
+                        else {
+                            workspace
```
Drakulix: This ignores sticky windows, which aren't part of a workspace, but of a workspace set.

src/wayland/handlers/pointer_constraints.rs
```diff
@@ -27,11 +27,78 @@ impl PointerConstraintsHandler for State {
...
+                                .mapped()
+                                .find(|window| {
+                                    window.wl_surface().as_deref() == Some(surface)
```
Drakulix: Same here, this needs to scan the whole surface tree

src/wayland/handlers/pointer_constraints.rs
```diff
@@ -27,11 +27,78 @@ impl PointerConstraintsHandler for State {
...
+                    // the offset from the output (monitor position)
+                    let workspace_origin = out
+                        .geometry()
+                        .loc
+                        .to_f64();
+
+                    pointer.set_location(Point::new(
+                        workspace_origin.x + origin.unwrap().0 + location.x,
+                        workspace_origin.y + origin.unwrap().1 + location.y
+                    ));
```
Drakulix: This is also sadly too simplistic (see everything that happens in `src/input.rs`-`State::process_input_events`):
- We need to inform the shell about the new position:
  - `Shell::update_cursor_position`
  - and `Shell::update_focal_point`
- We need to update the cursor position for on-going cursor-capture screencopy sessions

We also need to check if the position given by the client is actually inside the input-region of the surface. Otherwise we need to reject it.

Drakulix (commented):
Thanks for taking the time! I have been meaning to get around to this for quite a while, so this contribution is very welcome!

However this PR shows, that we should introduce two new utility function as this protocol will likely not be the only place in the future making use of this and it is difficult to get right:

- We should introduce the opposite of `surface_under` in the `input`-module. Something that given a surface and a relative location returns the global position.
- Additionally we should add a helper for wrapping the cursor via `pointer.set_location`, which also does update the rest of the associated state so we don't duplicate that code needlessly as well.

Would you be up to adjust the PR to do this?

RooXChicken:
Thank you so much for the feedback! I'll work towards fixing these issues and implementing the helper functions.

Would it be a good idea to only allow this for windows that have focus?

Drakulix:
> Would it be a good idea to only allow this for windows that have focus?

That is a neat idea, but no. If anything we would want to limit, that only focused windows can enable a constraint in the first place. And I think we might already do that.

But once a window has successfully locked the cursor, there is no reason to deny it the pointer-warp inside of it's own surface.

skygrango:
I've been playing some Proton fullscreen games recently (FF14 in particular) and also ran into the issue where the mouse becomes hard to control after entering a pointer constraint. I tested the changes in this PR, and they definitely improve the experience—thank you very much!

However, I found two issues with the current modification:

1. after leaving the constrained state, my mouse always ends up at the bottom-right corner of the screen ( resolution 200% )
2. for a game like FF14 where you frequently enter and leave the constrained state, having the mouse always start from the bottom-right corner every time you move it becomes quite inconvenient. I tried making it always recenter the cursor, but that also felt unnatural and affected the experience.

So I modified fn new_constraint to record the last cursor position instead
```rust
fn new_constraint(&mut self, surface: &WlSurface, pointer: &PointerHandle<Self>) {
    // XXX region
    if pointer
        .current_focus()
        .is_some_and(|x| x.wl_surface().as_deref() == Some(surface))
    {
        with_pointer_constraint(surface, pointer, |constraint| {
            constraint.unwrap().activate();
            self.last_pointer = Some(pointer.current_location());
        });
    }else{
        self.last_pointer = None;
    }
}

/// only apply last pointer when we are using fullscreen in my case
if let Some(fullscreen) = workspace.get_fullscreen() {
  if fullscreen.wl_surface().as_deref() == Some(surface) {
        if let Some(geometry) = workspace.fullscreen_geometry() {
            let pos = geometry.loc.to_f64();
            let workspace_origin = out.geometry().loc.to_f64();
            let last_point = self.last_pointer.unwrap_or(Point::new(0.0, 0.0));
            pointer.set_location(Point::new(
                workspace_origin.x + pos.x +  last_point.x,
                workspace_origin.y + pos.y + last_point.y,
            ));
        }
    }
}

```

I'm not entirely sure what the conventional or expected behavior is supposed to be, but the current experience feels good on my end. I'll continue testing with this approach for now.

update: correcting the behavior of the record pointer

RooXChicken:
I have implemented sticky window support as well as preventing cursors from leaving the window boundary. Next is the `pointer.set_location` function. (by the way i am still VERY new to rust, so i apologize for any noob-ish rust code!)

skygrango:
> I have implemented sticky window support as well as preventing cursors from leaving the window boundary. Next is the `pointer.set_location` function. (by the way i am still VERY new to rust, so i apologize for any noob-ish rust code!)

Thank you so much for making it possible for me to enjoy the game smoothly.

As a small way to give back, I’ve put together a personal template for reference and experimentation. Please note it is not an official solution.

Feel free to use or adapt it if you find it helpful: https://gist.github.com/skygrango/6a45ff0dd0e2e492a8d36dc8beba4703

RooXChicken:
This looks very helpful, thank you! I'm glad I was able to help, enjoy your game <3

skygrango:
Hey @RooXChicken , hope you're doing well. 

I’ve been working on a similar implementation and have something functional on my side. If you’re open to it, I’d be happy to share my work and explore how we might move this forward together.

Here’s my fork with the current implementation: https://github.com/pop-os/cosmic-comp/compare/master...skygrango:cosmic-comp:implement-pointer-constraints

If it’s helpful, feel free to use it as a reference, or I’d be happy to help build on top of it together.

Please let me know what you think :smile:

ids1024:
This should be superseded by https://github.com/pop-os/cosmic-comp/pull/2357. Closing.

PR #1863, opened 2025-12-05
toolchain: update to 1.90

misterclayt0n:
### Rationale 
Using a consistent Rust toolchain across all components simplifies development and increases reproducibility.

Since cosmic-applets, cosmic-notifications and cosmic-settings-daemon are in 1.90, we should probably follow that as well for the rest of the components.


git-f0x:
This can be closed now, since the toolchain was updated since.

PR #1871, opened 2025-12-08
Support for rounding windows and server-side shadows

Drakulix:
Can finally be tested, but also needs accompanying cosmic-settings work to expose the new experimental options.

### DIFF
--- a/src/backend/render/mod.rs
+++ b/src/backend/render/mod.rs
@@ -280,10 +280,10 @@ impl IndicatorShader {
                     Uniform::new(
                         "radius",
                         [
-                            outer_radius[0] as f32,
-                            outer_radius[1] as f32,
-                            outer_radius[2] as f32,
                             outer_radius[3] as f32,
+                            outer_radius[1] as f32,
+                            outer_radius[0] as f32,
+                            outer_radius[2] as f32,
                         ],
                     ),
                     Uniform::new("scale", scale as f32),
--- a/src/shell/element/window.rs
+++ b/src/shell/element/window.rs
@@ -581,6 +581,7 @@ impl CosmicWindow {
                         p.window.set_tiled(false);
                     }
                 }
+                p.window.send_configure();
             }
         })
     }


tobyrg:
I've noticed that the left and right sides of a window there is a small gap between window and the active hint. Seems to have been introduced with these changes as I don't see it on the latest commit on master.

image

image

git-f0x:
The shadows look great!
Some issues regarding clipping I noticed:
When all the options in Settings are turned off, SSD windows end up with weird borders (the rounding is on opposite sides than it should be):
Screenshot_2025-12-10_12-56-10

Clipping for the bottom corners of SSD apps doesn't work:
Screenshot_2025-12-10_12-58-47

Borders for SSD apps also stay present when the window is maximized, regardless of the options in Settings.

Clipping for floating non-libcosmic CSD apps makes them impossible to resize by grabbing an edge.
Edit: Qt apps seems to not be affected as much (potentially because of resize border placement?), since the grab handle appears and they can be resized. But GTK and others (e.g. Zed) are affected.

TheESN:
Do I understand correctly that rounding window edges here affect not only apps with SSDs, but with CSDs too?

tobyrg:
There is now a bright transparent strip at the bottom and sides of windows. Changing wallpaper to a colour makes it more obviously as some darker wallpapers it's harder to see.

image

image

Drakulix:
> There is now a bright transparent strip at the bottom and sides of windows. Changing wallpaper to a colour makes it more obviously as some darker wallpapers it's harder to see.

Pretty sure this was the shadow not being properly aligned with fractional scale values (which is hard to do). I pushed a commit that adds a physical pixel of safety margin, which seems to fix it.

git-f0x:
Other than the last 2 issues I mentioned (SSD maximized borders and clipped resize borders), this seems pretty good.
Though the border looks worse IMO since it was increased to 2(?) pixels, since it ends up with some aliasing artifacts in the corners (it seems to mostly affect the right top and bottom corners).
Screenshot_2025-12-15_14-44-38

Could the clipping be applied only to the visible part of the window, rather than also the resize border? It seems that's advertised by windows in some way, since the active hint doesn't include the resize border.

Drakulix:
@git-f0x Thanks for the in-depth review, I believe the bugs you found should all be fixed.

git-f0x:
Yep, seems to be working really well. There are still slight artifacts at the right top and bottom corners, but they're less noticable now, and the borders look better overall.
There is the issue of the SSD header bars not updating until being hovered over (e.g. if turning on/off tiled corner matching), but that's unrelated to this PR and has already been reported (I think, but I can't find it now 🙃).

hojjatabdollahi:
Is it intentional that "Round" and "Slightly round" are the same for window corners?

hojjatabdollahi:
This PR covers the gap differently. In the master branch choosing 24 for hint and 24 for the gap correctly fills the gap. But here the hint grows inward "over" the window instead of growing outward, leaving part of the gap empty and part of the window covered:
 
 
 Screenshot_2025-12-15_12-18-08
Screenshot_2025-12-15_12-23-55

git-f0x:
> Is it intentional that "Round" and "Slightly round" are the same for window corners?

Yes (though this is more a libcosmic thing). The corners are made to be (mostly) concentric with elements inside the window (e.g. navbars and context drawers).

hojjatabdollahi:
After closing a window I still see the shadow for a moment. If I disable shadow, I still see the 1 pixel wide frame (I'm assuming it's related to the 1 pixel wide safety that you mentioned0:
Screenshot_2025-12-15_12-38-26
Screenshot_2025-12-15_12-38-37

Drakulix:
> This PR covers the gap differently. In the master branch choosing 24 for hint and 24 for the gap correctly fills the gap. But here the hint grows inward "over" the window instead of growing outward, leaving part of the gap empty and part of the window covered.

Thanks! Should be fixed.

> After closing a window I still see the shadow for a moment. If I disable shadow, I still see the 1 pixel wide frame (I'm assuming it's related to the 1 pixel wide safety that you mentioned:

Nope, related to something completely different in the tiling mode code, but thanks anyway for noticing, this also caused some other problems with tiling animations specifically around swapping windows. Should also be fixed now.

hojjatabdollahi:
Thank you. The issue with the gap is solved. But when I enable shadow I see these little things in the corners:
image

Edit: I should mention, this only happens for tiled windows. Floating windows are good. So, I assume it's another tiling mode issue.

hojjatabdollahi:
Another issue I found:
If you set the hint size to 1, and resize a window, at some point when the window is wide enough, the right edge gets overlapped by something, and it keeps flashing in and out of existence! 

image

The same thing happens for the bottom edge too (this time it gets dimmer, not completely hidden):
image

For tiled windows the right and the bottom edge are separated from the window edges, so you can see a 1 pixel wide line of the background (the left and top edges are fine):

image




Edit: The dimming of the window hint is also present in the master branch, but it looks to bee more extreme in this PR to the point of vanishing:
image

Drakulix:
> Another issue I found [...]

I really struggle to reproduce/see these. Is this tiling or floating? Resizing with the mouse or keyboard? Scaling value? Would it be too much to ask to record a little video of this?

hojjatabdollahi:
@Drakulix 
- both tiling and floating
- I'm resizing with mouse
- I saw it on 125% scaling, but I tested it on 100% and it happens there too (video)

It's most visible when the hint size is 1, but it's hard to show it well in a video. Hopefully this video helps:

https://github.com/user-attachments/assets/fb0a73ad-4ef8-4519-9c62-65d3bec19296

hojjatabdollahi:
I was able to fix the vanishing border issue, and the weird corner by adding this:
```
--- a/src/backend/render/shaders/--- a/src/backend/render/shaders/rounded_outline.frag
+++ b/src/backend/render/shaders/rounded_outline.frag
@@ -1,4 +1,4 @@
-precision mediump float;
+precision highp float;
 uniform float alpha;
 #if defined(DEBUG_FLAGS)
 uniform float tint;
```

after
before

Drakulix:
> I was able to fix the vanishing border issue, and the weird corner by adding this:
> 
> ```
> --- a/src/backend/render/shaders/--- a/src/backend/render/shaders/rounded_outline.frag
> +++ b/src/backend/render/shaders/rounded_outline.frag
> @@ -1,4 +1,4 @@
> -precision mediump float;
> +precision highp float;
>  uniform float alpha;
>  #if defined(DEBUG_FLAGS)
>  uniform float tint;
> ```

Oh interesting, that is probably very driver dependant, but we should really use `highp` here. Out of curiousity, what gpu is this on?

hojjatabdollahi:
> Out of curiousity, what gpu is this on?

I'm on a Framework 13 AMD laptop with Ryzen 7 7840U APU ∴ AMD Radeon 780M Graphics (RADV PHOENIX)

Software:
 Mesa 25.3.1
 Linux 6.18.0
 Pop!_OS 24.04

hojjatabdollahi:
Thank you!

The latest changes fixes most of the issues I had.

One last little issue I see is the outline doesn't stick to the window border. Furthermore the border is drawn a pixel too far inside the window, so some of the window is visible outside the border. In this picture you can see for the bottom right window:

desktop -> window hint -> **desktop** -> **window** -> window border -> window

Instead of 

desktop -> window hint -> window border -> window 


image

I thought the issue is that the inside of the window hint is being AA'd (instead of being opaque) and since it's drawn before the window that's how some of the background gets exposed, but the bigger issue (I think) is the window border isn't matching the window. Fixing that would also fix the one pixel wide desktop background leaking through.

peterkrull:
This is really nice so far!

When tiling, where some windows can be forced into a smaller area than the application accepts, the corners may not get clipped and the shadow continues past the visible application.

Screenshot_2025-12-18_20-26-21

It might also be what is causing this, where the clipping mask and shadow are not always synchronized with the window hint.

https://github.com/user-attachments/assets/17b1f8cd-26fa-4576-922c-d5940c64a9bc

l-const:
It looks really nice. I am seeing the same issues reported by @hojjatabdollahi with the little border artifact.  Using a simple AMD Ryzen 9 7940HS (Radeon 780M Graphics)  for what its worth.  
floating-border-artifact

Edit: fyi i have an issue only with Brave where the clipping happens only when i hover over for the first time, it doesn't happen with chrome, discord or the cosmic apps. really weird , it eventually gets fixed when i hover with the mouse: 
brave_Artifact_clipping_until_pointer_hover

Brave version: [Brave 1.85.111 (Official Build) (64-bit)](https://brave.com/latest/)
Chromium: 143.0.7499.40


Edit: I am using this for 2 days now and I got a recent comp crash when using the workspace view:
[comp-crash.log](https://github.com/user-attachments/files/24279740/comp-crash.log)
Might not be related to this update and I can't replicate it easily.

`Dec 21 18:09:45 fedora cosmic-comp[304679]: Failed to force redraw for corner radius change.
Dec 21 18:10:00 fedora cosmic-comp[304679]: Failed to force redraw for corner radius reset.
Dec 21 18:10:03 fedora cosmic-comp[304679]: thread 'main' panicked at 'internal error: entered unreachable code': src/shell/mod.rs:3192`

Drakulix:
> One last little issue I see is the outline doesn't stick to the window border. Furthermore the border is drawn a pixel too far inside the window, so some of the window is visible outside the border. In this picture you can see for the bottom right window.

I overhauled the whole rendering code of the outline to
- Handle the inner and outer radius properly, meaning as separate variables and thus fix the misalignment in the corners
- Only apply the anti-aliasing in the corners not for the whole outline, resulting in crisper lines.
- Make sure the thickness of the outline is rounded to a physical pixel size in a way that makes sure we never expose the background image

Unfortunately there are still some issues with the exact positioning of the outline, which can cause it to slightly overlap the window when fractional scaling is in use. Fixing this requires a bigger overhaul of the rendering code.

> When tiling, where some windows can be forced into a smaller area than the application accepts, the corners may not get clipped and the shadow continues past the visible application.

Thanks for the hint. Turns out the corners and window border where always calculated using the window dimensions, not the tile dimensions. (Fixing this was a little annoying as clipping to a tile was supposed to happen after rendering the window, but alas I compute the tile size now before and pass the maximum size to window rendering code.)

So this should be fixed!

hojjatabdollahi:
@Drakulix Thank you. This looks really good.
With the latest changes it looks like SSDs are not updating when you change the roundness settings, until you hover your cursor over the window (It works fine for CSDs):

https://github.com/user-attachments/assets/91334692-87ab-4815-9c5f-26433b56c2c5

git-f0x:
That was the case before this PR as well, with SSDs not updating until hovered over.

hepp3n:
There is this bug with corners with some Snap packages. For example Firefox, Thunderbird and maybe more GTK apps. I wonder if it's only like snap issue or something with rendering can be improved. 

Screenshot_2025-12-22_23-05-32
Screenshot_2025-12-22_23-05-44
Screenshot_2025-12-22_23-09-08
Screenshot_2025-12-22_23-09-22

It happens on `Square` style in COSMIC Settings.

hepp3n:
Screenshot_2025-12-22_23-15-25


If window is maximized it's squared so maybe somehow we could "fake" this state in `Square` style.

Drakulix:
> There is this bug with corners with some Snap packages. For example Firefox, Thunderbird and maybe more GTK apps. I wonder if it's only like snap issue or something with rendering can be improved.
> It happens on `Square` style in COSMIC Settings.

> If window is maximized it's squared so maybe somehow we could "fake" this state in Square style.

I am not sure about the technical internals of snap, but that sounds like it is messing with the wayland connection somehow.

This PR sends the "tiled"-states from the xdg-shell protocol to instruct the client to render square corners: https://wayland.app/protocols/xdg-shell#xdg_toplevel:enum:state

Maximized has other side-effects, e.g. rendering potentially another maximize-button icon and sending `unset_maximize`-requests instead of `set_maximize` on click, so I don't want to abuse that state. If this only affects snap-packages, then Canonical needs to fix this.

Drakulix:
> That was the case before this PR as well, with SSDs not updating until hovered over.

Pushed another commit, that should damage the SSD headers when changing the radius.

hepp3n:
> If this only affects snap-packages, then Canonical needs to fix this.

Sorry, I found that Snap brave behave properly, so it's not like fully snap issue. I'm gonna test your latest commit and see how it behave for a moment. Will be back with some feedback :)

hepp3n:
> I am not sure about the technical internals of snap, but that sounds like it is messing with the wayland connection somehow.

You might be right. From what I see it has problems `squaring` apps which are `rounded` by default. So Firefox (GTK), Snap Store, etc. Brave is square by default, so it does not have problem with it. Rounding is on the other side simpler I guess?

After some digging this `wayland` thing, most of the snaps are built with `core22` runtime, so it uses Ubuntu 22.04 under the hood. Is it the source of the issue? That's why u left cosmic-comp for Pop 22.04 not updated? 

Maybe I will try to build for example firefox on `core24` at least to see if it's better :)

hepp3n:
Yes it's snap or rather runtime issue. Apps using `core22` which has older runtime are not aligned properly.
Thunderbird on stable release (with gnome42-2204) has this issue I reported. I just changed thunderbird channel for `monthly` so it's running on `gnome46-2404` and.... the issue is gone! There is a little delay for setting corner squaring, but it works :)


https://github.com/user-attachments/assets/1565f0b3-594a-4ca5-85e7-3b0673d89863

Drakulix:
> > I am not sure about the technical internals of snap, but that sounds like it is messing with the wayland connection somehow.
> 
> You might be right. From what I see it has problems `squaring` apps which are `rounded` by default. So Firefox (GTK), Snap Store, etc. Brave is square by default, so it does not have problem with it. Rounding is on the other side simpler I guess?

For rounding we literally cut of the square corners. So we need to force square corners in many cases.

> 
> After some digging this `wayland` thing, most of the snaps are built with `core22` runtime, so it uses Ubuntu 22.04 under the hood. Is it the source of the issue? That's why u left cosmic-comp for Pop 22.04 not updated?

That might cause the issue. cosmic-comp isn't updated on 22.04. anymore, because other dependencies we require aren't packages for 22.04 at all.

hojjatabdollahi:
> Pushed another commit, that should damage the SSD headers when changing the radius.

This did not fix the issue. And also introduced another issue with the headers when moving windows around, the header sticks out on the right side of the window:

https://github.com/user-attachments/assets/572164d2-9a7f-4421-a69c-ee7fa7012bb7

jacobgkau (approved):
Not seeing any real problems with this (after the community filtered out the initial kinks).

On upgrade, it looks like this will default to on for floating windows and off for tiled windows.

hepp3n:
Good job for all, and thanks for working on it @`Drakulix` :)

silvaivanilto:
Will context menus be supported?

Issue #1792, opened 2025-11-07
Crash when changing from windowed to fullscreen along with other windowing problems


iexavl:
I encountered multiple issues with windows in fullscreen/windowed mode. 

```
cosmic-comp 0.1.0 (git commit 6e4164643ea81d3a2c3b40ffc441950722a61bfb)
Arch linux
Linux 6.17.7-arch1-1
AMD Ryzen 9 5900X (24) @ 4.95 GHz
1920x1080 in 24", 144 Hz [External]
1920x1080 in 27", 60 Hz [External]
RTX 3060 (Driver Version: 580.95.05)
```

The most major of which was a cosmic-comp crash, which happened when I was changing a proton game from windowed to fullscreen (the game was warframe and it was presumably running with wayland since I had PROTON_ENABLE_WAYLAND=1 in launch options)

```
Nov 07 18:17:28 iexavl cosmic-comp[69047]: Toplevel for foreign-toplevel-list not registered for cosmic-toplevel-info.
Nov 07 18:17:28 iexavl cosmic-comp[69047]: Toplevel for foreign-toplevel-list not registered for cosmic-toplevel-info.
Nov 07 18:17:28 iexavl cosmic-comp[69047]: Toplevel for foreign-toplevel-list not registered for cosmic-toplevel-info.
Nov 07 18:17:28 iexavl cosmic-comp[69047]: Toplevel for foreign-toplevel-list not registered for cosmic-toplevel-info.
Nov 07 18:17:28 iexavl cosmic-comp[69047]: Toplevel for foreign-toplevel-list not registered for cosmic-toplevel-info.
Nov 07 18:17:28 iexavl cosmic-comp[69047]: Toplevel for foreign-toplevel-list not registered for cosmic-toplevel-info.
Nov 07 18:17:28 iexavl cosmic-comp[69047]: Toplevel for foreign-toplevel-list not registered for cosmic-toplevel-info.
Nov 07 18:17:28 iexavl cosmic-comp[69047]: Toplevel for foreign-toplevel-list not registered for cosmic-toplevel-info.
Nov 07 18:17:28 iexavl cosmic-comp[69047]: Toplevel for foreign-toplevel-list not registered for cosmic-toplevel-info.
Nov 07 18:17:28 iexavl cosmic-comp[69047]: Toplevel for foreign-toplevel-list not registered for cosmic-toplevel-info.
[...]

PR #1872, opened 2025-12-08
Fix toplevel disappearing after window gets unmapped and mapped 

iexavl:
Currently, when a window  is unmapped and at some point mapped again, it doesn't reappear as a toplevel.
This happens, because `new_toplevel` only inserts a toplevel state in the userdata of the window if it doesn't already exist, which is just the first time the window is mapped.
So when it gets unmapped, and [its foreign handle gets removed](https://github.com/pop-os/cosmic-comp/blob/8fc7f0809f32664864582ed39e40169cf6cfe8a0/src/wayland/protocols/toplevel_info.rs#L362)
it doesn't get it's foreign_handle reassigned.

Also, I am not entirely sure about this, but shouldn't the `state_inner.instances` get drained in remove_toplevel while sending out the closed events? If they aren't and for some reason the `ZcosmicToplevelHandleV1` objects there aren't destroyed by the time the surface is mapped again, aren't events are going to be sent for a presumably closed handle?

### DIFF
--- a/src/wayland/protocols/toplevel_info.rs
+++ b/src/wayland/protocols/toplevel_info.rs
@@ -335,9 +335,16 @@ where
         let toplevel_handle = self
             .foreign_toplevel_list
             .new_toplevel::<D>(toplevel.title(), toplevel.app_id());
-        toplevel
-            .user_data()
-            .insert_if_missing(move || ToplevelStateInner::from_foreign(toplevel_handle));
+
+        if let Some(toplevel_state) = toplevel.user_data().get::<ToplevelState>() {
+            let mut toplevel_state = toplevel_state.lock().unwrap();
+            toplevel_state.foreign_handle = Some(toplevel_handle);
+        } else {
+            toplevel
+                .user_data()
+                .insert_if_missing(move || ToplevelStateInner::from_foreign(toplevel_handle));
+        }
+
         for instance in &self.instances {
             send_toplevel_to_client::<D, W>(&self.dh, workspace_state, instance, toplevel);
         }


Drakulix:
> Currently, when a window is unmapped and at some point mapped again, it doesn't reappear as a toplevel. This happens, because `new_toplevel` only inserts a toplevel state in the userdata of the window if it doesn't already exist, which is just the first time the window is mapped. So when it gets unmapped, and [its foreign handle gets removed](https://github.com/pop-os/cosmic-comp/blob/8fc7f0809f32664864582ed39e40169cf6cfe8a0/src/wayland/protocols/toplevel_info.rs#L362) it doesn't get it's foreign_handle reassigned.

Thanks!

> Also, I am not entirely sure about this, but shouldn't the `state_inner.instances` get drained in remove_toplevel while sending out the closed events? If they aren't and for some reason the `ZcosmicToplevelHandleV1` objects there aren't destroyed by the time the surface is mapped again, aren't events are going to be sent for a presumably closed handle?

`closed` is not a destructor event: https://wayland.app/protocols/ext-foreign-toplevel-list-v1#ext_foreign_toplevel_handle_v1:event:closed

This we can't simply destroy the wayland object. But if we can end up still sending events to these objects, that would be an error on our side.

Drakulix (approved):

Drakulix:
@iexavl Did you test this with a specific client/application, that does unmap/map it's window? Just so @pop-os/quality-assurance knows what to look for, when testing the PR.

iexavl:
> Thanks! 

My pleasure, it is a pretty annoying bug.

> `closed` is not a destructor event: https://wayland.app/protocols/ext-foreign-toplevel-list-v1#ext_foreign_toplevel_handle_v1:event:closed
> 
> This we can't simply destroy the wayland object. But if we can end up still sending events to these objects, that would be an error on our side.

Right, there was that. I think it is a real possibility though, that if the handles aren't destroyed by the next time the surface is mapped, events are sent. I haven't found any evidence that would prevent this, but I also haven't tested it. At some point   I will and see what happens.

iexavl:
> @iexavl Did you test this with a specific client/application, that does unmap/map it's window? Just so @pop-os/quality-assurance knows what to look for, when testing the PR.

It happens in warframe, though any application that unmaps itself and maps itself again will do the job.
To get it to happen in warframe, just minimize the window (in fullscreen) and unminimize it. 

Edit: note that while running warframe, you may also experience [this](https://github.com/pop-os/cosmic-comp/issues/1792) unreachable code crash. I also discovered the cause of this and will probably open a pr for that too.

jacobgkau (approved):
I had some trouble recreating the minimization issue, as Warframe won't change from the default Borderless Fullscreen to regular Fullscreen on my hardware. However, after installing Proton-GE and adding the launch option to launch it with Wayland, I was able to see a similar issue when attempting to change the setting to regular Fullscreen (the window disappearing with the associated toplevel log lines). With this fix, I'm not seeing that sort of issue occur anymore.

jacobgkau:
I tried to revisit this to confirm what I saw wasn't the problem that https://github.com/pop-os/cosmic-comp/pull/1902 is supposed to fix instead, but I'm again not seeing the game ever enter actual fullscreen mode (only borderless fullscreen) and am now seeing no other problems aside from that (on the currently released version of cosmic-comp).

Without knowing what other apps would unmap and remap themselves like this, I'd still recommend just merging this if it makes sense from an engineering perspective. The fix doesn't seem to be causing any other problems.

Edit: I also tried this on a different system with Intel instead of NVIDIA graphics. On that one, the game does enter fullscreen okay, but it seems to launch minimized and won't un-minimize, with or without this fix. So I still can't say with certainty how this fix will affect behavior.

iexavl:
> Edit: I also tried this on a different system with Intel instead of NVIDIA graphics. On that one, the game does enter fullscreen okay, but it seems to launch minimized and won't un-minimize, with or without this fix. So I still can't say with certainty how this fix will affect behavior.

I also ran into this and tried to debug it, but it doesn't seem to stem from the compositor. The game minimizes and when you try to unminimize it, it just minmizes itself back. This minimization from what I saw at least is the game requesting it be minimized and not the compositor doing anything. To confirm this I checked if it happens on pop os 22.04 under X11 and it does, although there you can kind of get it to unminimize after spam retrying for a bit. On cosmic-comp I have also gotten it to unminimize by spamming, but it's harder. This could possibly be an issue with proton.

PR #1881, opened 2025-12-11
focus: Restore focus to swap group

Drakulix:
Fixes the second issue reported in https://github.com/pop-os/cosmic-epoch/issues/1269.

The issue was triggered, because the switch back from the second workspace to the first, did not take the swap groups into consideration, allowing the code to falsely select a child of the group as it's focus. Swapping then attempted to swap between a group and a child of said group, which resulted in cosmic-comp "losing" a window.

This fix makes sure, that we always select any active swap group for focus when switching back to the originating workspace, thus fixing the issue.

(A nicer fix would be to have an easier way to determine, if a potential selection would part of the tree of a current swap group. But that would require some more refactoring for quite a niche issue, the focus of this PR is just to avoid the error state.)

### DIFF
--- a/src/shell/focus/mod.rs
+++ b/src/shell/focus/mod.rs
@@ -1,5 +1,5 @@
 use crate::{
-    shell::{CosmicSurface, MinimizedWindow, Shell, element::CosmicMapped},
+    shell::{CosmicSurface, MinimizedWindow, Shell, Trigger, element::CosmicMapped},
     state::Common,
     utils::prelude::*,
     wayland::handlers::{xdg_shell::PopupGrabData, xwayland_keyboard_grab::XWaylandGrabSeatData},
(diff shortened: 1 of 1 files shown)


src/shell/focus/mod.rs
john-dc252, line 1: not a reviewer and asking just out of curiosity but, wouldn't the `else` block look cleaner the following way?

```rust
        let workspace = shell.active_space(output).unwrap();

        shell
            .overview_mode()
            .0
            .active_trigger()
            .and_then(|trigger| match trigger {
                Trigger::KeyboardSwap(_, desc)
                    if workspace.handle == desc.handle
                        && workspace.tiling_layer.has_node(&desc.node) =>
                {
                    workspace.tiling_layer.node_desc_to_focus(desc)
                }
                _ => None,
            })
            .or_else(|| {
                workspace
                    .focus_stack
                    .get(seat)
                    .last()
                    .cloned()
                    .map(Into::<KeyboardFocusTarget>::into)
            })
            .or_else(|| {
                workspace
                    .mapped()
                    .next()
                    .cloned()
                    .map(KeyboardFocusTarget::Element)
            })
            .or_else(|| {
                workspace
                    .get_fullscreen()
                    .cloned()
                    .map(KeyboardFocusTarget::Fullscreen)
            })
```

PR #1888, opened 2025-12-13
feat(kms): advertise texture formats for DMA-BUF to enable NV12 client rendering

Ericky14:
## Summary

This PR changes the DMA-BUF format advertisement from `render_formats` to `texture_formats` when creating the Wayland DMA-BUF socket for clients.

## Problem

Currently, cosmic-comp only advertises `render_formats` to clients via the `zwp_linux_dmabuf` protocol. These are formats the compositor can render TO (as framebuffer targets). However, clients that want to submit YUV formats like NV12 (common for hardware video decoders) cannot do so because NV12 is an "external only" format in EGL - it can be sampled but not rendered to.

This forces video players and other clients to perform unnecessary GPU color conversion (NV12→BGRx) before submitting buffers to the compositor.

## Solution

Advertise `texture_formats` instead, which includes all formats the compositor can sample FROM as textures. This includes:
- All formats from `render_formats`
- External-only formats: NV12, P010, and other YUV formats

## Benefits

- **Zero-copy video playback**: Clients can submit NV12 DMA-BUFs directly from hardware video decoders (NVDEC, VAAPI, etc.)
- **Reduced GPU load**: Eliminates unnecessary color conversion on the client side
- **Lower latency**: One less GPU operation in the video pipeline

## Testing

Tested with a custom GStreamer plugin that outputs NV12 DMA-BUFs from NVIDIA NVDEC. Before this change, the compositor did not advertise NV12 support. After this change, NV12 is properly advertised and the compositor correctly samples the NV12 textures.

## Format Comparison

| Type | Description | Includes NV12? |
|------|-------------|----------------|
| `render_formats` | Formats for rendering TO | ❌ No |
| `texture_formats` | Formats for sampling FROM | ✅ Yes |

## Sample Implementation

https://github.com/Ericky14/gst-cuda-dmabuf

### DIFF
--- a/src/backend/kms/device.rs
+++ b/src/backend/kms/device.rs
@@ -243,7 +243,7 @@ impl State {
 
         let gbm = GbmDevice::new(fd)
             .with_context(|| format!("Failed to initialize GBM device for {}", path.display()))?;
-        let (render_node, render_formats, is_software) = {
+        let (render_node, render_formats, texture_formats, is_software) = {
             let egl = init_egl(&gbm)?;
 
             let render_node = egl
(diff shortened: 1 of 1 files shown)


Drakulix:
This has potential complications for multi-gpu setups, where we might need to migrate buffer contents between different gpus, so this needs more testing.

Ericky14:
@Drakulix I have a system with two GPUs, one AMD and one Nvidia, so I could potentially test. Currently I use the Nvidia as the primary.
Do you know how I could go about testing cosmic-comp using both GPUs? Or is it something that it identifies and uses by default?
Because currently it is working well on my system. Also tested on my AMD Laptop and it worked well.

Drakulix:
Superseded and fixed by https://github.com/pop-os/cosmic-comp/pull/1928

Issue #1819, opened 2025-11-17
mpv window "jumps" when I start dragging it


fstanis:
Tested with `cosmic-comp` version `0.1~1763108062~24.04~1c74adf`.

When `mpv` is using Wayland (default without any config), if I start dragging the window with my mouse to move it, it first "jumps", usually so that the top of the window ends up where my mouse pointer started dragging.

Mouse pointer isn't visible on this screencap, but I started dragging from the bottom right when the window jumps:

![Image](https://github.com/user-attachments/assets/ee1b8029-3ef6-41df-81fe-7ecbdf50e6e7)

This doesn't happen if the `mpv` window is small. If I resize a `mpv` window so that it takes most of the screen, the issue starts happening. If I then resize it back to be smaller, the issue stops. I'd say the threshold is at 50% of the screen height, though I'm not sure.

Furthermore, per https://github.com/mpv-player/mpv/issues/17048, this doesn't happen if I start `mpv` with `--no-keepaspect`.

# TL; DR

1. Run `mpv --no-config --autofit=90% anyvideo.mp4`
2. Try moving the window by dragging it
3. Observe the window jumping when the drag starts
4. Resize the window to be much smaller or run `mpv --no-config --autofit=90% --no-keepaspect anyvideo.mp4`
5. Observe the jumping no longer happens

bobcat141:
I've experienced a similar behavior, although, for me, just maximizing and unmaximizing the mpv window is enough to fix the issue for as long as that specific window is open.

<details>

<summary>video evidence</summary>

https://github.com/user-attachments/assets/54dcf09c-7cdc-491d-a2ff-ae0363c3657c

</details>

git-f0x:
Maybe related to https://github.com/pop-os/cosmic-comp/issues/1645?

ids1024:
It looks like `initial_window_location` in the call to `MoveGrab::new()` is wrong when the keepaspect feature of mpv is used. Which causes the client to call `xdg_toplevel::set_geometry` with a different size than the one `xdg_toplevel::configure` sent to the client. (Which all appears to be correct on mpv's side.)

In particular, these lines seem to be at fault:

https://github.com/pop-os/cosmic-comp/blob/455abbb2f7705e92babd4a91ba18fc1a018d1a55/src/shell/mod.rs#L3608-L3610

I guess that comparison may not be quite right, nor the way `new_size` is then being used.

ids1024:
[...]

Issue #1645, opened 2025-09-16
Some apps jump on start of drag


git-f0x:
Some apps jump to another position when a drag is started. I only noticed it with LocalSend and LACT for now.
Not sure when it started happening, but it didn't happen some time back.

https://github.com/user-attachments/assets/ab0f9ce5-7adb-4c3f-adfd-d724dc746d28

GPU: 5070 Ti

ZeusCraft10:
have these pull requests addressed the issue? If not I Have a PR ready for yall.

jacobgkau:
https://github.com/pop-os/cosmic-comp/pull/1901 fixes this issue-- tested with LocalSend and LACT.

PR #1901, opened 2025-12-17
layout/floating: In `unmap`, don't alter size if no pending changes

ids1024:
If the intent here was that the `if let` would only handle the case where there are pending changes, then something like this should be right.

Seems to fix behavior in https://github.com/pop-os/cosmic-comp/issues/1819. https://github.com/pop-os/cosmic-comp/issues/1645 may be the same issue.

### DIFF
--- a/src/shell/element/surface.rs
+++ b/src/shell/element/surface.rs
@@ -173,6 +173,13 @@ impl CosmicSurface {
         }
     }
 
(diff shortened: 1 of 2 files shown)


Drakulix (dismissed):
That was the intent, seems reasonable to me.

ids1024:
Perhaps some other things could be improved in the relevant code, but this should be a good improvement for now anyway.

PR #1902, opened 2025-12-17
Fullscreen request unreachable code crash

iexavl:
This attempts to address the unreachable code described in #1792 by not allowing duplicate surfaces to be put in `shell.pending_windows`.
As mentioned in the issue, this crash occurs when running warframe. (found out later, but it happens when running fullscreen)

An application can cause a crash by doing something along these lines:

map a window, then unmap it -> window [inserted](https://github.com/pop-os/cosmic-comp/blob/455abbb2f7705e92babd4a91ba18fc1a018d1a55/src/shell/mod.rs#L2892) in pending
make a fullscreen request -> fullscreen [set](https://github.com/pop-os/cosmic-comp/blob/455abbb2f7705e92babd4a91ba18fc1a018d1a55/src/xwayland.rs#L1078) on the pending window

make a map request -> window [inserted](https://github.com/pop-os/cosmic-comp/blob/455abbb2f7705e92babd4a91ba18fc1a018d1a55/src/xwayland.rs#L762) in pending

Then at some point, Xwayland sends two  [map_window_notify](https://github.com/pop-os/cosmic-comp/blob/455abbb2f7705e92babd4a91ba18fc1a018d1a55/src/xwayland.rs#L770) events

At this point, the surface has been mapped twice. Once as a fullscreen surface and once as a floating/tiled window.
After that, if the application makes one more fullscreen request, a `CosmicMapped` [gets found](https://github.com/pop-os/cosmic-comp/blob/455abbb2f7705e92babd4a91ba18fc1a018d1a55/src/shell/mod.rs#L4577) for the specified window, which is not supposed to happen when the window is fullscreen and a little bit [below](https://github.com/pop-os/cosmic-comp/blob/455abbb2f7705e92babd4a91ba18fc1a018d1a55/src/shell/mod.rs#L4640) that the window gets unmapped from the workspace. The first thing the `Workspace::unmap_surface` function does is check for a fullscreen surface, which it finds and returns `WorkspaceRestoreData::Fullscreen`
And then we get the unreachable code a bit below the `unmap_surface` call..


Drakulix (commented):
This doesn't seem like a bad fix, but I wonder if this would be more concise and simpler, if we replaced the `pending_windows` vector with a `HashSet` of `PendingWindow` and make `PendingWindow` only hash the underlying `surface`.

Would you be up for trying this out and updating the PR accordingly? In any case thanks for the PR.

iexavl:
> This doesn't seem like a bad fix, but I wonder if this would be more concise and simpler, if we replaced the `pending_windows` vector with a `HashSet` of `PendingWindow` and make `PendingWindow` only hash the underlying `surface`.
> 
> Would you be up for trying this out and updating the PR accordingly? In any case thanks for the PR.

I also think this makes sense. In regards to that, I had a question. Right now CosmicSurface computes its hash based on it's [underlying Window field](https://github.com/pop-os/cosmic-comp/blob/d708e182782922370262c4a839fdbe109f689adc/src/shell/element/surface.rs#L67).
This makes comparing a CosmicSurface's hash with X11Surface and WlSurface hash pretty janky, because they aren't equal. Meaning, for example, 
```
let a : X11Surface = ...;
let b : CosmicSurface = CosmicSurface::from(a.clone());

a == b // true
hash(a.window_id()) == hash(b) // false
```

I reckon this can be worked around one way or another to make a hashset of `PendingWindow` work, but is this intended?

Drakulix:
> I reckon this can be worked around one way or another to make a hashset of `PendingWindow` work, but is this intended?

I had to actually dig a bit through smithay to figure out why, but in a nutshell:
- As I am sure you noticed, `Window`s Hash implementation is purely based on some internal id-field.
- The historic reason for this seems to be, that neither `X11Surface` nor `ToplevelSurface` were hashable.
  - This doesn't hold true for the `ToplevelSurface` anymore. The internal `ObjectId` used to compare as equal for any destroyed object, this isn't true anymore.
  - For X11Surface this unfortunately still is true. X11 re-uses object ids and we make no attempt to track their lifetime, nor do we guarantee `X11Wm` to have a value, but we need to know which connection a surface belongs to to have a meaningful hash.

As a result of this I unfortunately don't think the `HashSet` approach is a good idea anymore, working around both of these is too fragile and `pending_windows` shouldn't grow large enough for this to become a performance issue anyway. (Though we might want to add a little doc comment eluding to the nature of `Window`'s `Hash`-implementation.)

I am inclined to merge this PR now as is, but it needs to be updated after https://github.com/pop-os/cosmic-comp/pull/1975 changed the logic around `unmap_surface` a bit.

iexavl:
@Drakulix I forgot to comment and then I was away from my pc for a while. I updated it so it can be merged, let me know if there is anything else

Drakulix (approved):
Thanks!

Issue #1286, opened 2025-03-06
[Fullscreen][Gaming] Cursor moves slowly at the edges of game window


git-f0x:
When a game is fullscreen (not borderless window) and the cursor is at a screen edge, any motion in the direction of that screen edge cancels out most/all perpendicular motion.
E.g. the cursor is at the left screen edge, moving the mouse diagonally to the top-left doesn't produce any cursor motion.
This is especially an issue in games where moving and pushing-in the cursor at a screen edge moves the map (e.g. Civilization).

I've tested non-game Wayland and Xwayland clients (that can be fullscreened with F11), and I can't reproduce this, so it seems specific to games (in some way).

gabriele2000:
Just wanted to play Dying Light after months.
I couldn't, cursor is in top-left and while menus can be used with the keyboard, the game can't be played without a mouse, since the bug affects the player camera too (it starts rotating like crazy)

TLCPlusPlus:
I can confirm this issue is in Skyrim Special Edition as well, it makes the game completely unplayable.

TLCPlusPlus:
I found a weird workaround. I plugged in my controller (ps4 to be exact) and started up my game. When the game was at the main menu, I unplugged it and I was able to use my mouse again.

pugaizai:
dota2 also

nivaldoguedes:
Some solution?

PR #1907, opened 2025-12-19
[Input] Improve pointer clamping

### LINKED ISSUE #981: Interaction with UI is prevented when the cursor is touching the bottom or right screen edge
This follows up pop-os/cosmic-panel#149 (and other linked issues), where a panel-specific band-aid was applied to fix panel behavior at the bottom and right screen edges.

When the cursor is touching the bottom or right screen edges, nothing below the cursor can be clicked or interacted with.
A few examples:
- The main scrollbar of any maximized app can't be interacted with when the cursor is touching the right screen edge, so extra precision is required to grab a scrollbar. E.g. in Firefox, when the cursor is near the right edge, the scrollbar is revealed and can be interacted with, but when the cursor touches the screen edge, the scrollbar becomes uninteractable and hides.
- The close button in the top-right corner of maximized apps can't be clicked when the cursor is touching the top-right corner in apps that have this behavior, like Firefox (because the cursor also touches the right screen edge).
- When dragging an auto-tiled app with the cursor from e.g. the bottom left corner to the right edge (3 total tiled apps), the app won't move to the new location (and the preview won't be shown), if the cursor was moved fast enough and is touching the right screen edge. The cursor has to be moved slightly away from the edge for the preview to be shown, making the cursor tiling experience feel unresponsive in those cases.

There are likely more examples of this causing issues, but these are the most disruptive ones that I've noticed.

git-f0x:
This seems to match the behavior on other compositors and fixes #981.

I noticed that the cursor completely disappears when it's placed all the way to the right, rather than a small 1px portion still being visible like it is on other compositors or e.g. Windows, so it seemed like an off-by-one issue.

Moves pointer clamping to before  `new_under`, since it previously ignored any motion that goes off screen, causing issues in some fullscreen clients. Fixes https://github.com/pop-os/cosmic-comp/issues/1286.

The second (edit: now first) commit is optional, but some other components are already on 1.90, and it allows using `f64::next_down()`.

Not sure if similar parts of the zoom code should also have the same change, since I'm not fully sure if it's affected by this.
Edit: I think not, since it seems to make the issue Levi mentioned easier to reproduce.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -23,7 +23,7 @@ use crate::{
         },
         zoom::ZoomState,
     },
-    utils::{float::NextDown, prelude::*, quirks::workspace_overview_is_open},
+    utils::{prelude::*, quirks::workspace_overview_is_open},
     wayland::{
         handlers::{screencopy::SessionHolder, xwayland_keyboard_grab::XWaylandGrabSeat},
         protocols::screencopy::{BufferConstraints, CursorSessionRef},
@@ -356,6 +356,16 @@ impl State {
                         .cloned()
                         .unwrap_or(current_output.clone());
 
+                    let output_geometry = output.geometry();
+                    position.x = position.x.clamp(
+                        output_geometry.loc.x as f64,
+                        (output_geometry.loc.x + output_geometry.size.w - 1) as f64,
+                    );
+                    position.y = position.y.clamp(
+                        output_geometry.loc.y as f64,
+                        (output_geometry.loc.y + output_geometry.size.h - 1) as f64,
+                    );
+
                     let new_under = State::surface_under(position, &output, &shell)
                         .map(|(target, pos)| (target, pos.as_logical()));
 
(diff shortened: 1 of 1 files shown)


leviport:
I'm pretty sure this also fixes https://github.com/pop-os/cosmic-epoch/issues/2683. So far I haven't been able to recreate it after this update.

leviport (approved):
This seems good to me. Tested on multi-monitor and single monitor configurations. I'm not sure that https://github.com/pop-os/cosmic-epoch/issues/2683 is entirely solved, but this certainly makes it much harder to recreate.

git-f0x:
This now has an extra change that might fix some gaming issues. I haven't noticed issues from some testing (and it really improves behavior in affected games), but not sure if it might regress something.

Drakulix:
@leviport Can you give a quick re-approve after the rebase?

@git-f0x Thanks!

PR #1921, opened 2025-12-23
Fix/issue 1890 race condition

ZeusCraft10:
# fix: panic when quickly clicking after stacking windows (#1890)
## Description
This PR fixes a panic that occurs when a user triggers "Stack Windows" and then quickly clicks the mouse button again (Issue #1890).
## Issue Analysis
The panic was caused by a race condition in the `MoveGrab::drop` implementation:
1.  When a drag ends, [drop()](cci:1://file:///c:/Users/akhil/OneDrive%20-%20Roanoke%20College/Documents/Github/cosmic-epoch/cosmic-comp/src/shell/grabs/moving.rs:780:4-948:5) is called.
2.  The method scheduled an idle closure to perform the actual drop logic.
3.  Inside the idle closure, it retrieved the `SeatMoveGrabState` from the seat's user data.
4.  **Race Condition**: If the user clicked quickly enough (e.g., spamming clicks), a *new* grab session could start *before* the idle closure executed.
5.  This caused the idle closure to pick up the [MoveGrabState](cci:2://file:///c:/Users/akhil/OneDrive%20-%20Roanoke%20College/Documents/Github/cosmic-epoch/cosmic-comp/src/shell/grabs/moving.rs:56:0-66:1) of the *new* grab.
6.  The code then attempted to use this state (e.g., `window_id`) with logic intended for the *previous* grab's state, leading to a type mismatch (expecting `Data::Mapped` but finding `Data::Placeholder` or similar) and resulting in a panic: `internal error: entered unreachable code`.
## Fix
The fix ensures that the `SeatMoveGrabState` is captured **synchronously** in `MoveGrab::drop` before scheduling the idle callback. This guarantees that the drop logic operates on the state associated with the grab that is ending, regardless of any subsequent input events.
## verification
-   ** Reproduction**:
    1.  Open two tiled windows.
    2.  Drag one window over the other to stack them.
    3.  Spam left-click immediately upon release.
-   **Before Fix**: The compositor panics and crashes.
-   **After Fix**: The windows stack correctly, and subsequent clicks are handled as a new interaction without crashing.
Closes #1890

PR #1923, opened 2025-12-23
Fix/ime popup grab

### LINKED ISSUE #1763: [BUG] Qt application menus won't open when IME is active
Involved actors: `comsic-comp` from the latest git commit (5a2eca2), Qt 6.10.0, fcitx5 5.1.14

Find attached a minimal Qt application showing the bug. When a Wayland input method is active under cosmic-comp (`fcitx5-anthy` in this case, but it doesn't really matter which; and it only matters that it is active - no actual pre-edit is required), once a text input widget has gotten the focus in Qt, menus will no longer open when clicked on.

See the following video demonstrating this:

https://github.com/user-attachments/assets/647c8fee-0a05-4644-aa05-4237f5b973ef

(It is captured with cosmic-comp running nested inside KDE due to the lack of convenient screen capturing tool in COSMIC, but be assured that it behaves just the same in a regular COSMIC session).

I'm starting here as preliminary investigation suggests that it is more likely it is a `cosmic-comp` issue than a Qt issue. But it is possible that Qt is "painting outside the lines" here (especially in the `zwp_text_input_v3` implementation, which is the weakest of the three that it supports) in a way that is OK with KDE and GNOME but not with COSMIC. If so, please point to evidence and I'll take it to upstream.

[ime-issue.zip](https://github.com/user-attachments/files/23141258/ime-issue.zip)

ZeusCraft10:
# Description
Fixes #1763
This PR modifies the `XdgShellHandler::grab` method to allow XDG popup grabs to proceed even if the keyboard is currently grabbed by an input method (IME).
Previously, `cosmic-comp` would unconditionally reject popup grabs if any other keyboard grab was active. This prevented Qt application menus from opening when an IME (like `fcitx5`) held a grab for text input.
I added a check to identify if the conflicting grab is an `InputMethodKeyboardGrab` using `smithay::wayland::input_method::InputMethodKeyboardGrab`. If it is, the popup grab is allowed, enabling menus to function correctly during IME composition.
## Verification
- Verified manually by checking code logic against Smithay documentation.


hojjatabdollahi:
@ZeusCraft10 Are these PRs AI generated?

leviport:
All three PRs have the same 3 commits, with one additional commit on this PR. 

I would also appreciate an explanation @ZeusCraft10

ZeusCraft10:
the descriptions are AI generated. I didnt know if you guys had a contrbution template or anything so I asked gemini to write it for me.

leviport:
Why are https://github.com/pop-os/cosmic-comp/pull/1922 and https://github.com/pop-os/cosmic-comp/pull/1921 identical with the same 3 commits, and why are those commits also present in this PR?

ZeusCraft10:
There were 2 different issues submitted, I believe both of which had the same/similar  fixes. I submitted one PR for each.

src/wayland/handlers/xdg_shell/mod.rs
```diff
@@ -108,8 +108,18 @@ impl XdgShellHandler for State {
...
                         {
-                            grab.ungrab(PopupUngrabStrategy::All);
-                            return;
+                            // If the keyboard is grabbed by the input method, we should still allow the popup grab
+                            // to proceed. The input method grab is likely for the virtual keyboard or similar
+                            // and shouldn't prevent application menus from opening.
+                            let is_ime_grab = keyboard.grab().is_some_and(|g| {
+                                g.as_any()
+                                    .is::<smithay::wayland::input_method::InputMethodKeyboardGrab>()
+                            });
```
Drakulix, line 117: The problem here is, that the popup grab will override the active ime-grab. Which is also wrong, we need to truly allow both here.

There are more cases like these as seen in the discussion of https://github.com/Smithay/smithay/pull/1845. As such this needs to be addressed upstream first.

PR #1926, opened 2025-12-23
migrate from make to just and add mock command

FreddyFunk:
Biased, but I _just_ like it more. And most other cosmic projects also use just.

Also added a few commands like `just mock` to run the entire cosmic shell with in a winit window with the local build of cosmic-comp. This makes development of cosmic-comp easier, since you do not have to restart the host shell and potentially brick your pc temporary. Mock is not perfect, since for example app drawer and workspace commands are routed to the host but most things work as expected.


FreddyFunk:
there seems to be no interest in running cosmic comp its own window without needed to install it, so I close this PR

hojjatabdollahi:
Can you actually run `cosmic-comp cosmic-session` inside COSMIC DE? I would think all the dbus names are gonna collide, so pressing the workspaces button would open the workspaces on the host, opening launcher will open it in the host, etc....

You can probably run it nested with `cosmic-session` in a different DE though.

Issue #1661, opened 2025-09-23
[bug/regression] wine-wayland flickering and low fps


avnyu:
When running game with wine-wayland + dxvk, game fps is very low, mangohud is flickering and reported 0 fps. Wine logs constantly spammed with 
```
info:  Presenter: Actual swapchain properties:
info:    Format:       VK_FORMAT_R8G8B8A8_UNORM
info:    Color space:  VK_COLOR_SPACE_SRGB_NONLINEAR_KHR
info:    Present mode: VK_PRESENT_MODE_MAILBOX_KHR (dynamic: yes)
info:    Buffer size:  1920x1080
info:    Image count:  3
info:  Presenter: Got VK_SUBOPTIMAL_KHR, recreating swapchain
```
Also only when in workspace overview mode, game runs normally, the fps report correctly and logs no longer spammed.
The last work without issues commit is 28a9000833c5b3e347f07dcc44b48edc185b5a21. Current latest commit f339e21479c05153864af79c8fb5d005dcff8abc can also work if revert both b7fd4bc781914bf08c8e99b3fb4a061b784380fa and db9380e6dbb586a571c8f475c10e2a457028906d, for me.
OS: archlinux, cosmic-session-git, intel + nvidia dual gpu laptop

Issue #1915, opened 2025-12-22
Undesired GPU offloading behaviour with hybrid (NVIDIA) graphics


Ner0theHer0:
Apologies in advance as I'm sure part of this issue is the fault of NVIDIA.

Specs:
Host: Inspiron 16 Plus 7620 (no mux switch)
CPU: Intel i7-12700H
iGPU: Alder Lake-P Iris Xe
dGPU: GeForce RTX 3060 Mobile / Max-Q
NVIDIA drivers tried: 580.82.09, 580.119.02, 590.48.01

After a fresh COSMIC install, I noticed that a number of desktop related activities appeared to have poor performance on external monitors. This includes heavy stuttering/sub 30fps rendering when mousing over the desktop, mousing over full screen COSMIC apps (terminal, file explorer), window maximise/minimise/snap animations, top bar and dock auto-hide animations, etc.

Since I had come from 22.04 on x11, I assumed this was a wayland issue. However, I found I could not replicate the issue on gnome wayland and KDE plasma on Fedora 43. I did however find the same behaviour present in the Fedora COSMIC spin. 

After a fair bit of digging and playing around with different installs, it appears that this issue comes down to how aggressively COSMIC seems to offload workloads to the dGPU. I found that gnome and KDE _do not_ offload these desktop activities to the dGPU, whilst COSMIC does. When COSMIC offloads to the dGPU, it is often in the NVIDIA PowerMizer low performance level (level 0), which causes the aforementioned performance issues. When an application forces the dGPU to spin up a little more (performance level 2/3) the lag disappears. I noticed this when starting OBS to get a screen recording of the behaviour, which unfortunately pushed the GPU enough to keep it in levels 2 and 3. 

It is worth noting that the dGPU seems _very_ resistant to move to the performance level 2 and 3 even in COSMIC's high performance mode. It instead tends to stay in performance level 1 and results in constant sub 30fps outside of non COSMIC applications. In balanced mode, I cannot get the GPU to go higher than level 1 with normal desktop use, which means there is always stuttering. Worth noting here that between level 1 and 2 there is a huge jump in memory transfer rate, which appears to be part of the issue.

Right now there are a few soft workarounds, but nothing that is able to get a constant smooth experience in COSMIC. The following stops the GPU from going into level 0, but still has heavy stuttering if the GPU goes back to level 1 (which takes about 2 seconds):
[...]

PR #1928, opened 2025-12-23
Bunch of multigpu + dmabuf fixes

Drakulix:
Supersedes https://github.com/pop-os/cosmic-comp/pull/1888.
Ping @Ner0theHer0, this is the PR I talked in https://github.com/pop-os/cosmic-comp/issues/1915#issuecomment-3681787845 about.
Should fix https://github.com/pop-os/cosmic-comp/issues/1661.

This PR improves various edge-cases related to dmabufs in particular on multi-gpu setups.

- We track clients instead of buffers to determine, if a GPU context should be held active. This fixes vulkan apps, which drop all their buffers upon re-creating a suboptimal swapchain. Causing a huge lag spike and potentially broken imports after cosmic-comp has re-created the device..
- Advertising `texture_formats` instead of `render_formats`, recent changes to smithays multigpu-renderer make this possible without accidentally locking us out of copy-paths.
- Create dumber surface feedback, like most clients expect (*cries*) and be more stable with the feedback we are advertising as mesa is very quick to re-allocate the whole swapchain.
- Send initial surface feedback to further decrease the number of feedback changes in the lifetime of a program.

This fixes a bunch of issues with wine-wayland (tested with Proton-GE) in particular.

### DIFF
--- a/src/backend/kms/surface/mod.rs
+++ b/src/backend/kms/surface/mod.rs
@@ -118,7 +118,7 @@ pub struct Surface {
     known_nodes: HashSet<DrmNode>,
 
     active: Arc<AtomicBool>,
-    pub(super) feedback: HashMap<DrmNode, SurfaceDmabufFeedback>,
+    pub feedback: HashMap<DrmNode, SurfaceDmabufFeedback>,
     pub(super) primary_plane_formats: FormatSet,
     overlay_plane_formats: Option<FormatSet>,
 
--- a/src/wayland/handlers/dmabuf.rs
+++ b/src/wayland/handlers/dmabuf.rs
@@ -1,10 +1,18 @@
 // SPDX-License-Identifier: GPL-3.0-only
 
-use crate::state::State;
+use crate::{
+    state::{BackendData, State},
+    wayland::handlers::compositor::frame_time_filter_fn,
+};
 use smithay::{
-    backend::allocator::dmabuf::Dmabuf,
+    backend::{allocator::dmabuf::Dmabuf, renderer::element::Kind},
     delegate_dmabuf,
-    wayland::dmabuf::{DmabufGlobal, DmabufHandler, DmabufState, ImportNotifier},
+    desktop::WindowSurfaceType,
+    reexports::wayland_server::protocol::wl_surface::WlSurface,
+    wayland::{
+        compositor::with_states,
+        dmabuf::{DmabufFeedback, DmabufGlobal, DmabufHandler, DmabufState, ImportNotifier},
+    },
 };
 
 impl DmabufHandler for State {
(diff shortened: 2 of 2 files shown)


Drakulix:
Reverted to draft. Needs https://github.com/Smithay/smithay/pull/1888 to work correctly.

niklaskorz:
Tested this PR on a single-GPU desktop with Nvidia open kernel modules 590.48.01 and 580.119.02 (using cosmic-comp from this branch, all other cosmic packages from latest stable release) and it looks like most wgpu based applications (including all cosmic applications, as well as simple winit + wgpu tests) fail to render after the first frame (and thus look frozen) or crash (`failed to get surface texture: Outdated`). Firefox, Ghostty and Electron-based applications seem to work without issues though.

hepp3n:
I gave this PR a go since @niklaskorz comment. I have ThinkPad with 2x AMD gpus. I cannot observe issues like he describes but I'm testing my wine-wayland game I talked to you @Drakulix on mattermost.

I mean this topic: https://chat.pop-os.org/pop-os/pl/5867df17abra9bu1qwze3gz4ho

I'm testing this for few minutes but it looks like this wine-wayland game performs much better. I hope it's not placebo but experience much smoother gameplay and also keeping the game on another workspace so it can be in `idle` state I talked about, does not make discrete GPU behave like in this topic. When I go to game's workspace, the game runs still smoothly just like I would play it. Before it could potentially lag the game and `unfreeze` it after a second when switching workspaces. Totally improved.

Drakulix:
> Tested this PR on a single-GPU desktop with Nvidia open kernel modules 590.48.01 and 580.119.02 (using cosmic-comp from this branch, all other cosmic packages from latest stable release) and it looks like most wgpu based applications (including all cosmic applications, as well as simple winit + wgpu tests) fail to render after the first frame (and thus look frozen) or crash (`failed to get surface texture: Outdated`). Firefox, Ghostty and Electron-based applications seem to work without issues though.


I can re-create this using an nvidia-only setup, thanks for the heads up. I'll have to dissect this unfortunately, given that nvidia doesn't have sources for their vulkan-wsi code available.

Drakulix:
@niklaskorz Should be fixed! Can you verify as well?

niklaskorz (approved):
Works perfectly fine on Nvidia and Proton Wayland now runs without issues 🎊

niklaskorz:
One more thing I noticed just now, though probably not a blocker considering this PR is already a huge improvement: Proton Wayland games running in "exclusive fullscreen" mode still freeze, but run perfectly fine in borderless fullscreen. I was only able to reproduce this with DXVK and VKD3D games though, not with games that directly use Vulkan.

Edit: Doesn't seem to be strictly limited to exclusive fullscreen (which isn't really a thing anyway on Wayland afaik?), Oblivion Remastered also freezes in borderless and windowed mode if the whole screen is filled. In windowed mode with a resolution lower than the display, it runs without issues.

Edit 2: oddly enough that is only a problem at game start, if I change to fullscreen while already playing it doesn't freeze

Edit 3: same issue observed in Manor Lords (VKD3D), freezes at fullscreen / borderless fullscreen / window covering the whole screen, but works fine in windowed mode not covering the screen; unlike Oblivion it also freezes when switching to fullscreen at runtime

gabriele2000:
> Proton Wayland games running in "exclusive fullscreen" mode still freeze

Huge disappointment for me since I've been waiting for this PR to merge because I thought it'd fix this issue.
Since Proton-GE 10-26, starting a game with the Wayland driver makes it freeze, the exception is Helldivers 2 and a few others.

The vulkan issue that gets returned is VK_SUBOPTIMAL_KHR which makes everything freeze... I mean, this is what happens before the freeze.
Disabling Wayland driver fixes the issue, but I'm not going to play without Wayland driver anymore, I'd rather wait or stay with Proton-GE 10-25 until it's fixed.

Drakulix:
> > Proton Wayland games running in "exclusive fullscreen" mode still freeze
> 
> Huge disappointment for me since I've been waiting for this PR to merge because I thought it'd fix this issue. Since Proton-GE 10-26, starting a game with the Wayland driver makes it freeze, the exception is Helldivers 2 and a few others.
> 
> The vulkan issue that gets returned is VK_SUBOPTIMAL_KHR which makes everything freeze... I mean, this is what happens before the freeze. Disabling Wayland driver fixes the issue, but I'm not going to play without Wayland driver anymore, I'd rather wait or stay with Proton-GE 10-25 until it's fixed.

Don't worry, I'll look into that bug as well, however I strongly suspect this partially is due to the experimental status of the wine-wayland backend and perhaps some unintended interactions with the proton patchset.

VK_SUBOPTIMAL_KHR is a hint for the client to re-allocate the swapchain and very much intentional and correct. What previously happened (and was not correct) was that the client got this error for every single present event, thrashing framerates and sometimes outright freezing clients.

You can work around this in the meantime by un-fullscreening and re-fullscreening with Super+F11.

gabriele2000:
> Don't worry, I'll look into that bug as well, however I strongly suspect this partially is due to the experimental status of the wine-wayland backend and perhaps some unintended interactions with the proton patchset.
> 
> VK_SUBOPTIMAL_KHR is a hint for the client to re-allocate the swapchain and very much intentional and correct. What previously happened (and was not correct) was that the client got this error for every single present event, thrashing framerates and sometimes outright freezing clients.

This is very interesting... thank you by the way, I'll keep waiting, this is not a huge problem since I'm using COSMIC since October 2023 and I saw the whole ecosystem mature.
It's been an amazing journey and it still is.

> You can work around this in the meantime by un-fullscreening and re-fullscreening with Super+F11.

Roger that, thank you!

PR #1931, opened 2025-12-26
fix: Prevent system hang on double Caps Lock press

### LINKED ISSUE #2819: Hitting Caps Lock twice hangs my system (Razer Blade Stealth 2022)
Usage of Caps Lock reliably hangs my system, and not just the desktop environment as once hung I am unable to even break out into headless mode  Ctrl + Alt + F4 - the system becomes entirely unresponsive.

The issue stems from "Caps Unlock" meaning I can hit the key once with no issue - the key's indicator light will turn on and the system will behave as expected. However, when hitting the key a second time the system will hang, sometimes with a "graphical noise" pattern.

The issue seems to be with the COSMIC Desktop Environment as I have not had this issue running different DEs on this system and the issue only happens in the graphical user interface - when using headless mode. 

Going into COSMIC Settings and swapping Caps Lock with Escape shifts the broken behavior migrates to the Escape key - hitting the Caps Lock any number of times no longer hangs the system but hitting Escape twice now hangs the system on the second press.

What this implies to me is that an issue with the COSMIC Desktop Environment and the software functionality of Caps Lock rather than an OS level issue or a hardware/driver issue with the physical Caps Lock key.

**To Reproduce**
1. Boot the system
2. Enter disk encryption password (Caps Lock may be used on this screen without issue)
3. Log in not required to replicate - the issue is present on the log in screen or post-log in screen
4. Hit Caps Lock to turn Caps Lock on
5. Hit Caps Lock again to turn Caps Lock off
6. The system is now unresponsive

**Expected behavior**
I would expect the system to turn Caps Lock off and continue to run and function normally.

**Desktop (please complete the following information):**
 - OS: Pop!_OS 24.04

ZeusCraft10:
Fixes #2819. The issue was caused by a recursive call to keyboard.input within the filter_keyboard_input function when Caps Lock was used as a virtual modifier. This patch moves the re-injection of input events to an idle callback to avoid the deadlock.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -1774,25 +1774,31 @@ impl State {
                 && handle.modified_sym() == Keysym::Caps_Lock
                 && (modifiers.serialized.locked & 2) != 0
             {
-                let serial = SERIAL_COUNTER.next_serial();
-                let time = self.common.clock.now().as_millis();
-                keyboard.input(
-                    self,
-                    event.key_code(),
-                    KeyState::Pressed,
-                    serial,
-                    time,
-                    |_, _, _| FilterResult::<()>::Forward,
-                );
-                let serial = SERIAL_COUNTER.next_serial();
-                keyboard.input(
-                    self,
-                    event.key_code(),
-                    KeyState::Released,
-                    serial,
-                    time,
-                    |_, _, _| FilterResult::<()>::Forward,
-                );
+                let seat = seat.clone();
+                let key_code = event.key_code();
+                self.common.event_loop_handle.insert_idle(move |state| {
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
Thanks, LGTM

jacobgkau (approved):
I've never been able to replicate https://github.com/pop-os/cosmic-epoch/issues/2819, but I'm not seeing any problems with this fix. I can still enable and disable Caps Lock multiple times, and the settings in Input Devices -> Keyboard, such as Caps Lock Key settings and setting the Compose Key to Caps Lock, still work.

Smoke tested on three systems (Intel/hybrid laptop, AMD laptop, and NVIDIA desktop).

Issue #1816, opened 2025-11-16
[Bug] Telegram media viewer started to crash on beta.6


cdtj:
[Telegram Desktop](https://github.com/telegramdesktop/tdesktop) client (is a messaging app) has a built-in media viewer (probably uses QT and OpenGL to render), which started to crash after upgrading to beta.6, downgrading to previous release of cosmic-comp resumes normal behavior.

Version that causes crash: `cosmic-comp 0.1~1763108062~24.04~1c74adf`
Previous version that I had in apt archives that works properly: `0.1~1743697847~24.04~a8d12f3`

Telegram itself doesn't produce any useful (from my opinion) logs, also, there is nothing interesting in syslogs.

More details in telegram-desktop issues: https://github.com/telegramdesktop/tdesktop/issues/30002

alissonlauffer:
Bisected to commit e129094bfb480f713a52a44b897395c3e9e5cb7c, so the problem is probably in smithay (https://github.com/Smithay/smithay/compare/eb45814...d743e1a).

Joomzie:
For the time being, unsetting `WAYLAND_DISPLAY` seems to work as a viable workaround. Just modify `Exec` in the .desktop for Telegram.
```
Exec=env -u WAYLAND_DISPLAY Telegram
```
I'm not sure if any other issues come from doing this, but so far so good. I'm able to consecutively open the media viewer without causing a crash, and unlike when setting `QT_QPA_PLATFORM="xcb"`, the clipboard still works.

cdtj:
tested locally all commits between beta.5 and beta.6 and found that this commit cause this unexpected behavior:

https://github.com/pop-os/cosmic-comp/commit/e129094bfb480f713a52a44b897395c3e9e5cb7c

hardBSDk:
This bug still happens in the stable version, Telegram Flatpak will crash in the second video player execution (open the video two times).

Probably I had this bug in GNOME Wayland, a migration to KDE Plasma fixed it.

beemdvp:
> For the time being, unsetting `WAYLAND_DISPLAY` seems to work as a viable workaround. Just modify `Exec` in the .desktop for Telegram.
> 
> ```
> Exec=env -u WAYLAND_DISPLAY Telegram
> ```
> 
[...]

PR #1932, opened 2025-12-26
fix: prevent protocol error for unparented popups (#2827)

### LINKED ISSUE #2827: Crash in Workspace::unminimize at workspace.rs:1178 after minimized-window restore
## Summary

`cosmic-comp` crashed with an `Option::unwrap()` panic at `src/shell/workspace.rs:1178`. The panel and applets then reported broken Wayland connections.

## Version

- Distribution: Pop!_OS
- Installed `cosmic-comp` package: `0.1~1788462011~24.04~f834412`
- Source commit: `f834412`

## Error

The following is the relevant journal message. Journal metadata has been removed.

```text
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value': src/shell/workspace.rs:1178
```

The logged backtrace did not resolve the application frames to symbols.

## Reproduction

There are no confirmed reproduction steps. The application and exact user action that triggered the crash are not established.

The failure location is in `Workspace::unminimize`, in the `MinimizedWindow::Floating` branch when `previous.was_maximized` is true. This identifies the code path; it is not a verified reproduction procedure.

## Possible relationship to #2337

Related issue: #2332. Related fix: #2337.

The installed source already contains the early return in `FloatingLayout::remap_minimized` when `!mapped.alive()`:

https://github.com/pop-os/cosmic-comp/blob/f834412/src/shell/layout/floating/mod.rs#L624

However, its caller still requests the window geometry with an unchecked `unwrap()` after that function returns:

https://github.com/pop-os/cosmic-comp/blob/f834412/src/shell/workspace.rs#L1178

If the helper returns without mapping the window, the caller appears able to receive `None` and panic. This may be a case not covered by #2337. The log does not establish that a dead window caused this incident.

The same caller remains in the checked upstream revision `0f56d214b9ceaa53ca12ef6ce7043e07626bb42c`:

https://github.com/pop-os/cosmic-comp/blob/0f56d214b9ceaa53ca12ef6ce7043e07626bb42c/src/shell/workspace.rs#L1178

## Expected behavior

An unsuccessful window restoration should not terminate the compositor. Please check how the helper communicates failure to its callers, including the previously maximized case.

ZeusCraft10:
This PR fixes a crash in Telegram (and potentially other clients) where creating a popup without a resolvable parent caused a Wayland protocol error. fix #2827 https://github.com/pop-os/cosmic-comp/issues/1816
    
    Change:
    - Modified src/wayland/handlers/xdg_shell/mod.rs to always configure and track popups, even if unconstrain_popup is skipped due to a missing parent.
    
    Fixes #2827

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4991,7 +4991,7 @@ checksum = "67b1b7a3b5fe4f1376887184045fcf45c69e92af734b7aaddc05fb777b6fbd03"
 [[package]]
 name = "smithay"
 version = "0.7.0"
-source = "git+https://github.com/smithay/smithay.git?rev=cb9acba#cb9acba2775de4f2032e023520e007cd66909eda"
+source = "git+https://github.com/smithay/smithay.git?rev=8148d67#8148d67ea3ee3959a84970e8f80f591e068e65ce"
 dependencies = [
  "aliasable",
  "appendlist",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -146,4 +146,4 @@ cosmic-protocols = { git = "https://github.com/pop-os//cosmic-protocols", branch
 cosmic-client-toolkit = { git = "https://github.com/pop-os//cosmic-protocols", branch = "main" }
 
 [patch.crates-io]
-smithay = { git = "https://github.com/smithay/smithay.git", rev = "cb9acba" }
+smithay = { git = "https://github.com/smithay/smithay.git", rev = "8148d67" }
--- a/src/backend/kms/device.rs
+++ b/src/backend/kms/device.rs
@@ -38,12 +38,13 @@ use smithay::{
         drm::control::{Device as ControlDevice, ModeTypeFlags, connector, crtc},
         gbm::BufferObjectFlags as GbmBufferFlags,
         rustix::fs::OFlags,
-        wayland_server::{DisplayHandle, Weak, protocol::wl_buffer::WlBuffer},
+        wayland_server::DisplayHandle,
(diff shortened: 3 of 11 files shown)


ZeusCraft10:
> The fix looks good, can you please separate out commit [db5392d](https://github.com/pop-os/cosmic-comp/commit/db5392df6bd0e51f59b2b0675d5b5e70cae252e5) for this PR?

I went ahead and did this. Thank you!

leviport (approved):

jacobgkau:
https://github.com/pop-os/cosmic-comp/pull/1975 fixed the same issues. It looks like this one doesn't conflict. Should it still be merged? (@Drakulix)

mikairyuu:
Just a quick note to help distinguish the issues:
I tested this PR while creating #1975, and didn't see any effect. This PR targets `xdg_popup` behavior, but the Telegram media viewer actually uses `xdg_toplevel` (which is why my fix in #1975 resolved it).
So, this PR likely fixes a different issue (perhaps related to context menus or tooltips) and shouldn't be considered a duplicate or alternative to the toplevel lifecycle fix.

Also, the linked https://github.com/pop-os/cosmic-epoch/issues/2827 can probably be closed as well now, I couldn't recreate it with #1975 merged and the issue looks very similar.

jacobgkau:
Confirmed with @leviport that this did not fix #2827 or #1816 in his testing, either. But we didn't find any regressions, so it's good to go if @Drakulix wants it.

PR #1934, opened 2025-12-26
fix(grab): prevent window jump on late drag start

### LINKED ISSUE #1645: Some apps jump on start of drag

ZeusCraft10:
Updates MoveGrab initialization to use the current pointer location if the client initiates the move request after a threshold. This prevents windows from jumping to the original click position when there is a delay between the click and the move request. Fixes #1645.


jacobgkau:
Thank you for working on this. Do you think this PR would still be useful now that https://github.com/pop-os/cosmic-comp/pull/1901 has fixed the same issue this one was targeting, or can this now be considered obsolete? It looks like this PR might've also gotten a couple of unrelated commits on accident, so we can just close it if it's no longer needed.

PR #1944, opened 2025-12-28
Fix/xwayland fullscreen panels

ZeusCraft10:
fix: Correctly hide panels for fullscreen Xwayland apps

This PR addresses issue #2840 where COSMIC panels (top and bottom bars) would remain visible when an Xwayland application (such as Proton games) entered fullscreen mode.

The issue was caused by the `has_focused_fullscreen` check in `src/shell/focus/order.rs` strictly checking for `KeyboardFocusTarget::Fullscreen`. However, Xwayland windows are often focused as `KeyboardFocusTarget::Element` even when they are in fullscreen state.

This fix updates the logic to check if the currently focused element corresponds to the active fullscreen surface, regardless of the wrapper variant (`Element` or `Fullscreen`).

Closes: https://github.com/pop-os/cosmic-epoch/issues/2840

## Verification
- Confirmed that panels hide correctly when launching Proton games (e.g., via Steam) in fullscreen mode.
- Verified that panels reappear upon exiting fullscreen or Alt-Tabbing.
- `cargo check` passes.


src/shell/focus/order.rs
```diff
@@ -128,10 +128,14 @@ fn render_input_order_internal<R: 'static>(
...
         let current_focus = seat.get_keyboard().unwrap().current_focus();
-        matches!(current_focus, Some(KeyboardFocusTarget::Fullscreen(_)))
-            || (current_focus.is_none()
-                && focus_stack_is_valid_fullscreen
-                && !workspace_overview_is_open(output))
+        current_focus.as_ref().is_some_and(|target| match target {
+            KeyboardFocusTarget::Fullscreen(s) => Some(s) == fullscreen.map(|f| &f.surface),
+            KeyboardFocusTarget::Element(m) => fullscreen
+                .is_some_and(|f| m.windows().any(|(s, _)| s == &f.surface)),
+            _ => false,
```
Drakulix, line 135: This is weird. I'd assume those windows are override-redirect windows otherwise they really shouldn't be `Element`s but fullscreen.

Drakulix (commented):
Thanks, looks like an okay fix.

However please drop the other commits (including the merge) and rebase 55de0764 against master.

PR #1946, opened 2025-12-30
fix: clear keyboard and pointer grabs on session lock

ZeusCraft10:
This ensures that the compositor doesn't stay in a stale grab state when the session is locked or focus is lost, which can cause clients like Firefox to become unresponsive (e.g. issues with toolbars and right-click menus after idle).

Closes: https://github.com/pop-os/cosmic-epoch/issues/2851


src/shell/focus/mod.rs
```diff
@@ -384,13 +384,27 @@ fn update_focus_state(
             }
         }
 
-        if target.is_some_and(|t| {
+        if target.is_none() || target.is_some_and(|t| {
```
Drakulix, line 387: This could be written more concisely with `is_none_or`, but I am not sure we want to test for `target.is_none()`. I understand the reasoning for the LockSurface, but not for no target.

src/shell/focus/mod.rs
```diff
@@ -384,13 +384,27 @@ fn update_focus_state(
...
+            // Also dismiss popups if we haven't already
+            if let Some(mut popup_grab) = seat
+                .user_data()
+                .get::<PopupGrabData>()
+                .and_then(|x| x.take())
+            {
+                if !popup_grab.has_ended() {
+                    popup_grab.ungrab(PopupUngrabStrategy::All);
+                }
+            }
```
Drakulix, line 407: Dismissing popups on exclusive layer shell surfaces might not be the worse idea, but I wouldn't want to force unsetting pointer grabs, which can include moving or resizing windows or drag&drop operations (which might target the new exclusive layer shell surface).

As such the pointer grab code would need to be special cased for just `LockSurface`.

src/shell/focus/mod.rs
ZeusCraft10: Good Catch, this is a good change.

src/shell/focus/mod.rs
ZeusCraft10: Thanks for the feedback! Both points make sense:
I'll remove the target.is_none() check - agreed that it's too aggressive for cases where focus is just transitioning.
I'll special-case pointer grab clearing for LockSurface only, since we don't want to interrupt window move/resize or drag-and-drop operations for regular exclusive layer surfaces.
While implementing this, I was thinking of restructuring slightly to avoid redundant pattern matching. Would something like this be preferred?    

`if let Some(t) = target {
    let is_lock_surface = matches!(t, KeyboardFocusTarget::LockSurface(_));
    let is_exclusive_layer = matches!(
        t,
        KeyboardFocusTarget::LayerSurface(layer_surface) 
        if layer_surface.cached_state().keyboard_interactivity == KeyboardInteractivity::Exclusive
    );

    // Clear keyboard grabs for lock surface or exclusive layer surfaces
    if is_lock_surface || is_exclusive_layer {
        keyboard.unset_grab(state);
    }

    // Only clear pointer grabs and popups for LockSurface (security critical)
    if is_lock_surface {
        if let Some(pointer) = seat.get_pointer() {
            pointer.unset_grab(state);
        }
        if let Some(mut popup_grab) = seat
            .user_data()
            .get::<PopupGrabData>()
            .and_then(|x| x.take())
        {
            if !popup_grab.has_ended() {
                popup_grab.ungrab(PopupUngrabStrategy::All);
            }
        }
    }
}`

This way the conditions are explicit and we only match once. Let me know if you'd prefer a different approach!

juehang:
Hi all, any progress on this PR? It is a significant QOL improvement I think.

PR #1949, opened 2025-12-30
Implement max_window_width

mike-lloyd03:
This PR aims to address the issue of tiled windows becoming too wide on ultrawide monitors as suggested in https://github.com/pop-os/cosmic-epoch/issues/410.

It exposes a new config variable called `max_window_width` which limits the maximum size of any tiled window. It will ignore any values below 300. This seems like a reasonable minimum as anything smaller than that will be basically useless.

There are a lot of solutions for handling tiling on ultrawide monitors such as alternate layouts. But this is the simplest solution that I know of and has served me well on other tiling WMs (LeftWM did this first to my knowledge).

Additionally, a similar feature was added to pop-shell earlier this year and the devs merged the change [with the note](https://github.com/pop-os/shell/discussions/1752#discussioncomment-11670796) that future work on pop-shell would likely cease and efforts to implement these features into cosmic would be appreciated.

PR to pop-shell: https://github.com/pop-os/shell/pull/1762

To set the configuration variable, users would need to create a file at `$XDG_CONFIG_HOME/cosmic/com.system76.CosmicComp/v1/max_window_width`. Example:

```bash
echo 2400 > ~/.config/cosmic/com.system76.CosmicComp/v1/max_window_width
```


maria-komarova:
Sounds like a good addition to the configuration file options for now. We might want to expose a way to set it in the UI in the future but that might need more thought first.

mike-lloyd03:
Looks like this is actually causing `cosmic-greeter` to crash. I'm not sure why. I'm not seeing anything in system journal. But I get a black screen with blinking underscore when booting up and using this build of `cosmic-comp` installed at `/usr/bin/cosmic-comp`.

JMPMarshall:
Hello, I recently starting using Cosmic on a laptop and am liking it a lot.  I use a 55" 4k tv as a monitor on Gnome with Tiling Shell on most of my machines presently.  I currently have the default Tiling Shell configuration set to split the screen into a 2x4 grid.  I normally view web pages and documents in portrait and put anything I need to view in landscape across 2 squares.  I generally place windows from the inside out and have it set to auto-tile new windows into the grid.  I don't need a grid but something that would allow me to put an upper limit on default Window size in both planes would be helpful.  I question if putting this in the form of a percentage of the screen space in each plane might be more intuitive.  I question if these settings could also be tied to specific displays or resolutions.  I would love for my laptop to function as Cosmic does by default now until I plug it into my dock and 4k monitor upon which it would remember the specific configuration for the larger screen.

Frosthaven:
My neck is very happy to see this PR!

I put together a  [per-display implementation](
https://github.com/pop-os/cosmic-comp/compare/master...Frosthaven:cosmic-comp:feature/single-tile-max-width). No PR currently because I don't want to step on toes, but just in case there is interest it's what I've been using.

image

```ron
// ~/.config/cosmic/com.system76.CosmicComp/v1/single_tile_max_widths

[
      (
          output: "DP-3",
          max_width: 2560,
      ),
]
```

I also implemented the setting in a hastily put together [update to cosmic-settings](https://github.com/pop-os/cosmic-settings/commit/7c4121896287e9067d8b7b388eb5b6cd0f7fb5bd) - though it can still work as a hidden option.

Feel free to use whatever is useful, ignore it, or ask me to do something about it separately!

PR #1976, opened 2026-01-09
xwm: Set xcursor variables in Xresources db

Drakulix:
It seems that there isn't a standardized XSETTINGS variable for the cursor size and not even all apps reliably read `XCURSOR_SIZE` (which we wouldn't be able to update anyway). 

The only dynamic source for X11 apps for this information seems to be the Xresources database. Additionally many apps only read the size from there, if there is also a theme set...

We really don't want to implement a serializer for this cursed database though (see https://github.com/psychon/x11rb/issues/941), so we just call out to `xrdb` (which e.g. kde-gtk-config also does).

Fixes the cursor sizes on scaled X11 apps (e.g. Steam or various Electron apps).

### DIFF
--- a/src/xwayland.rs
+++ b/src/xwayland.rs
@@ -1,4 +1,10 @@
-use std::{ffi::OsString, os::unix::io::OwnedFd, process::Stdio};
+use std::{
+    ffi::OsString,
+    io::Write,
+    os::unix::io::OwnedFd,
+    process::Stdio,
+    sync::mpsc::{self, Receiver, Sender},
+};
 
 use crate::{
     backend::render::cursor::{Cursor, load_cursor_env, load_cursor_theme},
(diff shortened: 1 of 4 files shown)


src/xwayland.rs
```diff
@@ -665,6 +661,28 @@ impl Common {
...
+                            cosmic::icon_theme::default(),
+                            (new_scale * cursor_size as f64).round() as i32,
+                        );
+                        if let Some(mut stdin) = child.stdin.take() {
+                            if let Err(err) = stdin.write_all(resources.as_bytes()) {
+                                warn!("Failed to update xresources: {}", err);
+                            }
+                        }
+                        // drop the child, letting it run til completion.
+                        // TODO: When child.pidfd is standardized we can add it to the runloop and check the exit code
```
mmstick: Should we spawn a thread to call `.wait()` on it? https://doc.rust-lang.org/stable/std/process/struct.Child.html#warning

Drakulix: Good catch, indeed we don't want any zombies. I updated the commit to use a little thread for the xrdb calls, where we can wait.

PR #1988, opened 2026-01-14
fix: more stable output arrangement for identical external displays

hojjatabdollahi:
Monitors connected to an external USB C dock may get assigned different connectors (DP-5, DP-6...) on every reconnection or when the OS sleeps depending on the order the kernel sees them. This causes the outputs to rearrange often. This commit removes the connector as a part of the hash key for the outputs if and only if all outputs are unique (when their EDID is considered). This way no matter where the monitor is connected or if the USB C dock randomly assigns a different connector to it, the arrangements of the outputs stays the same.

If the external monitors have identical EDID info, then their connector is taken into account as a fallback, similar to what exists in cosmic-comp now.

The `outputs.ron` now looks like this:
```
        [
            (
                make: "LG Electronics",
                model: "LG HDR 4K",
                edid: Some((
                    manufacturer: ('G', 'S', 'M'),
                    product: 30470,
                    serial: Some(178300),
                    manufacture_week: 6,
                    manufacture_year: 2024,
                    model_year: None,
                )),
            ),
            (
                make: "LG Electronics",
                model: "LG HDR 4K",
                edid: Some((
                    manufacturer: ('G', 'S', 'M'),
                    product: 30470,
                    serial: Some(178311),
                    manufacture_week: 6,
                    manufacture_year: 2024,
                    model_year: None,
                )),
            ),
            (
                make: "BOE",
                model: "0x095F",
                edid: Some((
                    manufacturer: ('B', 'O', 'E'),
                    product: 2399,
                    serial: None,
                    manufacture_week: 23,
                    manufacture_year: 2019,
                    model_year: None,
                )),
            ),
        ]: [
<details>
        ],
```

As opposed to something like this:
```
        [
            (
                connector: "DP-5",
                make: "LG Electronics",
                model: "LG HDR 4K",
            ),
            (
                connector: "DP-6",
                make: "LG Electronics",
                model: "LG HDR 4K",
            ),
            (
                connector: "eDP-1",
                make: "BOE",
                model: "0x095F",
            ),
        ]: [
<details>
        ],
```

~~I've marked this as a draft since I guess we need to update cosmic-greeter and cosmic-randr because cosmic-greeter also reads outputs.ron and applies it using cosmic-randr. But I'm not sure if that's true and would like your input.~~

More importantly I'm interested to know what the team thinks of this approach? To me it makes sense since workspaces are assigned using this exact data structure. And in case you have completely identical monitors, then it just revert back to using the connector id as we do now. So, no regressions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -924,6 +924,7 @@ dependencies = [
  "libdisplay-info",
  "ron 0.12.0",
  "serde",
+ "slotmap",
  "tracing",
 ]
 
@@ -4787,7 +4788,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.4.15",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -5758,7 +5759,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c5f7c95348f20c1c913d72157b3c6dee6ea3e30b3d19502c5a7f6d3f160dacbf"
 dependencies = [
  "cc",
- "windows-targets 0.52.6",
+ "windows-targets 0.48.5",
 ]
 
 [[package]]
@@ -6626,15 +6627,6 @@ name = "windows-sys"
 version = "0.52.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "282be5f36a8ce781fad8c8ae18fa3f9beff57ec1b52cb3de0789201425d9a33d"
-dependencies = [
- "windows-targets 0.52.6",
-]
-
-[[package]]
-name = "windows-sys"
-version = "0.59.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1e38bc4d79ed67fd075bcc251a1c39b32a1776bbe92e5bef1f0bf1f8c531853b"
 dependencies = [
  "windows-targets 0.52.6",
 ]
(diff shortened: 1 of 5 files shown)


hojjatabdollahi:
I have been using this for the past few days and it has fixed the annoying random assignment of outputs to monitors after each sleep, or every time I connect my USB C dock. 

This saves the config the way @Drakulix mentioned a while back: https://github.com/pop-os/cosmic-comp/pull/1575#issuecomment-3175638122. And is fairly human readable.

If this approach is satisfactory I can update other packages that need to be updated to use this format too.

hojjatabdollahi:
I noticed that cosmic-greeter have to update to avoid ruining the outputs.ron file. I did that last week and I've had a very nice experience since then. I moved the logic for matching the outputs to their config to cosmic-comp-config to avoid duplicating the logic. The last commit needs some clean up. But I'll do that if you guys are onboard with this approach. This has fixed the most annoying issue I've had with COSMIC since last year that I got these external monitors.

hojjatabdollahi:
I clean up this PR. It's two small commits now, a lot more readable. The code in [6724d1f](https://github.com/pop-os/cosmic-comp/pull/1988/commits/6724d1f623c703e0e70632db77e8dc608758d0a8) is mostly coming from cosmic-greeter.

It's ready for review.

garrettjwilke (approved):
the new `outputs.ron` file correctly distinguishes between similar displays.

nicholaspsmith:
Hey, just wanted to share a workaround I've been using for this exact issue while waiting for this PR to land.

I have a 4-display setup (two identical Dell P2722H monitors + an HP Z27u G3 + laptop panel) and the connector shuffle on every reboot was driving me nuts. The two Dells are indistinguishable by make/model alone, so cosmic-comp can't tell them apart.

I wrote a small Python script + systemd user service that:
1. Reads each connected display's EDID serial from `/sys/class/drm/`
2. Maps known serials to fixed left-to-right positions
3. Applies positions via `cosmic-randr` at login
4. Retries until positions stick (to win the race against cosmic-comp's config)

It's been working reliably for me across reboots. Published it here in case it helps others: https://github.com/nicholaspsmith/cosmic_restore_monitors

Looking forward to this PR landing so the workaround isn't needed anymore!

PR #1997, opened 2026-01-15
Rename bare-session files not to overwrite cosmic-session

hojjatabdollahi:
Running `make install-bare-session` breaks your system (login loop and other issues) because it overwrites `cosmic-session`'s files with outdated files. ~~Instead of keeping these files updated and in sync with cosmic-session, we should probably just remove them.~~ Renaming them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,14 +53,14 @@ install:
 	install -Dm0644 "data/tiling-exceptions.ron" "$(TILING_EXCEPTIONS_CONF)"
 
 install-bare-session: install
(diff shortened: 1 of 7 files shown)


Drakulix:
They aren't meant to be in sync, this is (intentionally) a standalone session of just cosmic-comp. But I agree the breakage isn't very nice nor intended, so we probably should rename the files.

hojjatabdollahi:
@Drakulix I see. I renamed the files instead of deleting them. Suggesting to call it "Cosmic compositor-only session".

PR #2019, opened 2026-01-22
WIP Touch input changes

ids1024:
Set keyboard focus on touch down event, and use touch popup grab.

Requires https://github.com/Smithay/smithay/pull/1911.

This seems to match the behavior of Sway, but could use a bit more testing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1230,7 +1230,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users",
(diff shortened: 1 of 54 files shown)


Quackdoc:
> Set keyboard focus on touch down event

can this be made optional, I myself and at least 2 other users prefer keyboard focus doesn't get sent, since this can allow waydroid to operate independently, I also found use for this since it means I can play my games without tabbing out

PR #2020, opened 2026-01-23
Fix: Display Output config being applied to wrong output

stoorps:
## Problem
Displays of the same make & model can confuse cosmic-randr & cosmic-comp, leading to the wrong configuration being applied. This presents as one display's rotation, scaling, position, etc. being applied to another. This occurs (at least in my experience) on ~50% of boots & re-connections of outputs.

## Cause
The order of Display adapters is not guaranteed upon boot/connection, and therefore the names that are derived from this order  (eg. `DP-1` or `HDMI-2`) are not actually tied to a physical port, and can switch. 

## Solution
When restoring configuration from KDL, try to match via `output.serial_number` first, and fallback to name matching if serials are not present. 

## Changes
 - Add `serial_number` to `outputs.ron` configuration
 - Backwards compatibility & upgrade for old `outputs.ron` files 

## Dependencies

- Requires https://github.com/pop-os/cosmic-randr/pull/53 

 
 ### Test Plan
- [x] Builds successfully (with cosmic-randr on branch `serial-number`) 
- [x] Old outputs.ron file is successfully migrated
- [x] Displays always configure correctly on boot (10/10 tries)
- [x] Displays always configure correctly on re-connection 
 
 ## Outstanding Decisions
 There is a decision to make around backwards compatibility. Do we:
1.  Support old `outputs.ron` files, meaning that users facing this issue will have to change/delete their display config for the fix to have any effect
2. Break compatibility and regenerate a default configuration, forcing all users to reconfigure their displays.
 
 I have opted for option 1 at the moment (see suggestion in mod.rs), but will defer to maintainers to make the final decision.

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -422,6 +422,7 @@ impl Config {
         // Backward-compatible lookup: older `outputs.ron` files won't have `serial_number`.
         // If a strict match fails, retry with serial numbers ignored.
         // This prevents bricking a users config and resetting to default. 
+        let mut used_fallback_key = false;
         let mut infos_key = infos.clone();
         let mut configs_ref_opt = outputs_config.config.get(&infos);
         if configs_ref_opt.is_none() {
@@ -433,6 +434,7 @@ impl Config {
             if let Some(cfgs) = outputs_config.config.get(&infos_no_serial) {
                 infos_key = infos_no_serial;
                 configs_ref_opt = Some(cfgs);
+                used_fallback_key = true;
             }
         }
 
@@ -484,6 +486,7 @@ impl Config {
             }
 
             let mut backend = backend.lock();
+            let mut applied_config_ok = false;
             if let Err(err) = backend.apply_config_for_outputs(
                 false,
                 loop_handle,
@@ -531,6 +534,7 @@ impl Config {
                     }
                 }
             } else {
+                applied_config_ok = true;
                 for (output, enabled) in found_outputs {
                     if enabled == OutputState::Enabled {
                         output_state.enable_head(&output);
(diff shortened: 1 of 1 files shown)


src/config/mod.rs
```diff
@@ -417,11 +417,28 @@ impl Config {
...
+            for info in &mut infos_no_serial {
+                info.serial_number = None;
+            }
+            infos_no_serial.sort();
+            if let Some(cfgs) = outputs_config.config.get(&infos_no_serial) {
+                infos_key = infos_no_serial;
+                configs_ref_opt = Some(cfgs);
+                used_fallback_key = true;
+            }
+        }
```
stoorps, line 439: Optional backwards compatibility. Can be removed and a default display configuration will be created.

src/config/mod.rs
```diff
@@ -524,6 +546,13 @@ impl Config {
 
             output_state.update();
             self.write_outputs(output_state.outputs());
+
+            // One-time migration: if we only found a config by ignoring serial numbers,
+            // persist the now-known serial-aware key and remove the old key.
+            if used_fallback_key && applied_config_ok {
+                let mut outputs_mut = self.dynamic_conf.outputs_mut();
+                outputs_mut.config.remove(&infos_key);
+            }
```
stoorps, line 555: Optional backwards compatibility. Can be removed and a default display configuration will be created.

ids1024:
> try to match via output.serial_number first, and fallback to name matching if serials are not present.

I have a pair of Viewsonic monitors that report the same serial number, but can be distinguished by manufacture date. So I don't think comparing just serial number will work. We should compare all the EDID info.

We already have an `EdidProduct` type in cosmic-comp that is used to determine which output a workspace belongs to (when monitors are removed and re-attached), and is persisted for pinned outputs. We probably want to use that in the output configuration as well.

(KDE, for comparison, just MD5 hashes the whole EDID data, and compares that, which serves the same purpose.)

hojjatabdollahi:
@ids1024 my PR also is trying to solve this problem, and it uses the whole EDID. Would you mind taking a look:
https://github.com/pop-os/cosmic-comp/pull/1988

stoorps:
Apologies @hojjatabdollahi! I didn't see your PR otherwise I wouldn't have submitted mine.

@ids1024 That is a strange thing for them to do, it kind of defeats the purpose of a serial :laughing: 

I'll close this and my cosmic-randr change down now. I'm on the popos mattermost under the same user if you want anything :)

PR #2038, opened 2026-01-30
layout/floating: allow size altering if window was maximized

### LINKED ISSUE #1984: If I unmaximize a window by dragging its titlebar, it becomes the maximized size again instead of the older unmaximized size
What the title says, there is a video of the issue:

https://github.com/user-attachments/assets/dce5a8b2-2307-4492-a79a-4c30fdaa316e

I'm pretty sure this issue started happening after today's update to PopOS 24.04, as I've never seen this issue before

thegabriele97:
This should fix #1984 .

I took as reference edits made in 612ff2f.

https://github.com/user-attachments/assets/7900567d-c465-474e-8238-01ac0faa1a48

### DIFF
--- a/src/shell/element/surface.rs
+++ b/src/shell/element/surface.rs
@@ -45,7 +45,7 @@ use smithay::{
         IsAlive, Logical, Physical, Point, Rectangle, Scale, Serial, Size, user_data::UserDataMap,
     },
     wayland::{
-        compositor::{self, SurfaceData, TraversalAction, with_states, with_surface_tree_downward},
+        compositor::{SurfaceData, TraversalAction, with_states, with_surface_tree_downward},
         seat::WaylandFocus,
         shell::xdg::{
             SurfaceCachedState, ToplevelCachedState, ToplevelSurface, XdgToplevelSurfaceData,
@@ -184,17 +184,15 @@ impl CosmicSurface {
 
     pub fn last_server_size(&self) -> Option<Size<i32, Logical>> {
         match self.0.underlying_surface() {
-            WindowSurface::Wayland(toplevel) => {
-                compositor::with_states(toplevel.wl_surface(), |states| {
-                    let attributes = states
-                        .data_map
-                        .get::<XdgToplevelSurfaceData>()
-                        .unwrap()
-                        .lock()
-                        .unwrap();
-                    attributes.current_server_state().size
-                })
-            }
+            WindowSurface::Wayland(toplevel) => with_states(toplevel.wl_surface(), |states| {
+                let attributes = states
+                    .data_map
+                    .get::<XdgToplevelSurfaceData>()
+                    .unwrap()
+                    .lock()
+                    .unwrap();
+                attributes.current_server_state().size
+            }),
             WindowSurface::X11(_) => None,
         }
     }
--- a/src/shell/layout/floating/mod.rs
+++ b/src/shell/layout/floating/mod.rs
@@ -682,7 +682,9 @@ impl FloatingLayout {
                 mapped_geometry.size = last_size;
             }
         } else if !window.is_maximized(true) {
-            if window.active_window().has_pending_changes() && let Some(pending_size) = window.pending_size() {
+            if window.active_window().has_pending_changes()
+                && let Some(pending_size) = window.pending_size()
+            {
                 mapped_geometry.size = pending_size.as_local();
             } else if let Some(server_size) = window.active_window().last_server_size() {
                 mapped_geometry.size = server_size.as_local();


rhessus:
Tests: OK 👍

rhessus:
@leviport do you think it would be possible to take a look at this PR ? It’s very disruptive on a daily basis. I’ve been working with this commit at work for a week, and it fixes the problem well. Thanks in advance.

thegabriele97:
The CI failed, I guess I need to fix the code formatting, am I right?

leviport:
It only failed on Jammy, and cosmic-comp doesn't build for Jammy. That can be ignored.

leviport (approved):
Seems to be working very nicely, thanks!

src/shell/layout/floating/mod.rs
```diff
@@ -686,11 +686,11 @@ impl FloatingLayout {
                 mapped_geometry.size = last_size;
             }
         } else if !window.is_maximized(true) {
-            if window.active_window().has_pending_changes() {
+            if window.active_window().has_pending_changes() || window.active_window().was_maximized() {
```
Drakulix: Sorry, but this is a hacky workaround. The bigger question is, why `has_pending_changes` isn't returning `true` here.

After investigating it seems like the check introduced in https://github.com/pop-os/cosmic-comp/commit/612ff2f523580af2729cc7de1e73c334ea135726 isn't entirely right.

`has_pending_changes` only tests if there are changes not yet *send* to the client. But `element_geometry` returns the last acked state.

Meaning in the case of un-maximizing we send a new size to the client, but because it hasn't acknowledged it yet, we end up storing the old size in `last_geometry`.

So the correct fix would be to do something like (pseudo-code):

```
if !window.active_window().toplevel.pending_configures().is_empty() {
    mapped_geometry.size = window.active_window().toplevel.current_server_state().size;
}
```

I don't have a good idea how this would interact with the original issue in https://github.com/pop-os/cosmic-comp/issues/1819 though. @ids1024 Do you remember the exact configure sequence there in question?

thegabriele97: EDIT: ok I think I was able to reproduce the issue but I need your help here because it is hard for me to reproduce it systematically. I'll push asap. Thanks!

-------------------------
I went ahead and implemented your suggestion (second snippet). To check for regressions against #1819, I temporarily reverted the code to how it was before the commit that originally fixed #1819 (first snippet), so I could compare the old behavior against my new changes.

However, I wasn't able to reproduce #1819 with the old code either — and on top of that, the issue this PR is trying to fix also didn't appear. So I'm not confident my test environment is reliable enough to verify regressions.

Let me know if you'd like me to push the new code (second snippet) so you can test it on your end.

**First snippet:**
```rust
} else if !window.is_maximized(true) {
    if let Some(pending_size) = window.pending_size() {
        mapped_geometry.size = pending_size.as_local();
    }
    *window.last_geometry.lock().unwrap() = Some(mapped_geometry);
}
```

**Second snippet:**
```rust
} else if !window.is_maximized(true) {
    if window.active_window().has_pending_changes() && let Some(pending_size) = window.pending_size() {
        mapped_geometry.size = pending_size.as_local();
    } else if let Some(server_size) = window.active_window().unacked_server_size() {
        mapped_geometry.size = server_size.as_local();
    }
    *window.last_geometry.lock().unwrap() = Some(mapped_geometry);
}
```

src/shell/element/surface.rs
```diff
@@ -182,6 +186,31 @@ impl CosmicSurface {
         }
     }
 
+    pub fn unacked_server_size(&self) -> Option<Size<i32, Logical>> {
```
Drakulix: I'd prefer to name this something like `last_server_size` and always return the `current_server_state`. This might be equal to the last acked state, but that is of no concern to the floating-layout code. It always wants to last configured size.

I expect the same to be true in other places. Code either needs the pending_size of the last_server_size.

thegabriele97:
> LGTM
> 
> Does this still solve the issue in your testing?

Yep it solve the minimize bug (the object of this PR). However I can't test the bug mentioned in #1819.

src/shell/element/surface.rs
thegabriele97: Ok thanks, tomorrow I'll remove the TODO comments (I was waiting for your approval first) and I'll modify the code according to your comment.

> I expect the same to be true in other places. Code either needs the pending_size of the last_server_size.

Does this need to be part of this PR?

thegabriele97:
@Drakulix done! :+1:

thegabriele97:
CI failed. Should I split the if condition into two lines?

Drakulix:
> CI failed. Should I split the if condition into two lines?

Just run `cargo fmt`.

src/shell/element/surface.rs
```diff
@@ -45,7 +45,7 @@ use smithay::{
         IsAlive, Logical, Physical, Point, Rectangle, Scale, Serial, Size, user_data::UserDataMap,
     },
     wayland::{
-        compositor::{SurfaceData, TraversalAction, with_states, with_surface_tree_downward},
+        compositor::{self, SurfaceData, TraversalAction, with_states, with_surface_tree_downward},
```
git-f0x: Really minor, but this might not be needed, since `with_states` is already imported.

PR #2041, opened 2026-01-30
fix/cursor_floows_focus_setting_different_window

Kikanon:
This change ensures that changing focus to a different monitor respects the cursor_follows_focus setting.

Related issue: https://github.com/pop-os/cosmic-epoch/issues/2233


Lcstyle:
I just don't understand what's taking so long to merge a fix that is clearly the root cause and clean to merge.  thank you for the fix @Kikanon i am experiencing this issue.

leviport:
@Lcstyle relax, we're a small team with a lot going on.

leviport (approved):

Kikanon:
I can't see why arm and amd builds failed, anything else i should do @leviport ?

leviport:
cosmic-comp doesn't build for Jammy. Those can be ignored.

Drakulix (approved):
Thanks!

BlakeGardner:
Just noticed this bug is fixed for me on Pop OS after updating. Thank you!

gordan-ramsey:
the issue resurfaced after the round of updates that came with kernel 7.0

PR #2063, opened 2026-02-06
kms: Import on device supporting format, if not advertised device

ids1024:
Fixes https://github.com/pop-os/cosmic-epoch/issues/2978.

This reverses the part of https://github.com/pop-os/cosmic-comp/commit/ca00df0b37 that made it only try import on the advertised GPU. But this version avoids initializing an EGL context simply to re-check the supported texture formats.


Drakulix (approved):
LGTM, while we wait for v6

Issue #642, opened 2024-07-22
Shortcuts with Letter Keys don't work when non Latin based keyboard layout is active


quendimax:
Shortcuts with letter keys (e.g. `Super+Q`, `Super+T`) don't work when you use a non Latin based keyboard layout.

I suppose, it is caused by using Unicode instead of Keycode of pressed keys. Actually it is common issue of some rust GUI libraries (at least slint, iced, and perhaps gpui).

If my guess is true, it means that even with some Latin based keyboard layouts, some shortcuts can be activated with different key combinations (for example French _AZERTY_ layout).

Drakulix:
Hmm the matching is done through `libxkbcommon` which should gracefully handle non-latin keyboards and is used by other desktops as well.

That said we do currently convert back from text-descriptions of the keys to a keysym, instead of a keycode. (which I am also not entirely sure is possible in that direction.)

We should be saving keycode obviously, but making those configurable through the UI, requires functionality, where cosmic-settings can capture the keycode entered by a user, so we only have to convert from codes to text descriptions and never vise-versa.

That however will probably not happen for the first alpha release, as it requires more integration with the compositor (possibly through some additional protocol).

justF0T0N:
Any updates? Will it be fixed soon?

Jaymz665:
I'm not sure, but apparently this also opens up another bug, if you run, for example, the same "World of warcraft" in wine and use a non-Latin keyboard, you will see letters like Ü ü, Ö ö, Ä ä instead of those that you wrote. There was such a bug many years ago in Wine itself, then this bug was observed in Gnome, but that was a few years ago, I'm unlikely to be able to find links.

PR #2067, opened 2026-02-07
fix(keyboard): Fix keyboard shortcuts for non-Latin layouts

mayakwd:
Fixes https://github.com/pop-os/cosmic-comp/issues/642

Use `raw_latin_sym_or_raw_current_sym()` as a fallback when matching shortcuts, so bindings like `Super+Q` and `Ctrl+J` work regardless of active keyboard layout (e.g., Russian, Ukrainian).

The approach follows the GNOME/Mutter approach of resolving keysyms at matching time. 
Sway/i3 takes a [different approach](https://github.com/swaywm/sway/pull/3058) by translating keysyms to keycodes at config load time, which is more invasive but avoids per-event lookups.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -1581,6 +1581,13 @@ impl State {
         handle: KeysymHandle<'_>,
         serial: Serial,
     ) -> FilterResult<Option<(Action, shortcuts::Binding)>> {
+        // Pre-compute for layout-agnostic shortcut matching
+        let raw_syms = handle.raw_syms();
+        let latin_sym = handle.raw_latin_sym_or_raw_current_sym();
+        let key_matches = |binding_key: Keysym| -> bool {
+            raw_syms.contains(&binding_key) || latin_sym.is_some_and(|sym| sym == binding_key)
+        };
+
         let mut shell = self.common.shell.write();
 
         let keyboard = seat.get_keyboard().unwrap();
@@ -1632,7 +1639,7 @@ impl State {
                 || (action_pattern.modifiers.logo && !modifiers.logo)
                 || (action_pattern.modifiers.shift && !modifiers.shift)
                 || (action_pattern.key.is_some()
-                    && handle.raw_syms().contains(&action_pattern.key.unwrap())
+                    && key_matches(action_pattern.key.unwrap())
                     && event.state() == KeyState::Released)
             {
                 shell.set_overview_mode(None, self.common.event_loop_handle.clone());
@@ -1651,7 +1658,7 @@ impl State {
         if let Some(action_pattern) = shell.resize_mode().0.active_binding() {
             if action_pattern.key.is_some()
                 && event.state() == KeyState::Released
-                && handle.raw_syms().contains(&action_pattern.key.unwrap())
+                && key_matches(action_pattern.key.unwrap())
             {
                 shell.set_resize_mode(
                     None,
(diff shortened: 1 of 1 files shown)


deemoun:
Thanks for fixing that. When this fix is supposed to land for 24.04 users?

EvgeniyGRep:
Thanks for fixing the issue globally, but it seems the same problem still occurs in the Cosmic Terminal, Cosmic Files, and possibly other Cosmic apps.

For example:

In Cosmic Terminal, pasting via Ctrl + Shift + V works only when using the English keyboard layout.

Similarly, in Cosmic Files, Ctrl + C and Ctrl + V behave the same way.

deemoun:
Same. Still seeing it in Cosmic Edit

mayakwd:
> Thanks for fixing the issue globally, but it seems the same problem still occurs in the Cosmic Terminal, Cosmic Files, and possibly other Cosmic apps.
> 
> For example:
> 
> In Cosmic Terminal, pasting via Ctrl + Shift + V works only when using the English keyboard layout.
> 
> Similarly, in Cosmic Files, Ctrl + C and Ctrl + V behave the same way.

Thank you for the feedback! Yes, I also encountered this problem. Things are getting a little more complicated here.

**Explanation:**
I fixed the problem for the cosmic-comp (the server in Wayland terms), but client applications are on the other side of the protocol. 

Client applications, such as cosmic-term or cosmic-edit, use `Keybind::matches` from libcosmic, which should match using a non-localized (default Latin layout) Key, but receive only a localized Key from iced (for example, "м" instead of "v").

**Possible fix in libcosmic:**
We can make changes to libcosmic and make minor changes to all applications that use libcosmic. The change is not ideal either, but it covers, let's say, 95% of issues:

- Mark `Keybind::matches` as obsolete
- Add the `Keybind::matches_event(&self, event: &keyboard::Event)` method to `KeyBind` and update all applications to use this method. Under the hood, the method will use `physical_key` from the event and match it to the keybind.

**Possible fix in iced-rs:**
There is another, more universal way, but it requires changes in iced to also return a `latin_keysym` in the event. Which will bring us to the `keybind.matches(modifiers, &key) || keybind.matches(modifies, &latin_key)` so we'll be able to check and localized and non-localized keys. 

However, none of these methods are universal and require intervention into the upstream code. Therefore, I would really like some help from the maintainers in designing a solution and choosing the least evil. 

The ideal option for cosmic-native developers would be to fix it in libcosmic, but this would create technical debt everywhere, right down to smithay-client-toolkit.

For some unknown reason, the shortcut solution is based on the user's current locale rather than physical keys. 

@Drakulix, @garrettjwilke, and @leviport, you were active on this PR, so I'd really appreciate your help.

Drakulix:
> @Drakulix, @garrettjwilke, and @leviport, you were active on this PR, so I'd really appreciate your help.

I am not too familiar with the exact internal workings here, pinging @wash2 who should be able to guide you better in contributing a fix.

So conceptually from a high-level perspective, this should be solvable exactly the same way:
- wayland-clients receive the keymap and raw keycodes from the compositor
- they can thus use xkbcommon in the same way to get a non-localized keysym for shortcuts

Of course the various abstractions are what makes it difficult here:
- So the first question is whether `winit` exposes enough to implement this correctly.
- Which is to say:
  - either the keycode + keymap (imo the correct design)
  - or the non-localized/latin keysym (weird choice, but okay?)
- If not that needs to be fixed, but I would be surprised and your analysis also doesn't imply this.

The next question is whether iced:
- is forwarding that directly.
- is only forwarding localized-keycodes.
  - and in this case whether they are aware of this limitation and want to fix it or not.

Imo this should be fixed in `iced`, as a serious toolkit needs to support shortcuts on non-latin keymaps. If upstream turns out difficult to work with here, this could be added to our iced-fork used by libcosmic first.

mayakwd:
Great questions! I've created a [PR in libcosmic](https://github.com/pop-os/libcosmic/pull/1144) with a more detailed explanation and a potential fix. Which still requires migration from the cosmic-apps (which is quite trivial).

PR #2069, opened 2026-02-07
fix: Remove redundant configure in popup reposition_request

### LINKED ISSUE #1971: cosmic-comp[2058]: Client bug: Unable to re-configure repositioned popup.
Randomly happens when I hover the `Dock`  and move the mouse.

mayakwd:
Fixes: https://github.com/pop-os/cosmic-comp/issues/1971, https://github.com/pop-os/cosmic-epoch/issues/3151

`send_repositioned()` already sends a configure event internally via `send_configure_internal()`. The extra `send_configure()` call right after either sends a duplicate configure (for reactive popups) or fails with a misleading "Client bug: Unable to re-configure repositioned popup" warning (for non-reactive popups).

### DIFF
--- a/src/wayland/handlers/xdg_shell/mod.rs
+++ b/src/wayland/handlers/xdg_shell/mod.rs
@@ -164,12 +164,6 @@ impl XdgShellHandler for State {
 
         self.common.shell.read().unconstrain_popup(&surface);
         surface.send_repositioned(token);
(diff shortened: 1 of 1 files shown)


Drakulix (approved):

leviport (approved):
Not seeing any regressions after this change. Thanks!

Issue #2071, opened 2026-02-08
cosmic-comp panic: "Placeholder group without real window?" when moving window to tile


jmhunter83:
Summary
cosmic-comp crashes with a panic when attempting to move a window into a tiling position, causing complete desktop session restart.
 Steps to Reproduce
1. Open an application window
2. Attempt to move the window into a tile position
3. Desktop crashes immediately
 Error
thread 'main' panicked at 'Placeholder group without real window?': src/shell/layout/tiling/mod.rs:3623
## Environment
- **OS:** Fedora Linux 43 (COSMIC Edition)
- **Kernel:** 6.18.8-200.fc43.x86_64
- **COSMIC Version:** 1.0.4-1.fc43
- **Hardware:** Lenovo Yoga 7 2-in-1 14IML9
- **GPU:** Intel Meteor Lake-P
## Impact
- Occurred twice within 15 minutes
- Complete desktop session crash
- All unsaved work lost
- 18 core dumps generated
## Crash Logs
Feb 07 19:49:12 cosmic-comp8944: thread 'main' panicked at 'Placeholder group without real window?': src/shell/layout/tiling/mod.rs:3623
Feb 07 19:49:12 cosmic-session8870: cosmic-comp exited with error code 101
Reproducible when tiling windows. All COSMIC components at version 1.0.4-1.fc43.

caniko:
## Additional occurrence: NVIDIA GPU + Dual Monitor Setup

Experiencing the same panic on a different configuration:

**Environment:**
- **OS:** NixOS
- **cosmic-comp:** 1.0.5
- **GPU:** NVIDIA GTX 1080 Ti (proprietary driver 580.126.09)
- **Setup:** Dual monitors (DP-1 + HDMI-A-2), autotile enabled globally, per-output workspaces

**Crash:**
```
Feb 11 21:24:35 murph .cosmic-comp-wrapped[2221]: thread 'main' panicked at 'Placeholder group without real window?': src/shell/layout/tiling/mod.rs:3623
     7: cosmic_comp::shell::layout::tiling::TilingLayout::update_pointer_position::{{closure}}
     8: cosmic_comp::shell::layout::tiling::TilingLayout::update_pointer_position
     9: cosmic_comp::shell::Shell::update_pointer_position
    10: cosmic_comp::input::<impl cosmic_comp::state::State>::process_input_event
```

**Impact:**
[...]

PR #2077, opened 2026-02-10
layout/tiling: Fix panic from tiling placeholders not being cleaned up

ids1024:
Reproduction steps:
- Run `cosmic-comp`
- Open a terminal
- Make workspace tiling
- Drag terminal to other output
- Make that workspace tiling too
  * The window now have an empty placeholder next to it
- Try to drag window again
  * Compositor panics

The particular instance I've seen of this seems to start with fa88002ba41d2edec25dd7ffdee9719fbb928fc0, after 67d0a825d1ab16df866219329b012994bb8688f5 fixed another case of the same panic. It seems simplest to expand the call to `cleanup_drag()` added there to run unconditional.

I wonder if there's a better way to to things that would be less fragile and not require such care about making sure `cleanup_drag()` is called in all the right places... but anyway, this also has a change that should make that extremely low cost to call, and not unnecessarily push an extra tree.

https://github.com/pop-os/cosmic-comp/issues/2071 is the same panic, though the description there doesn't seem quite detailed enough to tell if it means the same circumstances. If it's a slightly different edge case, this may or may not also fix it.

### DIFF
--- a/src/shell/grabs/moving.rs
+++ b/src/shell/grabs/moving.rs
@@ -906,19 +906,21 @@ impl Drop for MoveGrab {
                         }
                     }
                 } else {
-                    let mut shell = state.common.shell.write();
-                    shell
-                        .workspaces
-                        .active_mut(&cursor_output)
-                        .unwrap()
-                        .tiling_layer
-                        .cleanup_drag();
-                    shell.set_overview_mode(None, state.common.event_loop_handle.clone());
                     None
                 }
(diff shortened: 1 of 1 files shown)


garrettjwilke (approved):
works with the following setup:
open terminal, set tiling
drag terminal to external display, then enable tiling on the external display
attempt to drag terminal on newly tiled external display -> works

PR #2079, opened 2026-02-10
fix(dbus): handle PrepareForSleep signal for clean suspend/resume

### LINKED ISSUE #2078: [BUG] cosmic-comp does not handle PrepareForSleep D-Bus signal — causes NVIDIA suspend failures and 60s freeze timeout
### Summary

cosmic-comp does not listen for the `org.freedesktop.login1.Manager.PrepareForSleep` D-Bus signal. During system suspend, this causes a race condition where the NVIDIA driver revokes DRM access while cosmic-comp is still actively rendering. The compositor enters a tight error-retry loop of failing page flips, which causes a 60-second `user.slice` freeze timeout that delays or prevents proper S3 entry.

Other major Wayland compositors (Mutter, KWin) subscribe to `PrepareForSleep` and stop rendering before the NVIDIA driver takes control.

### Environment

- **OS:** Fedora 43, COSMIC Desktop (alpha 6)
- **Kernel:** 6.18.7 / 6.18.8
- **GPU:** NVIDIA GeForce RTX 3090 Ti
- **Driver:** 580.119.02 (Open kernel module)
- **Session:** Wayland (cosmic-comp)
- **Sleep mode:** S3 deep

### NVIDIA Module Configuration

```
options nvidia_drm modeset=1 fbdev=1
options nvidia NVreg_PreserveVideoMemoryAllocations=1 NVreg_TemporaryFilePath=/tmp
```

### What Happens

1. User triggers `systemctl suspend`
2. `nvidia-suspend.service` runs (ordered `Before=systemd-suspend.service`)
3. `nvidia-sleep.sh` executes `chvt 63` then `echo "suspend" > /proc/driver/nvidia/suspend`
4. NVIDIA revokes DRM access from userspace
5. **cosmic-comp is still rendering** — gets "Permission denied" errors
6. Failed page flips trigger immediate retry via `queue_redraw(true)` with no backoff
7. `systemd-suspend.service` starts — tries to freeze `user.slice`
8. Freeze times out after 60 seconds because cosmic-comp is in a bad state
9. System either enters a degraded suspend or fails to suspend entirely

### Journal Evidence

```
Feb  9 22:42:36 ICESWORD systemd-logind[6634]: The system will suspend now!
Feb  9 22:42:36 ICESWORD systemd[1]: Reached target sleep.target - Sleep.
Feb  9 22:42:36 ICESWORD systemd[1]: Starting nvidia-suspend.service - NVIDIA system suspend actions...

Feb  9 22:42:36 ICESWORD cosmic-comp[7334]: Failed to submit rendering: Failed to submit result for display
    Caused by:
        0: The underlying drm surface encountered an error: DRM access error: Page flip commit failed on device `Some("/dev/dri/card1")` (Permission denied (os error 13))
        1: DRM access error: Page flip commit failed on device `Some("/dev/dri/card1")` (Permission denied (os error 13))
        2: Permission denied (os error 13)

Feb  9 22:42:36 ICESWORD cosmic-comp[7334]: Failed to submit rendering: [repeated 3x]

Feb  9 22:42:41 ICESWORD systemd[1]: Finished nvidia-suspend.service - NVIDIA system suspend actions.
Feb  9 22:42:41 ICESWORD systemd[1]: Starting systemd-suspend.service - System Suspend...
Feb  9 22:43:41 ICESWORD systemd-sleep[65872]: Failed to freeze unit 'user.slice': Connection timed out
Feb  9 22:43:41 ICESWORD systemd-sleep[65872]: Performing sleep operation 'suspend'...
```

Note: No kernel `PM: suspend entry (deep)` message appears — the system may never have fully entered S3.

### Root Cause Analysis

After investigating the cosmic-comp source, I identified three contributing issues:

#### 1. No `PrepareForSleep` D-Bus signal handling (primary cause)

cosmic-comp does not subscribe to `org.freedesktop.login1.Manager.PrepareForSleep`. The only logind integration is for lid switch inhibition (`src/dbus/logind.rs`), which calls `ManagerProxy::inhibit(HandleLidSwitch, ...)` — nothing related to system suspend.

cosmic-comp *does* have a session pause/resume mechanism via libseat (`src/backend/kms/mod.rs:101-119`), which correctly handles `SessionEvent::PauseSession` by suspending input, DRM devices, and render surfaces (`src/backend/kms/mod.rs:419-431`). However, this only fires on VT switching, not on system suspend.

**How other compositors handle this:**

| Compositor | PrepareForSleep? | Source |
|------------|-----------------|--------|
| **Mutter** | Yes | `src/backends/meta-backend.c` — `prepare_for_sleep_cb()`, registered via `g_dbus_connection_signal_subscribe()` |
| **KWin** | Yes | `src/logind.cpp` — logind class handler |
| **Sway** | Delegated | Handled by `swayidle` daemon rather than compositor core |
| **cosmic-comp** | **No** | Only lid switch inhibition in `src/dbus/logind.rs` |

#### 2. Race between `chvt 63` and NVIDIA GPU takeover

The NVIDIA suspend mechanism is handled by `/usr/bin/nvidia-sleep.sh`, shipped by the `xorg-x11-drv-nvidia-power` RPM package (Fedora's packaging of NVIDIA's official power management scripts). This is not a local workaround — it is NVIDIA's documented suspend mechanism. The relevant excerpt:

```bash
# From /usr/bin/nvidia-sleep.sh (nvidia-sleep.sh, xorg-x11-drv-nvidia-power-580.119.02)
suspend|hibernate)
    mkdir -p "${RUN_DIR}"
    fgconsole > "${XORG_VT_FILE}"
    chvt 63
    if [[ $? -ne 0 ]]; then
        exit $?
    fi
    echo "$1" > /proc/driver/nvidia/suspend
    ;;
```

**What `chvt 63` does:** Linux supports 63 virtual terminals (VT1–VT63). The Wayland session typically runs on a low VT (VT1 or VT2). `chvt 63` switches to an unused/empty terminal, which forces the current session to lose DRM master status — this is how libseat knows to emit `SessionEvent::PauseSession`. After the VT switch, NVIDIA writes `"suspend"` to `/proc/driver/nvidia/suspend` to save VRAM and power down the GPU. On resume, `RestoreVT()` switches back to the original VT, restoring DRM master.

**The race:** `chvt 63` and the `/proc` write happen back-to-back in the same script with no wait for the session to actually process the VT switch. The `chvt 63` triggers an async D-Bus notification chain (kernel → logind → libseat → cosmic-comp's calloop), but NVIDIA takes exclusive GPU control synchronously on the very next line. cosmic-comp never gets a chance to process the pause event.

The `chvt 63` *should* eventually trigger `SessionEvent::PauseSession` through the libseat chain, which would call `pause_session()` and properly suspend all render surfaces (`src/backend/kms/surface/mod.rs:666-686`). However, NVIDIA writes to `/proc/driver/nvidia/suspend` immediately after `chvt`, taking exclusive GPU control before the async session pause event propagates through the event loop. By the time cosmic-comp could process the pause event, DRM access has already been revoked.

#### 3. Render loop retries failed DRM commits with no backoff

When `redraw()` fails at `src/backend/kms/surface/mod.rs:953-956`:
```rust
if let Err(err) = state.redraw(estimated_presentation) {
    let name = state.output.name();
    warn!(?name, "Failed to submit rendering: {:?}", err);
    state.queue_redraw(true);  // Immediate retry, no backoff
}
```

Failed page flips immediately schedule another render attempt via `queue_redraw(true)`. There is no check for `EACCES`/`Permission denied`, no exponential backoff, and no circuit breaker. The `redraw()` function also does not check `self.active` before attempting DRM operations — it only checks `self.compositor.as_mut()` (`src/backend/kms/surface/mod.rs:992`). This creates a tight loop of failing DRM commits during the window between NVIDIA taking control and the system entering suspend.

#### Fix 1: Subscribe to `PrepareForSleep` (recommended)

Add a D-Bus signal subscription to `org.freedesktop.login1.Manager.PrepareForSleep` in the KMS backend initialization. When `PrepareForSleep(true)` is received (suspend imminent), call the existing `pause_session()`. When `PrepareForSleep(false)` is received (resume), call `resume_session()`.

The infrastructure already exists — `logind-zbus` is already a dependency (optional, feature-gated as `systemd`), and the `pause_session()`/`resume_session()` handlers properly suspend and resume all render surfaces, input devices, and DRM devices. This is straightforward to wire up.

#### Fix 2: Add backoff/circuit breaker for DRM errors

When `redraw()` returns an error, especially `EACCES`, the retry should use exponential backoff or stop entirely rather than immediately rescheduling. This would prevent the tight error loop that contributes to the freeze timeout.

#### Fix 3: Check session active state in render path

The `redraw()` function could check `self.active.load(Ordering::SeqCst)` before attempting DRM operations. Currently `active` is set to `false` during `suspend()` (`src/backend/kms/surface/mod.rs:667`), but `redraw()` does not check this flag. Similarly, `queue_redraw()` could skip scheduling if the surface is inactive.

### Current End-User Workaround

To work around this issue, I have created systemd drop-in overrides that send `SIGSTOP` to cosmic-comp before NVIDIA takes GPU control, and `SIGCONT` after resume:

**`/etc/systemd/system/nvidia-suspend.service.d/10-freeze-cosmic-first.conf`:**
```ini
# Workaround for cosmic-comp DRM permission errors during suspend
#
# Problem: nvidia-suspend.service runs BEFORE systemd-suspend freezes user
# processes. When NVIDIA takes exclusive GPU control, cosmic-comp is still
# trying to render, causing "DRM access error: Permission denied" errors
# and a 60-second freeze timeout that prevents proper S3 entry.
#
# Solution: Send SIGSTOP to cosmic-comp before NVIDIA takes control.

[Service]
ExecStartPre=/bin/sh -c 'pkill -STOP -x cosmic-comp 2>/dev/null || true'
ExecStartPre=/bin/sleep 0.3
```

**`/etc/systemd/system/nvidia-resume.service.d/10-resume-cosmic.conf`:**
```ini
[Service]
ExecStartPost=/bin/sh -c 'pkill -CONT -x cosmic-comp 2>/dev/null || true'
```

This workaround is effective but fragile — it relies on process naming, timing heuristics (`sleep 0.3`), and external service orchestration to work around what should be compositor-internal behavior. It must be re-applied after every NVIDIA driver update that regenerates the service files, and it can silently break if the cosmic-comp binary is renamed or wrapped.

The fact that end users must create custom systemd drop-in overrides to `SIGSTOP` the compositor process — essentially force-freezing it from the outside because it has no internal suspend awareness — speaks to the severity of this gap. This is a fundamental lifecycle event that the compositor should handle natively.

### Related Issues

- #1113 — Wake from sleep on NVIDIA (blank screens, Xid 13 errors, texture loss)
- #1798 — cosmic-comp panic when resume from suspend
- #694 — NVIDIA desktop suspend/resume failure (panic at `surface/mod.rs:726`)
- #1208 — Intermittent blank monitor resuming from suspend
- #872 — "Failed to submit rendering" errors (Intel GPU variant)

### References

- [NVIDIA Linux Driver Power Management Documentation](https://download.nvidia.com/XFree86/Linux-x86_64/580.126.09/README/powermanagement.html)
- [org.freedesktop.login1 D-Bus Interface — PrepareForSleep](https://www.freedesktop.org/software/systemd/man/latest/org.freedesktop.login1.html)
- [Smithay Session Management](https://docs.rs/smithay/latest/smithay/backend/session/)

Lcstyle:
## Summary

- Subscribe to logind's `PrepareForSleep` D-Bus signal to cleanly pause and resume the compositor around system sleep
- Take a delay-type sleep inhibitor lock at startup so the compositor has time to pause DRM surfaces before the kernel suspends
- On wake, let `SessionEvent::ActivateSession` (libseat) be the sole resume path — calling `resume_session` from both `PrepareForSleep(false)` and `ActivateSession` caused a race that left NVIDIA primary-plane outputs blank
- Add post-resume connector re-probe timer (1s, 2s, 3s) for DisplayPort link training

## Test plan

- [x] Tested on NVIDIA (RTX 3090) with 4 monitors (3x DP, 1x HDMI)
- [x] Suspend/resume from logged-in session: all displays restore correctly including lock screen
- [x] Suspend/resume from greeter: displays restore correctly
- [x] Sleep inhibitor acquired at startup, released on suspend, re-acquired on wake
- [x] No duplicate `resume_session` calls (verified via journal logs)

Fixes #2078

### DIFF
--- a/src/backend/kms/mod.rs
+++ b/src/backend/kms/mod.rs
@@ -9,6 +9,7 @@ use crate::{
 
 use anyhow::{Context, Result};
 use calloop::LoopSignal;
+use calloop::timer::{TimeoutAction, Timer};
 use cosmic_comp_config::output::comp::{AdaptiveSync, OutputState};
 use indexmap::IndexMap;
 use render::gles::GbmGlowBackend;
@@ -47,6 +48,7 @@ use std::{
     collections::{HashMap, HashSet},
     path::Path,
     sync::{Arc, RwLock, atomic::AtomicBool},
+    time::{Duration, Instant},
 };
 
 mod device;
@@ -73,6 +75,7 @@ pub struct KmsState {
     libinput: Libinput,
 
     pub syncobj_state: Option<DrmSyncobjState>,
+    pub udev_dispatcher: Dispatcher<'static, UdevBackend, State>,
 }
 
 pub struct KmsGuard<'a> {
@@ -130,6 +133,7 @@ pub fn init_backend(
         libinput: libinput_context,
 
         syncobj_state: None,
+        udev_dispatcher: udev_dispatcher.clone(),
     });
 
     // manually add already present gpus
@@ -347,7 +351,7 @@ fn init_udev(
 }
 
 impl State {
-    fn resume_session(
+    pub(crate) fn resume_session(
         &mut self,
         dispatcher: Dispatcher<'static, UdevBackend, Self>,
         loop_handle: LoopHandle<'static, State>,
@@ -413,10 +417,53 @@ impl State {
             }
             state.common.refresh();
         });
+
+        // Schedule delayed re-probes for connectors that need link training
+        // after resume (e.g., DisplayPort takes 1-3s to re-establish).
+        let resume_start = Instant::now();
+        if let Err(err) = loop_handle.insert_source(
+            Timer::from_duration(Duration::from_secs(1)),
+            move |_, _, state| {
+                let nodes = match &mut state.backend {
+                    BackendData::Kms(kms) => kms.drm_devices.keys().cloned().collect::<Vec<_>>(),
+                    _ => Vec::new(),
+                };
+                let mut added = Vec::new();
+                for node in nodes {
+                    match state.device_changed(node.dev_id()) {
+                        Ok(outputs) => added.extend(outputs),
+                        Err(err) => {
+                            error!(?err, "Failed to re-probe drm device {}.", node);
+                        }
+                    }
+                }
+                if let Err(err) = state.refresh_output_config() {
+                    warn!("Unable to load output config during re-probe: {}", err);
+                    if !added.is_empty() {
+                        for output in added {
+                            output.config_mut().enabled = OutputState::Disabled;
+                        }
+                        if let Err(err) = state.refresh_output_config() {
+                            error!("Unrecoverable config error during re-probe: {}", err);
+                        }
+                    }
+                }
+                state.common.refresh();
+
+                if resume_start.elapsed() < Duration::from_secs(3) {
+                    TimeoutAction::ToDuration(Duration::from_secs(1))
+                } else {
+                    TimeoutAction::Drop
+                }
+            },
+        ) {
+            error!(?err, "Failed to schedule post-resume re-probe timer");
+        }
+
         loop_signal.wakeup();
     }
 
-    fn pause_session(&mut self) {
+    pub(crate) fn pause_session(&mut self) {
         let backend = self.backend.kms();
         backend.libinput.suspend();
         for device in backend.drm_devices.values_mut() {
--- a/src/dbus/logind.rs
+++ b/src/dbus/logind.rs
@@ -1,7 +1,10 @@
 use std::os::fd::OwnedFd;
 
 use anyhow::{Context, Result};
-use logind_zbus::manager::{InhibitType::HandleLidSwitch, ManagerProxyBlocking};
+use logind_zbus::manager::{
+    InhibitType::{self, HandleLidSwitch},
+    ManagerProxyBlocking,
+};
 use zbus::blocking::Connection;
 
 pub fn inhibit_lid() -> Result<OwnedFd> {
@@ -14,6 +17,18 @@ pub fn inhibit_lid() -> Result<OwnedFd> {
         "block",
     )?;
 
+    Ok(fd.into())
+}
+
+pub fn inhibit_sleep() -> Result<OwnedFd> {
+    let conn = Connection::system()?;
+    let proxy = ManagerProxyBlocking::new(&conn)?;
+    let fd = proxy.inhibit(
+        InhibitType::Sleep,
+        "cosmic-comp",
+        "Pausing rendering before suspend",
+        "delay",
+    )?;
     Ok(fd.into())
 }
 
--- a/src/dbus/mod.rs
+++ b/src/dbus/mod.rs
@@ -16,6 +16,8 @@ pub mod a11y_keyboard_monitor;
 pub mod logind;
 mod name_owners;
 mod power;
+#[cfg(feature = "systemd")]
+mod sleep;
 
 pub fn init(
     evlh: &LoopHandle<'static, State>,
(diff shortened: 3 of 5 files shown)


Lcstyle:
Feel free to pull this in or rework this to your liking. Using it everyday and works well.

PR #2089, opened 2026-02-12
adding ShowDesktop functionality

Areopagitics:
This adds Action::ShowDesktop functionality so that one can show/hide desktop applications similar to Windows.

It also needs https://github.com/pop-os/cosmic-settings-daemon/pull/129 and https://github.com/pop-os/cosmic-settings/pull/1836 for building (referenced already in Cargo.toml for building)


Drakulix:
As stated in the previous PR, we wouldn't want this functionality to rely on minimizing. Implementing this first needs a workflow design from the UX team.

Areopagitics:
> As stated in the previous PR, we wouldn't want this functionality to rely on minimizing. Implementing this first needs a workflow design from the UX team.

Sounds good!

Issue #1388, opened 2025-05-01
cosmic-comp crashing immediately after login on bare metal, fresh pop install


8L4YD:
![20250501_234056.JPG](https://github.com/user-attachments/assets/fc449502-50be-4095-938d-d8c76ed8ed33)

This is my first foray into cosmic. I did a fresh bare metal install, wiping the NVMe drive and partitioning with efi (fat32), boot (ext4), swap, root and home (btrfs), and mounting my other drives (ext4) to specific user folders. I have an all AMD system, 5700X and RX 6800, I have tried with one and two monitors. I have tried launching from TTY and the login manager and both exhibit the same results. The live USB worked fine. While in TTY I ran apt update/upgrade hoping the new version of comp would fix the issue but no dice. 

I'm happy to provide any more information needed, I wasn't able to find another issue where its completely broken on bare metal but fine live. Will try a reinstall now and update if fixed. TIA :)

8L4YD:
UPDATE: Reinstalled using ext4 for home and root this time, same results. Saw some graphical glitches in the live environment, unsure of root cause.

![20250502_001255.JPG](https://github.com/user-attachments/assets/cc4188d7-53fa-44c6-abe4-deacee5efc81)

8L4YD:
UPDATE 2: Formatting without the extra drives got me a bootable system, it seems the home directory was being left empty if mounts to home subfolders are created during partitioning setup. I was able to copy the fstab entries to the working system.

vfarrell-mayday:
@8L4YD It looks like you're using an ultrawide monitor, and this is the same error I'm encountering when using the full native resolution of the monitor (5120x1440). What are your display specs?
#1331 Related?

PR #2093, opened 2026-02-12
Add EIS input receiver for RemoteDesktop portal support

### LINKED ISSUE #584: [Feature Request] Implement RDP Server
## Description
This feature request proposes implementing a Remote Desktop Protocol (RDP) server directly into cosmic-comp. This feature would allow remote access and control of COSMIC desktops, similar to the RDP server functionality already present in the Weston Wayland compositor.

## Motivation
Adding RDP server capabilities to cosmic-comp would significantly enhance the remote accessibility and management of COSMIC desktops. This feature would be particularly beneficial for:

1. Remote work scenarios
2. IT support and remote administration
3. Collaborative work environments
4. Access to desktop environments from mobile devices or less powerful machines

## Proposed Implementation
The implementation could follow a similar approach to Weston's RDP server:

1. Integrate an RDP server library (e.g., FreeRDP) into cosmic-comp
2. Add configuration options for enabling/disabling the RDP server and setting parameters like port number, authentication methods, etc.

This feature would greatly enhance the functionality and appeal of cosmic-comp and the COSMIC desktop environment as a whole.

## Inspiration and Related Resources
For further inspiration and technical details, the following resources may be helpful:

- [Remote Desktop using RDP protocol Plasma Wayland](https://quantumproductions.info/articles/2023-08/remote-desktop-using-rdp-protocol-plasma-wayland)
- [Remote Desktop Protocol (RDP) 10 AVC/H.264 improvements in Windows Server 2019/Windows 10](https://techcommunity.microsoft.com/t5/security-compliance-and-identity/remote-desktop-protocol-rdp-10-avc-h-264-improvements-in-windows/ba-p/249588)
- [Microsoft WSLg - Weston RDP Backend](https://github.com/microsoft/wslg?tab=readme-ov-file#weston)
- [GNOME Merges RDP Graphical Remote Login Support](https://www.phoronix.com/forums/forum/software/desktop-linux/1433786-gnome-merges-rdp-graphical-remote-login-support)

olafkfreund:
## Summary

- Adds compositor-side EIS (Emulated Input Server) support via a D-Bus interface (`com.system76.CosmicComp.RemoteDesktop`) that accepts EIS socket fds from `xdg-desktop-portal-cosmic`
- Implements a full EIS protocol server using the `reis` crate, supporting keyboard, pointer (relative + absolute), button, scroll, and touch input injection into Smithay's input stack
- Includes D-Bus caller authentication, connection limits, poll timeouts, and input validation

## Context

This provides the compositor half of the [RemoteDesktop portal](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.impl.portal.RemoteDesktop.html) flow, as discussed in #584. The portal creates a socket pair, sends one end to the compositor via D-Bus, and gives the other end to the RDP/VNC client via the portal API.

I'm aware of the existing `upstream/reis` branch (from May 2024) which takes a Smithay-native approach via `smithay::backend::libei::EiInput`. That approach depends on Smithay changes (`ids1024/smithay` `reis` branch) that haven't been merged upstream yet. This implementation works with the current production Smithay and can serve as a bridge until native Smithay EIS support lands.

### Architecture

```
RDP/VNC client
    ↓ (portal API)
xdg-desktop-portal-cosmic
    ↓ (D-Bus: AcceptEisSocket fd)
cosmic-comp  ← this PR
    ├── src/dbus/eis.rs     (D-Bus interface, auth, socket validation)
    └── src/input/eis.rs    (EIS protocol server, event injection)
    ↓ (Smithay input APIs)
Wayland clients
```

### Files changed

| File | Change |
|------|--------|
| `src/dbus/eis.rs` (new) | D-Bus interface with caller auth and socket validation |
| `src/input/eis.rs` (new) | EIS protocol server with full input support |
| `src/dbus/mod.rs` | Register EIS module and init |
| `src/input/mod.rs` | Register EIS module |
| `src/state.rs` | Add `eis_state` field to `Common` |
| `Cargo.toml` | Add `reis` dep, `rustix` event feature |

## Test plan

- [ ] Verify `cargo check` passes with default features
- [ ] Verify `cargo check --no-default-features` passes
- [ ] Verify `cargo fmt --all -- --check` passes
- [ ] Integration test: portal sends EIS socket fd via D-Bus, compositor accepts and processes input events
- [ ] Verify unauthorized D-Bus callers are rejected

Closes #584

### DIFF
--- a/src/dbus/eis.rs
+++ b/src/dbus/eis.rs
@@ -1,0 +1,183 @@
+// SPDX-License-Identifier: GPL-3.0-only
+
+//! D-Bus interface for accepting EIS socket connections from the RemoteDesktop
+//! portal. The portal creates a UNIX socket pair and sends the server-side fd
+//! to the compositor via this interface.
+
+use calloop::channel;
+use futures_executor::ThreadPool;
+use std::os::unix::net::UnixStream;
+use tracing::{error, info, warn};
+use zbus::message::Header;
+
+/// Channel sender for delivering EIS sockets to the compositor's calloop.
+#[derive(Clone)]
+pub struct EisSocketSender {
+    tx: channel::Sender<UnixStream>,
+}
+
+impl EisSocketSender {
+    pub fn new(tx: channel::Sender<UnixStream>) -> Self {
+        Self { tx }
+    }
+}
+
+/// Allowed D-Bus well-known names that may call `AcceptEisSocket`.
+const ALLOWED_CALLERS: &[&str] = &["org.freedesktop.impl.portal.desktop.cosmic"];
+
+/// D-Bus interface for the compositor to accept EIS socket fds.
+pub struct CosmicCompEis {
+    sender: EisSocketSender,
+}
+
+impl CosmicCompEis {
+    pub fn new(sender: EisSocketSender) -> Self {
+        Self { sender }
+    }
+}
+
+#[zbus::interface(name = "com.system76.CosmicComp.RemoteDesktop")]
+impl CosmicCompEis {
+    /// Accept an EIS socket fd from the RemoteDesktop portal.
+    /// The portal sends the server-side of a UNIX socket pair; the compositor
+    /// will run an EIS receiver on it to accept emulated input events.
+    ///
+    /// Only callers that own an allowed D-Bus well-known name (currently the
+    /// COSMIC portal) may invoke this method.
+    async fn accept_eis_socket(
+        &self,
+        #[zbus(header)] header: Header<'_>,
+        #[zbus(connection)] connection: &zbus::Connection,
+        fd: zbus::zvariant::OwnedFd,
+    ) -> zbus::fdo::Result<()> {
+        // Verify caller identity: resolve sender's unique name to well-known names
+        let sender = header
+            .sender()
+            .ok_or_else(|| zbus::fdo::Error::AccessDenied("no sender in D-Bus message".into()))?;
+
+        let dbus_proxy = zbus::fdo::DBusProxy::new(connection)
+            .await
+            .map_err(|e| zbus::fdo::Error::Failed(format!("D-Bus proxy error: {e}")))?;
+
+        // Check if the sender owns any of the allowed well-known names
+        let mut authorized = false;
+        for allowed in ALLOWED_CALLERS {
+            let bus_name: zbus::names::BusName<'_> = (*allowed)
+                .try_into()
+                .map_err(|e| zbus::fdo::Error::Failed(format!("invalid bus name: {e}")))?;
+            if let Ok(owner) = dbus_proxy.get_name_owner(bus_name).await
+                && owner.as_str() == sender.as_str()
+            {
+                authorized = true;
+                break;
+            }
+        }
+
+        if !authorized {
+            warn!(
+                sender = sender.as_str(),
+                "Rejected AcceptEisSocket call from unauthorized D-Bus sender"
+            );
+            return Err(zbus::fdo::Error::AccessDenied(
+                "caller is not an authorized portal process".into(),
+            ));
+        }
+
+        // Verify the fd is a UNIX stream socket (not a file, pipe, etc.)
+        let raw_fd = std::os::fd::OwnedFd::from(fd);
+        {
+            use std::os::fd::AsRawFd;
+            let mut sock_type: libc::c_int = 0;
+            let mut len = std::mem::size_of::<libc::c_int>() as libc::socklen_t;
+            let ret = unsafe {
+                libc::getsockopt(
+                    raw_fd.as_raw_fd(),
+                    libc::SOL_SOCKET,
+                    libc::SO_TYPE,
+                    std::ptr::addr_of_mut!(sock_type).cast(),
+                    std::ptr::addr_of_mut!(len),
+                )
+            };
+            if ret != 0 || sock_type != libc::SOCK_STREAM {
+                warn!(
+                    sender = sender.as_str(),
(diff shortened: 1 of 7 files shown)


src/dbus/eis.rs
```diff
@@ -0,0 +1,183 @@
...
+            warn!(
+                sender = sender.as_str(),
+                "Rejected AcceptEisSocket call from unauthorized D-Bus sender"
+            );
+            return Err(zbus::fdo::Error::AccessDenied(
+                "caller is not an authorized portal process".into(),
+            ));
+        }
+
+        // Verify the fd is a UNIX stream socket (not a file, pipe, etc.)
```
mmstick, line 86: Wouldn't it be better to use the standard library's own function for this? https://doc.rust-lang.org/stable/std/os/unix/fs/trait.FileTypeExt.html#tymethod.is_socket.

mmstick:
> I'm aware of the existing upstream/reis branch (from May 2024) which takes a Smithay-native approach via smithay::backend::libei::EiInput. That approach depends on Smithay changes (ids1024/smithay reis branch) that haven't been merged upstream yet. This implementation works with the current production Smithay and can serve as a bridge until native Smithay EIS support lands.

This is still being actively worked on by @ids1024. The `reis` crate is being developed by @ids1024 for use in smithay. It would be better to wait for this feature to be implemented properly from the start with Smithay instead of using a workaround.

src/input/eis.rs
```diff
@@ -0,0 +1,655 @@
...
+            active_connections: Arc::new(AtomicUsize::new(0)),
+        })
+    }
+
+    /// Accept a new EIS client connection from a UNIX socket fd.
+    ///
+    /// Spawns a background thread that reads EIS protocol events from the
+    /// socket and forwards them as `EisInputEvent` through the calloop channel.
+    /// Rejects the connection if the maximum number of concurrent connections
+    /// has been reached.
```
ids1024, line 101: Spawning a thread per EI connection is unnecessary. As for Wayland sockets, we want to use a `calloop` source so to just add the FD to to list of sockets we're polling in a single epoll instance. Which is what https://docs.rs/reis/0.6.0/reis/calloop/struct.EisRequestSource.html is for.

ids1024:
I just released a new version of `reis` yesterday, addressing a couple things I wanted for Smithay integration. So at the moment my focus is on https://github.com/Smithay/smithay/pull/1388.

Once that part is merged in Smithay, the hard parts of this should be done. Adding a (better) way to pass an FD to the portal and adding a UI are things I haven't prioritized since they aren't needed to test the reis/smithay parts, and they're fairly straightforward.

src/input/eis.rs
```diff
@@ -0,0 +1,655 @@
...
+                }
+                EisRequest::Bind(bind) => {
+                    // Client bound to seat capabilities - add device and resume
+                    let capabilities = capabilities_from_mask(bind.capabilities);
+                    debug!("EIS client bound with capabilities: {:?}", capabilities);
+                    let device = bind.seat.add_device(
+                        Some("remote-input"),
+                        eis::device::DeviceType::Virtual,
+                        &capabilities,
+                        |_| {},
```
ids1024, line 349: This seems to be incorrect. In several ways.

* `Bind` should create and remove devices based on the changes from what, if anything was previously bound. It should not just unconditionally create devices.
* `libei` seems to be designed with the expectation that the mouse, keyboard, etc. will be on separate devices, though for virtual devices it's a bit vague what's correct
* For keyboard input to work, the device with the keyboard capability needs to call `.keymap` in this callback so it's sent before device creation is done. We also need to re-create the keyboard device for a keymap change in the compositor.

Have you tested this at all? Keyboard emulation is presumably broken without a keymap here.

olafkfreund: Yes, this has been tested end-to-end with xdg-desktop-portal-cosmic RemoteDesktop sessions. Keyboard input works because the KeyboardKey handler injects raw evdev keycodes directly into Smithay's KeyboardHandle::input(), which applies the compositor's own XKB keymap - the EIS device keymap isn't in the processing path for key events.

  That said, you're right that this is incomplete per the EIS protocol - the keymap should be sent to the client via Keyboard::keymap() before Device::done(). I'll fix the Bind handler to:
  - Send the compositor's current XKB keymap via memfd in the add_device callback
  - Track bind/unbind state rather than unconditionally creating devices
  - Handle keymap changes

olafkfreund: The code has been tested end-to-end and works because keyboard.input() applies the compositor's own XKB keymap to raw evdev keycodes. The missing EIS keymap didn't break functionality for our use case (raw scancode
  forwarding from RDP), but you're right it was protocol-incomplete. The Bind handler now properly sends the compositor's XKB keymap via sealed memfd before device.done().

olafkfreund:
Fair enough - makes sense to wait for the proper Smithay integration in #1388 rather than merge an interim approach that would need to be replaced anyway.

This was primarily built for our RDP remote desktop use case (cosmic-ext-comp-rdp + cosmic-ext-rdp-server) where I needed EIS input injection working now. Happy to keep it in our fork for that purpose.

Looking forward to the Smithay PR landing - happy to help test it when it's ready. Closing this in the meantime.

src/input/eis.rs
ids1024: Yeah, I guess if libei and reis don't error when the server fails to send a keymap (since I think that's supposed to be required by the protocol), not all clients necessarily rely on it.

https://github.com/ids1024/reis/blob/master/examples/type-text.rs is an example of a simple test case that does need a valid keymap.

Issue #2084, opened 2026-02-11
Memory leaks report + closing windows leak fix


MartinKavik:
Hello, I'm currently running a changed `cosmic-comp` in my Pop!_OS that fixes the VRAM leak on closing windows: https://github.com/pop-os/cosmic-comp/issues/1179 (https://github.com/Smithay/smithay/issues/1562). 

I also found several other potential VRAM/RAM leaks, so I've written a report with fix branches  that can be transformed into PRs after discussion: https://github.com/MartinKavik/popos_fix_vram_leak

Main fix I'm currently testing on my machine:
- `cosmic-comp` changes: https://github.com/pop-os/cosmic-comp/compare/master...MartinKavik:cosmic-comp:weak_window
- `smithay` changes: https://github.com/Smithay/smithay/compare/master...MartinKavik:smithay:weak_window

Drakulix:
Could you open pull requests to the respective repositories with your changes? The proposed solution seems good.

ids1024:
Is this being leaked because of a reference cycle in cosmic-comp (that this fixes), or is there still a weak reference to a closed window that probably shouldn't exist anymore?

Using weak references where a strong reference isn't needed is generally good either way, but in the latter case there may be another issue to address.

Drakulix:
> Is this being leaked because of a reference cycle in cosmic-comp (that this fixes), or is there still a weak reference to a closed window that probably shouldn't exist anymore?

From what I gathered looking at our code:

A window is removed from `ToplevelInfoState` on `unmap_surface`, which for xdg-toplevel windows is called via `XdgSurfaceHandler::toplevel_destroyed`, which is in turn called by smithay's XdgToplevel-dispatcher when that object is destroyed. Which does not happen because the `ToplevelInfoState` holds the window, which holds a reference to the `XdgToplevel`.

I looked around to figure out, if we want to remove it from the `ToplevelInfoState` on close instead, but it seemed cumbersome to make sure, that every code path that might transition the window over to be unmapped would consistently clean it up. It seemed easier to continue tracking all toplevel (no matter their state) in the `ToplevelInfoState` and then let that handle updating the client state accordingly for unmapped windows, etc.

So imo this is a good fix for a reference cycle in cosmic-comp with few other viable alternatives and no stale weak reference.

[...]

PR #2095, opened 2026-02-13
Fix VRAM leak: use weak references in ToplevelHandleState and stop capture sessions

MartinKavik:
## Summary

  - Fix VRAM leak when closing windows: store `Weak` references instead of strong `CosmicSurface` clones in `ToplevelHandleState`, so protocol
  handles don't keep `WindowInner` (and its GPU textures) alive
  - Fix VRAM leak when using workspace overview (Super+W): use `WeakCosmicSurface` in `ImageCaptureSourceKind::Toplevel` and explicitly drop
  capture sessions on toplevel removal so `Session::drop()` fails active frames and releases GPU buffers
  - Bump smithay to upstream `599857c` (includes `mem::forget` fix for `Frame::success()`/`Frame::fail()`)

  More details and investigation notes: https://github.com/pop-os/cosmic-comp/issues/2084

  ## Test results

  5 rounds of 20 cosmic-term windows + 5 workspace overview (Super+W) open/close cycles each. NVIDIA RTX 2070.

  | Round | cosmic-comp | cosmic-workspaces | Total |
  |-------|------------|-------------------|-------|
  | 0 (baseline) | 32 MiB | 14 MiB | 508 MiB |
  | 1 | 207 MiB | 142 MiB | 893 MiB |
  | 2 | 73 MiB | 142 MiB | 759 MiB |
  | 3 | 248 MiB | 142 MiB | 893 MiB |
  | 4 | 140 MiB | 142 MiB | 868 MiB |
  | 5 | 290 MiB | 14 MiB | 765 MiB |
  | 6 | 248 MiB | 142 MiB | 893 MiB |

  VRAM fluctuates but does not accumulate across rounds. Before this fix, cosmic-comp leaked ~200 MiB per round, growing linearly.

  Actively using the fixed `cosmic-comp` binary in live OS right now to make sure it's stable and not leaking.

  ## Resolves

  - Closing windows VRAM leak (stale `ToplevelHandleState.window` references)
  - Workspace overview VRAM leak (capture session / `ImageCaptureSource` reference cycle)

### DIFF
--- a/src/wayland/handlers/image_copy_capture/user_data.rs
+++ b/src/wayland/handlers/image_copy_capture/user_data.rs
@@ -35,10 +35,22 @@ pub struct ImageCopySessions {
     cursor_sessions: Vec<CursorSession>,
 }
 
+/// Drop all capture sessions stored in the given `UserDataMap`.
+///
+/// When a toplevel is destroyed, its owned `Session` objects must be dropped
+/// so that `Session::drop()` fires — this fails active frames and sends
+/// `stopped` to the client, releasing GPU buffers.
+///
+/// Smithay doesn't automatically stop sessions when their capture source's
+/// underlying toplevel dies. A cleaner long-term fix would be in smithay's
+/// `ImageCopyCaptureState::cleanup()` — e.g. stopping sessions whose
+/// `source().alive()` is false — so all compositors benefit without manual
+/// session management. For now, we handle it on the cosmic-comp side.
 pub fn stop_all_capture_sessions(user_data: &UserDataMap) {
     if let Some(data) = user_data.get::<ImageCopySessionsData>() {
-        let _sessions: Vec<_> = data.borrow_mut().sessions.drain(..).collect();
-        let _cursor_sessions: Vec<_> = data.borrow_mut().cursor_sessions.drain(..).collect();
+        let mut data = data.borrow_mut();
+        data.sessions.clear();
+        data.cursor_sessions.clear();
     }
 }
 
--- a/src/wayland/protocols/image_capture_source.rs
+++ b/src/wayland/protocols/image_capture_source.rs
@@ -30,6 +30,12 @@ pub struct WorkspaceImageCaptureSourceManagerGlobalData {
     filter: Box<dyn for<'a> Fn(&'a Client) -> bool + Send + Sync>,
 }
 
+/// What a capture source targets.
+///
+/// `Toplevel` uses `WeakCosmicSurface` (not a strong `CosmicSurface`) so
+/// that capture sources created by cosmic-workspaces for window thumbnails
+/// don't prevent the underlying window from being dropped. `Output` already
+/// uses `WeakOutput` for the same reason.
 #[derive(Debug, Clone)]
 pub enum ImageCaptureSourceKind {
     Output(WeakOutput),
--- a/src/wayland/protocols/toplevel_info.rs
+++ b/src/wayland/protocols/toplevel_info.rs
@@ -31,6 +31,11 @@ use tracing::error;
 
 pub trait Window: IsAlive + Clone + PartialEq + Send {
     /// A weak reference type that does not keep the window alive.
+    ///
+    /// Used by `ToplevelHandleStateInner` to avoid preventing window cleanup.
+    /// Protocol handle objects persist until the *client* destroys them, so
+    /// a strong reference here would keep `Arc<WindowInner>` (and its GPU
+    /// textures) alive indefinitely.
     type Weak: Clone + Send + 'static;
 
     fn title(&self) -> String;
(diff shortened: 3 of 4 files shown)


src/wayland/handlers/image_copy_capture/mod.rs
```diff
@@ -234,7 +240,10 @@ impl ImageCopyCaptureHandler for State {
 
                 workspace.add_cursor_session(session);
             }
-            ImageCaptureSourceKind::Toplevel(mut toplevel) => {
+            ImageCaptureSourceKind::Toplevel(weak) => {
+                let Some(mut toplevel) = weak.upgrade() else {
+                    return;
```
Drakulix, line 245: I don't think this can just return, we need to invalidate the session here.

src/wayland/protocols/image_capture_source.rs
```diff
@@ -30,11 +30,17 @@ pub struct WorkspaceImageCaptureSourceManagerGlobalData {
     filter: Box<dyn for<'a> Fn(&'a Client) -> bool + Send + Sync>,
 }
 
-#[derive(Debug, Clone, PartialEq)]
+/// What a capture source targets.
+///
+/// `Toplevel` uses `WeakCosmicSurface` (not a strong `CosmicSurface`) so
+/// that capture sources created by cosmic-workspaces for window thumbnails
+/// don't prevent the underlying window from being dropped. `Output` already
+/// uses `WeakOutput` for the same reason.
```
Drakulix, line 38: Please drop this comment.

src/wayland/protocols/toplevel_info.rs
```diff
@@ -29,6 +30,14 @@ use cosmic_protocols::toplevel_info::v1::server::{
 use tracing::error;
 
 pub trait Window: IsAlive + Clone + PartialEq + Send {
+    /// A weak reference type that does not keep the window alive.
+    ///
+    /// Used by `ToplevelHandleStateInner` to avoid preventing window cleanup.
+    /// Protocol handle objects persist until the *client* destroys them, so
+    /// a strong reference here would keep `Arc<WindowInner>` (and its GPU
+    /// textures) alive indefinitely.
+    type Weak: Clone + Send + 'static;
```
Drakulix, line 39: Please drop this comment.

src/wayland/protocols/toplevel_info.rs
```diff
@@ -90,6 +101,14 @@ impl ToplevelStateInner {
...
 }
 
+/// Per-handle state stored in each `ZcosmicToplevelHandleV1`'s user data.
+///
+/// The `window` field stores a *weak* reference (`W::Weak`) to avoid keeping
+/// the underlying window alive. Protocol handles persist until the client
+/// destroys them — long-lived clients (cosmic-panel, cosmic-workspaces, and
+/// even cosmic-term itself) hold handles indefinitely. A strong reference
+/// here would prevent `WindowInner` from dropping, trapping GPU textures
+/// in `SurfaceUserData::data_map` → `MultiTextureInternal`.
```
Drakulix, line 111: Half of this text isn't accurate and repetitive. Please drop this comment from the commit.

src/wayland/handlers/image_copy_capture/user_data.rs
```diff
@@ -33,6 +35,25 @@ pub struct ImageCopySessions {
...
+/// underlying toplevel dies. A cleaner long-term fix would be in smithay's
+/// `ImageCopyCaptureState::cleanup()` — e.g. stopping sessions whose
+/// `source().alive()` is false — so all compositors benefit without manual
+/// session management. For now, we handle it on the cosmic-comp side.
+pub fn stop_all_capture_sessions(user_data: &UserDataMap) {
+    if let Some(data) = user_data.get::<ImageCopySessionsData>() {
+        let mut data = data.borrow_mut();
+        data.sessions.clear();
+        data.cursor_sessions.clear();
+    }
```
Drakulix, line 54: This whole function seems unnecessary and the documentation wrong?
The user-data of the toplevel is dropped, when the wayland-object is destroyed.

As long as `ImageCopySessionsData` doesn't hold any recursive structures, this isn't necessary.

(Looking at our code this, it might be that output-session which are created with a `DamageTracker` holding a reference to an output might be self-referencial, but that doesn't affect this code path. @ids1024 We need to make sure to manually drop sessions on `remove_output`.)

delicious-pancakes:
Heads up: Drakulix's comment about not just returning when the weak upgrade fails is worth looking at closely. The `Destroyed` / failed-upgrade paths in `new_session`, `new_cursor_session`, and `frame` currently just `return`, but the session is still alive from the client's perspective. It needs to be invalidated (e.g. `session.stop()`) so the client knows to stop submitting frames, otherwise it'll keep sending requests against a source that can never produce output.

The `session_destroyed` and `cursor_session_destroyed` paths are fine since the session is already going away there.

ids1024:
> Heads up: Drakulix's comment about not just returning when the weak upgrade fails is worth looking at closely. The Destroyed / failed-upgrade paths in new_session, new_cursor_session, and frame currently just return, but the session is still alive from the client's perspective. It needs to be invalidated (e.g. session.stop()) so the client knows to stop submitting frames, otherwise it'll keep sending requests against a source that can never produce output.

For reference, this shouldn't be necessary since `Session` implements `Drop` to call `stopped()` and fail frames:
https://github.com/Smithay/smithay/blob/188b805427172659c3b9d574f5b3ab8348fa6e17/src/wayland/image_copy_capture/mod.rs#L335-L356

Issue #1234, opened 2025-02-23
The highlight from the workspaces doesn't disappear sometimes


italoghost:
Hi! Sometimes, when I close an app and click in the Workspace button, the highlight from the application doesn't disappear (even though I have disabled the highlight).

![Image](https://github.com/user-attachments/assets/91cb3e36-9757-4265-be7a-ac91b653b71f)

![Image](https://github.com/user-attachments/assets/d12de905-1b4d-4d8d-a14c-f9c615cbb4b4)

[cosmic-session.txt](https://github.com/user-attachments/files/18931761/cosmic-session.txt)

git-f0x:
I think this is likely a duplicate of https://github.com/pop-os/cosmic-files/issues/757. It happens in some cases of drag and drop over the desktop.

iamkartiknayak:
I think this issue can be closed

Issue #51, opened 2022-12-28
Popups in bottom anchored panel are repositioned off screen


wash2:
I have set the popups in all cosmic-panel applets to allow repositioning by cosmic-comp. Side anchored panel popups and top anchored panel popups all seem to be positioned correctly when necessary, but the bottom panel popup, are re-positioned to appear off screen below the panel, even when they don't seem to have needed re-positioning in the first place.

XV-02:
@wash2 I'm not entirely sure what you mean by this. As far as I can tell, menus spawned by interacting with panel applets are being positioned correctly regardless of where on the panel they are (centre, left, or right) and regardless of the position of the panel itself (top, left, bottom, right) and regardless of other factors like whether the applet is near the border of multiple displays. 

Can you confirm if this is still an issue or not?

wash2:
Oh this should be closed

PR #2097, opened 2026-02-13
fix(kms): re-configure lock surfaces after resume to prevent blank output

### LINKED ISSUE #2096: [BUG] Lock surface blank on one output after extended S3 resume — no re-configure sent to prompt buffer re-commit
## Summary

After extended S3 suspend (3+ hours), one DisplayPort output (DP-3) shows a blank screen at the lock screen — backlight on, hardware cursor visible, but no lock screen content. The other three outputs (DP-1, DP-2, HDMI-A-1) render correctly. The `LockSurface` object exists in `session_lock.surfaces` for DP-3, but `render_elements_from_surface_tree()` returns **0 elements** — the underlying buffer is missing or not importable.

**Short sleeps (~1 min) work fine. Only extended sleeps reproduce this.** The greeter session (which doesn't use `ext-session-lock`) works fine even after 8+ hours of sleep, confirming the issue is specific to the session lock surface rendering path.

## System

- **GPU:** NVIDIA RTX 3090 Ti, driver 580.119.02 (open kernel module)
- **Displays:** 4 outputs — DP-1, DP-2, DP-3, HDMI-A-1
- **Kernel:** 6.18.8, S3 deep sleep
- **NVIDIA config:** `modeset=1`, `fbdev=1`, `PreserveVideoMemoryAllocations=1`

### The Invariant Violation

**Invariant:** *"Every lock surface in `session_lock.surfaces` must have a valid, renderable `wl_buffer` at all times while the session is locked."*

This invariant is violated after resume because:
- The compositor doesn't verify buffer validity after resume
- The compositor has no mechanism to request buffer re-submission
- `send_configure()` is only called once at surface creation time (`session_lock.rs:64`)

### Root Cause Chain (5 links)

```
1. GPU context recreation
   On resume, new GbmDrmOutput compositor created
   → renderer context_id may change
   → RendererSurfaceState.textures cache misses for old context

2. DMA-BUF import failure
   import_surface() attempts re-import of old buffer
   → if NVIDIA didn't fully restore that specific DMA-BUF allocation, import fails silently
   → from_surface() returns Ok(None) — "unmapped" surface

3. 0 render elements
   render_elements_from_surface_tree() → from_state() finds no valid texture
   → returns empty Vec
   → session_lock_elements() returns 0 elements for that output
   → output renders only background fill (appears black)

4. Frame callback deadlock
   send_frames() checks surface_primary_scanout_output()
   → surface was never rendered (0 elements) → primary scanout output never set
   → frame callbacks suppressed for this lock surface
   → client never learns it needs to re-commit

5. No recovery mechanism
   send_configure() is only called once at lock surface creation time
   → no post-resume re-configure to prompt client re-commit
   → deadlock: no buffer → no render → no frame callback → no re-commit → no buffer
```

### Invariant Layer Analysis

| Layer | Question | Answer |
|-------|----------|--------|
| **L1** (Function) | Can `render_elements_from_surface_tree` fix this? | No — it correctly returns empty for invalid buffers |
| **L2** (Module) | Can `session_lock_elements` fix this? | No — it's on the render thread, can't send Wayland events |
| **L3** (Contract) | Can the compositor-to-client contract fix this? | **YES — send configure to prompt re-commit** |

### Why Only After Extended Sleep?

The exact mechanism that makes buffer import fail after long (but not short) sleep is not yet confirmed, but the leading theories are:

1. **NVIDIA VRAM restoration timing:** Longer VRAM save/restore cycles may cause specific DMA-BUF allocations to be restored in a different order or fail to restore entirely.
2. **Lock client freeze duration:** The lock screen client (cosmic-greeter in lock mode) is frozen by the kernel during S3. After an extended freeze, internal client-side state (iced/libcosmic rendering pipeline) may need a configure event to re-initialize.

Regardless of the upstream trigger, the compositor-side fix is the same: detect and recover by re-configuring lock surfaces after resume.

### Diagnostic Evidence

**Phase 5c (3-hour sleep, reproduced):**
```
DP-3: session_lock=true, lock_surface_for_output=true → 1 element (background only)
DP-2: session_lock=true, lock_surface_for_output=true → 3 elements (background + lock + cursor)
DP-1: session_lock=true, lock_surface_for_output=true → 3 elements
HDMI: session_lock=true, lock_surface_for_output=true → 2 elements
```

**Phase 7 (overnight 8h50m, greeter session — did NOT reproduce):**
```
All outputs: session_lock=false (greeter, no lock surfaces) → 3 elements → OK
device_changed: "no connector changes detected" (rules out Output recreation)
```

## The Fix

After resume, force re-configure all lock surfaces to prompt the client to create fresh buffers. This breaks the deadlock at link 5 (no recovery mechanism).

**Complication:** Smithay's `LockSurface::send_configure()` deduplicates when the size hasn't changed (comparing `server_pending` against `current_server_state()`). Since the output size didn't change during suspend, a naive `send_configure()` call is silently dropped.

**Solution:** Bypass the dedup by temporarily bumping the height by 1px, sending configure, then correcting to the real size and sending again. The client processes both events in a single `wl_display_dispatch`, acks the correct size, and re-commits a fresh buffer.

Two locations:
1. **Immediate (idle callback):** In `resume_session()`'s idle callback, after `refresh_output_config()` + `refresh()`, re-configure all lock surfaces
2. **Safety net (re-probe timer):** In the delayed re-probe timer (1s/2s/3s), check if any lock surface still has no renderer buffer and re-configure again

## Related Issues

- #2078 — Parent investigation (PrepareForSleep handler)
- PR #2079 — PrepareForSleep fix (prerequisite)
- #1463 — Output destruction/recreation invalidating surfaces (investigated, ruled out for this case)
- #1113, #1208, #1332 — NVIDIA + DP blank-after-resume reports

Lcstyle:
## Summary

- After extended S3 suspend, re-send configure events to all session lock surfaces during the resume idle callback to prompt the lock client to re-commit fresh buffers
- Add a safety net in the post-resume re-probe timer that detects lock surfaces with no renderer buffer and re-sends configure

Fixes #2096

## Root Cause

After extended S3 sleep (3+ hours), one or more lock surface buffers can become unimportable by the GPU renderer. The `LockSurface` object exists in `session_lock.surfaces`, but `render_elements_from_surface_tree()` returns 0 elements because the DMA-BUF backing the buffer failed to re-import after GPU context recreation during resume.

This creates a deadlock with 5 links:

1. GPU context recreation → renderer `context_id` changes → texture cache misses
2. `import_surface` re-import fails (DMA-BUF invalidated by GPU driver during suspend)
3. `render_elements_from_surface_tree` returns empty → 0 render elements for that output
4. `surface_primary_scanout_output` never set (surface was never rendered) → `send_frames` suppresses frame callbacks
5. Lock client never receives frame callback → never re-commits → deadlock

The compositor currently only calls `send_configure()` on lock surfaces at creation time (`session_lock.rs:64`). There is no post-resume re-configure mechanism, so the client has no way to know it needs to re-commit.

## The Fix

After `resume_session()` completes output re-configuration, iterate through all lock surfaces and send configure events. Since Smithay deduplicates `send_configure()` when the size hasn't changed (comparing `server_pending` against `current_server_state()`), the configure is sent twice: first with `height+1` to bypass dedup, then with the correct size. The client processes both events in a single `wl_display_dispatch`, acks the correct size, and re-commits a fresh buffer with new DMA-BUF backing.

A safety net in the post-resume re-probe timer (running at 1s/2s/3s after resume) checks if any lock surface still has no renderer buffer via `with_renderer_surface_state()` and re-sends configure if needed.

## Investigation Context

This issue was discovered during the investigation of #2078 (missing PrepareForSleep handler). After fixing the suspend/resume lifecycle (#2079), the blank screen persisted intermittently on extended sleep cycles. Extensive diagnostic instrumentation across 7 investigation phases narrowed the root cause from DRM race conditions → damage tracker issues → lock surface buffer invalidation → the deadlock described above. Full analysis in #2096.

## Test Plan

- [ ] Suspend/resume from locked user session — short sleep (~1 min): all displays restore correctly
- [ ] Suspend/resume from locked user session — extended sleep (3+ hours): all displays restore correctly (previously failed)
- [ ] Suspend/resume from greeter session: displays restore correctly (no lock surfaces, no-op)
- [ ] No regressions: lock screen renders correctly on all outputs after normal screen lock (no suspend)
- [ ] Safety net: if lock client is slow to respond, re-probe timer re-sends configure

### DIFF
--- a/src/backend/kms/mod.rs
+++ b/src/backend/kms/mod.rs
@@ -416,6 +416,37 @@ impl State {
                 }
             }
             state.common.refresh();
+
+            // Re-configure lock surfaces to prompt the client to re-commit
+            // fresh buffers. During S3 suspend, the GPU context may change
+            // (new EGL context / DRM device reset), invalidating cached textures
+            // in RendererSurfaceState. The lock client was frozen during suspend
+            // and never re-committed, so its old buffer may be unimportable.
+            // Without this, a deadlock occurs: no valid buffer → 0 render
+            // elements → primary scanout never set → frame callbacks suppressed
+            // → client never learns it needs to re-commit.
+            //
+            // Smithay deduplicates send_configure() when the size hasn't changed,
+            // so we bump height by 1, send, then correct and send again. The
+            // client processes both in one dispatch, acks the correct size, and
+            // re-commits.
+            {
+                let shell = state.common.shell.read();
+                if let Some(session_lock) = &shell.session_lock {
+                    for (output, lock_surface) in &session_lock.surfaces {
+                        let size = output.geometry().size;
+                        let (w, h) = (size.w as u32, size.h as u32);
+                        lock_surface.with_pending_state(|states| {
+                            states.size = Some(Size::from((w, h.saturating_add(1))));
+                        });
+                        lock_surface.send_configure();
+                        lock_surface.with_pending_state(|states| {
+                            states.size = Some(Size::from((w, h)));
+                        });
+                        lock_surface.send_configure();
+                    }
+                }
+            }
         });
 
         // Schedule delayed re-probes for connectors that need link training
@@ -449,6 +480,40 @@ impl State {
                     }
                 }
                 state.common.refresh();
+
+                // Safety net: if any lock surface still has no renderer buffer,
+                // re-send configure. Handles the case where the initial
+                // re-configure was sent before the client processed its events.
+                {
+                    let shell = state.common.shell.read();
+                    if let Some(session_lock) = &shell.session_lock {
+                        for (output, lock_surface) in &session_lock.surfaces {
+                            let has_buffer =
+                                smithay::backend::renderer::utils::with_renderer_surface_state(
+                                    lock_surface.wl_surface(),
+                                    |rs| rs.buffer().is_some(),
+                                )
+                                .unwrap_or(false);
+                            if !has_buffer {
+                                let size = output.geometry().size;
+                                let (w, h) = (size.w as u32, size.h as u32);
+                                lock_surface.with_pending_state(|states| {
+                                    states.size =
+                                        Some(Size::from((w, h.saturating_add(1))));
+                                });
+                                lock_surface.send_configure();
+                                lock_surface.with_pending_state(|states| {
+                                    states.size = Some(Size::from((w, h)));
+                                });
+                                lock_surface.send_configure();
+                                warn!(
+                                    "Lock surface for {} had no buffer after resume, re-sent configure",
+                                    output.name(),
+                                );
+                            }
+                        }
+                    }
+                }
 
                 if resume_start.elapsed() < Duration::from_secs(3) {
                     TimeoutAction::ToDuration(Duration::from_secs(1))


Drakulix:
No. This is a bug of the nvidia-driver, which clears it's VRAM contents without informing the client (or us for that matter). (A lost-context would be appropriate here.)

Our compositor/DE isn't tested or supported without nvidia's `NVreg_PreserveVideoMemoryAllocations` setting being enabled, which mitigates this issue.

If you want to work on an upstream solution, the [surface-invalidation](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/180) protocol would provide a proper way to solve this from our side.

Lcstyle:
Unfortunately, PreserveVideoMemoryAllocations=1 is both configured AND active in the running kernel. I already have it enabled and the bug still occurs. 

Look at the memory peaks across suspend events:


| Time  | Duration | Memory Peak | Bug?                     |
|-------|----------|-------------|--------------------------|
| 11:52 | ~30s     | 830.7M      | No                       |
| 12:15 | ~30s     | 903.4M      | No                       |
| 12:51 | ~30s     | 901.2M      | No                       |
| 13:14 | 3h04m    | 3.7G        | YES — blank DP-3         |
| 18:56 | ~22m     | 2.0G        | No (greeter)             |
| 23:02 | 8h50m    | 505.9M      | No (greeter)             |



None of those implementations are actually merged:

| Project            | PR/MR                                                                 | Status                                                                 |
|--------------------|-----------------------------------------------------------------------|------------------------------------------------------------------------|
| wayland-protocols  | https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/180 | NOT merged — open since Nov 2022, last edit Feb 2023                  |
| wlroots            | https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3921  | NOT merged — depends on !3910 (GPU reset recovery), also unmerged     |
| wlroots            | https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3910  | NOT merged — "Automatic GPU reset recovery", open since Dec 2022, 29 commits, still open as of Sept 2024 |
| foot               | https://codeberg.org/dnkl/foot/pulls/1234                             | NOT merged — draft/WIP, unresolved concerns, merge conflicts          |
| swaybg             | https://github.com/swaywm/swaybg/pull/51                              | NOT merged — reviewer said implementation was incomplete              |

Lcstyle:
Thanks for the pointer to surface-invalidation — I've done a deep dive into the protocol, the ecosystem status, and what the right upstream path forward looks like. Here's what I found:

## Why this happens despite `PreserveVideoMemoryAllocations=1`

The nvidia-suspend service IS saving VRAM (3.7G on the reproduction event, 4.5s CPU time), and nvidia-resume runs without error. But one lock surface's DMA-BUF still becomes unimportable after extended sleep.

Notably, the **greeter session survives 8+ hours of sleep with no issues** on the same hardware. The difference is architectural:

- **Greeter session**: Renders as normal surfaces with an active render loop. After resume, frame callbacks flow → the client re-renders → fresh buffers are created → self-heals even if the old buffer was stale.
- **Locked user session**: Lock surfaces are created via ext-session-lock-v1, the client commits a buffer once and waits. After resume, if the buffer can't be imported → 0 render elements → primary scanout never set → frame callbacks suppressed → client never learns it needs to re-commit → **deadlock**.

The greeter survives because its rendering pipeline is active and self-healing. The lock surface fails because it's passive — it committed a buffer and went silent, and nothing tells it to try again. This same hardware has never exhibited this behavior in Windows 10, where DWM's tight integration with the driver means there's no equivalent deadlock path.

So the root issue is an interaction between NVIDIA's imperfect VRAM restoration on extended sleep and the compositor having **no recovery path** for invalidated buffers on lock surfaces. Either side fixing it would resolve the problem.

## Surface-invalidation protocol status

The protocol would indeed be the proper solution, but the entire ecosystem is stalled:

- [wayland-protocols !180](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/180): Open since Nov 2022, last edit Feb 2023, not merged
- All referenced implementations (wlroots !3921, foot #1234, swaybg #51): None merged

The protocol doesn't exist in any released software, and Smithay has no implementation.

## What wlroots shipped instead

wlroots 0.18 shipped GPU reset recovery ([MR #4606](https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4606), merged July 2024) **without** the surface-invalidation protocol. Sway 1.10 uses it. The compositor-side recovery (detect reset → destroy/recreate renderer → clear stale texture caches) works independently — surface-invalidation is an optional client-side complement.

## The right upstream target: Smithay #978

[Smithay #978: "Detect and recover from GPU resets"](https://github.com/Smithay/smithay/issues/978) is the equivalent gap in Smithay. It's open, has no PRs, and no assignee. This is the foundation that cosmic-comp needs — you can't send invalidation events if the compositor doesn't know buffers are invalid.

I'd like to contribute to this. The path I see:

```
Smithay #978 (GPU reset detection/recovery)
    ├── Detect buffer invalidation after resume / GPU reset
    ├── Destroy/recreate renderer, clear stale texture caches
    ├── Reference: wlroots 0.18 MR #4606
    │
    └── surface-invalidation-v1 (follow-up)
        ├── Take protocol XML from wayland-protocols !180
        ├── Compositor → client invalidation events
        └── iced/libcosmic client-side support
```

Our bug (NVIDIA VRAM loss on extended S3 with `PreserveVideoMemoryAllocations=1` confirmed active) provides a concrete real-world reproduction case with diagnostic data. Would a contribution toward GPU reset recovery in Smithay be welcome, using wlroots 0.18's approach as reference?

Issue #1642, opened 2025-09-15
[BUG] Maximized chromium browsers: clicking video fullscreen makes browser fullscreen instead


jsanchezba:
# How to reproduce

1. Click on the fullscreen icon of a video (Doesn't matter which one, youtube or reddit video)
2. Click again on the fullscreen icon to change the video into fullscreen

# What happens
When clicking for the first time the browser goes into fullscreen and when clicking a second time the video finally changes to fullscreen.

# What should happen
Should change the video into fullscreen mode, not the browser

# Notes
Tried with Brave and Vivaldi, Flatpak and native versions. Seems like happens with chromium based browsers. With Firefox works as expected.

rafaelcga:
Curiously, Steam, which uses the [Chromium Embedded Framework](https://github.com/chromiumembedded/cef) for its store does not suffer from this issue in video playback within game pages.

WatchMkr:
Are the browsers running in Wayland or XWayland? If `xprop` shows information when clicking inside the browser, it's running in XWayland.

rafaelcga:
I am no longer on COSMIC on this machine. But at the time Chrome was already being built with flags to use Wayland if available, so I assume it wasn't using XWayland. References: [Chromium PR](https://chromium-review.googlesource.com/c/chromium/src/+/6819616) and [Chromium discussion](https://issues.chromium.org/issues/40083534#comment593).

WatchMkr:
Ah, figured out how to reproduce the issue. In my testing, it only occurs when the window is maximized.

dries007:
I can confirm it happens only for maximized windows, but if you fullscreen a not-maximized video and then shift focus on another monitor, the fullscreen video does not stay on top. 
I don't know if this is related. I can make a video if needed.

rafaelcga:
I have been testing this for a while, now using beta-4 in NixOS. The issue for some reason only occurs when using floating windows. However, when in tiling mode, exiting a full-screen takes a significant amount of time in which the window seems to minimize / disappear.

Xemnas97:
Any updates on this? There's a workaround until it's fixed?

leviport:
It seems that fullscreening twice still fullscreens the video, so that's a potential workaround. You'll have to un-fullscreen the browser after, but that's easily doable with F11. Either that or only fullscreen a video when the browser is not maximized.

joshua-hvmn:
[...]

Issue #528, opened 2024-05-31
window movement


rrahl0:
When I move a maximized window from one virtual desktop to another, it should keep it's maximized status. As of right now, it gets unmaximized (but it stays the same size (gets the active hint borders)

RafaelKr:
This also happens when moving windows between displays (via Shortcut `Shift + SUPER + Alt + Direction keys`) or Workspaces (via Shortcut `Shift + SUPER + Ctrl + Direction keys` or `Shift + SUPER + 0-9 keys`)

RafaelKr:
@ids1024 you fixed #1645 and #1819 via https://github.com/pop-os/cosmic-comp/commit/612ff2f523580af2729cc7de1e73c334ea135726

This issue is somehow related, it would be nice if you could have a look at this also. Would be a great usability improvement, as I'm often moving maximized windows between my screens.

leviport:
I don't think those are related to this in any meaningful way.

RafaelKr:
From my perspective as a user, it seemed as though there might be some overlap in the underlying code paths, since all of the issues relate to the size/status of windows after moving them.

Therefore, I thought it would be useful to have someone who is currently familiar with this part of the code take a look at it.

I apologise if my comments are distracting.

notlaast:
TLDR: two behaviors should be fixed:
1. I agree that a keyboard shortcut which pushes a maximized window to a new monitor should retain its maximized state; but,
2. dragging a maximized window should indeed produce a non-maximized window with some particular size, and that size should *not* change after drag-and-drop is ended.

I don't mean to change scope but the title of this ticket is vague and I think clarifications are worth noting here. I agree with others above about only one aspect of the general behavior, and just hope the other undesirable behavior is also addressed.

To describe the second at the risk of being pedantic: I think dragging should indeed retain the window size but not the maximized state (and it's anyhow relatively easy to drag it to the top of the new screen to become maximized again). This behavior is, however, not perfect: the dragged window initially takes on its previous non-maximized size while being dragged, but then it gets auto-resized when I stop dragging so that it then jumps to its previous size while it was still maximized (except now no longer in the maximized state). Case in point: when I drag a maximized window from my smaller laptop screen directly to the bottom-right corner of larger screen, the window inevitably snaps to some other size which is no longer flush with the screen edges.

PR #2099, opened 2026-02-14
fix: window state preservation

mayakwd:
Fixes a bunch of issues where the maximized/snapped state of a window was lost in scenarios involving:
- transitions to full-screen mode,
- from full-screen mode to snapped state,
- moving full-screen/snapped window between workspaces.

**Exact fixes:**

- For Chromium and similar clients, implements strict protocol compliance via pre-set fullscreen before `unmap`. `Unmap` sends intermediate `configure`s considered by Chromium as a refusal to transition to full-screen mode, as they did not contain full-screen. This fixes the issue: https://github.com/pop-os/cosmic-comp/issues/1642 (for both maximized and snapped windows).

- Preserving snapped state when transitioning to fullscreen/minimizing. `FloatingRestoreData` now stores  `was_snapped: Option<TiledCorners>`. The information is captured before `floating_layer.unmap()` clears it and is used later for restoration.
- Preserving maximized/snapped state when moving between workspaces. After moving the window, the maximized or snapped state from `FloatingRestoreData` is re-applied in the target workspace (https://github.com/pop-os/cosmic-comp/issues/528, https://github.com/pop-os/cosmic-epoch/issues/2608)

**What I tested:**

- Maximize Chrome > Click fullscreen on any video > Single-click fullscreen
- Maximize Firefox > Click fullscreen on any video > Single-click fullscreen (regression check)
- Snap Chrome to a corner/side > Click fullscreen on any video >  Single-click fullscreen
- Snap window to a corner/side > Go fullscreen > Exit fullscreen > Window returns to the snap position
- Maximize window > Move window to the other workspace > Window stays maximized
- Snap window to a corner/side > Move window to the other workspace > Window stays snapped to the same corner/side

I tried to make the changes as small as possible.

I also wanted to fix the issue with returning the window to the stack from full-screen state in this PR, but it appears to be much more complicated than I initially imagined and has a lot of edge cases.


src/shell/mod.rs
```diff
@@ -4657,6 +4657,8 @@ impl Shell {
                 stack.remove_window(&surface);
                 surface
             } else {
+                mapped.set_fullscreen(true);
```
Drakulix, line 4685: Hmm, I wonder if it would make more sense to change `FloatingLayout::unmap` to not call configure and give the caller control to as to when this should be set. Though `map_internal` seems to have the same issue and this might end up being a bigger refactor than necessary to fix this case.

Could you add a comment to this line explaining, that this needs to be called before `sticky_layer.unmap`?

src/shell/mod.rs
```diff
@@ -4698,6 +4700,8 @@ impl Shell {
                 return None;
             }
 
+            mapped.set_fullscreen(true);
```
Drakulix, line 4732: Same here.

src/shell/mod.rs
```diff
@@ -2593,6 +2594,12 @@ impl Shell {
                         fullscreen_geometry,
                         true,
                     );
+                } else if let Some(corners) = was_snapped {
+                    *window.floating_tiled.lock().unwrap() = Some(corners);
+                    window.set_tiled(true);
+                    let snapped_geo = workspace.floating_layer.snapped_geometry(&corners);
+                    window.set_geometry(snapped_geo.to_global(&workspace.output));
+                    window.configure();
```
Drakulix: I'd prefer to not duplicate this logic. Could we add this as a method to `FloatingLayout`? Something like `FloatingLayout::snap_to_corner(&self, mapped: &CosmicMapped)`? (And then use this in other places as well, but feel free to leave that as a TODO for us.)

src/shell/workspace.rs
```diff
@@ -1133,6 +1138,12 @@ impl Workspace {
                     });
                     std::mem::drop(state);
                     self.floating_layer.map_maximized(window, geometry, true);
+                } else if let Some(corners) = previous.was_snapped {
+                    *window.floating_tiled.lock().unwrap() = Some(corners);
+                    window.set_tiled(true);
+                    let snapped_geo = self.floating_layer.snapped_geometry(&corners);
+                    window.set_geometry(snapped_geo.to_global(&self.output));
+                    window.configure();
```
Drakulix: If you move the logic above, please re-use `self.floating_layer.snap_to_corner` here as well.

src/shell/mod.rs
```diff
@@ -3348,13 +3348,31 @@ impl Shell {
...
+                });
+                to_workspace
+                    .floating_layer
+                    .map_maximized(mapped.clone(), geometry, false);
+            } else if let Some(corners) = was_snapped {
+                *mapped.floating_tiled.lock().unwrap() = Some(corners);
+                mapped.set_tiled(true);
+                let snapped_geo = to_workspace.floating_layer.snapped_geometry(&corners);
+                mapped.set_geometry(snapped_geo.to_global(&to_workspace.output));
+                mapped.configure();
```
Drakulix: Same here (`self.floating_Layer.snap_to_corner`)

Drakulix (commented):
Thanks! These are very welcome and easy to review bugfixes.

I added some comments, as I would like to see some of the new code to be moved around slightly, if you don't mind.

src/shell/mod.rs
mayakwd: Of course, added the comment for the later refactoring.

src/shell/mod.rs
mayakwd: Done.

src/shell/mod.rs
mayakwd: Done, extracted logic to FloatingLayout, and replaced duplicated code with the usage of the new method everywhere.

src/shell/workspace.rs
mayakwd: Done :-)

src/shell/mod.rs
mayakwd: And done :-)

Drakulix (approved):
Thanks! Code-wise this looks great now. Just needs a QA review.

leviport (approved):

rhessus:
@Drakulix There is a small issue with this fix. When I move a maximized window between two workspaces, the destination workspace does not apply the maximized style to the panel and dock, and the source workspace keeps that style instead (opaque/transparent panel and dock). Not sure if that’s very clear…

https://github.com/user-attachments/assets/3b2e41fc-ceaa-4b64-87a6-95c1aa1f863f

Milli189:
there is an issue with this pull request, whenever I fullscreen a chromium based browser, then the scrolling stops being smooth and goes back to 60hz, and also when I maximize a chromium based browser or I fullscreen a video and I move the cursor to the top of the monitor to activate the "X" that appears on chromium based browsers to remove the fullscreen, it doesn't always happen and the anymation can be laggy, which doesn't happen on KDE Plasma

Issue #2035, opened 2026-01-29
Window resizing is weird and backwards in JetBrains IDEs


dkonigsberg:
So I've reproduced this issue in Android Studio and CLion, both of which are (or are based on) development IDEs originally from JetBrains.

When I try resizing the window from the upper-left corner, the window actually resizes from the lower-right.
Along the same lines, grabbing the left edge resizes the right edge, and grabbing the top edge resizes the bottom edge.

Its also quite difficult in general to grab the window edge for resizing in the first place.

Here's a little video clip that should hopefully illustrate the issue:

https://github.com/user-attachments/assets/057bc128-0a3b-4fd8-ac1e-4341e5a73b48

I'm running the latest cosmic-comp as of today (0.1~1769425738~24.04~1dfc948) on Pop!_OS 24.04.

LiamillionSS:
The rest of the Jetbrains Suite too most likely, since Rider and RustRover do the same.

Until now, the Steam client was the only one reported doing this https://github.com/pop-os/cosmic-comp/issues/1526

dkonigsberg:
Just wanted to add another observation here...
If I make the CLion (the JetBrains IDE that I use the most) run in "experimental" Wayland mode, this problem goes away.  The rendering still feels glitchy when I'm resizing the window, but at least resizing itself isn't wonky anymore.
I've also learned that JetBrains is planning to make Wayland mode enabled by default (when running on Wayland) in their upcoming 2026 releases, so this problem (as written) could end up going away on its own.

However...  This probably is uncovering some sort of weird bug between XWayland and Cosmic, so its still worth at least investigating.

peterkrull:
The cause seems to be different than what was happening to Steam. JetBrains products resize by requesting a specific position, which is handled in:

https://github.com/pop-os/cosmic-comp/blob/c569a35fc9d829874594366e77ae2ded55aa0eaa/src/xwayland.rs#L901

But, due to:

https://github.com/pop-os/cosmic-comp/blob/c569a35fc9d829874594366e77ae2ded55aa0eaa/src/xwayland.rs#L911-L912

The position is ignored, and only the size is changed.

PR #2107, opened 2026-02-16
(fix) Reject resize_request on surfaces which are already being resized

peterkrull:
Addresses #1526 ~~and #2035~~

I added some log messages, and saw that certain applications (like Stream) would start two resize requests, creating a corrupted state. This patch just rejects the second resize event, if we are already resizing.


Drakulix (approved):
Seems sensible. Thanks!

peterkrull:
Hmm, I had only tested this on the steam client, where it solves the issue, but it still seems broken on Clion..

peterkrull:
On further review it seems that JetBrains products handle resizing differently, but the symptom of top/left resizing acting oppositely is similar. Specifically, they resize by requesting a specific position, but due to

https://github.com/pop-os/cosmic-comp/blob/c569a35fc9d829874594366e77ae2ded55aa0eaa/src/xwayland.rs#L911-L912

The position is ignored, and only the size is changed.

Drakulix:
> On further review it seems that JetBrains products handle resizing differently, but the symptom of top/left resizing acting oppositely is similar. Specifically, they resize by requesting a specific position, but due to
> 
> https://github.com/pop-os/cosmic-comp/blob/c569a35fc9d829874594366e77ae2ded55aa0eaa/src/xwayland.rs#L911-L912
> 
> The position is ignored, and only the size is changed.

So what is the effect here? Does the window still change it's size? Or is the JetBrains window waiting for the new position? We obviously don't want the window to change it's position in tiling mode.

peterkrull:
> > On further review it seems that JetBrains products handle resizing differently, but the symptom of top/left resizing acting oppositely is similar. Specifically, they resize by requesting a specific position, but due to
> > https://github.com/pop-os/cosmic-comp/blob/c569a35fc9d829874594366e77ae2ded55aa0eaa/src/xwayland.rs#L911-L912
> > 
> > The position is ignored, and only the size is changed.
> 
> So what is the effect here? Does the window still change it's size? Or is the JetBrains window waiting for the new position? We obviously don't want the window to change it's position in tiling mode.

It is visually the same effect. Top left corner is anchored since its requested position is ignored, but the size changes. 

https://github.com/user-attachments/assets/05eb02d1-ff88-433a-8d3a-378926749901

Drakulix:
> > > On further review it seems that JetBrains products handle resizing differently, but the symptom of top/left resizing acting oppositely is similar. Specifically, they resize by requesting a specific position, but due to
> > > https://github.com/pop-os/cosmic-comp/blob/c569a35fc9d829874594366e77ae2ded55aa0eaa/src/xwayland.rs#L911-L912
> > > 
> > > The position is ignored, and only the size is changed.
> > 
> > 
> > So what is the effect here? Does the window still change it's size? Or is the JetBrains window waiting for the new position? We obviously don't want the window to change it's position in tiling mode.
> 
> It is visually the same effect. Top left corner is anchored since its requested position is ignored, but the size changes.
> 2026-02-17.16-40-38.mp4

Hmm, maybe we can relax the rule a bit in that case. Something like if the size changes, also allow position changes for floating windows, but reject pure position updates?

We don't want X11 apps to arbitrarily move themselves, as that has been abused in the past, but I'd like to make sure JetBrains IDEs are working somewhat reasonably well.

hojjatabdollahi:
For what it's worth you can enable wayland support in Jetbrain IDEs which solves this issue (and also fractional scaling issues). And they're gonna make that the default in the upcoming version:
https://blog.jetbrains.com/platform/2026/02/wayland-by-default-in-2026-1-eap/

(Not saying we shouldn't fix the issue)

leviport (approved):

PR #2111, opened 2026-02-17
(fix) Do not create duplicate PendingWindow

### LINKED ISSUE #2074: Steam Window Duplication When Moving Between Monitors
- Steam windows sometimes duplicates.  Both windows can move different spaces but windows appear to be linked to the same process: they show the same content, scale together, and display the same pages simultaneously.

- Even when clicking different sections (for example, the Steam Store), both windows update to the same view instead of behaving independently.

- i dont know how it happens but it usually happens when i take the steam from second monitor to main monitor and its usually waiting in fullscreen 
- it happens once or twice a week so its pretty rare but its with me from the 1.0.0 release and so on

Image

peterkrull:
Add a guard to reject map_window_request for X11 windows which are already pending
Resolves #2074 & https://github.com/pop-os/cosmic-epoch/issues/1151

The removal of the `set_mapped(false) -> set_mapped(true)` cycle fixed some parts of the Steam UI being black on restore. I do not think it handled the rapid unmap/remap cycle nicely. I can see that [this hack](https://github.com/pop-os/cosmic-comp/commit/3b9d0ce3) was introduced as part of a larger PR #603 so the change may introduce regressions. However, testing TF2, it seems to (most of the times?) restore to fullscreen just fine after being minimized.

### DIFF
--- a/src/xwayland.rs
+++ b/src/xwayland.rs
@@ -806,7 +806,8 @@ impl XwmHandler for State {
         }
 
         let seat = shell.seats.last_active().clone();
-        if let Some(context) = window.startup_id()
+        if let Some(context) = window
+            .startup_id()
             .map(XdgActivationToken::from)
             .and_then(|token| self.common.xdg_activation_state.data_for_token(&token))
             .and_then(|data| data.user_data.get::<ActivationContext>())


Drakulix:
I think there is already https://github.com/pop-os/cosmic-comp/pull/1902 covering the pending_windows part, that just needs a review by QA.

src/shell/element/surface.rs
```diff
@@ -429,10 +429,6 @@ impl CosmicSurface {
             .store(minimized, Ordering::SeqCst);
         if let WindowSurface::X11(surface) = self.0.underlying_surface() {
             let _ = surface.set_hidden(minimized);
-            if !minimized {
-                let _ = surface.set_mapped(false);
-                let _ = surface.set_mapped(true);
-            }
```
Drakulix, line 435: Hmm, we can't simply remove this. Many windows need an unmap/map cycle during minimize, the reason we delay this til restore is, that Xwayland stop rendering unmapped windows, so we don't have a preview for the minimized window.

Have to tried to take a look into what other WMs are doing here?

peterkrull:
> I think there is already #1902 covering the pending_windows part, that just needs a review by QA.

You are right that does seem to cover it. Did not see that one

src/shell/element/surface.rs
peterkrull: Do I understand correctly that it is primarily because the "Minimized Windows" applet will not have a preview if we unmap immediately? Since the compositor should just be able to grab the latest image prior to unmapping, which is what it seems like some other compositors do to hold on to previews.

Drakulix: Yes, we would have to add caching logic to cosmic-comp for that though, which we are currently lacking (and is also not entirely straight forward for multi-gpu scenarios) and then wire that up to our ext-image-capture implementation.

peterkrull: I see, it is not a quick fix then.

Issue #2115, opened 2026-02-17
fix: preserve wl_output across connector disconnect/reconnect


modelmiser:
## Problem

When an HDMI monitor power-cycles (or cable is unplugged/replugged), the kernel HPD de-assert produces the same udev event as a cable unplug. cosmic-comp destroys the Output (wl_output global, workspace state, client bindings) on disconnect and rebuilds everything on reconnect. This causes browsers' D-Bus sessions to get stuck in Auth state, leaving tabs unable to render.

The kernel cannot distinguish HDMI HPD de-assertion (monitor sleep) from cable unplug — both produce the same udev "Changed" event. So the fix must handle both identically.

## Fix

Track disconnected connectors as "suspended" instead of removing their Output. The Surface (render thread, DRM resources) is still destroyed since the hardware is gone, but the wl_output global, workspace assignments, and client bindings survive. On reconnect, `connector_added()` finds the existing Output and creates a fresh Surface.

| Component | Before | After |
|-----------|--------|-------|
| wl_output global | Destroyed + recreated | **Persists** |
| Client surface bindings | Severed | **Intact** |
| Workspace assignments | Lost + migrated | **Preserved** |
| Window positions | Rearranged | **Preserved** |
| Surface (render thread) | Destroyed + recreated | Destroyed + recreated (unchanged) |

## Implementation

Two files, ~28 lines net. Full diff: https://github.com/modelmiser/cosmic-comp/commit/HEAD

*Note: I'm unable to create a PR because GitHub has restricted forking on my account. The fix is at https://github.com/modelmiser/cosmic-comp — the single commit on top of current master. Happy to collaborate on getting this merged however works best.*

## Testing

- [x] `cargo build --release` — clean, zero warnings
- [x] Power-cycle HDMI monitor via external power button — Firefox/Librewolf tabs continue rendering
- [x] Unplug/replug HDMI cable — Firefox/Librewolf tabs continue rendering

modelmiser:
Superseded by PR #2116 — forking restriction has been resolved.

Issue #906, opened 2024-10-05
Crash when disconnecting/reconnecting monitor


rmrfbugs:
Distro: Arch Linux
Version: cosmic-comp 1.0.0.alpha.2-1

When reconnecting the monitor (happens when turning monitor back on according to timestamp) cosmic-comp crashes, taking all open programs with it.

Logs:

```
Oct 05 13:45:24 cosmic cosmic-comp[1501]: thread 'main' panicked at 'called `Option::unwrap()` on a `None` value': src/shell/mod.rs:973
                                             0: <unknown>
                                             1: <unknown>
                                             2: <unknown>
                                             3: <unknown>
                                             4: <unknown>
                                             5: <unknown>
                                             6: <unknown>
                                             7: <unknown>
                                             8: <unknown>
                                             9: <unknown>
                                            10: <unknown>
                                            11: <unknown>
                                            12: <unknown>
                                            13: <unknown>
                                            14: <unknown>
                                            15: <unknown>
                                            16: <unknown>
                                            17: <unknown>
                                            18: <unknown>
                                            19: <unknown>
                                            20: __libc_start_main
                                            21: <unknown>
Oct 05 13:45:24 cosmic cosmic-comp[1501]: thread 'surface-HDMI-A-1' panicked at 'called `Result::unwrap()` on an `Err` value: PoisonError { .. }': src/backend/kms/surface/mod.rs:1076
                                             0: <unknown>
                                             1: <unknown>
                                             2: <unknown>
                                             3: <unknown>
                                             4: <unknown>
                                             5: <unknown>
                                             6: <unknown>
                                             7: <unknown>
                                             8: <unknown>
                                             9: <unknown>
                                            10: <unknown>
                                            11: <unknown>
                                            12: <unknown>
                                            13: <unknown>
                                            14: <unknown>
                                            15: <unknown>
                                            16: <unknown>
                                            17: <unknown>
Oct 05 13:45:24 cosmic xdg-desktop-por[1752]: Error reading events from display: Broken pipe
Oct 05 13:45:24 cosmic cosmic-session[1476]: Io error: Broken pipe (os error 32)
Oct 05 13:45:24 cosmic systemd[1463]: xdg-desktop-portal-gtk.service: Main process exited, code=exited, status=1/FAILURE
Oct 05 13:45:24 cosmic systemd[1463]: xdg-desktop-portal-gtk.service: Failed with result 'exit-code'.
[...]

PR #2116, opened 2026-02-18
fix: preserve wl_output across connector disconnect/reconnect

modelmiser:

## Summary

- When HDMI monitors power-cycle (or cable is unplugged/replugged), the kernel HPD de-assert produces the same udev event as a cable unplug. Previously cosmic-comp destroyed the Output (wl_output global, workspace state, client bindings) on disconnect and rebuilt everything on reconnect, causing browsers' D-Bus sessions to get stuck in Auth state with unrecoverable rendering failures.
- Track disconnected connectors as "suspended" instead of removing their Output. The Surface (render thread, DRM resources) is still destroyed since the hardware is gone, but the wl_output global, workspace assignments, and client bindings all survive.
- On reconnect, `connector_added()` finds the existing Output (via `outputs.get(&conn).cloned()` at device.rs:547) and creates a fresh Surface for it.

## What changes

| Component | Before | After |
|-----------|--------|-------|
| wl_output global | Destroyed + recreated | **Persists** |
| Client surface bindings | Severed | **Intact** |
| Workspace assignments | Lost + migrated | **Preserved** |
| Layer surfaces (panels) | Closed + respawned | **Preserved** |
| Window positions | Rearranged | **Preserved** |
| Surface (render thread) | Destroyed + recreated | Destroyed + recreated (unchanged) |

## Implementation

A `suspended_connectors: HashSet<connector::Handle>` field is added to `InnerDevice`. When a connector disappears:
- Surface is destroyed (DRM resources are invalid) — unchanged
- Output is **kept alive** and connector is marked suspended — **new**

When the connector reappears, it's reported as "added", `connector_added()` reuses the existing Output, a new Surface is created, and the suspended flag is cleared.

Suspend-without-timeout was chosen over a debounce approach (wait N seconds, then destroy). A timeout adds complexity and race conditions for minimal benefit — permanently-removed connectors are cleaned up when `device_removed()` handles full GPU removal, which is the only case where the hardware is truly gone rather than power-cycled.

Two files changed, ~28 lines added (~18 net). Edge cases handled:
- Repeated udev events while disconnected (skipped)
- Full GPU device removal (`device_removed()` cleans up suspended outputs)
- `enumerate_surfaces()` skips suspended connectors (avoids `bail!("Missing crtc")`)

## Test plan

- [x] `cargo build --release` — clean, zero warnings
- [x] Power-cycle HDMI monitor via external power button — Firefox/Librewolf tabs continue rendering
- [x] Unplug/replug HDMI cable — Firefox/Librewolf tabs continue rendering
- [ ] VT switch while monitor is disconnected
- [ ] Different monitor plugged into same connector

Supersedes #2115, which was filed as an issue with the patch linked externally because GitHub forking was restricted at the time. This PR contains the same fix submitted as a proper PR. Related: #1463, #906 — previous fixes made the destroy/recreate path crash-free; this PR avoids the destroy path entirely for transient disconnects.

### DIFF
--- a/src/backend/kms/device.rs
+++ b/src/backend/kms/device.rs
@@ -120,6 +120,7 @@ pub struct InnerDevice {
     pub leasing_global: Option<DrmLeaseState>,
     pub active_leases: Vec<DrmLease>,
     pub active_clients: HashSet<ClientId>,
+    pub suspended_connectors: HashSet<connector::Handle>,
 }
 
 impl fmt::Debug for InnerDevice {
@@ -136,6 +137,7 @@ impl fmt::Debug for InnerDevice {
             .field("leasing_global", &self.leasing_global)
             .field("active_leases", &self.active_leases)
             .field("active_clients", &self.active_clients.len())
+            .field("suspended_connectors", &self.suspended_connectors)
             .finish()
     }
 }
@@ -347,6 +349,7 @@ impl State {
                 leasing_global,
                 active_leases: Vec::new(),
                 active_clients: HashSet::new(),
+                suspended_connectors: HashSet::new(),
             },
 
             supports_atomic,
@@ -405,7 +408,7 @@ impl State {
         }
 
         let drm_node = DrmNode::from_dev_id(dev)?;
-        let mut outputs_removed = Vec::new();
+        let outputs_removed = Vec::<Output>::new();
         let mut outputs_added = Vec::new();
 
         {
@@ -415,6 +418,11 @@ impl State {
 
                 let mut w = self.common.shell.read().global_space().size.w as u32;
                 for conn in changes.removed {
+                    // Skip already-suspended connectors (repeated udev events)
+                    if device.inner.suspended_connectors.contains(&conn) {
+                        continue;
+                    }
+
                     // contains conns with updated crtcs, just drop the surface and re-create
                     if let Some(pos) = device
                         .inner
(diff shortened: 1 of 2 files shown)


hojjatabdollahi:
How do you replicate this error? My browser or the layout is fine if I disconnect/reconnect or power cycle my external monitors. Is this happening on a desktop or a laptop?

modelmiser:
Minisforum MS-01 mini workstation, Intel 13th gen i9-13900H (Iris Xe, Raptor Lake-P), mainboard HDMI. NVIDIA RTX 4000 SFF Ada that is compute-only — not driving any displays.

GPU topology (no cross-GPU interaction):
```
card1 (i915, Intel Iris Xe):  HDMI-A-1 connected/enabled (only connector)
card2 (nvidia, RTX 4000 SFF): DP-5..8 all disconnected, Display Active: Disabled
```
The NVIDIA GPU has no display role — all rendering goes through i915/HDMI-A-1.

Two reproduction paths:
1. **Physical power-cycle**: monitor has a physical power button — off, wait 30 seconds to a minute, back on. After reconnect, **new** Gecko browser tabs (Firefox, Librewolf) render blank. Existing tabs are fine.
2. **COSMIC power settings**: Settings → Power & Battery → Power Saving Options → "Turn off the screen after" set to anything other than Never. When the screen turns off, wait 30 seconds to a minute. After it turns back on, same result — **new** tabs render blank.

If your monitors recover cleanly, it could be connection type (DP vs HDMI), GPU driver, or how the monitor handles the power cycle. Happy to gather logs or test further if that would help narrow it down.

Drakulix:
So this just indefinitely keeps wl_outputs alive? I am sorry this is no solution, clients need to be able to handle wl_outputs appearing and disappearing, that is part of the core wayland protocol.

modelmiser:
Thanks for the feedback — understood that preserving `wl_output` indefinitely isn't the right compositor-side fix, and that clients are expected to handle output removal and re-addition per the Wayland protocol. I'll close this PR.

Wanted to close the loop on the client side: I've investigated the Firefox/Gecko failure and filed it upstream with full diagnostic evidence:

**https://bugzilla.mozilla.org/show_bug.cgi?id=2018866**

The failure mode on my hardware (i915/HDMI-A-1, details in the bug): after the `wl_output` destroy/recreate sequence, Firefox permanently disables hardware compositing for the session (`layers.acceleration.disabled=true`), and the new `wl_output`'s geometry/scale events appear not to be applied — `Display0` ends up recorded as `0x0@60Hz scales:0.000000|0.000000`. New tabs render blank; only a restart recovers. The GPU process also appears to crash and restart at the moment of reconnect.

Note that @hojjatabdollahi doesn't reproduce this, so it may be hardware- or connection-dependent. Linking here in case other COSMIC users report similar symptoms — the Bugzilla bug has the details.

PR #2117, opened 2026-02-18
fix(input): Fix pointer sticking to screen edges during cursor movement

### LINKED ISSUE #2102: Cursor getting stuck at display edge when moving from left to right
When I move my cursor from my left to my right display, it seems to get caught on the display's edge. If I move my mouse more quickly it does not happen.

Maybe it's a `< 0` vs `<= 0` issue?

### Setup
- Cosmic 1.0.0 on Fedora 43.
- My monitors have the same resolution (1920x1080) and are at 100% scale. They are arranged next to each other.
-  It only seems to happen from left to right.

rhessus:
This PR fixes the pointer behavior reported in the linked issue where the cursor becomes stuck on screen edges during normal movement when multiple displays are connected.

The problem can be observed when moving the mouse left ↔ right as well as up ↕ down. Under certain display arrangements, the pointer incorrectly clamps to output boundaries instead of transitioning smoothly, resulting in the cursor appearing to "stick" to the edges.

Fixes #2102

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -59,7 +59,7 @@ use smithay::{
     reexports::{
         input::Device as InputDevice, wayland_server::protocol::wl_shm::Format as ShmFormat,
     },
-    utils::{Point, Rectangle, SERIAL_COUNTER, Serial},
+    utils::{Coordinate, Point, Rectangle, SERIAL_COUNTER, Serial},
     wayland::{
         image_copy_capture::{BufferConstraints, CursorSessionRef},
         keyboard_shortcuts_inhibit::KeyboardShortcutsInhibitorSeat,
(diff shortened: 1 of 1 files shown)


rhessus:
**Mouse position clamping at output edges**

- When the pointer goes past the **right** edge of an output, it is clamped to the rightmost column of that output (right edge − 1.0) **only if** there is no other output to the right of it. If there is an output on the right, the pointer is allowed to move into it.
- When the pointer goes past the **bottom** edge of an output, it is clamped to the bottommost row (bottom edge − 1.0) **only if** there is no other output below. If there is an output below, the pointer is allowed to move into it.

So clamping to the right/bottom edge (−1.0) is applied only when that edge is the actual boundary of the output layout (no adjacent output on that side), avoiding the pointer getting stuck at the edge when multiple outputs are arranged side by side or stacked.

---

Drakulix:
Can you post an example of an outputs.ron file that re-produces this bug? Outputs should be placed without any gaps between them, so I am failing to see how this would occur without explicitly messing with the configuration file.

rhessus:
> Can you post an example of an outputs.ron file that re-produces this bug? Outputs should be placed without any gaps between them, so I am failing to see how this would occur without explicitly messing with the configuration file.

This PR is related to the change introduced in PR 1907 ([https://github.com/pop-os/cosmic-comp/pull/1907](https://github.com/pop-os/cosmic-comp/pull/1907)), which modified the mouse clamping on the right and bottom edges of the screen to use max - 1.0.

image

I am finishing resolving the issue and will provide full reproduction details. But to reproduce the issue, simply use two displays placed side by side and move the mouse very slowly from the left display to the right display (or from the top display to the bottom display). You will notice that the mouse cursor sticks to the boundary between the two displays. This does not occur if the mouse is moved very quickly (delta.x|y > 1.0).

Correcting this issue is somewhat more complex than the changes I have made so far. I will return with an appropriate solution as soon as possible.

szpolny:
Addresses #2407

PR #2131, opened 2026-02-23
fix: use text widget instead of custom tab_text

hojjatabdollahi:
~~Draft while waiting for https://github.com/pop-os/libcosmic/pull/1140~~

Ellipsizes the window title in SSD and tab text for stacked windows.

image

___


src/shell/element/stack/tab.rs
```diff
@@ -194,9 +193,13 @@ impl<Message: TabMessage + 'static> Tab<Message> {
                 .padding([2, 4])
                 .center_y(Length::Fill)
                 .into(),
-            tab_text(self.title, self.active)
+            cosmic::widget::text(self.title)
                 .font(self.font)
-                .font_size(14.0)
+                .size(14.0)
```
git-f0x: Minor nit, it might be better to add the text widget next to the existing `cosmic::widget` imports at the start (just so the line isn't as long), and instead of specifying size here, use `text::body`.

git-f0x: Also, maybe the `font` field isn't needed for the Tab struct, since I don't think it can be modified externally (and the text widget should automatically use the default font).

hojjatabdollahi: I'll apply the changes when libcosmic PR goes through, and update the dependencies too.

src/shell/element/stack/tab.rs
```diff
@@ -194,9 +192,12 @@ impl<Message: TabMessage + 'static> Tab<Message> {
                 .padding([2, 4])
                 .center_y(Length::Fill)
                 .into(),
-            tab_text(self.title, self.active)
-                .font(self.font)
-                .font_size(14.0)
+            text_body(self.title)
+                .size(14.0)
```
git-f0x: You don't need to specify the size here when using `text::body`.
Also, why did you add `text` as an `iced_core` import by itself, rather than keeping it as it was?
It would be cleaner to have it as it was before, but with importing `text` in `cosmic::widget` (and then using `text::body`, rather than importing `text::body as text_body`).

hojjatabdollahi: We probably should've reexprted the stuff in `libcosmic::widget::text` so that you don't need to configure `text::body` with `iced_core::text::Wrapping`!

git-f0x:
Actually, specifying the font with `.font(self.font)` might be needed, since it's used in `tabs.rs` to [set the font to semibold](https://github.com/pop-os/cosmic-comp/blob/253b44f0b8920a251c8970d463d9fe3995ecf0c2/src/shell/element/stack/tabs.rs#L151). Sorry about that.

hojjatabdollahi:
superseded by https://github.com/pop-os/cosmic-comp/pull/2180

PR #2161, opened 2026-03-06
feat: add nearest-neighbor filtering option for accessibility zoom

### LINKED ISSUE #1671: [Feature Request] Screen Magnifier: Option to disable bilinear filtering
Hi!

I was delighted when I tried COSMIC for the first time and found that the zoom gesture (Super+scroll) JUST WORKS by default! That's such a good move. It is a real pain for someone that depends on this feature to find and enable it otherwise. (I was ready to get my physical magnifier to find the setting).

By the way: often times when people see me using this feature on macOS, they ask how to enable it. It turns out it's very useful for pixel artists/UI designers/illustrators as well.

One tiny thing would make it much more useful and pleasant to use: The ability to get crisp, unfiltered (unsmoothed) pixels. Looking at the blurry character in my terminal for a long time makes me feel dizzy/drunk. And for pixel artists it's super important to be able to see the actual pixels. Please don't do it like Plasma though, where at higher zoom levels a small gap appears between pixels, like a gird. (not sure if that's intentional or a bug)

Thanks for reading and considering!
Jan

modelmiser:
## Summary

Adds a `smooth_images` option to `ZoomConfig` (default: `true`) that controls texture filtering during zoom rendering.

- **`true` (default)**: bilinear filtering (current behavior, smooth but blurry)
- **`false`**: nearest-neighbor filtering (crisp pixel boundaries)

Both upscale and downscale filters are set, since on HiDPI displays surface buffers may be at higher resolution than the zoomed viewport at lower zoom levels.

Filter state is restored after rendering to avoid affecting other render contexts.

### Live toggle support

Changing `smooth_images` at runtime (via the companion settings toggle) takes effect immediately. This required two additional fixes:

- **Schedule render on config change**: the config watcher now triggers a redraw for all outputs when zoom settings change.
- **Force full re-render on filter change**: GL texture filter state is invisible to the damage tracker (no element geometry or content changes), so the damage tracker would skip rendering entirely. Fixed by resetting buffer ages when `smooth_images` changes, which forces the damage tracker to treat all buffers as having no history.

Companion PR for the settings UI toggle: pop-os/cosmic-settings#1892.

Closes #1671.

## Test plan

- [x] Tested on 3840x2160 display at 175% fractional scale
- [x] Verified crisp pixels visible when zoomed above native scale
- [x] Verified smooth (default) behavior unchanged
- [x] Verified config backward compatibility (existing configs without `smooth_images` deserialize correctly via serde default)
- [x] Verified non-zoomed rendering unaffected when crisp mode enabled (no pixelation on other windows/outputs)
- [x] Verified live toggle via settings UI switches between smooth and crisp immediately

### DIFF
--- a/cosmic-comp-config/src/lib.rs
+++ b/cosmic-comp-config/src/lib.rs
@@ -181,6 +181,10 @@ fn default_repeat_delay() -> u32 {
     600
 }
 
+fn default_smooth_images() -> bool {
+    true
+}
+
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Deserialize, Serialize)]
 pub struct ZoomConfig {
     pub start_on_login: bool,
@@ -188,6 +192,10 @@ pub struct ZoomConfig {
     pub increment: u32,
     pub view_moves: ZoomMovement,
     pub enable_mouse_zoom_shortcuts: bool,
+    /// Use bilinear filtering when zoomed (smooth but blurry).
+    /// When false, use nearest-neighbor filtering (crisp pixels).
+    #[serde(default = "default_smooth_images")]
+    pub smooth_images: bool,
 }
 
 impl Default for ZoomConfig {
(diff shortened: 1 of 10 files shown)


jackpot51:
If I wanted Claude to make this change I would direct it myself

PR #2163, opened 2026-03-07
fix: use f64 arithmetic for zoom focal point calculation

### LINKED ISSUE #1503: Magnifier breaks at >2x zoom when Cursor Tracking is set to Center — cursor no longer centered, screen reframes incorrectly
When using the magnifier with Cursor Tracking set to "Center", zoom works correctly up to 2x.

At 3x zoom and beyond, the viewport breaks:
The screen abruptly reframes or shifts.
The cursor is no longer centered.
In some cases, the cursor ends up stuck at the screen edge or even outside the visible magnified area.
Tracking behavior becomes unreliable.

---

To Reproduce

1. Go to COSMIC Settings → Accessibility → Magnifier.
2. Set Cursor Tracking Mode to: Center.
3. Increase zoom level to 3x or higher.
4. Move the mouse.

modelmiser:

## Summary

Fixes the magnifier viewport calculation for **Centered** and **OnEdge** cursor tracking modes at zoom levels other than 2x.

**Root cause:** The focal point scale factor was computed using integer division (`i32 / i32`), which truncates when the result isn't an exact integer. The divisor is `output_width - zoomed_width` (where `zoomed_width = output_width / level`). At 2x zoom this happened to work (e.g. `1920 / (1920 - 960) = 2`), but at 3x: `1920 / (1920 - 640) = 1` instead of the correct `1.5`. With fractional display scaling (e.g. 175%), odd logical output dimensions cause truncation at even lower zoom levels.

A second bug in `OutputZoomState::new()` used `.upscale(level)` instead of the correct `.upscale(level / (level - 1.0))` for the initial Centered focal point — also only correct at exactly 2x (where `level == level / (level - 1.0)`).

## Changes

- `OutputZoomState::new()` Centered branch: correct the scale factor from `level` to `level / (level - 1.0)`
- `update_focal_point()` OnEdge branch: replace integer division with f64 division
- `update_focal_point()` Centered branch: compute focal point directly from f64 cursor position, eliminating i32 intermediate conversions

## Test plan

- [x] Tested with Centered cursor tracking at 125%, 150%, 175%, 200%, 250%, 300% zoom on a 175% HiDPI display
- [x] Verified cursor stays centered in viewport (previously drifted at all levels except 200%)
- [x] Verified OnEdge mode still works correctly
- [x] Verified Continuously mode unaffected (it doesn't use this code path)
- [x] `cargo check` passes

## Note

After this fix, a small residual cursor jitter remains visible in Centered mode (cursor briefly glitches off-center, even during slow movement). This appears to be a separate issue in the rendering pipeline, not related to the focal point calculation fixed here. The focal point itself is now mathematically correct.

Closes #1503

---

- [x] I understand the changes in this pull request.
- [x] I have tested the changes.
- [x] I certify that this contribution is made under the terms of the applicable license and I agree to the Developer Certificate of Origin (https://developercertificate.org/).
- [x] This PR was developed with AI assistance (Claude).

### DIFF
--- a/src/shell/zoom.rs
+++ b/src/shell/zoom.rs
@@ -86,10 +86,11 @@ impl OutputZoomState {
                     zoomed_output_geometry.loc =
                         cursor_position - zoomed_output_geometry.size.downscale(2.).to_point();
 
+                    let scale = if level <= 1.0 { 1.0 } else { level / (level - 1.0) };
                     let mut focal_point = zoomed_output_geometry
                         .loc
                         .to_local(output)
-                        .upscale(level)
+                        .upscale(scale)
                         .to_global(output);
                     focal_point.x = focal_point.x.clamp(
                         output_geometry.loc.x,
@@ -260,6 +261,7 @@ impl ZoomState {
         original_position: Point<f64, Global>,
         movement: ZoomMovement,
     ) {
+        let cursor_f64 = cursor_position;
         let cursor_position = cursor_position.to_i32_round();
         let original_position = original_position.to_i32_round();
         let output_geometry = output.geometry();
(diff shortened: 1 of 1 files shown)


modelmiser:
**Instrumentation data confirming root cause**

I built an instrumented binary with per-frame logging and captured 10 seconds of cursor movement in Centered mode at 1.8x zoom on a 175% HiDPI display.

Key findings from ~460 frames of data:

**1. Integer division produces wrong tracking gain**
The scale factor `output_w / (output_w - zoomed_w)` = `2257 / 1003` = **2** (truncated) instead of the correct **2.25** (`zoom/(zoom-1)`). This means the focal point only moves 89% of what it should per cursor pixel.

**2. Systematic off-center — not a rendering artifact**
The cursor-to-focal-point offset was consistently **120–250px** depending on cursor position, with the viewport shifted left/up. This matches the integer division error: `(correct_scale - truncated_scale) * viewport_loc` = `0.25 * viewport_loc`.

**3. "Jitter" is actually speed-proportional tracking lag**
Faster cursor movement produces larger per-frame undershoot, causing visible content shifting. It's not oscillation — it's the viewport systematically falling behind. When the cursor reverses direction, the lag reverses too, which looks like jitter.

The predicted focal point using integer factor 2 matched the actual focal point within ±2px (residual from `to_i32_round()` in `zoomed_geometry()`). The f64 fix in this PR should eliminate the off-center behavior and the speed-proportional tracking lag.

The small residual jitter mentioned in the PR note is likely from the remaining i32 roundtrips (`cursor_position.to_i32_round()` at the top of `update_focal_point()`, and `zoomed_geometry()` returning `Rectangle<i32>`), which introduce sub-pixel rounding error amplified by the zoom factor — a separate, much smaller issue.

jackpot51:
If I wanted Claude to make this change I would direct it myself

Issue #715, opened 2024-08-12
[Req] Make xwayland optional


Quackdoc:
Currently if cosmic-comp is launched without xwayland it will crash out, it would be nice if xwayland was optional. cosmic-comp is quite low memory usage which makes it fairly nice for embedded usage, it's not as low as weston, but running a cosmic-comp with weston-terminal of a 256M system gave impressive results, so it would be nice if xwayland was optional

![image](https://github.com/user-attachments/assets/f626ebe7-dca1-48c0-9748-2ade40044a1e)

qoijjj:
+1 to this, other DEs have support for this. For example, for GNOME:

```
/usr/bin/gnome-shell --no-x11
```

```
/usr/bin/kwin_wayland_wrapper
```
(removed --xwayland)

It would be good for cosmic to have parity.

Don-Keehotee:
Not sure if this topic is discussed further elsewhere, I'm looking to build Cosmic sans Xwayland.

I've been building my compositors (`mutter` -> `sway/wlroots` -> `hyprland`) for 3 years now without Xwayland.

I'd really like to try out Cosmic, but unwilling to install Xwayland and all its deps.

RoyalOughtness:
@Don-Keehotee IMO those are two different requests. This is about allowing the compositor to start without xwayland, which doesn't necessarily mean making it buildable without xwayland

ids1024:
`cosmic-comp` appears to start fine without `Xwayland` present. If `XWayland::spawn` errors, it just logs `Failed to start Xwayland`.

Is anyone still seeing issues with this?

> which doesn't necessarily mean making it buildable without xwayland

We use `x11rb`, which provides a pure-Rust implementation of of an X11 client, so I don't think there are any X11-related system dependencies needed for building. (Not counting `libxkbcommon`, which is not X11 specific and is generally required by Wayland compositors.)

RoyalOughtness:
@ids1024 It would be useful to also have a flag to run it without xwayland even if xwayland is available, like kde or gnome (--no-x11)

Don-Keehotee:
> cosmic-comp appears to start fine without Xwayland present

Oh jeez, thanks for clarifying. I'd not even tried.

Built, installed, and ran Cosmic session successfully without Xwayland.

Arch still has it as a dep. If I pursue this further I'll open an issue with Arch packaging to make the dep optional.

[...]

PR #2171, opened 2026-03-09
feat: add --no-xwayland argument for running without Xwayland

RoyalOughtness:

Tested with an unmodified `/usr/bin/start-cosmic`, and with --no-xwayland appended like so:

```bash
# Run cosmic-session
if [[ -z "${DBUS_SESSION_BUS_ADDRESS}" ]]; then
    exec /usr/bin/dbus-run-session -- /usr/bin/cosmic-session cosmic-comp --no-xwayland
else
    exec /usr/bin/cosmic-session cosmic-comp --no-xwayland
fi
```

When `--no-xwayland` is appended, Xwayland is not started:

image

When `--no-xwayland` is not appended, Xwayland is started:

image

### DIFF
--- a/src/backend/kms/mod.rs
+++ b/src/backend/kms/mod.rs
@@ -168,9 +168,13 @@ pub fn init_backend(
         }
     }
 
-    // start x11
-    let primary = *state.backend.kms().primary_node.read().unwrap();
-    state.launch_xwayland(primary);
+    if state.common.with_xwayland {
+        // start x11
+        let primary = *state.backend.kms().primary_node.read().unwrap();
+        state.launch_xwayland(primary);
+    } else {
+        state.notify_ready();
+    }
 
     Ok(())
 }
--- a/src/backend/winit.rs
+++ b/src/backend/winit.rs
@@ -241,7 +241,12 @@ pub fn init_backend(
         }
         state.common.refresh();
     }
-    state.launch_xwayland(None);
+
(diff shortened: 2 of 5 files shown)


RoyalOughtness:
https://github.com/pop-os/cosmic-comp/issues/715

RoyalOughtness:
> Has no effect on the default COSMIC session in Pop!_OS, but does work to launch a COSMIC session without Xwayland (in which X11-only apps fail to start) when using the flag.
> 
> (While testing, a COSMIC session failed to launch after modifying `/usr/bin/start-cosmic` as described in the PR; but running `dbus-run-session` or `cosmic-session` manually with the flag did work.)

Odd, I wonder why that would be. You're on 1.0.8 I presume? And you added `cosmic-comp --no-xwayland` right? Not just `--no-xwayland`?

jacobgkau:
> And you added `cosmic-comp --no-xwayland` right? Not just `--no-xwayland`?

Ah, that's what I did wrong. Thanks for clarifying.

I'll go ahead and merge this, since that mystery's solved and this has both necessary approvals.

RoyalOughtness:
@jacobgkau np, I hit that as well. It's caused by this logic here: https://github.com/pop-os/cosmic-session/blob/master/src/main.rs#L124

It looks like it might be intentional behavior but it seems idiosyncratic if not buggy and error prone.

PR #2179, opened 2026-03-10
Frosted Glass (Blur) effect

Drakulix:
Draft for supporting ext-background-effect-v1 in cosmic-comp.

Depends on ongoing smithay work:
- [x] https://github.com/Smithay/smithay/pull/1910
- [x] https://github.com/Smithay/smithay/pull/1946

This also introduces a config value (for *testing*, you need to remove that manually) in `XDG_STATE/cosmic-comp/appearance_conf` for the blur-strength, but I am pretty sure we want to move this into the CosmicTheme. So this is also partially blocked on support in libcosmic.

This ended up being much more complex than anticipated, due to necessary and quite invasive changes to smithay's public api and damage-tracking.

Additionally this reworks all of the element-collecting code with heavy inspiration of https://github.com/niri-wm/niri/pull/3113. The blur-effect of the main-surface needs to not be affected by drop-shadows, but drop-shadows should still render below the main-surface. Which was easy enough for the floating layout, but when tiling we additionally want to make sure no drop-shadow can extend on-top of another tiled window, so I had to split the element-generation into "lower" and "top" surfaces to order all the drop-shadows in the middle.

This definitely needs wider testing with all the code this touches, but I cannot find any new rendering glitches at this point.

---


git-f0x:
This seems to make the panel, applets and layer shell surfaces (e.g. workspace overview) very sluggish. Is that expected for now?

Drakulix:
> This seems to make the panel, applets and layer shell surfaces (e.g. workspace overview) very sluggish. Is that expected for now?

Should be fixed now with the latest commits.

ryzendew:
can you explain more about the appearance_conf and how the syntax for blur goes? the details are a littler lacking to test.

Drakulix:
> can you explain more about the appearance_conf and how the syntax for blur goes? the details are a littler lacking to test.

The syntax is `blur_strength: 1` with values from 0-14. (Though 0 doesn't actually disable it at the moment.)

The value is 9 by default, so you don't have to set it. If you aren't seeing any blur, the reason is simply, that you aren't running any application supporting the `ext-background-effect-v1` protocol, as those are pretty rare at the moment. Our apps also don't make use of this yet.

Drakulix:
This is now only really blocked figuring out how we want to configure this / make the necessary changes to cosmic-theme.

WatchMkr:
Fullscreen video (tested with YouTube) isn't working with this branch. The desktop freezes while audio continues.

WatchMkr:
Suspend and resume isn't working with this branch on a Thelio Prime with AMD Radeon RX 7600.

Drakulix:
> Fullscreen video (tested with YouTube) isn't working with this branch. The desktop freezes while audio continues.

Should be fixed by the latest commit.

> Suspend and resume isn't working with this branch on a Thelio Prime with AMD Radeon RX 7600.

Cannot reproduce on my systems, how exactly does this manifest? Does the system simply not suspend? Does it wake up immediately? Does it not wake up correctly? Anything in the kernel logs?

WatchMkr:
Full screen video is fixed :tada: 

I'm not seeing the suspend issue any longer.

hojjat-works:
The `waylnad-protocols` package currently installed in pop_os (Version 1.45-1~ubuntu0.24.04.1) lists "blur" as 0. But cosmic-comp is based on a newer version listing "blur" as 1.

This causes `foot` compiled on PopOS 24.04 for example to not blur and say `warn: wayland.c:2546: background blur requested, but compositor does not support it`.

https://gitlab.freedesktop.org/wayland/wayland-protocols/-/commit/efbc060534be948b63e1f395d69b583eebba3235

Drakulix:
> The `waylnad-protocols` package currently installed in pop_os (Version 1.45-1~ubuntu0.24.04.1) lists "blur" as 0. But cosmic-comp is based on a newer version listing "blur" as 1.
> 
> This causes `foot` compiled on PopOS 24.04 for example to not blur and say `warn: wayland.c:2546: background blur requested, but compositor does not support it`.
> 
> https://gitlab.freedesktop.org/wayland/wayland-protocols/-/commit/efbc060534be948b63e1f395d69b583eebba3235

As seen in the commit message, the original value was never correct to begin with, we cannot advertise 0 for a bitfield. As such foot should really add a workaround.

We cannot simply update wayland-protocols, as foot (and every other application relying on the protocol definitions at build time) would need to be re-build as well.

If ubuntu 24.04 isn't updating wayland-protocols anymore, this will only resolve itself with either a bugfix release of foot (that might be backported to 24.04) or *much more likely* upgrading to 26.04 once that is released.

There isn't really anything actionable for us here to fix this.

hojjat-works:
I'm seeing visual glitches on my external 4k monitor when interacting with a pop up (for example an applet).

To repro:

1- On an AMD laptop with 2 4k monitors
2- open the date-time applet on one monitor and alternate hovering over next/previous month buttons
2- or alternatively, open the screenshot UI and move the cursor between the two external displays back and forth.
3- You'd see visual corruption on one or both monitors.

- It doesn't show up in screen recordings.
- The internal display doesn't show any distortions no matter if external displays are connected or not.

hojjatabdollahi:
Looks like blur has an offset now (tested with different fractional scaling):
image

Drakulix:
> Looks like blur has an offset now (tested with different fractional scaling)

Should've only happened if the window is close to the left/upper edge of the screen, but should be fixed now (turns out math is hard).

hojjatabdollahi:
After days of investigation, trying to fix the glitches I mentioned in https://github.com/pop-os/cosmic-comp/pull/2179#issuecomment-4164603760. Nothing worked. I suspected the driver. Updated to Linux 7.0 and all the glitches are gone!

My guess is [this](https://github.com/torvalds/linux/commit/1a38ded4bc8ac09fd029ec656b1e2c98cc0d238c) MPO patch fixed it. Sigh.

src/wayland/protocols/corner_radius.rs
```diff
@@ -254,66 +350,388 @@ where
...
+        }
+    });
+}
+
+fn pad_rect(
+    mut rect: Rectangle<i32, Logical>,
+    padding: &Padding,
+) -> Option<Rectangle<i32, Logical>> {
+    rect.size.h = rect.size.h.checked_sub(padding.top)?;
+    rect.loc.x += padding.top;
```
wash2:
```suggestion
    rect.loc.x += padding.left;
```

src/wayland/protocols/corner_radius.rs
```diff
@@ -254,66 +350,388 @@ where
...
+    rect.size.w = rect.size.w.checked_sub(padding.left)?;
+    rect.loc.y += padding.left;
```
wash2:
```suggestion
    rect.loc.y += padding.top;
```

src/wayland/handlers/corner_radius.rs
```diff
@@ -34,16 +40,83 @@ impl CornerRadiusHandler for State {
...
+        padding.left.min(half_min_dim),
+    ])
+}
+
+pub fn pad_rect(
+    mut rect: Rectangle<i32, Logical>,
+    padding: &[i32; 4],
+) -> Option<Rectangle<i32, Logical>> {
+    rect.size.h = rect.size.h.checked_sub(padding[0])?;
+    rect.loc.x += padding[0];
```
wash2:
```suggestion
    rect.loc.x += padding[3];
```

src/wayland/handlers/corner_radius.rs
```diff
@@ -34,16 +40,83 @@ impl CornerRadiusHandler for State {
...
+    rect.size.w = rect.size.w.checked_sub(padding[1])?;
+    rect.size.h = rect.size.h.checked_sub(padding[2])?;
+    rect.size.w = rect.size.w.checked_sub(padding[3])?;
+    rect.loc.y += padding[3];
```
wash2:
```suggestion
    rect.loc.y += padding[0];
```

wash2 (dismissed):
I think this fixes a few typos that should fix the launcher blur location

adil192:
The blur effect gets a bit glitchy on a rotated portrait monitor. Some areas are fully transparent while others get "twisted"/diagonal blocky blurred sections.
imageimage
I'm on the latest `theme-v2` and `frosted-glass_noble` commits as of writing.
(I'm not expecting any support, just thought I'd report this here if it's helpful)

Update: Also the SSDs seem to have the wrong corner radii on my rotated monitor (upside down?). The decorations are correct on my non-rotated monitor. CSDs are unaffected.
Screenshot_2026-05-20_19-08-10 Screenshot_2026-05-20_19-08-39

Drakulix:
> The blur effect gets a bit glitchy on a rotated portrait monitor. Some areas are fully transparent while others get "twisted"/diagonal blocky blurred sections.

Thanks, this might be an issue with damage tracking on a rotated output. I'll make sure to test that specific case once we get closer to merging this feature.

src/backend/render/wayland/blur_effect.rs
```diff
@@ -0,0 +1,694 @@
...
+        let radius = BLUR_PARAMS[(strength + 2).min(MAX_STEPS - 1)].extended_radius as f64;
+        extended_geo.loc -= Point::<f64, Physical>::new(radius, radius);
+        extended_geo.size += Size::<f64, Physical>::new(radius, radius).upscale(2.);
+
+        // compute input_to_geo so that it crops the extended capture radius
+        let geo_scale = {
+            let Scale { x, y } = geo.size / extended_geo.size;
+            Matrix3::from_nonuniform_scale(x as f32, y as f32)
+                .invert()
+                .unwrap()
```
ids1024, line 236: On Mattermost, Carl reported an unwrap panic that seems to be on this line.

I think the homogeneous transformation matrix here should be invertible unless `x` or `y` are `0`? Not sure why `geo` would have a zero size.

src/backend/render/wayland/blur_effect.rs
```diff
@@ -0,0 +1,694 @@
...
+        };
+        let geo_translation = {
+            let offset = geo.loc - extended_geo.loc;
+            Matrix3::from_translation(Vector2::new(
+                (offset.x / extended_geo.size.w) as f32,
+                (offset.y / extended_geo.size.h) as f32,
+            ))
+            .invert()
+            .unwrap()
```
ids1024: This unwrap should never panic, but it could be avoided by negating the `Vector2` (or the args passed to create it) which produces the same result.

shynekomaid:
Nice!

Drakulix:
Okay, no temp branches anymore (assuming https://github.com/pop-os/libcosmic/pull/1196 gets merged as is) and the git history is reasonably cleaned up.

I think this is ready for review/merge.

isFakeAccount:
I am not sure if it is intentional or not but the frosting effects gets disabled when you maximize a window. 

If it is a intentional then I hope there is a way to configure it to not do that.

hojjatabdollahi:
I'm testing frosted glass and I'm seeing a band on the right side of maximized windows that isn't blurred. If you guys don't see it too then it's probably something I poked!

It only happens when I resize a window to almost the full 
image


It depends on the width, for example I made the cosmic-settings window wider than my display and the unblurred area is now wider:
image

git-f0x:
> I'm testing frosted glass and I'm seeing a band on the right side of maximized windows that isn't blurred.

Does changing the frost thickness setting while in such a state cause the unblurred area to change in size? Extended panels have a similar issue at some frost thickness values.

Drakulix:
> I'm testing frosted glass and I'm seeing a band on the right side of maximized windows that isn't blurred. If you guys don't see it too then it's probably something I poked!

Can definitely reproduce it by resizing the window more. Thanks for the report, working on a fix! :)

Drakulix:
Fixed in smithay https://github.com/Smithay/smithay/pull/2086

hojjatabdollahi:
Is the blurred rectangle around windows in workspaces view intentional? I'd say blurring the whole background would look better, or not blurring anything but the windows:
image

EDIT: my bad, I reported here because I thought everything is already merged except this PR!

git-f0x:
I think that should be asked in https://github.com/pop-os/cosmic-workspaces-epoch/pull/312.

hojjatabdollahi:
I see these little corners for SSDs when windows are tiled.

image

It feels like the header is flipped upside down. You can see a faint little curved piece missing from the bottom. Probably some shader needs a `-1*` somewhere! :smile: 

It's more visible here, the little corner sticking out is blurred, and the little missing area is just transparent:

image

Drakulix:
> I see these little corners for SSDs when windows are tiled.

Should be fixed with 176329cd

hojjatabdollahi:
I'm not sure if this is a regression but the border for windows with SSD is thicker than other windows when you turn off "Match system corners"
image

hojjatabdollahi:
The controls for resizing (`super+R`) are drawn weirdly. 

image

Drakulix:
> The controls for resizing (`super+R`) are drawn weirdly.

The background is simply frosted as well. That is intended, though it definitely looks interesting in your case.

hojjatabdollahi:
> > The controls for resizing (`super+R`) are drawn weirdly.
> 
> The background is simply frosted as well. That is intended, though it definitely looks interesting in your case.

Alright. IMO the arrows with just transparent (not blurred) background would look better. And I think they used to be the same color as the theme accent and now they're black.

wash2 (approved):

leviport (approved):

fuzun:
@Drakulix Hello,

It seems there is a difference in implementation of this effect with Mutter, which is an hassle to adjust the behavior in the client depending on the compositor. With a surface that has xdg surface role, when there is no region provided, Cosmic compositor does not blur the frame extents (such as the area the client provides shadows), I think considering surface roles when no region is provided is reasonable and should be fine. However, when a region is provided, Cosmic compositor (unlike Mutter) still considers the frame extents.

The protocol states: "The blur region is specified in the surface-local coordinates", and I interpreted this as the Wayland surface itself, not its extension or redefinition by surface roles. Even xdg_surface::set_window_geometry does not seem to redefine the surface coordinates, it just states "The window geometry of a surface is its "visible bounds" from the user's perspective".

Do you have plans to change the behavior? Or do you think the behavior is justified and won't be changed?

The discussion in the client can be found here: https://code.videolan.org/videolan/vlc/-/merge_requests/9367.

fuzun:
@Drakulix Hello again, for reference, KWin's fix can be found here: https://invent.kde.org/plasma/kwin/-/merge_requests/9910. I guess a similar thing needs to be done here.

PR #2185, opened 2026-03-12
image-copy: Store offscreen buffer for shm capture in session 

ids1024:
Requires https://github.com/Smithay/smithay/pull/1960.

Seems okay, but I'm still seeing cursor capture using full damage when nothing has changed. Damage tracking still needs improvement...

### DIFF
--- a/src/wayland/handlers/image_copy_capture/render.rs
+++ b/src/wayland/handlers/image_copy_capture/render.rs
@@ -5,8 +5,8 @@ use smithay::{
     backend::{
         allocator::{Buffer, Fourcc, format::get_transparent},
         renderer::{
-            BufferType, Color32F, ExportMem, ImportAll, ImportMem, Offscreen, buffer_dimensions,
-            buffer_type,
+            BufferType, Color32F, ExportMem, ImportAll, ImportMem, Offscreen, Renderer,
+            buffer_dimensions, buffer_type,
             damage::{Error as DTError, OutputDamageTracker, RenderOutputResult},
             element::{
                 AsRenderElements, RenderElement,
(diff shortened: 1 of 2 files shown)


src/backend/render/element.rs
```diff
@@ -289,9 +292,15 @@ where
     }
 }
 
-pub trait AsGlowRenderer
```
Drakulix, line 292: Not saying this is a bad idea, but at that point we might as well rename this to CosmicRenderer.

The original idea was to keep the requirements per function very explicit for the future where a new renderer might cover 99% of our needs, but we need to special case a single function or two.

Perhaps the reverse would work? Introduce a `ImageCopyRenderer`-trait, that requires `AsGlowRenderer` and the rest of the `Renderer`-traits and possibly do the same for other modules?

Though I agree that `AsGlowRenderer` might as well require `ImportAll` and `ImportMem`, given those are needed to render wayland surfaces.

src/backend/render/mod.rs
```diff
@@ -1365,7 +1365,7 @@ where
     match result {
         Ok((res, mut elements)) => {
             for (session, frame) in output.take_pending_frames() {
-                if let Some(pending_image_copy_data) = render_session::<_, _, GlesTexture>(
+                if let Some(pending_image_copy_data) = render_session(
```
Drakulix, line 1368: I am trying to figure out, why this used `GlesTexture` in the first place. I guess getting rid of the generic might bite us in the future, when we actually do need to sample from the framebuffer for some post-processing effects? But right now the texture doesn't seem to be required by anything, so we certainly should use the `GlesRenderbuffer` for now.

ids1024: I think something else in this function requires `Offscreen<GlesTexture>`, so maybe this was using `GlesTexture` to avoid needing both `Offscreen<GlesTexture> + Offscreen<RenderBuffer>`?

In any case, the offscreen buffer type here is only used for shm capture, and isn't exposed to the caller, so just always using `GlesRenderbuffer` should be fine (until we want to support Vulkan or Pixman; then everything will need changes to be more generic.)

src/backend/render/element.rs
ids1024: My logic was that `AsGlowRenderer` kind of implies all of these traits, since any renderer we can get a `GlowRenderer` from should work with GLES textures/renderbuffers.

In the future, `PixmanRenderer` supports `Bind<Dmabuf>`, and a Vulkan renderer will too. So `Offscreen<GlesTexture>` and `Offscreen<GlesRenderbuffer>` are the only part here that should need to change to be more generic; but I guess we haven't figured out exactly how we want to make it all generic (and it will probably involve not having an `AsGlowRenderer` trait at all, unless we intend for certain things to always use GL to render.)

Drakulix: Fair. Let's leave it like this then.

ids1024:
Any further improvements to damage tracking can be addressed in a later PR. Avoid extra buffer allocations should be a performance improvement regardless, as long as we see no issues here.

I wanted to test this on the darp10 to make sure re-using the buffer didn't get the same graphical corruption issues we we're getting there with dmabuf capture, but it seems I got https://github.com/pop-os/cosmic-workspaces-epoch/pull/280 wrong. And https://github.com/pop-os/cosmic-workspaces-epoch/pull/292 is the correct fix.

So a good way to test this may be on a Meteor Lake system like the darp10 alongside https://github.com/pop-os/cosmic-workspaces-epoch/pull/292, with a window playing an animation. (Which I've done; and it seems good now.)

hojjatabdollahi:
This causes huge gpu memory leak.

To reproduce just minimize and restore a window a couple of times.

image

PR #2197, opened 2026-03-15
Move initization of pointer state to Common state

chrisglass:
- [] I understand these changes in full and will be able to respond to review comments.

This branch seems to fix https://github.com/pop-os/cosmic-epoch/issues/2021 from my personal testing.

I purposefully have not ticked the "I understand all the changes in this branch fully" since I'm not 100% certain of the side effects this patch can have (I do not know why the initialization was done how it was done before). I noticed this particular state wrapper was handled differently from the others.

What I think was causing the linked issue is the libinput context initialization being triggered once again when entering 3d state, which reset the state for relative pointer management tracking, resulting in it thinking the cursor was not captured.

Instead, this patch makes the state for relative cursor part of the overall state, ensuring it is not reset (?). This follows the same pattern as other State wrappers. This seems to solve my particular issue: relative movement being broken in Helldivers2 in 3d mode (2d mode - hence, not "relative movement" related, worked fine).

I would like this branch to undergo more manual testing for other games before being merged. While I can see it fixed my own problem (I have been running my branch successfully for a while), I am unsure of the implications on the rest of the compositor - it's my first time looking at the code base and I would be lying if I said I understand how it all works at this point (it's... complex).

### DIFF
--- a/src/backend/kms/mod.rs
+++ b/src/backend/kms/mod.rs
@@ -37,7 +37,6 @@ use smithay::{
     wayland::{
         dmabuf::DmabufGlobal,
         drm_syncobj::{DrmSyncobjState, supports_syncobj_eventfd},
-        relative_pointer::RelativePointerManagerState,
     },
 };
 use surface::GbmDrmOutput;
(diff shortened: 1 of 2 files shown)


chrisglass:
While this seems like it fixes the issue, reverting back to master seems to have the issue fixed as well (!). I am now unsure whether my fix changes anything, or if this is a side effect of something else. Will Repoen after more investigation.

BlakeGardner:
@chrisglass thanks for digging into this bug. I cannot reproduce it consistently. Only about 20% of the time does it happen for me. I find logging out and back in fixes it.

PR #2190, opened 2026-03-12
image-copy: Do not panic on `ImageCaptureSourceKind::Destroyed`

ids1024:
The way this was handled previously is incorrect. We should still handle creation of a capture session, just send `stopped` when it is created.

This can be tested by creating a capture source and session for a workspace that has been removed. Toplevel and output sources have a different issue in `smithay`: https://github.com/Smithay/smithay/pull/1961

Should fix https://github.com/pop-os/cosmic-epoch/issues/3319.

https://gist.github.com/ids1024/808358ed4671af7a8b31edc97a45cccc has a minimal test client using `cosmic-client-toolkit` to reproduce the issue; by trying to create a capture session for a workspace as soon as it is removed. Prior to this PR, that crashed the compositor when a workspace is removed. After it, the compositor runs fine, and that shows the capture session is `stopped`.

- [x] I have disclosed use of any AI generated code in my commit messages.
  - If you are using an LLM, and do not fully understand the changes it is making to the code base, do not create a PR.
  - In our experience, AI generated code often results in overly complex code that lacks enough context for a proper fix or feature inclusion. This results in considerably longer code reviews. Due to this, AI authored or partially authored PRs may be closed without comment.
- [x] I understand these changes in full and will be able to respond to review comments.
- [x] My change is accurately described in the commit message.
- [x] My contribution is tested and working as described.
- [x] I have read the [Developer Certificate of Origin](https://developercertificate.org/) and certify my contribution under its conditions.

PR #2205, opened 2026-03-18
fix: handle destroyed image capture sources gracefully

Aethelos-Mneia:
## Summary

When a capture source (output, workspace, or toplevel) is destroyed before a session interacts with it, the `Destroyed` variant is reached in `ImageCopyCaptureHandler` methods. Previously this triggered `unreachable!()`, crashing the compositor.

This patch handles the `Destroyed` case gracefully:
- `new_session` / `new_cursor_session`: return without setup
- `session_destroyed` / `cursor_session_destroyed`: return without cleanup  
- `frame`: explicitly fail the frame with `CaptureFailureReason::Unknown` to notify the client rather than silently dropping

## Architectural note

The `Output` variant already handles lifecycle races cleanly via `WeakOutput` — when the output is gone, `weak.upgrade()` returns `None` and the handler returns gracefully. The `Workspace` and `Toplevel` variants use strong references, which is why the `Destroyed` fallback variant exists.

A deeper fix would be to use weak references for `Workspace` and `Toplevel` as well, making the `Destroyed` variant unnecessary. This has broader implications across the codebase, so it is left as a recommendation for maintainers to evaluate.

## Test plan

- [ ] Reproduce by destroying a capture source while a session is active
- [ ] Verify compositor no longer crashes on `unreachable!()`
- [ ] Verify client receives frame failure notification

### DIFF
--- a/src/wayland/handlers/image_copy_capture/mod.rs
+++ b/src/wayland/handlers/image_copy_capture/mod.rs
@@ -21,8 +21,9 @@ use smithay::{
         dmabuf::get_dmabuf,
         image_capture_source::ImageCaptureSource,
         image_copy_capture::{
-            BufferConstraints, CursorSession, CursorSessionRef, DmabufConstraints, Frame, FrameRef,
-            ImageCopyCaptureHandler, ImageCopyCaptureState, Session, SessionRef,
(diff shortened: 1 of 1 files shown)


Drakulix:
Thanks, but https://github.com/pop-os/cosmic-comp/pull/2190 is a better solution for this.

Issue #376, opened 2024-03-21
[feature request] Ability to change animation speed and disable animations entirely


thejevans:
Animations when moving windows, adding and removing windows, and changing workspaces are slow and generally unnecessary, even though they are pretty. It would be great if there was a setting in `config.ron` to change the animation durations or set them to 0 to disable them.

Drakulix:
Yes, this is also an accessibility feature. Not the top priority right now, but we will very likely include options to "reduce animations" and to "disable animations" in the future. (Though precisely controlling the animation duration might be out of scope.)

vnelis:
Drakulix please prioritize this accessibility feature! 
Forget about tweaking the animation speed, just allow us to remove the animations altogether. I guess I speak for all the devs in the world here when I say that mapping all my windows to different workspaces is crucial. I personally switch between workspaces every minute (or every few seconds sometimes). These animations are a guaranteed motion sickness. Removing them is literally the first I do in every distro.

B83C:
Please allow disabling animations

pspsdev:
Bumping this up. Having animations while switching workspaces are extremely annoying

FinGu:
+1

alvaroaleman:
Please do not respond with `+1`, it adds literally nothing, just add a thumbs up emoji to the original request. GitHub allows to [sort issues by number of reactions](https://github.com/pop-os/cosmic-comp/issues?q=%20is%3Aissue%20is%3Aopen%20sort%3Areactions-%2B1-desc), sorting by number of comments on the other hand is mostly useless, as there can be a number of reasons for that including a feature being contentious.

pudrio:
+1

SeniorMatt:
+1

sudomateo:
> +1

@SeniorMatt as a fellow Matthew I have to educate you that commenting with "+1" helps no one. Instead, react to the initial message with 👍🏼 that way maintainers can [sort issues by reactions](https://github.com/pop-os/cosmic-comp/issues?q=is%3Aissue%20state%3Aopen%20sort%3Areactions-%2B1-desc).

This was already addressed in https://github.com/pop-os/cosmic-comp/issues/376#issuecomment-3418734444 when someone else commented with a "+1". Please also read the issue in its entirety before commenting.

SeniorMatt:
Oh thank you my fellow Matthew, still learning GitHub a little)

sudomateo:
[...]

PR #2209, opened 2026-03-19
feat: add ability to disable animations

spinualexandru:
This PR adds the ability to toggle on and off the animations via an `enabled` flag. Partially addresses https://github.com/pop-os/cosmic-comp/issues/376

Demo:

[Kooha-2026-03-19-23-41-31.webm](https://github.com/user-attachments/assets/c20b817e-9847-49e9-8f11-fb0c9b4a0971)


spinualexandru:
Rebased

awalvie:
Would love to get a few more eyes on this! Really looking forward to using it : )

jklgrasso (dismissed):

src/shell/focus/order.rs
```diff
@@ -149,13 +150,17 @@ fn render_input_order_internal<R: 'static>(
...
+                        ease(
+                            EaseInOutCubic,
+                            0.0,
+                            1.0,
+                            Instant::now().duration_since(*st).as_millis() as f32
+                                / ANIMATION_DURATION.as_millis() as f32,
+                        )
+                    } else {
+                        1.0
+                    },
```
Drakulix: Hmm, I am not sure, I am too happy with this approach.

If we disable animations, why do we end up with a `previous` workspace anyway? (Except for potentially disabling animations mid-way through an animation, but I'd argue, that it wouldn't be an issue to finish these normally.)

In general, I would *not* want this to essentially scale the animations to 0-time, but skip all the animation specific code instead.

This includes skipping storing previous workspaces, the `Started`/`Ended` variants of fullscreen surfaces or directly applying any new tiling trees.

src/shell/layout/tiling/mod.rs
```diff
@@ -121,8 +122,15 @@ impl TreeQueue {
...
-            .push_back((tree, duration.into().unwrap_or(Duration::ZERO), blocker))
+        self.trees.push_back((
+            tree,
+            if self.animations_enabled {
+                duration.into().unwrap_or(Duration::ZERO)
+            } else {
+                Duration::ZERO
+            },
+            blocker,
+        ))
```
Drakulix, line 169: If we don't have a blocker, we can directly apply the new tree.

src/shell/layout/tiling/mod.rs
```diff
@@ -369,6 +379,19 @@ impl TilingLayout {
...
+        self.queue.animations_enabled
+    }
+
+    pub fn set_animations_enabled(&mut self, enabled: bool) {
+        self.queue.animations_enabled = enabled;
+        if !enabled {
+            for (_, duration, _) in self.queue.trees.iter_mut().skip(1) {
+                *duration = Duration::ZERO;
+            }
+        }
```
Drakulix: I am not sure that adjusting already on-going animations is worthwhile the additional code to maintain.

src/shell/mod.rs
```diff
@@ -118,6 +118,34 @@ const GESTURE_VELOCITY_THRESHOLD: f64 = 0.02;
...
+}
+
+pub(crate) fn animation_progress(start: Instant, duration: Duration) -> f32 {
+    if duration.is_zero() {
+        1.0
+    } else {
+        (Instant::now().duration_since(start).as_secs_f32() / duration.as_secs_f32())
+            .clamp(0.0, 1.0)
+    }
+}
```
Drakulix: I don't think `animation_duration` and `animation_progress` here shouldn't exist. An animation with a duration of zero shouldn't exist in the first place.

Drakulix (commented):
In general: thanks for tackling to implement this!

But... I am not super stoked about a few design decisions in this implementation. In general I think the current animation code shouldn't be modified as much as possible, but simply skipped. I strongly believe that approach would make the result much easier to maintain.

src/shell/focus/order.rs
spinualexandru: Agreed, I removed the disabled-animation fallback here. With animations disabled, it no longer creates a previous workspace for shortcut transitions so this path only handles transitions inflight.

src/shell/layout/tiling/mod.rs
spinualexandru: Done. I made it so if animations are disabled there isn o blocker, the new tree is applied directly instead of queueing a zero-duration transition. Blocked trees still wait for the configure blocker, then apply without animation

src/shell/layout/tiling/mod.rs
spinualexandru: True... Thanks for this! I removed the part rewriting queued durations when toggling them off and made it so the existing animations finish on their original timing

src/shell/mod.rs
spinualexandru: That makes sense - just a habit from working with TUIs :D Removed the zero duration approach and updated the state so disabled animations skip creatin started/ended animations states(where applicable).

spinualexandru:
Thanks for the review and the patience to explain each comment, Drakulix.

I reworked this so disabled animations are not following a model as zero duration animations anymore.

The new animation state is skipped when animations are disabled - workspace switches do not store previously active, overview/resize enter Active/None directly, full screen enter/exit avoids Start/Ended state and tiling applies the trees without the blocker

src/shell/element/stack/tabs.rs
```diff
@@ -268,6 +287,12 @@ impl State {
     }
 
     pub fn cleanup_old_animations(&mut self) {
+        if !self.animations_enabled {
+            self.scroll_animation = None;
+            self.tab_animations.clear();
+            return;
+        }
```
Drakulix: Do we need this, if we already do the same cleanup in `set_animations_enabled`? (And additionally, sorry for not noticing this earlier. Do we really need the cleanup?)

src/shell/layout/floating/mod.rs
```diff
@@ -265,18 +266,28 @@ impl FloatingLayout {
...
         layout.space.map_output(output, (0, 0));
         layout
     }
 
+    pub fn set_animations_enabled(&mut self, enabled: bool) {
+        self.animations_enabled = enabled;
+        if !enabled && !self.animations.is_empty() {
+            self.animations.clear();
+            self.dirty.store(true, Ordering::SeqCst);
+        }
```
Drakulix: Same potentially unnecessary cleanup.

src/shell/layout/tiling/mod.rs
```diff
@@ -4183,9 +4226,13 @@ impl TilingLayout {
             .then(|| &self.queue.trees.front().unwrap().0);
 
         let percentage = if let Some(animation_start) = self.queue.animation_start {
```
Drakulix: Please write this instead as:

```suggestion
        let percentage = if let Some(animation_start) = self.queue.animation_start.filter(|_| !duration.is_zero()) {
```

src/shell/zoom.rs
```diff
@@ -155,33 +163,46 @@ impl OutputZoomState {
...
+            .clamp(0.0, 1.0);
             ease(EaseInOutCubic, *old_level, self.level, percentage)
         } else {
             self.level
         }
     }
 
     pub fn is_animating(&self) -> bool {
-        self.previous_point.is_some() || self.previous_level.is_some()
+        self.animations_enabled && (self.previous_point.is_some() || self.previous_level.is_some())
```
Drakulix: I don't think we want to change this method.

When animations are disabled `previous_point` and `previous_level` should never be `Some(_)` in the first place. However if they somehow end up being `Some`, this will just break rendering the animation potentially showing outdated contents.

src/shell/zoom.rs
```diff
@@ -155,33 +163,46 @@ impl OutputZoomState {
...
+            if self.previous_level.as_ref().is_some_and(|(_, start)| {
+                Instant::now().duration_since(*start) > ANIMATION_DURATION
+            }) {
+                self.previous_level.take();
+            }
+            if self.previous_point.as_ref().is_some_and(|(_, start)| {
+                Instant::now().duration_since(*start) > ANIMATION_DURATION
+            }) {
+                self.previous_point.take();
+            }
```
Drakulix: Again I am not sure if the complexity is worthwhile.

Drakulix (commented):
This looks already *much* better than the previous function. I did notice some more spots I didn't before (sorry!), but I am pretty sure we are very close to a mergeable state now. Thanks for spending the time!

awalvie:
So happy to see progress on this! Really excited to see it merged!

Logarithmus:
Why not have a slider to adjust animation speed instead?

src/shell/element/stack/tabs.rs
spinualexandru: Removed

src/shell/layout/floating/mod.rs
spinualexandru: Removed

src/shell/layout/tiling/mod.rs
spinualexandru: Thanks for this, applied your suggestion

src/shell/zoom.rs
spinualexandru: Fair point! Made it so is_animating no longer gates on animations_enabled

src/shell/zoom.rs
spinualexandru: Simplified this complexity

spinualexandru:
> Why not have a slider to adjust animation speed instead?

This is a pre-requisite of adding whatever UI implementation on top of it.

spinualexandru:
Thanks for the comments Drakulix, sorry for taking so long to get back to them

Logarithmus:
@spinualexandru my point is that it would be nice to have a float value instead of boolean here, so that I could still have some animations, just faster ones. I like animations, but the default speed in Cosmic is very slow in my opinion, especially coming from Sway, which has no animations at all :) So is replacing the boolean with the float planned in the future, or there are no such plans?

awalvie:
Has there been any progress on this by any chance?

PR #2215, opened 2026-03-22
Add setting to toggle workspace wrapping

olekawaii:

This adds a setting to disable workspace wrapping.

issue: https://github.com/pop-os/cosmic-settings/issues/1050

changes:
* Adds workspace_wraparound flag to WorkspaceConfig, similar to WorkspaceLayout
* Changes to_next_workspace and to_previous_workspace to use this value

I tested that it works with keyboard shortcuts and gestures, but did not test it with multiple monitors.

There's a visual bug where if you use a gesture while an animation is still playing, it will play the workspace switching animation despite being on the last workspace. I'm pretty sure this was also the case before wrapping was added.

### DIFF
--- a/cosmic-comp-config/src/workspace.rs
+++ b/cosmic-comp-config/src/workspace.rs
@@ -4,13 +4,30 @@ use serde::{Deserialize, Serialize};
 
 use crate::EdidProduct;
 
-#[derive(Debug, Clone, Default, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub struct WorkspaceConfig {
     pub workspace_mode: WorkspaceMode,
     #[serde(default)]
     pub workspace_layout: WorkspaceLayout,
     #[serde(default)]
     pub action_on_typing: Action,
+    #[serde(default = "default_wraparound")]
+    pub workspace_wraparound: bool,
+}
+
+fn default_wraparound() -> bool {
+    true
+}
+
+impl Default for WorkspaceConfig {
+    fn default() -> Self {
+        Self {
+            workspace_mode: WorkspaceMode::default(),
+            workspace_layout: WorkspaceLayout::default(),
+            action_on_typing: Action::default(),
+            workspace_wraparound: default_wraparound(),
+        }
+    }
 }
 
 #[derive(Debug, Clone, Copy, Default, PartialEq, Eq, Serialize, Deserialize)]
--- a/src/input/actions.rs
+++ b/src/input/actions.rs
@@ -102,6 +102,12 @@ impl State {
 
     pub fn handle_swipe_action(&mut self, action: gestures::SwipeAction, seat: &Seat<State>) {
         use gestures::SwipeAction;
+        let wraparound: bool = self
+            .common
+            .config
+            .cosmic_conf
+            .workspaces
+            .workspace_wraparound;
 
         match action {
             SwipeAction::NextWorkspace => {
@@ -109,6 +115,7 @@ impl State {
                     &mut self.common.shell.write(),
                     seat,
                     true,
+                    wraparound,
                     &mut self.common.workspace_state.update(),
                 );
             }
@@ -117,6 +124,7 @@ impl State {
                     &mut self.common.shell.write(),
                     seat,
                     true,
+                    wraparound,
                     &mut self.common.workspace_state.update(),
                 );
             }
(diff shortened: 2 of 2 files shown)


Drakulix:
Code-wise this is pretty clean. Ping @pop-os/ux to see if there is any reason we don't want this option? (Either as something configurable in cosmic-settings or purely on the filesystem.)

jacobgkau:
I think UX has already been on-board with the setting in https://github.com/pop-os/cosmic-settings/issues/1050? The wording of the mockup will need to change, but we do have a mockup.

> There's a visual bug where if you use a gesture while an animation is still playing, it will play the workspace switching animation despite being on the last workspace. I'm pretty sure this was also the case before wrapping was added.

This was indeed already an issue before wrapping was implemented. We had been tracking it in https://github.com/pop-os/cosmic-comp/issues/725, so we can re-open that.

Drakulix (dismissed):

jacobgkau (changes requested):
The new setting's working great. However, after installing this, the setting defaulted to `false` (no wrapping). While I personally prefer no wrapping and will be utilizing this setting, we need to preserve the current behavior of wrapping as the default. So either the implicit default needs to change (probably the cleaner option), or some mechanism needs to set it to `true` in the config by default.

ghost:
When could we see this getting implemented in cosmic?

leviport:
@Unknownoof you're watching it get implemented right now, in this very pull request.

ghost:
Sorry when can we see this be in the official release? Will this come in a package update very soon or will it be in one of the next epochs? (Not really sure how this stuff works)

leviport:
Once this is completed and merged, it will be released within a week or so. It will not need to wait for an Epoch release.

olekawaii:
I added a manual Default implementation for WorkspaceConfig. It now sets workspace_wraparound to true by default.

ghost:
> Once this is completed and merged, it will be released within a week or so. It will not need to wait for an Epoch release.

That's amazing, thanks for the quick reply. Open source projects are great!

cosmic-comp-config/src/workspace.rs
```diff
@@ -4,13 +4,26 @@ use serde::{Deserialize, Serialize};
...
 
-#[derive(Debug, Clone, Default, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub struct WorkspaceConfig {
     pub workspace_mode: WorkspaceMode,
     #[serde(default)]
     pub workspace_layout: WorkspaceLayout,
     #[serde(default)]
     pub action_on_typing: Action,
+    #[serde(default)]
```
Drakulix: This line would probably set `workspace_wraparound` to `false`, if a config file exists missing this key. (So potentially every existing install.)

This also needs to be updated to use serde's [default-path](https://serde.rs/field-attrs.html#default--path) to default it to `true` in these cases as well.

Drakulix (commented):
LGTM, thanks! Just a minor oversight.

cosmic-comp-config/src/workspace.rs
olekawaii: If so, then maybe a manual Default impl is not the best idea, since we now have to worry about `#[serde(default)]` and `Default::default()` being the same. I think a simpler solution would be to just wrap the bool in a struct like the `enabled` flag in https://github.com/pop-os/cosmic-comp/blob/7fa5bd6b91096b064cd787c8fccb312b63e9640f/cosmic-comp-config/src/input.rs#L68-L74

```rs
struct WorkspaceWraparound {
    enabled: bool
}

impl Default for WorkspaceWraparound { ... }
```

olekawaii: I added a default_wraparound function to use in `#[serde(default = "default_wraparound")]` and in `Default::default`.

Drakulix (dismissed):
Thanks!

jacobgkau (changes requested):
The new default is working as expected.

On 1739652, I'm seeing that setting this to `false` stops wraparound for gestures, but doesn't stop it for the Ctrl + Super + Up/Down keyboard shortcuts. We want it to apply to both, and it sounds from the PR description like it was supposed to.

olekawaii:
It should now work with Ctrl+Super+Up/Down and Ctrl+Super+Shift+Up/Down. I just didn't realize they would each have their own wraparound code. Hopefully everything should be good now.

jacobgkau (approved):
Working as expected now with both keyboard shortcuts and gestures. The setting defaults to wrapping when it's not specified in the config.

All other Workspaces settings are still working as expected, so no regressions there.

Looks good to me.

Drakulix (approved):
Thanks!

PR #2225, opened 2026-03-25
session: Make sure to mark the socket as CLOEXEC early

### LINKED ISSUE #2156: Xwayland inherits COSMIC_SESSION_SOCK fd, breaking keyboard input and DRM master
## Description

After upgrading `cosmic-comp` (first noticed with `0.1~1771890949`, still present in `0.1~1772192261`), keyboard input is almost entirely broken — only roughly every 5th keypress gets through. Mouse input works normally.

## Root cause

When `cosmic-comp` spawns Xwayland, the `COSMIC_SESSION_SOCK` file descriptor (fd 12) is not marked `CLOEXEC`. Xwayland inherits the fd, closes it, and reuses the fd number for its own memfd (`/memfd:xwayland-shared`). After that, `cosmic-comp` can no longer communicate with `cosmic-session` over the session socket.

This causes:

- `Error reading from session socket`
- `Unable to become drm master, assuming unprivileged mode`
- Near-total loss of keyboard input (occasional keypresses leak through)

## Evidence

```
$ cat /proc/<cosmic-comp-pid>/environ | tr '\0' '\n' | grep COSMIC_SESSION_SOCK
COSMIC_SESSION_SOCK=12

$ ls -la /proc/<cosmic-comp-pid>/fd/12
lrwx------ 1 user user 64 ... /proc/<cosmic-comp-pid>/fd/12 -> '/memfd:xwayland-shared (deleted)'
```

fd 12 should be the session socket but has been replaced by Xwayland's shared memory segment.

Journal output at login:

```
cosmic-comp[3750]: Unable to become drm master, assuming unprivileged mode
cosmic-comp[3750]: Error reading from session socket
```

## Workaround

Disabling Xwayland restores full keyboard input:

```bash
echo "false" > ~/.config/cosmic/com.system76.CosmicComp/v1/xwayland_enabled
```

## Suggested fix

Set `FD_CLOEXEC` on the `COSMIC_SESSION_SOCK` fd before spawning child processes (Xwayland, etc.).

## Environment

- **OS:** Pop!_OS 24.04 (COSMIC)
- **cosmic-comp:** `0.1~1772192261~24.04~253b44f` (also affected: `0.1~1771890949~24.04~3f68eee`)
- **cosmic-session:** `1.0.0~1771619805~24.04~618624b`
- **Last known working:** cosmic-comp `0.1~1770407907~24.04~2ea1186` (pre-March 2 upgrade)
- **Keyboard:** Logitech K750
- **Mouse:** Logitech MX Master 2S
- **GPU:** Advanced Micro Devices, Inc. [AMD/ATI] Ellesmere [Radeon RX 470/480/570/570X/580/580X/590] (rev ef)

Drakulix:
Presumably fixes #2156 by making sure we set `cloexec` on the inherited socket before launching Xwayland.

Not sure why this would seemingly only a single person, but the previous logic definitely wasn't correct.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -73,7 +73,7 @@ impl State {
 
             // potentially tell the session we are setup now
             if let Err(err) =
-                session::setup_socket(self.common.event_loop_handle.clone(), &self.common)
+                session::run_socket(self.common.event_loop_handle.clone(), &self.common)
             {
                 warn!(?err, "Failed to setup cosmic-session communication");
             }
@@ -147,6 +147,11 @@ pub fn run(hooks: crate::hooks::Hooks) -> Result<(), Box<dyn Error>> {
     tracy_client::Client::start();
 
     utils::rlimit::increase_nofile_limit();
+    // This needs to be done before any potential program launches
+    // (e.g. Xwayland) as it handles passed file descriptors.
+    if let Err(err) = session::setup_socket() {
+        warn!("Session error: {:?}", err);
+    };
 
     // init hook globals
     hooks::HOOKS.set(hooks)
--- a/src/session.rs
+++ b/src/session.rs
@@ -72,18 +72,24 @@ pub fn get_env(common: &Common) -> Result<HashMap<String, String>> {
     Ok(env)
 }
 
-pub fn setup_socket(handle: LoopHandle<State>, common: &Common) -> Result<()> {
+pub fn setup_socket() -> Result<()> {
+    if let Ok(fd_num) = std::env::var("COSMIC_SESSION_SOCK")
+        && let Ok(fd) = fd_num.parse::<RawFd>()
+    {
+        let res = unsafe { set_cloexec(fd) }.with_context(|| "Failed to setup session socket");
+        if res.is_err() {
+            unsafe { rustix::io::close(fd) };
+        }
+        res
+    } else {
+        Ok(())
+    }
+}
+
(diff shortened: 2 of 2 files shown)


src/lib.rs
```diff
@@ -147,6 +147,9 @@ pub fn run(hooks: crate::hooks::Hooks) -> Result<(), Box<dyn Error>> {
     tracy_client::Client::start();
 
     utils::rlimit::increase_nofile_limit();
+    if let Err(err) = session::setup_socket() {
```
ids1024, line 152: It might be best to have comment here noting that it's important this is called before any subprocess is spawned, just to make sure we remember this in any future changes to this function. (Not that we're especially likely to add anything that starts a process to the top of `run`, but still.)

Otherwise, this change seems correct. If a bit awkward (as dealing with fd numbers from env vars will tend to be).

jacobgkau (dismissed):
I've tested this on a few different systems (Intel, NVIDIA, AMD graphics). Input still works fine both before and after launching an Xwayland application (seemed like a semi-relevant thing to check based on the issue this presumably fixes).

Can re-approve after that comment's added if it's the only thing that changes.

src/lib.rs
Drakulix: done.

Drakulix:
Comment added

jacobgkau (approved):

ids1024 (approved):

PR #2237, opened 2026-03-30
fix: remove double slashes from cosmic-protocols patch URLs

MasterofNull:
## Summary

Fixes double slashes in the Cargo.toml patch section URLs that were causing cargo to fail when resolving dependencies.

## Changes

- Fixed `https://github.com/pop-os//cosmic-protocols` → `https://github.com/pop-os/cosmic-protocols`
- Fixed `https://github.com/pop-os//cosmic-text` → `https://github.com/pop-os/cosmic-text`

## Testing

- Verified cargo fetch succeeds after the fix
- Nix flake build completes without URL resolution errors

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -144,8 +144,8 @@ inherits = "release"
 lto = "fat"
 
 [patch."https://github.com/pop-os/cosmic-protocols"]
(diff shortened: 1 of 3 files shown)


jackpot51:
The PR template is mandatory

PR #2243, opened 2026-04-01
fix: remove keyboard grab serial check that blocks popup menus

### LINKED ISSUE #1211: Menus not appearing inside a variety of applications
Since upgrading to COSMIC Alpha 5, menus (top bar and right-click context menus) no longer show up across a wide variety of applications.

In some cases, a menu will appear the first time I click, but if I try moving to another menu, or click off and then try clicking back on, the menu won't appear.

First Click—Menu appears
Second Click—No Menu

In these cases, switching focus to another window and then going back to the original application will "reset" this behavior (I'll be able to get into a new menu).

In other cases—namely LibreOffice and every right-click context menu I've tried—menus won't appear even on first click.

This issue is present across apps made with most—but not all—toolkits (I have this issue in GNOME and KDE apps, but not Vala apps like Pantheon files, Electron apps like Discord or whatever toolkit Firefox uses), and I've observed it in Flatpak applications as well as system apps.

This issue is only present in COSMIC—the applications behave normally in Pantheon Wayland.

Any help would be greatly appreciated.

### System Details
OS: Arch Linux x86_64
Kernel: Linux 6.13.2-arch1-1
GPU: AMD Radeon RX 6700XT
DE/ Compositor: COSMIC 1.0.0.alpha.5.1-1 (extra, not AUR)

### LINKED ISSUE #1763: [BUG] Qt application menus won't open when IME is active

akelmanson:
## Summary

- Remove the keyboard grab serial validation in the `xdg_popup` grab handler that incorrectly rejects popup grabs when an unrelated keyboard grab is active
- The check was blocking dropdown menus and context menus in apps using input methods (IBus, Fcitx) or after a previous popup was closed

## Root Cause

The keyboard grab serial check (`keyboard.is_grabbed() && !keyboard.has_grab(serial)`) rejects popup grabs whenever **any** keyboard grab is active with a non-matching serial. Two common scenarios trigger this:

1. **Input method keyboard grab** (`zwp_input_method_v2::GrabKeyboard`): Always active when an IME is running. Its serial never matches the popup's button-press serial, so every popup grab is rejected.

2. **Stale `PopupKeyboardGrab`**: After a popup is closed, the `PopupKeyboardGrab` lingers on the keyboard until the next keyboard event (Smithay only cleans it up on `input()` or `set_focus()`). Any subsequent popup grab is rejected because the stale grab's serial doesn't match.

## Why this is safe

Smithay's `grab_popup()` already validates the popup chain (parent hierarchy, topmost popup check, dismissed parent check) **before** this code runs. The keyboard grab state is orthogonal to popup validity — it controls keyboard event routing, not whether a popup should open. The new `PopupKeyboardGrab` replaces whatever keyboard grab was active.

Other compositors (wlroots, niri) do not perform this check. wlroots disabled serial validation entirely ([wlroots#755](https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/755)).

## Affected apps

- Any app when an input method is active (Qt, GTK, wxWidgets)
- LibreOffice Writer, Calc, Math (menus stop working after first use)
- OrcaSlicer and other wxWidgets apps with focusable child widgets
- Any app where menus work once then fail on subsequent attempts

## Reproducer

```cpp
// wxWidgets minimal reproducer — menus fail after first open
#include <wx/wx.h>
class App : public wxApp {
public:
    bool OnInit() override {
        auto *f = new wxFrame(nullptr, wxID_ANY, "Test", wxDefaultPosition, wxSize(800, 600));
        auto *mb = new wxMenuBar;
        auto *fm = new wxMenu;
        fm->Append(wxID_EXIT, "Exit");
        mb->Append(fm, "&File");
        f->SetMenuBar(mb);
        new wxWindow(f, wxID_ANY);  // focusable child triggers the bug
        f->Bind(wxEVT_MENU, [f](wxCommandEvent&) { f->Close(); }, wxID_EXIT);
        f->Show();
        return true;
    }
};
wxIMPLEMENT_APP(App);
```

## Test plan

- [x] wxWidgets reproducer — File menu opens reliably on repeated attempts
- [x] LibreOffice Writer — dropdown menus work with IME active
- [x] Click-outside-to-dismiss still works (pointer grab check is unchanged)
- [x] `cargo clippy --all-features -- -D warnings` passes
- [ ] Nested submenus (File → Recent Documents → ...) work correctly
- [ ] Native COSMIC app menus still work

Fixes #1211
Fixes #1763

### DIFF
--- a/src/wayland/handlers/xdg_shell/mod.rs
+++ b/src/wayland/handlers/xdg_shell/mod.rs
@@ -107,13 +107,11 @@ impl XdgShellHandler for State {
             match ret {
                 Ok(mut grab) => {
                     if let Some(keyboard) = seat.get_keyboard() {
-                        if keyboard.is_grabbed()
-                            && !(keyboard.has_grab(serial)
-                                || keyboard.has_grab(grab.previous_serial().unwrap_or(serial)))
-                        {
-                            grab.ungrab(PopupUngrabStrategy::All);
-                            return;
-                        }
+                        // Do not reject the popup grab based on keyboard grab state.
+                        // Smithay's grab_popup() already validated the popup chain.
+                        // Keyboard grabs from other subsystems (input method, stale
+                        // PopupKeyboardGrab, etc.) are unrelated to popup validity and
+                        // the new PopupKeyboardGrab will replace them.
                         Shell::set_focus(
                             self,
                             grab.current_grab().as_ref(),


Drakulix:
Duplicate of https://github.com/pop-os/cosmic-comp/pull/1923, see attached discussion. This needs fixes in smithay to allow multiple grabs, since we don't want the popup grab to simply replace the IME.

Issue #1874, opened 2025-12-09
Fullscreen XWayland apps mouse capture issue


Adam-Cosner:
This is a recurring issue that has been fixed multiple times and has returned again. Mouse movement in 3D games which usually relies on resetting mouse cursor position is not able to do so as of a couple days ago, so when you're in a game and you turn your view too far in one direction, it ends up hitting a wall that you can't move past.

Adam-Cosner:
Seems to be gone again

Adam-Cosner:
Nevermind, it's doing it again. I'm not exactly sure what is causing it, I'll try and see if I can consistently recreate the behavior

Adam-Cosner:
Okay, it seems that there's a trigger of some sort that causes it to become broken for the rest of the session and is only able to be fixed by restarting the compositor

Hyperchaotic:
I wonder is that the problem I've been seeing for a week or so in GuildWars2 (Proton+XWayland), when I adjust camera angle. 

At first everything works fine but some time into the game (it could be after a few minutes or 15, etc) when I try to adjust the view angle it snaps down and stays there, it just flips around there when I try to adjust the view. As can be seen at the end of the video if I bring up a menu with a hotkey the mouse works fine while visible.

Happens in windowed mode as well.

https://github.com/user-attachments/assets/2d3f1cda-5e2f-453b-95fc-17396002e321

jstelzer:
Just chiming in with some of my own observations about games interacting with the compositor. What's worked and whats been weird. 

**Environment**

* COSMIC (Wayland compositor)
* NVIDIA proprietary driver
* gamescope involved (sometimes)
* XWayland for most FPS titles

**Works**

* Controller-driven games (SDL / evdev path)
* WoW-style *modal* relative mouse (enter/exit camera mode)
* games that don t require continuous relative motion

**Fails**

* FPS games requiring **continuous relative mouse motion**
* Symptoms:

  * invisible left/right rotation boundary
  * partial pointer confinement
  * grab appears active but motion clamps
  * KDE/KWin does not exhibit the issue

**Key insight**

> This is not a game bug, not a Wine bug, not a gamescope flag issue   it s a compositor-level gap in handling continuous relative pointer semantics under Wayland. Specifically Xwayland.

*  FPS games using continuous relative mouse hit invisible rotation bounds 
*  Does not reproduce on KDE Plasma Wayland (KWin) 
[...]

PR #2255, opened 2026-04-05
fix(input): eagerly activate pointer constraints on motion events

RALaBarge:
Hey I'm Ryan -- thanks for PopOS, it is pretty slick.  I have ran into very few issues using it thus far.  I mostly game, but have also spent a lot of time developing after ditching windows...too much BS to program in general, let alone for linux on windows.

Anyways, I was playing the new Guild Wars remake and I couldn't use the camera properly.  My mouse dragging was not registering in the way it should, and I have noticed this in one other game but cannot recall what it was.  I have played about 20 others in the past couple of months since switching and have only noticed it once before.

The issue comes down to the way the pointer is registered in the compositor.  There is a race condition where a constraint is being called prior to be fully active which would block proper mouse input for moving the camera in the environment. There is also a fix to the logic to ensure that the data behind the current_focus() call is fresh.  We also will fetch the current physical position of the pointer when we notice the cache is stale.

## Problem

FPS/3D games on XWayland hit an invisible camera rotation wall in one direction. The cursor reaches the screen edge and stays there — subsequent XWarpPointer calls can no longer produce a delta, so the camera stops rotating.

Reported in #1874 and pop-os/cosmic-epoch#2871.

## Root cause

A one-frame race condition in `PointerMotion`: the constraint activation check ran **after** `ptr.motion()` was dispatched. On the very first mouse movement after a constraint was created, the constraint was not yet active, so the motion was processed normally and the cursor could reach the screen edge. The constraint then activated at the end of that same event — one frame too late.

`new_constraint()` is supposed to activate immediately, but only does so when `current_focus()` matches the surface. If the pointer hasn't moved since the window received focus (e.g. game launches fullscreen, user hasn't wiggled the mouse yet, or after suspend/resume), `current_focus()` is stale and the constraint is left inactive.  

## Fix

**`src/input/mod.rs`**: At the top of the `PointerMotion` handler, before checking whether a constraint is active, attempt to activate any inactive constraint if the pointer is already within the constraint region. This ensures no motion event is ever processed while a valid constraint is waiting to activate.

**`src/wayland/handlers/pointer_constraints.rs`**: `new_constraint()` now also checks physical pointer position via `surface_under()` as a fallback when `current_focus()` doesn't match. This covers stale-focus scenarios: post-suspend, after alt-tab, or when a constraint is requested before the first mouse movement.

## Testing

Tested on Pop!_OS 24.04 with COSMIC DE, NVIDIA 4070ti Super, games via Proton/XWayland. Camera rotation confirmed working without gamescope workaround on previously affected titles.  Specifically here, Guild Wars: Reforged.  Working great with the patched file.

> **Note:** This fix was developed with AI assistance (Claude). I understand the changes and can respond to review feedback.

- [ ] I have disclosed use of any AI generated code in my commit messages.

Thanks! -- Ryan

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -324,27 +324,42 @@ impl State {
                         .as_ref()
                         .and_then(|(target, l)| Some((target.wl_surface()?, l)))
                     {
-                        with_pointer_constraint(&surface, &ptr, |constraint| match constraint {
-                            Some(constraint) if constraint.is_active() => {
-                                // Constraint does not apply if not within region
-                                if !constraint.region().is_none_or(|x| {
-                                    x.contains(
-                                        (ptr.current_location() - *surface_loc).to_i32_round(),
-                                    )
-                                }) {
-                                    return;
-                                }
-                                match &*constraint {
-                                    PointerConstraint::Locked(_locked) => {
-                                        pointer_locked = true;
-                                    }
-                                    PointerConstraint::Confined(confine) => {
-                                        pointer_confined = true;
-                                        confine_region = confine.region().cloned();
-                                    }
+                        with_pointer_constraint(&surface, &ptr, |constraint| {
+                            let Some(constraint) = constraint else { return };
+                            // Eagerly activate if the pointer is already within the constraint
+                            // region but the constraint hasn't been activated yet (e.g., the
+                            // constraint was created before the first pointer motion event).
+                            if !constraint.is_active() {
+                                let region = match &*constraint {
+                                    PointerConstraint::Locked(locked) => locked.region(),
+                                    PointerConstraint::Confined(confined) => confined.region(),
+                                };
+                                let point =
+                                    (ptr.current_location() - *surface_loc).to_i32_round();
+                                if region.is_none_or(|r| r.contains(point)) {
+                                    constraint.activate();
                                 }
                             }
-                            _ => {}
+                            if !constraint.is_active() {
+                                return;
+                            }
+                            // Constraint does not apply if not within region
+                            if !constraint.region().is_none_or(|x| {
+                                x.contains(
+                                    (ptr.current_location() - *surface_loc).to_i32_round(),
+                                )
+                            }) {
+                                return;
+                            }
+                            match &*constraint {
+                                PointerConstraint::Locked(_locked) => {
+                                    pointer_locked = true;
+                                }
+                                PointerConstraint::Confined(confine) => {
+                                    pointer_confined = true;
+                                    confine_region = confine.region().cloned();
+                                }
+                            }
                         });
                     }
                     let original_position = position;
--- a/src/wayland/handlers/pointer_constraints.rs
+++ b/src/wayland/handlers/pointer_constraints.rs
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: GPL-3.0-only
 
-use crate::state::State;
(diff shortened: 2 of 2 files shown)


RALaBarge:
Upon further testing, we found a place where our fix was not being called when it should be.

**Scenario:** The player alt-tabs away from the game and back. Smithay deactivates the pointer constraint when focus is lost. When focus returns, `update_pointer_focus()` in `src/shell/focus/mod.rs` checks if `pointer.current_focus()` differs from what's physically under the pointer. If the pointer was already over the game window (common for fullscreen games), the focus already matches — so `pointer.motion()` is skipped, and neither of our previous fix points is reached. The constraint stays inactive until the next physical mouse movement, which is exactly the one-frame window the original fix was designed to close.

**Additional fix** — added to `update_pointer_focus()` in `src/shell/focus/mod.rs`:

After the `pointer.motion()` check, we now do an explicit constraint re-activation pass on the current focus surface. If there is an inactive constraint on that surface, it gets activated immediately. This is a no-op when motion was already dispatched (the motion handler handles it), but covers the case where motion was skipped due to matching focus.

This makes the fix hold across the full lifecycle: initial constraint creation → stale focus on first motion → alt-tab away and back with pointer staying over the game window.

Drakulix:
I am not convinced there is an issue with the constraint being activated a frame late.

If it would be simply constrained to the surface, the cursor can still be moved into the surface and moved freely.

If we are talking about a pointer lock, then the locked pointer will not emit any motion events anyway, but only relative motion events, so it doesn't matter where the cursor sits physically.

>  The cursor reaches the screen edge and stays there — subsequent XWarpPointer calls can no longer produce a delta, so the camera stops rotating.

XWarpPointer should use a pointer lock. I am pretty sure that issue would be solved by implementing cursor-hints correctly (see [this PR](https://github.com/pop-os/cosmic-comp/pull/1861)) and/or cosmic-comp implementing the new pointer-warp protocol (which may just reuse the code used for cursor-hints).

> If the pointer hasn't moved since the window received focus (e.g. game launches fullscreen, user hasn't wiggled the mouse yet, or after suspend/resume), current_focus() is stale and the constraint is left inactive.

I don't really see how this would cause an actual issue (after all the cursor lock becomes obvious the second the user *moves* the cursor), but the if one would want to fix this, the solution would be to make sure current_focus is updating correctly instead of implementing a secondary focus check. That just introduces technical debt.

Additionally please refrain from using LLMs for the issue/PR descriptions. They are quite noisy and difficult to read through. If you understand the changes made, you should have no trouble explaining the reasoning yourself, which I would very much appreciate for my own sanity.

RALaBarge:
Hey Drakulix -- Ryan here, no worries, I understand.  My context window is full too.

I can say that the changes are minute and it fixes the issue for me, and it seems like based on the description it was the same as the ones in the description from this repo...probably the one you are referencing.   I've had a few other issues during usage, but I am unsure if it is the compositor, the wm, or what.

Admittedly I don't specialize in rust but my bg is reading and troubleshooting code/systems and via testing it has been resolved for the example that spawned my desire to help fix this issue..  No crashes post install.

There are probably better methods, but this one works for me thus  far on my main machine.  If you had a suggestion that I can take to perform further testing, I'd be happy to!

edit also happy to close too if an overarching or upstream change should be made instead of it being done here, thanks for checking it out either way

PR #2268, opened 2026-04-08
fix: implement key repeat for system media/brightness keys

### LINKED ISSUE #384: Change volume by multiple steps with volume keys
When I hold down the keyboard volume control I expect there to be a slight pause and then for the sound to gradually continue to raise or lower the volume. Currently on COSMIC, I have to release and re-press the volume control to adjust the volume for each step.

I am on a Lemp10 and adjusting the volume with Fn + F5/F6.

### LINKED ISSUE #2703: On some hardware, pressing and holding volume keys doesn't make volume go up repeatedly
When pressing and holding the volume keys the volume only goes up once so i have to repeatedly spam the volume keys to make the volume go up. I don't know if this is intentional or a bug but i hope it gets fixed soon.

Sejder:
Wayland delegates key repeat to clients via wl_keyboard.repeat_info, meaning libinput drops EV_REP events by design. This adds a calloop timer on KeyState::Pressed for repeatable system actions that re-fires handle_action at 50ms intervals after a 400ms initial delay.

AI-Assisted: The calloop timer logic was drafted with the assistance of AI.

Fixes pop-os/cosmic-epoch#384
Fixes pop-os/cosmic-epoch#2703

- [x]  I have read the [Developer Certificate of Origin](https://developercertificate.org/) and certify my contribution under its conditions.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -1891,7 +1891,60 @@ impl State {
                     && cosmic_modifiers_eq_smithay(&binding.modifiers, modifiers)
                 {
                     modifiers_queue.clear();
-                    seat.supressed_keys().add(&handle, None);
+
+                    // For repeatable system actions (volume, brightness), libinput
+                    // does not forward EV_REP events to the compositor — Wayland
+                    // delegates key repeat to clients via wl_keyboard.repeat_info,
+                    // so shortcut actions would otherwise fire only once per hold.
+                    // Drive repeats ourselves via a calloop timer until the key is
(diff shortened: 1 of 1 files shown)


Drakulix:
No, sorry, we don't want to repeat command invocations linked to shortcuts, that has a whole host of problems.

The correct solution is to handle this in a separate process through some global-shortcuts protocol, which has been discussed at length in various issues and will be implemented eventually.

Sejder:
That's alright, thank you for your time:)

Issue #2272, opened 2026-04-09
[Bug] Electron CSD window (Spotify) jumps/teleports on drag after toggling "Floating windows gravitate to nearby edges"


Adiar1:
When attempting to drag the Spotify window (which uses Client-Side Decorations / Electron), the window immediately jumps to the left or teleports downward to align the top bar with the cursor. This behavior began specifically after toggling the "Floating windows gravitate to nearby edges" setting ON and then persisted after toggling it back OFF in the desktop settings. The window coordinate math seems permanently desynced for this specific application. 

**Steps to Reproduce**
1. Install Spotify (issue persists across Flatpak, APT, and Snap packages).
2. Go to **Settings** > **Desktop** > **Window Management**.
3. Toggle **"Floating windows gravitate to nearby edges"** ON, then toggle it OFF.
4. Left-click and drag the Spotify window using its top title bar.
5. Alternatively, attempt to move the window using `Super` + Left-click drag.

**Expected Behavior**
The window should drag smoothly from the exact coordinates the cursor grabbed it, maintaining its relative position to the mouse pointer without jumping.

**Actual Behavior**
The window instantly teleports to the left or jumps down, forcefully snapping a different window coordinate to the cursor the moment movement is registered.

**Troubleshooting Attempted (Did Not Resolve)**
* Forcing Wayland (`--enable-features=UseOzonePlatform --ozone-platform=wayland`)
* Disabling hardware acceleration (`--disable-gpu`)
* Moving the window via `Super` + Left-Click or `Alt` + Left-Click (still causes teleportation)
* Forcing window state resets via `Super` + `G` (toggling tiling on/off)
* Deleting Spotify's internal preference files (`prefs`) to clear cached coordinates
* Logging out and rebooting the system to clear the `cosmic-comp` cache

**System Information**
* **OS:** Pop!_OS (COSMIC Desktop Environment)
* **Hardware:** Lenovo Legion Pro 7i
* **Application:** Spotify 
* **Window Manager:** `cosmic-comp`

LiamillionSS:
Also occurs with Arch, started only after the 1.0.9 rebase

adil192:
This is fixed for me with 1.0.11 (https://github.com/pop-os/cosmic-comp/pull/2290) in Spotify and other SSD apps 🎉

PR #2290, opened 2026-04-15
Fix the windows shrinking after a drag-drop

### LINKED ISSUE #2285: Kitty floating window height reduces when moved
I belive it happened after update to 1.0.9 (arch linux), though i'm not sure.
While other widows moves normally (as shown with nemo in video), kitty snaps to cursor with it's title bar and gets smaller in height after moved. Video attached.
https://github.com/user-attachments/assets/78c19871-5d70-495c-bc10-16a8926b3e90

hojjatabdollahi:
SSD height is not currently taken into account, so dragging and dropping a window around will keep shrinking it.

https://github.com/user-attachments/assets/ffc6e53e-34f7-44a3-9b05-7ae8b119b924

Fixes https://github.com/pop-os/cosmic-comp/issues/2285 and https://github.com/pop-os/cosmic-comp/issues/2272
___

### DIFF
--- a/src/shell/layout/floating/mod.rs
+++ b/src/shell/layout/floating/mod.rs
@@ -686,7 +686,7 @@ impl FloatingLayout {
                 && let Some(pending_size) = window.pending_size()
             {
                 mapped_geometry.size = pending_size.as_local();
-            } else if let Some(server_size) = window.active_window().last_server_size() {
+            } else if let Some(server_size) = window.last_server_size() {
                 mapped_geometry.size = server_size.as_local();
             }
             *window.last_geometry.lock().unwrap() = Some(mapped_geometry);
--- a/src/shell/element/mod.rs
+++ b/src/shell/element/mod.rs
@@ -408,6 +408,14 @@ impl CosmicMapped {
         }
     }
 
+    pub fn last_server_size(&self) -> Option<Size<i32, Logical>> {
+        match &self.element {
+            CosmicMappedInternal::Stack(s) => s.last_server_size(),
+            CosmicMappedInternal::Window(w) => w.last_server_size(),
+            _ => unreachable!(),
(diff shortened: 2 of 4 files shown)


src/shell/layout/floating/mod.rs
```diff
@@ -686,7 +686,10 @@ impl FloatingLayout {
                 && let Some(pending_size) = window.pending_size()
             {
                 mapped_geometry.size = pending_size.as_local();
-            } else if let Some(server_size) = window.active_window().last_server_size() {
+            } else if let Some(mut server_size) = window.active_window().last_server_size() {
+                if let Some(ssd) = window.ssd_height(false) {
+                    server_size.h += ssd;
+                }
```
Drakulix: I looked at this and wondered "shouldn't this also apply to `pending_size()` in the branch above.

Then I realized the issue here is, that the former uses a method on the `CosmicWindow`/`CosmicStack` (dispatched through `CosmicMapped`), which takes SSDs into account, while `last_server_size` gets called directly on the `CosmicSurface`.

Could you instead please add a `last_server_size` method to the former, which handle SSDs internally, rather than add the fix here?

Drakulix (changes requested):

src/shell/layout/floating/mod.rs
hojjatabdollahi: Thank you. The previous fix wouldn't correctly handle stacked windows if tab height is different from SSD height.

Drakulix (approved):
Looks good now. Thanks!

jacobgkau (approved):
Was previously seeing this with Slack, and with any app while stacked; windows shrank after drag/drop, and after enabling/disabling tiling. All of those scenarios work properly with this fix.

PR #2321, opened 2026-04-25
feat: window snap action + pointer multiplier + larger resize border

jibsta210:
## Summary

Three small input/UX tweaks that I run as part of my daily COSMIC setup. They're independent — happy to split into separate PRs if maintainers prefer.

* **`Action::Snap(Direction)` handler** — implements keyboard window snapping (e.g. `Super+Arrow`) using the existing `floating_layer.move_element` primitive that drag-to-edge snapping already uses. Chained quarter snaps (Left then Down → BottomLeft quarter) work for free because that's how `move_element` already behaves.
* **Pointer sensitivity multiplier** — applies a configurable `COSMIC_POINTER_MULT` env var (default `1.25`) to relative pointer deltas. libinput's built-in acceleration is hard-capped at `[-1.0, 1.0]`, which feels sluggish on high-DPI mice; this lets users scale beyond that without patching libinput.
* **`RESIZE_BORDER` 10 → 20** — the 10px hit zone is fiddly to grab on a 4K display; 20px is comfortable and not visually noticeable.

## Motivation

* **Snap action**: there's no built-in keyboard window-snap shortcut in COSMIC today. The Action enum side of this lives in `cosmic-settings-daemon` (pop-os/cosmic-settings-daemon#144); this PR is the compositor side that handles the action. Without both, users either patch the compositor with hardcoded keybinds or write a sidecar.
* **Pointer multiplier**: an env-var escape hatch for the libinput cap. Cheap, opt-in, defaults to a near-no-op (1.25×).
* **Resize border**: a tiny constant change. Nothing in the codebase grep references the value other than its declaration and one consumer.

## Implementation

* `src/input/actions.rs` — new `Action::Snap(direction)` arm. Looks up the focused window via `seat.get_keyboard().current_focus()`, scoped to `KeyboardFocusTarget::Element`, and calls `workspace.floating_layer.move_element(direction, &seat, ManagedLayer::Floating, &theme, &window)`.
* `src/input/mod.rs` — pointer motion handler reads `COSMIC_POINTER_MULT` once per event and scales `event.delta()` before adding to `position`. Fall-through behaviour identical when env var is unset.
* `src/shell/element/window.rs` — `pub const RESIZE_BORDER: i32 = 10;` → `20;`.
* `Cargo.toml` — `[patch]` block points the two `cosmic-settings-daemon` dependencies at the `feat/lilypad-patches` branch on a fork. **This is a temporary patch and would be dropped once pop-os/cosmic-settings-daemon#144 lands** — the `Snap` variant comes from there.

## Testing

Daily-driven on COSMIC for the past several weeks. Snap behaves identically to drag-to-edge snapping (because it uses the same primitive). Pointer multiplier defaults preserve existing behaviour at the documented value. Resize border change is purely a hit-zone tweak.

## Open Questions

* The `[patch]` block in `Cargo.toml` is the awkward part — if you'd prefer I land the snap handler in a follow-up after the daemon PR merges, happy to split this PR into two (resize+pointer first, snap handler later).
* `COSMIC_POINTER_MULT` as an env var is a pragmatic choice; if you'd rather expose it through `cosmic-comp-config` I can do that instead.

## AI Disclosure

Patches were drafted with assistance from Claude (Anthropic). Commits include `Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>` trailers. I understand each change and will respond to review feedback. I'm aware of the AI-PR policy in the template — happy to rework or close if maintainers prefer.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -967,7 +967,7 @@ dependencies = [
 [[package]]
 name = "cosmic-settings-config"
 version = "0.1.0"
-source = "git+https://github.com/pop-os/cosmic-settings-daemon#e37160f14d1e7ee428f973cd2848b4e95f83dfe1"
+source = "git+https://github.com/jibsta210/cosmic-settings-daemon?branch=feat%2Flilypad-patches#2505b3b359645f5080896741bf18a672c72076c8"
 dependencies = [
  "cosmic-config",
  "ron 0.11.0",
@@ -988,7 +988,7 @@ dependencies = [
 [[package]]
 name = "cosmic-settings-daemon-config"
 version = "0.1.0"
-source = "git+https://github.com/pop-os/cosmic-settings-daemon#e37160f14d1e7ee428f973cd2848b4e95f83dfe1"
+source = "git+https://github.com/jibsta210/cosmic-settings-daemon?branch=feat%2Flilypad-patches#2505b3b359645f5080896741bf18a672c72076c8"
 dependencies = [
  "cosmic-config",
  "cosmic-theme",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -141,5 +141,11 @@ lto = "fat"
 cosmic-protocols = { git = "https://github.com/pop-os//cosmic-protocols", branch = "main" }
 cosmic-client-toolkit = { git = "https://github.com/pop-os//cosmic-protocols", branch = "main" }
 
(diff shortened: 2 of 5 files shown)


mmstick:
Same as https://github.com/pop-os/cosmic-settings-daemon/pull/144#issuecomment-4330157752

PR #2346, opened 2026-05-06
image-copy: Fixes for cursor capture sessions

ids1024:
We should probably handle this a bit differently, so we don't either need to duplicate the cursor session code every time `pointer.motion` is called, or have it miss out on pointer motions from "focus follows cursor" etc.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -640,11 +640,13 @@ impl State {
 
                     for session in cursor_sessions_for_output(&shell, &output) {
                         if let Some((geometry, offset)) = seat.cursor_geometry(
-                            position.as_logical().to_buffer(
-                                output.current_scale().fractional_scale(),
-                                output.current_transform(),
-                                &output_geometry.size.to_f64().as_logical(),
-                            ),
+                            (position - output_geometry.loc.to_f64())
+                                .as_logical()
+                                .to_buffer(
+                                    output.current_scale().fractional_scale(),
+                                    output.current_transform(),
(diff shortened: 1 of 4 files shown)


skygrango:
my [branch](https://github.com/pop-os/cosmic-comp/commit/29c6ca5172c51ed1cf5ed592294dfc039e1a8cc9) also references the code here, and I am indeed considering writing a helper function for reuse in cursor system synchronization

ids1024:
This now contains a fix for another issue (the definition of `surface_tree_offset` added by @skygrango in https://github.com/pop-os/cosmic-comp/pull/2357), so maybe it's a good time to merge the changes here. More improvements can be added in a later PR.

I assume the `surface_tree_offset` fix could affect a client that uses subsurfaces with pointer constraints, though I don't know what client would reproduce any issue there. The other changes here fix various issues with https://github.com/pop-os/xdg-desktop-portal-cosmic/pull/175, but otherwise generally don't have an impact because with don't use cursor capture sessions prior to that PR. The `surface_tree_offset` change was needed to get window capture to have the correct metadata cursor position for the `themed_window` sctk example.

skygrango:
The fix for `surface_tree_offset` is in line with my expectation. Thank you very much for the fix.

This also gave me ideas for improving the performance of `surface_geometry_offset_from_toplevel`.

Drakulix (approved):
LGTM

jklgrasso (approved):

PR #2353, opened 2026-05-07
kms: Try to re-open devices after resume on permission errors

Drakulix:
Draft until https://github.com/Smithay/smithay/pull/2025 is merged.

---

`libseat` [recommends to re-open file descriptors](https://git.sr.ht/~kennylevinsen/seatd/tree/master/item/include/libseat.h#L17) after suspend/resume, since they might have been permanently revoked.

We currently don't do this, which might lead to permanently blocked drm devices. Instead if we encounter any errors on resume for a particular device, try to re-open it.

Special care has been taken to:
1. Continue using the fast path, which destroys fewer resources, where possible.
2. Try to make 100% sure we can close the old file descriptor, since otherwise we fail the open call.
3. Try to keep as many resources (globals, outputs, shell state) untouched as possible to not potentially disrupt workspace arrangements or similar on resume.

### DIFF
--- a/src/backend/kms/device.rs
+++ b/src/backend/kms/device.rs
@@ -28,7 +28,7 @@ use smithay::{
         },
         egl::{EGLContext, EGLDevice, EGLDisplay, context::ContextPriority},
         renderer::glow::GlowRenderer,
-        session::Session,
+        session::{Session, libseat::LibSeatSession},
     },
     desktop::utils::OutputPresentationFeedback,
     output::{Mode as OutputMode, Output, PhysicalProperties, Scale, Subpixel},
@@ -40,7 +40,10 @@ use smithay::{
         wayland_server::DisplayHandle,
     },
     utils::{Clock, DevPath, DeviceFd, Monotonic, Point, Transform},
-    wayland::drm_lease::{DrmLease, DrmLeaseState},
+    wayland::{
+        drm_lease::{DrmLease, DrmLeaseState},
+        drm_syncobj::supports_syncobj_eventfd,
+    },
 };
 use tracing::{error, info, warn};
 use wayland_backend::server::ClientId;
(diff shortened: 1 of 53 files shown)


src/backend/kms/device.rs
```diff
@@ -489,6 +492,307 @@ impl State {
...
+                    leasing_global,
+                    active_leases: Vec::new(),
+                    active_clients,
+                },
+
+                supports_atomic,
+                texture_formats,
+                event_token: Some(token),
+                socket,
+            };
```
ids1024: This part, at least, is basically just duplicating `device_added`, right? Could we split it out somehow to avoid duplication?

ids1024:
Other than a bit of duplication, everything here looks reasonable. (Though I haven't tested it, or seen the behavior it is meant to address.)

src/backend/kms/device.rs
Drakulix: fixed

ids1024 (dismissed):
Moving the shared code into a `Device::new()` method is nice.

Everything here looks reasonable.

awilkins:
I built this locally yesterday to see if it would fix the hard crashes when my middle monitor disconnects itself (like other Samsung monitors I suspect it's powering off rather than just suspending).

Today I think it may have survived a resume that I think would have killed it previously ...

Drakulix:
Okay this is finally ready for review.

Due to https://github.com/Smithay/smithay/pull/2025 taking a bit of time, this ended up pulling in a bunch of other smithay changes, which required updates.

- smithay now has support for synchronizing geometry updates for resizing X11 surfaces. Cosmic-comp now includes support for it and seems to work as expected. Resizing x11 clients should be smoother now.
- @ids1024 has updated how minimizing x11 clients in smithay works now to align with what other compositors are doing.
  - This means we can finally get rid of our `set_mapped`-workaround, I cannot reproduce any clients turning into black squares after un-minimizing.
  - This seems to have the unfortunate side-effect of some clients stopping rendering, once they get minimized. This causes their window preview in the dock to be a white square. (e.g. steam) Fixing this properly would require cosmic-comp to keep a buffer from before minimizing and feeding that into window capture... This is out-of-scope for this PR and the new code definitely solves a few issues around minimizing x11 clients
- smithay includes new x11 window states. We use this to correctly report to clients when they are sticky. I also allow clients to set themselves sticky before they appear on screen for the first time. I have not found a client yet, that does that. Other states aren't integrated yet, though we might want to support `_NET_WM_STATE_ABOVE` at some point as this is apparently used by some game-overlay clients.
- Some focus issues with x11 clients are fixed, which might help games that were previously minimizing themselves on launch due to focus loss.

All of this is on top of the suspend-fix, which is mentioned in the original description and still is the main goal of this PR.

If any of the new features don't work properly, we should still go ahead with merging as long as nothing is regressing (apart from the minimized window preview).

ids1024 (approved):

jklgrasso (approved):

PR #2357, opened 2026-05-08
shell: implement pointer-constraints-unstable-v1

skygrango:
Needs https://github.com/Smithay/smithay/pull/2027
Inspired by: https://github.com/pop-os/cosmic-comp/pull/1861

This PR primarily implement the specification required for zwp_locked_pointer_v1, adds force unlock when focus change, and improves the user experience when the cursor is at the edge of the window.

I did the following:
- Constraints are enabled when all conditions are met, like PointerTargetFocus, KeyboardTargetFocus, Constraint Region
- Cursor Position Hint is applied when the constraint is removed
- Support sliding confinement: original approach made the mouse easily stick to the edge of the window, I changed it to check the possibility of the x or y axis, which greatly improves the user experience.
- Save cursor position hint in the seat to support Multiseats
- Support client_scale for xwayland

This protocol feature is a very important to games, especially first-person and third-person games, many games control mouse behavior through pointer-constraints, some professional software also requires this feature, but I don't have any of them to test it.

If there are still issues with camera control with xwayland, then it's probably not a problem with this protocol.

I still recommend using `PROTON_ENABLE_WAYLAND=1` whenever possible.

the games I have already tested:
- Persona 5 Royal
- Gray Zone Warfare
- ARC Raiders
- Crimson Desert
- Minecraft
- Ready or Not
- Broken Arrow
- Slay the Spire 2
- Clair Obscur: Expedition 33
- R.E.P.O
- FINAL FANTASY XIV
- HELLDIVERS 2

Many thanks to the System76 team for their hard work. cosmic-comp has already demonstrated impressive performance and stability.

With this implementation, no longer need to rely on gamescope to correct mouse behavior, which avoids unnecessary latency and performance overhead. the cursor is now correctly positioned according to the cursor hints provided by games, and pointer constraints are properly released when focus changes. Correct cursor behavior in games is essential for a smooth gaming experience, and this improvement helps make cosmic-comp a competitive gaming compositor alongside KWin from KDE.

I know this PR is a bit too large, so please let me know if there's anything I need to add.

:beers:  

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -2168,14 +2168,14 @@ impl State {
                     }
                     Stage::StickyPopups(layout) => {
                         if let Some(element) =
-                            layout.popup_element_under(global_pos.to_local(output))
+                            layout.popup_element_under(global_pos.to_local(output), seat)
                         {
                             return ControlFlow::Break(Ok(Some(element)));
                         }
                     }
                     Stage::Sticky(layout) => {
                         if let Some(element) =
-                            layout.toplevel_element_under(global_pos.to_local(output))
+                            layout.toplevel_element_under(global_pos.to_local(output), seat)
                         {
                             return ControlFlow::Break(Ok(Some(element)));
                         }
@@ -2330,7 +2330,7 @@ impl State {
                     }
                     Stage::StickyPopups(floating_layer) => {
                         if let Some(under) = floating_layer
-                            .popup_surface_under(relative_pos)
+                            .popup_surface_under(relative_pos, seat)
                             .map(|(target, point)| (target, point.to_global(output)))
                         {
                             return ControlFlow::Break(Ok(Some(under)));
@@ -2338,7 +2338,7 @@ impl State {
                     }
                     Stage::Sticky(floating_layer) => {
                         if let Some(under) = floating_layer
-                            .toplevel_surface_under(relative_pos)
+                            .toplevel_surface_under(relative_pos, seat)
                             .map(|(target, point)| (target, point.to_global(output)))
                         {
                             return ControlFlow::Break(Ok(Some(under)));
@@ -2392,11 +2392,8 @@ impl State {
                 let window_size = geometry.size.to_f64();
 
                 let is_legal = |p: Point<f64, Logical>| {
-                    let in_window = p.x >= 0.0
-                        && p.y >= 0.0
-                        // hack: prevent the cursor from touching the edge of the window
-                        && p.x <= window_size.w - 1.
-                        && p.y <= window_size.h - 1.;
+                    let in_window =
+                        p.x >= 0.0 && p.y >= 0.0 && p.x < window_size.w && p.y < window_size.h;
                     if !in_window {
                         return false;
                     }
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -505,12 +505,12 @@ impl State {
                             let new_under = State::surface_under(pos, &output, shell)
                                 .map(|(target, pos)| (target, pos.as_logical()));
 
-                            //We should only check size, not the surface, so that we can move the mouse over the OSD in confined mode
-                            // if new_under.as_ref().and_then(|(under, _)| under.wl_surface())
-                            //     != surface.wl_surface()
-                            // {
-                            //     return (false, None);
-                            // }
+                            // TODO: We might need a solution that allows constraints to bypass the surface without affecting the constraints themselves
+                            if new_under.as_ref().and_then(|(under, _)| under.wl_surface())
+                                != surface.wl_surface()
+                            {
+                                return (false, None);
+                            }
 
                             match surface {
                                 PointerFocusTarget::WlSurface { surface, .. } => {
@@ -2375,11 +2375,10 @@ impl State {
         pointer: &PointerHandle<Self>,
         mut location: Point<f64, Logical>,
     ) {
-        if let Some(client) = surface.client() {
-            let scale = self.client_compositor_state(&client).client_scale();
-            location.x /= scale;
-            location.y /= scale;
-        }
+        let Some(client) = surface.client() else {
+            return;
+        };
+        location = location.downscale(self.client_compositor_state(&client).client_scale());
 
         let point_and_output = {
             let shell = self.common.shell.read();
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -52,16 +52,21 @@ use smithay::{
             AxisFrame, ButtonEvent, GestureHoldBeginEvent, GestureHoldEndEvent,
             GesturePinchBeginEvent, GesturePinchEndEvent, GesturePinchUpdateEvent,
             GestureSwipeBeginEvent, GestureSwipeEndEvent, GestureSwipeUpdateEvent, MotionEvent,
-            PointerGrab, RelativeMotionEvent,
+            PointerGrab, PointerHandle, RelativeMotionEvent,
         },
         touch::{DownEvent, MotionEvent as TouchMotionEvent, UpEvent},
     },
     output::Output,
     reexports::{
-        input::Device as InputDevice, wayland_server::protocol::wl_shm::Format as ShmFormat,
+        input::Device as InputDevice,
+        wayland_server::{
+            Resource as _,
+            protocol::{wl_shm::Format as ShmFormat, wl_surface::WlSurface},
+        },
     },
-    utils::{Point, Rectangle, SERIAL_COUNTER, Serial},
+    utils::{Logical, Point, Rectangle, SERIAL_COUNTER, Serial},
     wayland::{
+        compositor::CompositorHandler,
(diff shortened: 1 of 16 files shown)


skygrango:
clean up commits and split them into smaller commits

move `find_window` to `src/shell/mod.rs` and rename it to `surface_geometry_offset_from_toplevel`
move `apply_cursor_hint` to `src/input/mod.rs`

this will be helpful for future support of `pointer_warp_v1`.

jackpot51:
https://github.com/Smithay/smithay/pull/2027 has been merged, can this be updated to use upstream smithay?

Also it needs to have conflicts resolved.

src/shell/mod.rs
```diff
@@ -727,6 +730,81 @@ impl WorkspaceSet {
...
+                        .data_map
+                        .get::<XdgPopupSurfaceData>()
+                        .and_then(|m| m.lock().unwrap().parent.as_ref().cloned())
+                })
+            {
+                root = parent;
+                continue;
+            }
+            break;
+        }
```
ids1024: `XdgPopupSurfaceData::parent` presumably is never a subsurface? So this should be possible to change to just a loop to get the root of the subsurface tree, then a loop through any `XDG_POPUP_ROLE`s, instead of nested loops.

skygrango: Thank you for pointing that out. Indeed, my original assumption was that `XDG_POPUP_ROLE` might be a subsurface of surface, but since that's impossible, we can simplify things here.

I added a commit for review, if you agree to my changes, I will rebase again.

src/input/mod.rs
```diff
@@ -493,50 +496,90 @@ impl State {
...
+                        let is_legal = |pos: Point<f64, Global>, shell: &Shell| {
+                            let new_under = State::surface_under(pos, &output, shell)
+                                .map(|(target, pos)| (target, pos.as_logical()));
+
+                            //We should only check size, not the surface, so that we can move the mouse over the OSD in confined mode
+                            // if new_under.as_ref().and_then(|(under, _)| under.wl_surface())
+                            //     != surface.wl_surface()
+                            // {
+                            //     return (false, None);
+                            // }
```
Drakulix: I am not sure we really want to remove this check. If we want to special case OSDs we should instead give layer-surfaces and other shell components a way to bypass constraints. This seems inconsistent and would for example not catch the Zoom UI, which we should probably also allow to be controlled among other things.

src/input/mod.rs
```diff
@@ -2309,6 +2360,130 @@ impl State {
...
+        &mut self,
+        surface: &WlSurface,
+        pointer: &PointerHandle<Self>,
+        mut location: Point<f64, Logical>,
+    ) {
+        if let Some(client) = surface.client() {
+            let scale = self.client_compositor_state(&client).client_scale();
+            location.x /= scale;
+            location.y /= scale;
+        }
```
Drakulix:
```suggestion
        let Some(client) = surface.client().unwrap() else { return };
        location = location.downscale(self.client_compositor_state(&client).client_scale());
```

src/input/mod.rs
```diff
@@ -2309,6 +2360,130 @@ impl State {
...
+            if let Some((output, geometry, surface_offset)) = found {
+                let mut pos_in_element = location + surface_offset.to_f64();
+                let window_size = geometry.size.to_f64();
+
+                let is_legal = |p: Point<f64, Logical>| {
+                    let in_window = p.x >= 0.0
+                        && p.y >= 0.0
+                        // hack: prevent the cursor from touching the edge of the window
+                        && p.x <= window_size.w - 1.
+                        && p.y <= window_size.h - 1.;
```
Drakulix: why? If this is necessary to be able to move the cursor right or down, wouldn't we also need to make sure we don't let the cursor touch the left or top edge to continue moving it?

If we simply want to make sure, that the cursor is inside the region (which isn't inclusive for the width/height), shouldn't we simply do this instead?

```rust
&& p.x < window_size.w
&& p.y < window_size.h;
```

Drakulix (commented):
Looks mostly good now, thanks!

Just a couple of questions/remarks.

src/input/mod.rs
skygrango: In my tests, this problem does not occur on the left and top sides, only right and bottom side triggered window interaction events, causing the cursor to no longer be considered within the surface. This might be an issue with other window handling, which is why I listed it as a "hack" here.

src/input/mod.rs
skygrango: In my original opinion, if any window can partially cover the current window, whether it's a window on top or a surface with special functions like a layer-shell (`wlr-layer-shell` is a special protocol that allows a program to determine whether it has an interactive surface, and it can change the geometry and position as needed at any time), then even with confined constraints, the constrains shouldn't prevent the user from moving the mouse over other surface. Even if this results in the mouse constraint being deactivated, this should be the behavior intended by the user. If there were this check, the mouse wouldn't be able to move over them.

Zoom UI sounds like a case I haven't considered. To ensure maximum compatibility with commonly used solutions, we could indeed bring this check back. We already have a way to forcibly remove constraints through focus switching, so let's do this for now. the constraint bypass methods you mentioned, can be addressed later.

src/input/mod.rs
Drakulix: So can you walk me through the code and explain to me, where what exactly is going wrong, when we don't substract one here, but p.x equals window_size.w?

skygrango: I tried setting it to 0.5 a few weeks ago, but the problem persisted. I didn't investigate further then; I need some time to confirm the solution you provided.

skygrango: I directly placed `apply_cursor_hint` into `cursor_position_hint` and applied your solution to windowed Minecraft, but it still causes the mouse to leave the game window.

skygrango: Based on my previous understanding, this might be related to focus_under function like: https://github.com/pop-os/cosmic-comp/blob/cefe8472b2df6051ea42045ea133aa4bc334130d/src/shell/mod.rs#L3638 

but at the time I thought it was beyond the scope of this PR, so I didn't address it.

git-f0x: Maybe somewhat related to #1907? That also fixed the pointer going outside the screen by 1 pixel on the right and bottom.

skygrango: Let me explain it more clearly.

From the perspective of the interactive geometry of the surface,@Drakulix is right. That's what I originally did too.

However, according to my understanding, `PointerFocusTarget` has a condition for determining whether it is a `WindowUI` element. In the current implementation, if `p.x` or `p.y` isn't subtracted by 1 px, compositor will assume the pointer has touched the window edge, then window resizing function is activated. This steals the focus that the Surface originally had, then triggers the removal of pointer constraints.

This is related to the window management strategy. It may also be related to the maximum size strategy, which is why I decided to give up 1 px of space, because I didn't want to break the original window management functionality in this PR, as this could have side effects.

If the strategy is that the window resizing function must respect the surface geometry, and this is considered a bug, then the best approach is to use a separate PR to achieve this goal.

skygrango: One solution is to check if the pointer has pointer constraints before checking the conditions of `WindowUI`; if so, skip that branch. However, this may require modifying some function interface. If this is the desired approach, we can use it to improve our work.

skygrango: @Drakulix I add commit https://github.com/pop-os/cosmic-comp/pull/2357/changes/1bd9bc9961bf3045fe5bd248cdb097acdbe35e71 to disable WindowUI check if constraint is active. In my tests, this fixed the issue of the cursor leaving the window. This proves that my original assumption was correct. we don't need 1 px hack now. ~~However, this may increase the difficulty for users to adjust the window size.~~ I think it's fine.

btw, before this fix, you need to set fractional scaling to over 150% to reproduce this bug.

Drakulix: Thank you, yeah this is a much better solution. Seems to be a combination of rounding inaccuracies at fractional scales and the resize handling for server-side decorated windows. You hack would've probably broken apart, once you needed to subtract more than one pixel (e.g. at 250% scaling). So thanks for getting to the bottom of this.

Drakulix (dismissed):
This looks good to me now! Thanks for your patience :)

skygrango:
Rebased on the latest main

ids1024:
I've added a commit to bump `smithay` again with https://github.com/Smithay/smithay/pull/2038, which apparently fixes a regression with game controller input. May as well test that along with this.

We have some more testing to do, but it will be good to get this merged.

WatchMkr (approved):

ids1024 (approved):

ids1024:
Things seem to be working well in our testing. Thanks for helping with this! It's good to finally get this working properly. Along with other recent fixes, gaming on Cosmic should be improved.

BlakeGardner:
My games are working perfectly now, no more issues with not being able to camera look and it hitting an invisible wall.

Issue #2363, opened 2026-05-11
Focus stealing prevention


NatoBoram:
> **Is your feature request related to a problem? Please describe.**
> A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]

When I start Discord, it takes a few different windows before it finally becomes usable. During that time, I'd like to be able to use my computer. If I write something in an app, start Discord on the side then go back to writing, Discord appears on the screen where my cursor is instead of on the monitor where I asked it to start on (https://github.com/pop-os/cosmic-comp/issues/2162) and repeatedly steals the focus of what I'm doing, so random keypresses are sent to Discord instead of what I was currently doing.

Similarly, in a Zoom call, its subtitle window appears and disappears when a different person speaks. Every time a new window is created, it steals the focus, so it becomes incredibly difficult to work.

> **Describe the solution you'd like**
> A clear and concise description of what you want to happen.

A setting similar to Xfce's "Focus stealing prevention" that, when enabled:

* Prevents any new windows from other applications from appearing on top of the application that you're currently using
* Prevents any window from stealing the focus and receiving inputs from the task you're currently doing

> **Describe alternatives you've considered**
> A clear and concise description of any alternative solutions or features you've considered.

Removing the ability to steal focus without user interaction entirely.

> **Additional context**
> Add any other context or screenshots about the feature request here.

When I used it in Xfce, it quickly became my favourite feature. It was strange at first to see new windows appear under my current one and not act like they own the place, but it was such a relief from normal. It made using a computer much more comfortable.

Hyperchaotic:
KDE Plasma has a setting "Focus Stealing Prevention". The setting ranges from "None" to "Extreme" and is [defined in this document.](https://docs.kde.org/stable_kf6/en/kwin/kcontrol/windowbehaviour/index.html#focus):

**_Focus stealing prevention_**

This option specifies how much KWin will try to prevent unwanted focus stealing caused by unexpected activation of new windows.

**None**: Prevention is turned off and new windows always become activated.

**Low**: Prevention is enabled; when some window does not have support for the underlying mechanism and KWin cannot reliably decide whether to activate the window or not, it will be activated. This setting may have both worse and better results than the medium level, depending on the applications.

**Medium**: Prevention is enabled.

**High**: New windows get activated only if no window is currently active or if they belong to the currently active application. This setting is probably not really usable when not using mouse focus policy.

**Extreme**: All windows must be explicitly activated by the user.

[...]

PR #2367, opened 2026-05-12
feat: add config option for xdg activation behavior

man0lis:
Currently, when a surface requests activation, cosmic-comp switches focus to it (in most cases/if a valid activation token exists). Even when the surface is on another workspace. This behavior can be quite unexpected and has caused me to enter a password into an IRC Chat window more than once.
I could not find a config option to make the focus switching less aggressive so I implemented this patch. I have daily driven cosmic-comp with this patch applied in the Urgent configuration for about 2 weeks now and did not notice any problems.  The FocusIfActiveWorkspace option was only tested once but seems to work.

This patch adds 3 ActivationPolicy options:
- `Focus`, the current behavior
- `FocusIfActiveWorkspace`, activates the surface if it is on the current workspace, else marks its workspace as urgent
- `Urgent`, does not change focus, just marks the target workspace as urgent

To test, create `~/.config/cosmic/com.system76.CosmicComp/v1/activation_policy` and put `Focus`, `FocusIfActiveWorkspace` or `Urgent` into it.

This might help to fix issue https://github.com/pop-os/cosmic-comp/issues/2363

I used an LLM to get an overview of the code base and find relevant parts of the code, but did not use it to write the code.

- [?] My change is accurately described in the commit message.

### DIFF
--- a/cosmic-comp-config/src/lib.rs
+++ b/cosmic-comp-config/src/lib.rs
@@ -99,6 +99,7 @@ pub struct CosmicCompConfig {
     pub edge_snap_threshold: u32,
     pub accessibility_zoom: ZoomConfig,
     pub appearance_settings: AppearanceConfig,
+    pub activation_policy: ActivationPolicy,
 }
 
 impl Default for CosmicCompConfig {
@@ -135,6 +136,7 @@ impl Default for CosmicCompConfig {
             edge_snap_threshold: 0,
             accessibility_zoom: ZoomConfig::default(),
             appearance_settings: AppearanceConfig::default(),
+            activation_policy: ActivationPolicy::default(),
         }
     }
 }
@@ -222,6 +224,14 @@ pub enum EavesdroppingKeyboardMode {
     Modifiers,
     Combinations,
     All,
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone, Copy, Default, PartialEq, Eq)]
+pub enum ActivationPolicy {
+    #[default]
+    Focus,
+    FocusIfActiveWorkspace,
+    Urgent,
 }
 
 #[derive(Debug, Deserialize, Serialize, Clone, Copy, Default, PartialEq, Eq)]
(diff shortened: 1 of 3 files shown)


Drakulix (commented):
This looks pretty good!

Ping @pop-os/ux. I think this is an excellent case for a config file option, since I have been running into this myself a bunch. I am however not sure, if we want to have this in cosmic-settings to some extend.

Additionally I am slightly worried the "Urgent"-state (although it has existed in the codebase for quite some time), isn't super visible/intuitive. More specifically currently the only thing using the urgent state, if I remember correctly (ping @wash2 / @ids1024  ?) is the workspace-applet. Neither the workspace overview show the state, nor is anything highlighting the urgent window in question. (I am not even sure we can currently communicate this through any of our protocols.) It would certainly be nice to maybe add a little dot to the icon in the dock, or highlight the urgent-tab of a stack somehow. (The latter is a particularily annoying issue right now, as the urgent state disappears as soon as you enter the workspace and if the stack window in question isn't the active one, you might completely miss, that something happened/finished.) This is more a question for @pop-os/ux to ponder about though, rather than an issue of this PR.

cosmic-comp-config/src/lib.rs
```diff
@@ -224,6 +226,15 @@ pub enum EavesdroppingKeyboardMode {
     All,
 }
 
+#[derive(Debug, Deserialize, Serialize, Clone, Copy, Default, PartialEq, Eq)]
+#[serde(rename_all = "lowercase")]
```
Drakulix: I don't think renaming to lowercase is necessarily what we want here though. I don't think any of our other config files (including existing enums) use it and `focusifactiveworkspace` isn't very readable. I'd suggest to just drop this line.

man0lis: Makes sense. I will change it. Probably just copy pasted it from below

man0lis:
Maybe some time in the future, we could have per surface urgent states, that propagate to the workspace urgent state. Surfaces with the urgent state could have a different outline color or tab color when in a stack. This state could then be cleared when the surface is focused (maybe with a configurable delay).

Drakulix (dismissed):
Code-wise this looks good to me. Thanks! :)

jklgrasso (dismissed):

man0lis:
git rebase upstream/master

Drakulix (approved):
Looks the same code-wise. No functional change due to rebase. @jklgrasso Can you give a quick re-approve?

leviport (approved):

PR #2376, opened 2026-05-14
fix(kms): defer process::exit() until surface threads are joined

### LINKED ISSUE #2375: cosmic-comp 1.0.12&1.0.13 trigger AMD panel refresh bug leading to compositor freeze
## Summary
I experienced cosmic-comp freezes since the upgrade to 1.0.12 on Arch on my amdgpu, RTX 4060 hybrid setup. 1.0.11 was rock-stable, 1.0.12 lead to freezes after seconds to minutes of usage. Some usage was always required, I never really got to a freeze idling on desktop.
Symptoms were: Image was frozen. VT-switch was not working, Music kept playing, Software volume keys on the keyboard still worked. I could pinpoint this down to a amdgpu or amd firmware bug:
Adding 
`
amdgpu.dcdebugmask=0x10
`
(Panel-Self Refresh turned off) to the kernel boot options resolved the issue.

So why do I open an issue against cosmic-comp, if this is caused by my driver/firmware/hardware combo? While debugging this, I found two things (not even calling them bugs here) in 1.0.12, one of which imho triggers this bug a lot more than necessary and I removed a SIGABRT on compositor teardown, which had nothing to do with the freeze.

The issues are (Disclaimer, these are written via Claude Code):

  1. ~~Spurious DRM cleanup commit on every udev event. `display_configuration()`
     runs on every udev `CHANGED` event and issues an `atomic_commit(ALLOW_MODESET)`
     even when its cleanup set is empty - a full modeset commit with nothing to do.
     On this firmware every `ALLOW_MODESET` commit is an opportunity to hit the
     hang, and most of these commits are unnecessary. It also treats `EPERM`/
     `EACCES` (no DRM master - routine on a hybrid GPU) as fatal, aborting
     `device_changed()`~~
     **Update: In 1.3.0 only the EPERM/EACCES errors are left. Check the PR below.**

  2. **SIGABRT on compositor teardown.** Unrelated to the freeze, but found while
     debugging: cosmic-comp aborts with SIGABRT when the compositor exits (greeter
     to session handoff, logout). The exit path calls `process::exit()`, which runs
     Mesa's `util_queue` atexit handler while KMS render threads are still in EGL
     calls. Confirmed by coredumps - every backtrace is a SIGABRT inside
     `exit()`. This is exit-path only and does not affect a running session.

## System Information

| | |
|---|---|
| **Machine** | ASUS ROG Zephyrus G14 GA402XV |
| **CPU** | AMD Ryzen 9 7940HS (Phoenix) |
| **iGPU** | AMD Radeon 780M / Phoenix1 — `driver=amdgpu` — primary display |
| **dGPU** | NVIDIA GeForce RTX 4060 Max-Q — `driver=nvidia-open 595.71.05` |
| **RAM** | 32 GiB DDR5-4800 |
| **OS** | Arch Linux |
| **Kernel** | 7.0.5-arch1-1 (also reproducible on LTS 6.18.29) |
| **Mesa** | libgallium-26.0.6-arch1.1 |
| **Desktop** | COSMIC (cosmic-greeter) |
| **Graphics mode** | Hybrid — AMD iGPU drives display; NVIDIA dGPU loaded via nvidia-open |

## Version that crashes

```
cosmic-comp 1:1.0.12-1
```

## Version that works

```
cosmic-comp 1:1.0.11-1
```
- #1953 — Hard freeze from amdgpu page faults on ASUS laptop with AMD Radeon 680M + NVIDIA RTX 3050 hybrid
- #2227 — cosmic-comp degraded/crashing on every boot with NVIDIA RTX 5080
- #2341 — Hard crash on COSMIC login within first second of cosmic-comp init with NVIDIA PRIME hybrid

timostrunk:
Before reviewing, please check #2375 for the problem analysis..

  When the kiosk child exits, calling process::exit() directly from the
  calloop callback causes Mesa's util_queue atexit handler to run
  concurrently with the KMS surface thread, which may still be mid-frame
  calling eglCreateSync(). This races on Mesa's internal heap state,
  producing "free(): corrupted unsorted chunks" and SIGABRT.

~Replace the three process::exit() calls with should_stop + a new
  kiosk_exit_code field, so the event loop exits cleanly and state is
  dropped — joining surface threads — before process::exit() is called.~
This is untrue, there is no explicit join in those codepaths.

  The race window is only reliably hit when a GPU-heavy client (Firefox,
  Thunderbird) is driving continuous rendering, explaining why the crash
  never occurred without one present.

  Introduced in 1.0.12 by the Smithay update that added export_sync_point()
  to GlesFrame::finish_internal; in 1.0.11 the surface thread did not call
  eglCreateSync() during frame finish, so the atexit was safe.

  **AI Usage:
  Claude Code did the root cause analyis especially of all the stacktraces  and all code for this bugfix.**

  Fixes #2375

  **Disclaimer here: I understand the changes, but I do not understand the complete effect this has on the teardown logic. Teardown logic is inherently complex and an expert needs to review this again. This would be true, even if I had written this code.**
  - In our experience, AI generated code often results in overly complex code that lacks enough context for a proper 
  - fix or feature inclusion. This results in considerably longer code reviews. Due to this, AI authored or partially authored PRs may be closed without comment.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -216,19 +216,17 @@ pub fn run(hooks: crate::hooks::Hooks) -> Result<(), Box<dyn Error>> {
                 // Kiosk child exited with status
                 Ok(Some(exit_status)) => {
                     info!("Command exited with status {:?}", exit_status);
-                    match exit_status.code() {
-                        // Exiting with the same status as the kiosk child
-                        Some(code) => process::exit(code),
-                        // The kiosk child exited with signal, exiting with error
-                        None => process::exit(1),
-                    }
+                    let code = exit_status.code().unwrap_or(1);
+                    state.common.kiosk_exit_code = Some(code);
+                    state.common.should_stop = true;
                 }
                 // Command still running
                 Ok(None) => {}
                 // Kiosk child disappeared, exiting with error
                 Err(err) => {
                     warn!(?err, "Failed to wait for command");
-                    process::exit(1);
+                    state.common.kiosk_exit_code = Some(1);
+                    state.common.should_stop = true;
                 }
             }
         }
@@ -239,9 +237,16 @@ pub fn run(hooks: crate::hooks::Hooks) -> Result<(), Box<dyn Error>> {
         let _ = child.kill();
     }
 
-    // drop eventloop & state before logger
+    let kiosk_exit_code = state.common.kiosk_exit_code;
+
+    // drop eventloop & state before logger — this joins surface threads
+    // before any Mesa atexit handlers can run
     std::mem::drop(event_loop);
     std::mem::drop(state);
+
+    if let Some(code) = kiosk_exit_code {
+        process::exit(code);
+    }
 
     Ok(())
 }
--- a/src/state.rs
+++ b/src/state.rs
@@ -241,6 +241,7 @@ pub struct Common {
     pub clock: Clock<Monotonic>,
     pub startup_done: Arc<AtomicBool>,
     pub should_stop: bool,
+    pub kiosk_exit_code: Option<i32>,
 
     pub gesture_state: Option<GestureState>,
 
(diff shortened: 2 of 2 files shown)


timostrunk:
The fix works, but is incorrect

PR #2381, opened 2026-05-15
fix(magnifier): smooth mouse wheel zoom and respect natural scroll direction

### LINKED ISSUE #2380: Accessibility: Magnifier zoom animation lags with mouse wheel, and zoom in/out is backwards when natural scrolling is enabled
Hi, I am a legally blind Rust developer and now using COSMIC Desktop a lot. The magnifier is essential to me. I cannot use the desktop without it.

There are two accessibility UX issues with mouse wheel zoom in the Magnifier, which I believe are easy to fix.

## Observed Behaviors

1. When zooming **with Super + the mouse wheel**, the zoom is "laggy." There is very little response while the mouse wheel is turning, then suddenly, the zoom jumps in or out when the mouse wheel stops, like a car that hesitates then lurches. **This lagging does not occur when zooming with Super + two-finger scroll on the touchpad.**
 
**Expectation:** Super + mouse wheel zoom should be as smooth as it is on the touchpad.

2. When **natural scrolling** is enabled, the **zoom direction is reversed**, making it very "unnatural." On both macOS and Windows, the wheel zoom in/out is always consistent: forward to zoom in, reverse to zoom out.

**Expectation:** Mouse wheel zoom should remain consistent, or be user-configurable independently of the system-wide "natural scroll."

## Proposed Fixes

I propose the following tiny fixes, both in [src/input/mod.rs](https://github.com/pop-os/cosmic-comp/blob/master/src/input/mod.rs):

1. **Disable animation for mouse wheel zoom.** The animation causes a timing race in which each "tick" of the wheel starts a new animated transition **from the current frame of the animation,** but before that animation can finish, subsequent ticks arrive and start the animation over again. This forces the displayed zoom to chase a moving target that lags behind the cumulative input. When the wheel stops, the animation is finally allowed to complete.

The animation is fine and smooth for keyboard zoom, but unnecessary for mouse wheel zoom, because each mouse wheel "tick" already produces a small, discrete (5%) zoom step. Adding animation only introduces lag.

Near [src/input/mod.rs:922](https://github.com/pop-os/cosmic-comp/blob/master/src/input/mod.rs#L922), simply change:
```rs
self.update_zoom(&seat, change, event.source() == AxisSource::Wheel);
```
to:
```rs
self.update_zoom(&seat, change, false);
```

2. **Short-term fix: Invert the effective wheel direction when natural scrolling is enabled.** 

Near [src/input/mod.rs:917](https://github.com/pop-os/cosmic-comp/blob/master/src/input/mod.rs#L917), add:
```rs
if event.relative_direction(Axis::Vertical)
    == AxisRelativeDirection::Inverted
{
    percentage *= -1.;
}
```

And add `AxisRelativeDirection` to imports from `smithay::backend::input`.

### Long-Term Fix

2b. If there is any objection to having mouse wheel zoom ignore the natural scrolling direction, and some users in the community prefer the inverted natural scroll zoom, I propose a configuration option in the **Accessibility** > **Magnifier** COSMIC Settings page, directly under "Enable mouse or touchpad zoom with Super + scroll":

**Zoom scroll direction follows natural scroll** (true/false, default: false).

Thank you for your commitment to accessibility in COSMIC!

tgrushka:
Closes #2380 .

- Disable animation for mouse wheel zoom to eliminate the timing race.
- Invert zoom direction when natural scrolling is enabled.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -37,11 +37,12 @@ use cosmic_settings_config::shortcuts;
 use cosmic_settings_config::shortcuts::action::{Direction, ResizeDirection};
 use smithay::{
     backend::input::{
-        AbsolutePositionEvent, Axis, AxisSource, Device, DeviceCapability, GestureBeginEvent,
-        GestureEndEvent, GesturePinchUpdateEvent as _, GestureSwipeUpdateEvent as _, InputBackend,
-        InputEvent, KeyState, KeyboardKeyEvent, PointerAxisEvent, ProximityState, Switch,
-        SwitchState, SwitchToggleEvent, TabletToolButtonEvent, TabletToolEvent,
-        TabletToolProximityEvent, TabletToolTipEvent, TabletToolTipState, TouchEvent,
+        AbsolutePositionEvent, Axis, AxisRelativeDirection, AxisSource, Device, DeviceCapability,
+        GestureBeginEvent, GestureEndEvent, GesturePinchUpdateEvent as _,
+        GestureSwipeUpdateEvent as _, InputBackend, InputEvent, KeyState, KeyboardKeyEvent,
+        PointerAxisEvent, ProximityState, Switch, SwitchState, SwitchToggleEvent,
+        TabletToolButtonEvent, TabletToolEvent, TabletToolProximityEvent, TabletToolTipEvent,
+        TabletToolTipState, TouchEvent,
     },
     desktop::{PopupKeyboardGrab, WindowSurfaceType, utils::under_from_surface_tree},
     input::{
@@ -914,12 +915,18 @@ impl State {
                             .or_else(|| event.amount(Axis::Vertical))
                             .map(|val| val * scroll_factor)
                         {
+                            if event.relative_direction(Axis::Vertical)
+                                == AxisRelativeDirection::Inverted
+                            {
+                                percentage *= -1.;
+                            }
+
                             if event.source() == AxisSource::Wheel {
                                 percentage *= 5.;
                             }
 
                             let change = -(percentage / 100.);
-                            self.update_zoom(&seat, change, event.source() == AxisSource::Wheel);
+                            self.update_zoom(&seat, change, false);
                         }
                     } else {
                         let mut frame = AxisFrame::new(event.time_msec()).source(event.source());
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -926,7 +926,7 @@ impl State {
                             }
 
                             let change = -(percentage / 100.);
-                            self.update_zoom(&seat, change, false);
+                            self.update_zoom(&seat, change, event.source() == AxisSource::Wheel);
                         }
                     } else {
                         let mut frame = AxisFrame::new(event.time_msec()).source(event.source());
--- a/src/shell/zoom.rs
+++ b/src/shell/zoom.rs
@@ -9,7 +9,7 @@ use cosmic::{
 };
 use cosmic_comp_config::ZoomMovement;
 use cosmic_config::ConfigSet;
-use keyframe::{ease, functions::EaseInOutCubic};
+use keyframe::{ease, functions::Linear};
 use smithay::{
     backend::renderer::{ImportMem, Renderer, element::AsRenderElements},
     desktop::space::SpaceElement,
@@ -135,7 +135,7 @@ impl OutputZoomState {
             let percentage =
                 duration_since.as_millis() as f32 / ANIMATION_DURATION.as_millis() as f32;
             ease(
-                EaseInOutCubic,
+                Linear,
                 EasePoint(*old_point),
                 EasePoint(self.focal_point),
                 percentage,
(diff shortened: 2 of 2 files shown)


tgrushka:
@Drakulix @ids1024 , I am new to this repo, and the guidelines say I should request reviewers. Not sure if you are the right people to ask, but the sidebar won't let me add any. Thank you for your commitment to a11y!

leviport:
I added the requests already. We most likely won't get to reviewing until next week though. Thank you for the contribution!

Drakulix:
>     * Disable animation for mouse wheel zoom to eliminate the timing race.

Hmm, what timing race is this referring to? I am worried that this might look very erratic with a high step size.

tgrushka:
> Hmm, what timing race is this referring to? I am worried that this might look very erratic with a high step size.

@Drakulix , I see your point. I am using an MX Master 3S, which uses high-resolution scrolling. I am willing to get a lower resolution mouse to test this. Is there a way to tell from the API if the wheel has a higher or lower step rate? For example, would those send low res instead of high res scroll events?

leviport:
I think the delay mentioned in https://github.com/pop-os/cosmic-comp/issues/2380 is more due to the free spinning nature of the MX Master 3S's scroll wheel, but I do see an improvement with this change.

Here is what it looks like when I give my MX Master 3's scroll wheel a good spin in either direction:

Before this change:
[mx master 3 without patch.webm](https://github.com/user-attachments/assets/f6188f2c-a980-4ab8-abe2-2b93c80c0199)

After this change:
[mx master 3 with patch.webm](https://github.com/user-attachments/assets/b4fe2602-ae81-49ce-a365-9c88aed684a9)

This change definitely improves the situation with this MX Master 3. I also tried my M570 trackball, which has a more notchy/traditional scroll wheel that does not keep spinning freely like the MX Master, and the difference is much less pronounced, but definitely still present if I scroll harder.

M570 trackball before:
[m570 without patch.webm](https://github.com/user-attachments/assets/592e57f8-4e17-4a71-b71b-2e73c45be912)

M570 trackball after:
[m570 with patch.webm](https://github.com/user-attachments/assets/20bc2a53-1abc-42c0-88d4-5ce2b75ba99a)

---

For the second part of the PR though, I'm not seeing the scroll direction change, either on a mouse or on a touchpad, when I enable or disable natural scrolling in cosmic-settings for either device.

I'm also a little less clear on the purpose of that change. Natural scrolling changes whether scrolling up/down moves the view up or down, or moves the content on the screen up or down. Since zoom is in and out and both my scroll devices are horizontal on my desk, scrolling "up" moves my finger in towards the screen on a scroll wheel, and my scrolling hand in towards the screen on a touchpad. The scroll "down" motion feels like scrolling out and away from my screen.

Admittedly I don't make super regular use of the magnifier or of natural scrolling, but I'm wondering if a natural scrolling/scroll direction inverting switch would be better in Settings > Accessibility > Magnifier next to the "Enable mouse or touchpad zoom with Super + scroll". That way people could set it separately from natural scrolling, depending on their preferences. We should probably ask @pop-os/ux for input since we're delving into UX territory at this point.

tgrushka:
Thanks @leviport  for helping test and capturing the videos. I just tried something else, can you please check it out? Instead of disabling it for the wheel entirely, I tried changing the easing function from `EaseInOutCubic` to `Linear`. I understand the issue better now, that basically, every tick of the zoom wheel was starting the easing function over again. The `EaseInOutCubic` always starts out slowly, speeds up, slows down, then ends slowly. The "starting up slowly" every time was causing the lag. Testing it again, now I don't really notice any difference between `Linear` animation and no animation at all with my MX Master 3S. I hope this might address @Drakulix 's concern, which is valid, especially for users with basic office mice, which often have a coarser wheel click.

tgrushka:
> For the second part of the PR though, I'm not seeing the scroll direction change, either on a mouse or on a touchpad, when I enable or disable natural scrolling in cosmic-settings for either device.

@leviport , do you mean when testing my PR you don't notice a change? That was intentional -- I posit that the zoom direction should not change when natural scrolling is enabled or disabled. In the existing code, the zoom direction gets inverted because "natural scrolling" is a system-wide change. So the intention of the second part was to make turning the wheel forward always zoom in, and turning it backwards always zoom out, no matter what the natural scrolling setting is. So the code looks like it "reverses" it, but actually it has to, because the compositor treats natural scrolling globally. Does this make sense? Or, are you somehow seeing no change in zoom direction based on the natural scrolling in the release build?

EDIT: Consistent zoom direction is commensurate with Zoom behavior on both Mac and Windows. On either of those OSes, when the user has enabled natural scrolling, the zoom direction remains consistent. See my original issue #2380 .

Drakulix (approved):
Code wise this looks good to me now, thanks! Using linear-interpolation for the zoom animation makes sense.

So from the engineering side this is good to go, but of course QA (and potentially UX) still need to sign it off.

leviport:
Oh I see, it _currently_ inverts and this PR makes it _not_ invert. Sorry, the way I read the second bullet point in your description made me think the inverse is true. Sounds like we're on the same page then; natural scrolling should _not_ invert the scroll direction for Super + scroll.

I'll do a little more testing with the new animation and approve as long as that looks good.

leviport (approved):
The linear ease looks great! Thank you for these improvements!

tgrushka:
Thank you @leviport  and @Drakulix !

PR #2382, opened 2026-05-16
Two independent KMS correctness fixes, relevant to hybrid-GPU systems

timostrunk:
**Update 2026-07-20:** I rebased to current master, the commits are a lot cleaner now, because master also contained two of my fixes, surviving are:

Fix commit 1: Exit via `libc::_exit()` on the kiosk-child exit path, so Mesa's atexit handlers cannot run while KMS surface threads are still inside EGL calls,  removes a reliable SIGABRT on greeter to session handoff and at logout.

Fix commit 2: Treat EPERM/EACCES from the DRM cleanup `atomic_commit()` as a skip rather than a hard error, so device enumeration continues on hybrid-GPU systems where the compositor does not hold DRM master on a secondary node.

## Old PR description:
This PR fixes on SIGABRT on teardown and skips DRM cleanup, when it is not needed or master is missing. Both are separated by commit. By testing a lot, I found that fix number 2 (the skipping of extra cleanup commits) triggers the AMD panel refresh freeze bug mentined in #2375 a lot less often. (Freezes go from happening after seconds to hours, but they still occur after going to 1.0.12).
Both commits should be reviewed independently. If required I can also split the PR.

**All fixes were done using Claude Code, as was the analysis. I also want to point out that I do understand the fixes, but I do not understand their complete fallout. I am still sending the PR, because it makes my system so much more stable without the explicit amdgpu boot option mentioned in #2375 .**

## Commits

**`fix(kms): exit via _exit() to avoid SIGABRT on compositor shutdown`**
cosmic-comp aborts with SIGABRT when the compositor exits — at the
greeter→session handoff and at logout. Those paths call `process::exit()`,
which runs Mesa's `util_queue` atexit handler while the KMS render threads
are still in EGL calls; the process aborts (confirmed by coredumps -
all SIGABRT inside `exit()`). This is an exit-path bug; it does not affect a
running session.
Exit with `libc::_exit()`, which skips atexit handlers, after `pause()`-ing
the DRM devices to release master. Rust destructors still run via explicit
`drop()`; only libc atexit handlers are skipped. `_exit()` also can't hang on
a stuck render thread, which a join-based fix would. The kiosk exit code is
carried out of the event loop in `Common::kiosk_exit_code` so the `_exit()`
happens once, after the loop returns.
*Trade-off:* `_exit()` skips atexit/stdio-flush. cosmic-comp registers no
atexit handlers of its own, and logging goes through journald (eager), so
nothing meaningful is lost; render-thread GPU resources are reclaimed by the
kernel, as in 1.0.11.
This is most probably the most controversial. I tried around a lot here by explicitly joining, but nothing worked.

**`fix(kms): skip DRM cleanup commit when not needed or master is missing`**
`display_configuration()` runs on every udev `CHANGED` event. It issued a
no-op `atomic_commit(ALLOW_MODESET)` even with an empty cleanup set, and
treated `EPERM`/`EACCES` (no DRM master - routine on a hybrid GPU) as fatal,
aborting `device_changed()`. Guard the cleanup block on a non-empty set, and
treat `PermissionDenied` as a skip so the rest of `device_changed()` runs.
*(Info after rebase: the atomic_commit stuff has been fixed on master already)*

## Scope
Commit 1 is confirmed by coredumps. Commit 2 is a correctness fix evident
from the code path.

Neither addresses the display freezes seen on this hardware. Those are a
separate AMD DMCUB/DCN3.1 firmware hang (`flip_done timed out` after a
hotplug-driven modeset), unrelated to this PR and mitigated at the kernel
level by disabling panel self-refresh (`amdgpu.dcdebugmask`).

## Testing

Commit 1: across repeated greeter-login and logout-login cycles, the
exit-path SIGABRT no longer occurs, whereas they always occured before.

Commit 2: external-display hotplug exercised over ~10 connect/disconnect
cycles on a DisplayPort MST dock, driving `device_changed()` /
`display_configuration()` repeatedly - no crash, no regression. Not run as a
controlled before/after against an unpatched build.

- [ ] I understand these changes in full and will be able to respond to review comments.
  - I cannot in good faith put an X here. The commits are simple to understand on their own, but I do not understand their full fallout. I can respond to review comments and provide coredumps.

### DIFF
--- a/src/backend/kms/drm_helpers.rs
+++ b/src/backend/kms/drm_helpers.rs
@@ -75,6 +75,7 @@ pub fn display_configuration(
     // And then cleanup
     if device.is_atomic() {
         let mut req = AtomicModeReq::new();
+        let mut has_changes = false;
         let plane_handles = device.plane_handles()?;
 
         // We cannot just shortcut and use the legacy api for all cleanups because of this.
@@ -95,10 +96,15 @@ pub fn display_configuration(
                     let fb_id = get_prop(device, plane, "FB_ID")?;
                     req.add_property(plane, crtc_id, property::Value::CRTC(None));
                     req.add_property(plane, fb_id, property::Value::Framebuffer(None));
+                    has_changes = true;
                 }
             }
         }
-        device.atomic_commit(AtomicCommitFlags::ALLOW_MODESET, req)?;
+        // Skip an empty commit: a no-op modeset that also fails with EPERM
+        // without DRM master (e.g. a render-only secondary GPU).
+        if has_changes {
+            device.atomic_commit(AtomicCommitFlags::ALLOW_MODESET, req)?;
+        }
     } else {
         for crtc in res_handles.crtcs() {
             #[allow(deprecated)]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,7 +16,7 @@ use smithay::{
 };
 
 use anyhow::{Context, Result};
-use state::{LastRefresh, State};
+use state::{BackendData, LastRefresh, State};
 use std::{
     env,
     ffi::OsString,
@@ -222,19 +222,17 @@ pub fn run(hooks: crate::hooks::Hooks) -> Result<(), Box<dyn Error>> {
                 // Kiosk child exited with status
                 Ok(Some(exit_status)) => {
                     info!("Command exited with status {:?}", exit_status);
-                    match exit_status.code() {
-                        // Exiting with the same status as the kiosk child
-                        Some(code) => process::exit(code),
-                        // The kiosk child exited with signal, exiting with error
-                        None => process::exit(1),
-                    }
+                    // Stop cleanly so surface threads are joined before exit() (signal -> 1).
+                    state.common.kiosk_exit_code = Some(exit_status.code().unwrap_or(1));
+                    state.common.should_stop = true;
                 }
                 // Command still running
                 Ok(None) => {}
                 // Kiosk child disappeared, exiting with error
                 Err(err) => {
                     warn!(?err, "Failed to wait for command");
-                    process::exit(1);
+                    state.common.kiosk_exit_code = Some(1);
+                    state.common.should_stop = true;
                 }
             }
         }
@@ -245,9 +243,31 @@ pub fn run(hooks: crate::hooks::Hooks) -> Result<(), Box<dyn Error>> {
         let _ = child.kill();
     }
 
+    let kiosk_exit_code = state.common.kiosk_exit_code;
+
+    // Join surface threads before exit() so no thread is mid-eglCreateSync when
+    // Mesa's atexit handlers run and corrupt the heap (issue #2375). Safe here
+    // because the event loop has stopped; an unconditional join in Surface::Drop
+    // would instead deadlock against apply_config_for_outputs.
+    if let BackendData::Kms(kms) = &mut state.backend {
+        // Release master first so the surface drop path skips its blocking commit.
+        for device in kms.drm_devices.values_mut() {
+            device.drm.pause();
+        }
+        for device in kms.drm_devices.values_mut() {
+            for (_, surface) in device.inner.surfaces.drain() {
+                surface.drop_and_join();
+            }
+        }
+    }
+
     // drop eventloop & state before logger
     std::mem::drop(event_loop);
     std::mem::drop(state);
+
+    if let Some(code) = kiosk_exit_code {
+        process::exit(code);
+    }
 
     Ok(())
 }
--- a/src/state.rs
+++ b/src/state.rs
@@ -244,6 +244,7 @@ pub struct Common {
     pub clock: Clock<Monotonic>,
     pub startup_done: Arc<AtomicBool>,
     pub should_stop: bool,
+    pub kiosk_exit_code: Option<i32>,
 
     pub gesture_state: Option<GestureState>,
 
@@ -754,6 +755,7 @@ impl State {
                 clock,
                 startup_done: Arc::new(AtomicBool::new(false)),
                 should_stop: false,
+                kiosk_exit_code: None,
                 gesture_state: None,
 
                 kiosk_child: None,


timostrunk:
I think this is by now fixed on main. I will retry and close.

timostrunk:
Not fixed yet, but hit much rarer, I will rebase to main.

timostrunk:
I did not test this yet after rebase, therefore
- [x] Actually test this before review.

Tested that it starts and works

timostrunk:
Rebase summary:

1. Commit 1 survived the rebase - applied cleanly and error condition still exists (on paper, cannot trigger it anymore in 1.3.0, but still worth a fix.
2. Commit 2 got a lot smaller, because two of the error conditions are just gone. The only leftover is not treating EPERM as fatal and rather just skip.

src/backend/kms/drm_helpers.rs
```diff
@@ -98,7 +98,18 @@ pub fn display_configuration(
...
+        if let Err(err) = device.atomic_commit(AtomicCommitFlags::ALLOW_MODESET, req) {
+            if err.kind() == std::io::ErrorKind::PermissionDenied {
+                tracing::warn!(
+                    ?err,
+                    "Failed to clean up DRM device state (no master), skipping"
+                );
+                return Ok(map);
+            }
+            return Err(err.into());
+        }
```
Drakulix, line 107: This isn't correct, as it would allow to partially apply a screen configuration, which we definitely want to abort on. We should rather check for empty commits instead.

src/lib.rs
```diff
@@ -245,11 +243,21 @@ pub fn run(hooks: crate::hooks::Hooks) -> Result<(), Box<dyn Error>> {
...
+    // pause() does not block.
+    if let BackendData::Kms(kms) = &mut state.backend {
+        for device in kms.drm_devices.values_mut() {
+            device.drm.pause();
+        }
+    }
 
-    Ok(())
+    // _exit() skips libc atexit handlers, avoiding the Mesa util_queue/EGL
+    // teardown race that aborts the process: https://github.com/pop-os/cosmic-comp/issues/2375
```
Drakulix: Dropping the KMS backend should join the surface-threads. We should rather figure out why this isn't working.

Drakulix (changes requested):

src/backend/kms/drm_helpers.rs
timostrunk: Ok, I will rewrite it this way.

src/lib.rs
timostrunk: I tried a few ways to explicitly join again. I will outline what I found and set the PR to draft until then.

timostrunk: TL;DR:
I tried at first to explicitly join, but that lead to issues, such as a deadlock. I tried around a bit more and now have the following apparently working version:

I stop the event loop via a should_stop member, pause all devices (this is new, didn't try this before), drop_and_join the surfaces and then exit. At that point the event loop is stopped so there's no outstanding handshake to deadlock on, and master is already released so there's no blocking commit.

The shutdown join is currently unbounded. Given the flip_done experience I explain below in the non TL;DR section, I'm slightly wary of a surface thread wedged in a DRM ioctl hanging shutdown, but I couldn't produce that after the `pause()` change, so I left the timeout out. I did see it in the version without the pause.

End of TL;DR

As to my previous tries and why they didn't work:
1. First I tried to simply drop. This lead to a deadlock. Disclaimer: Opus explanations (which I reviewed and which make sense to me): 
_Unconditional join in `Surface::Drop` lead to deadlock.** This is why the `join()` in `Drop` is commented out today. `Surface::Drop` also runs during `apply_config_for_outputs`, i.e. live output reconfiguration, and there the main thread is driving the surface thread through the synchronous `ThreadCommand` handshakes (`NodeAdded`/`NodeRemoved`/`Suspend` all do a `send` followed by a blocking `rx.recv()`), while the surface thread itself takes `shell.read()` in `redraw`. If `Drop` joins in that context, the main thread ends up blocked on a thread that is in turn waiting on the main thread. It leads to a hang every time you reconfigure outputs._
2. Explicit `drop_and_join()` at shutdown, but while still holding DRM master: the SIGABRT was gone, but I traded it for a ~7-minute freeze at the *next* session's startup.** The surface thread's teardown runs `AtomicDrmSurface::drop() to clear_state() to atomic_commit(ALLOW_MODESET)`, and that commit blocks waiting on a GPU fence / flip completion. So the greeter's compositor sits inside that blocking commit while still holding master. Meanwhile the user session's compositor starts up, calls `set_master()`, fails because the old process hasn't released it, and falls back to unprivileged mode  and logind never re-sends `SessionResumed`, so there's no retry. The session just sits there with no rendering until the fence finally unblocks. In my case that was around seven minutes, lining up with an NVIDIA power-state sync timeout. The fix was to call `DrmDevice::pause()` on each device *before* joining: `pause()` flips `active` to false so the drop path skips `clear_state()` entirely and releases master immediately, so the incoming compositor gets it right away and the join no longer runs any blocking DRM work
3. Joining on connector change (`device_changed`) could hang on its own. Separately from shutdown, when a connector goes away and comes back, the old surface has to be joined before the replacement is created on the same CRTC  otherwise the old render thread's `DrmCompositor` drop issues an `atomic_commit(ALLOW_MODESET)` disabling the CRTC while the *new* surface is already driving it, which gave me `flip_done timed out` and a display-engine hang. But a plain join there can itself block if the render thread is stuck in an ioctl, so at the time I bounded it with a 2s timeout and detach-on-timeout.

src/backend/kms/drm_helpers.rs
timostrunk: Done

src/lib.rs
Drakulix: > The surface thread's teardown runs AtomicDrmSurface::drop() to clear_state() to atomic_commit(ALLOW_MODESET), and that commit blocks waiting on a GPU fence / flip completion.

The problem here is, that we likely want *some* cleanup.

E.g. we want to remove any overlay planes in use, in case the launched compositor (not cosmic-comp) doesn't know how to handle them.

The timing of greetd + cosmic-greeter here is really unfortunate. Just because the child has existed, we shouldn't immediately proceed to the next session while cosmic-comp still runs. But I guess greetd doesn't wait for the process to exit, once the ipc connection is complete.

Well, it probably makes sense to make cleanup the responsibility of the next compositor until we can fix this, given upstream discussions like https://lore.kernel.org/all/20260708-drm-reset-state-flag-v1-0-c37dc985485d@kernel.org/ also point to that direction...

Drakulix (approved):
LGTM now. Thanks!

timostrunk:
Thanks for the approval. I think the clippy-check might not be related to the PR here. Let me know, if I should fix that or whether it is known breakage.

Drakulix:
> Thanks for the approval. I think the clippy-check might not be related to the PR here. Let me know, if I should fix that or whether it is known breakage.

You don't need to care about it. The last thing missing for merge is the QA team taking a look at it.

jacobgkau (approved):
We don't have any hybrid-GPU hardware with an AMD GPU to test with. But I gave this a spin on a hybrid Intel/NVIDIA system, a dedicated AMD GPU system, and a dedicated NVIDIA GPU system, and didn't see any problems with basic desktop or gaming usage.

Issue #204, opened 2023-10-25
Resizing with the mouse in tiling mode causes windows to overlap 


hojjatabdollahi:
steps to reproduce

1- spawn 3 windows (let's say gnome-terminal)
2- Put them in 3 columns like this: [[A,B], [C]] 

Notice windows A and B are grouped together.

3- grab the right edge of window B and try to increase its size, it will overlap window C
4- This doesn't happen if all windows have their own space so [[A], [B], [C]] works fine.

This happens on the master branch and the latest version of the ws-refactor.

Here is a video:

https://github.com/pop-os/cosmic-comp/assets/2222199/c946b427-abad-4a0f-bd06-6a1e45f865e2

Drakulix:
Oh this is weird. Thanks a bunch for reporting!

Drakulix:
So I have been tracking down this issue to a floating point rounding error, when propagating the new size from the outer group to the inner group on the left.

This has been fixed on the ws-refactor branch with [this commit](https://github.com/pop-os/cosmic-comp/pull/194/commits/33688e9bcf948a92fe21e6a2f9c2c1fa0f6c3baa), which at least prevents the overlap.

However the resizing is still not behaving very well, as this rounding error accumulates, causing the left window inside the group to be resized by a far greater amount, destroying the initial ratio of the group. I am still trying to come up with a better solution, that takes the original state before the resize into account, but in the meantime this doesn't result in broken behaviour any more at least.

hojjatabdollahi:
Thank you. It's fixed.

PR #2386, opened 2026-05-17
feat(input): add compositor-side kinetic scrolling

damianvander:
## Summary

Add an opt-in `KineticScroller` that buffers finger-source axis-event
deltas, computes a velocity-weighted lift, and emits a decaying synthetic
axis-event stream until the fling decays below a minimum velocity.

Config: `kinetic: Option<bool>` on `ScrollConfig` (per-device or global
default), defaults to **off**. No cosmic-settings UI yet.

## Why compositor-level, given the prior discussion in #204(https://github.com/pop-os/cosmic-epoch/issues/204)

GNOME (GTK) and KDE (Qt/QtQuick, system-wide as of Plasma 6.4) implement
kinetic at the toolkit. macOS does the same via AppKit. That works
because most apps in those environments share a toolkit. It does not
work for us:

1. **iced/libcosmic has no kinetic today.** COSMIC's own apps are the
   immediate gap. Doing this in the compositor closes it now without
   waiting on upstream iced.
2. **Per-app scroll behaviour is only possible at the compositor.** Each
   toolkit only sees its own surfaces. The compositor is the one place
   where "kinetic on for app X, off for app Y, custom decay for Z" can
   exist at all. That's the longer-term opportunity I'd like to set up
   here — a follow-up PR can add per-app rules + a cosmic-settings page.
3. Apps with no kinetic story at all (Java/Swing, some terminals, older
   toolkits) get the feature for free.

I'm aware this direction is contested. If maintainers would rather see
the effort go into iced upstream instead, I'd rather hear that now than
after the per-app config work.

## Known issues (carried into follow-up PRs)

The compositor-toolkit interaction warned about in (cosmic-epoch #204) is real:

- **Firefox**: scroll appears to "stick" until 2-finger lift. Caused by
  duplicate `axis_stop` frames Gecko emits when it sees finger-source
  events. Workaround written, kept out of this PR for size.
- **Some Java/Swing apps**: same, plus duplicate `hold_start`/`hold_end`
  frames. Workaround written, same reason.

These are precisely why the default is off and why per-app rules are the
intended next step. Kinetic-aware toolkits would land in the off-by-
default list; the user gets one consistent fling for everything else.

I'll open the workaround PRs separately if this approach is accepted.

## Protocol-semantics decision

At finger-lift when a fling is starting, we do **not** emit
`wl_pointer.axis_stop`. Kinetic-aware toolkits interpret `axis_stop` as
"user input ended, apply your own inertia," and would compound theirs
with ours. The synthetic events use `AxisSource::Continuous` so apps
treat them as already-smoothed motion. `axis_value120` is emitted
alongside `value` with fractional-remainder accumulation so the
high-res wheel stream stays smooth as the fling decays.

Cancellation paths: button press, key press, 2-finger touchpad land,
keyboard focus change, and (via `push()`) the first delta of a renewed
finger scroll.

## Files changed

- `src/input/kinetic.rs` (new, ~330 lines): state machine + plumbing
- `src/input/mod.rs` (+47): mod decl + thin hooks in PointerAxis /
  Button / Keyboard / GestureHoldBegin arms
- `src/shell/focus/mod.rs` (+8): cancel on focus change
- `src/shell/seats.rs` (+1): register on seat user_data
- `src/config/mod.rs` (+8): `kinetic_scroll_enabled` accessor
- `src/config/input_config.rs` (+1): default init
- `cosmic-comp-config/src/input.rs` (+2): field

## Testing

Tested on Pop!_OS 24.04 across a mix of libcosmic, GTK, Qt, Electron,
Firefox, and Java apps. libcosmic/iced apps gain inertia. GTK and Qt
apps work but with the compounding caveat noted above on aggressive
flings. Firefox and Java exhibit the documented bugs.

## AI disclosure

Claude Code was used for project exploration, code-generation
assistance, and drafting commit messages. I have read and understood
every line of the diff and am able to defend or revise any of it in
review.

- [x] I have read the Developer Certificate of Origin and certify my
      contribution under its conditions.

### DIFF
--- a/cosmic-comp-config/src/input.rs
+++ b/cosmic-comp-config/src/input.rs
@@ -47,6 +47,8 @@ pub struct ScrollConfig {
     pub natural_scroll: Option<bool>,
     pub scroll_button: Option<u32>,
     pub scroll_factor: Option<f64>,
+    #[serde(skip_serializing_if = "Option::is_none", default)]
+    pub kinetic: Option<bool>,
 }
 
 #[derive(Copy, Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
--- a/src/config/input_config.rs
+++ b/src/config/input_config.rs
@@ -65,6 +65,7 @@ pub fn for_device(device: &InputDevice) -> InputConfig {
                     None
                 },
                 scroll_factor: None,
+                kinetic: None,
             })
         } else {
             None
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -622,6 +622,14 @@ impl Config {
         .map_or(1.0, |x| x.0)
     }
 
+    pub fn kinetic_scroll_enabled(&self, device: &InputDevice) -> bool {
+        let (device_config, default_config) = self.get_device_config(device);
+        input_config::get_config(device_config.as_ref(), default_config, |x| {
+            x.scroll_config.as_ref()?.kinetic
+        })
+        .is_some_and(|x| x.0)
+    }
+
     pub fn map_to_output(&self, device: &InputDevice) -> Option<String> {
         let (device_config, default_config) = self.get_device_config(device);
         Some(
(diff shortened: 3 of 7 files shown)


damianvander:
Closing as I have found a better implementation that is simpler. Have opened a different PR to keep things clean:
https://github.com/pop-os/cosmic-comp/pull/2389

PR #2389, opened 2026-05-18
feat(input): add compositor-side kinetic scrolling

damianvander:
Opt-in inertial fling for two-finger touchpad scroll. Buffered finger-source axis deltas feed an exponential-decay fling on lift, with a short cubic ease-out tail for a smooth stop. The fling stream uses AxisSource::Continuous, the same source is used for live finger scroll when kinetic is enabled, so clients see one uninterrupted stream. Have tested on my own setup and works flawlessly, the next step would be adding per app overrides for the scroll speed to get consistency across different applications. Happy to open another PR if this one gets accepted

Cancellation paths: button press, key press, 2-finger touchpad land, pointer surface boundary crossing, keyboard focus change, renewed finger scroll.

Config: per-device kinetic: Option<bool> on ScrollConfig, defaults to off.

## Testing

Tested on Pop!_OS 24.04 across a mix of libcosmic, GTK, Qt, Electron,
Firefox, and Java apps.

## AI disclosure

Claude Code was used for project exploration, code-generation
assistance, and drafting commit messages. I have read and understood
every line of the diff and am able to defend or revise any of it in
review.

### DIFF
--- a/cosmic-comp-config/src/input.rs
+++ b/cosmic-comp-config/src/input.rs
@@ -47,6 +47,8 @@ pub struct ScrollConfig {
     pub natural_scroll: Option<bool>,
     pub scroll_button: Option<u32>,
     pub scroll_factor: Option<f64>,
+    #[serde(skip_serializing_if = "Option::is_none", default)]
+    pub kinetic: Option<bool>,
 }
 
 #[derive(Copy, Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
--- a/src/config/input_config.rs
+++ b/src/config/input_config.rs
@@ -65,6 +65,7 @@ pub fn for_device(device: &InputDevice) -> InputConfig {
                     None
                 },
                 scroll_factor: None,
+                kinetic: None,
             })
         } else {
             None
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -622,6 +622,14 @@ impl Config {
         .map_or(1.0, |x| x.0)
     }
 
+    pub fn kinetic_scroll_enabled(&self, device: &InputDevice) -> bool {
+        let (device_config, default_config) = self.get_device_config(device);
+        input_config::get_config(device_config.as_ref(), default_config, |x| {
+            x.scroll_config.as_ref()?.kinetic
+        })
+        .is_some_and(|x| x.0)
+    }
+
     pub fn map_to_output(&self, device: &InputDevice) -> Option<String> {
         let (device_config, default_config) = self.get_device_config(device);
         Some(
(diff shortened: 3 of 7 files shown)


Drakulix:
System-wide kinetic scrolling causes various side-effects. This is something, which should be implemented in toolkits.

damianvander:
Hey Drakulix, thanks for having a look at this. I understand the common consensus is that inertial scrolling is something that should be handled by the UI toolkit in Linux, I can also understand that this is something you team may not want to maintain. However, I think theres a unique opportunity here to solve a pain point that laptop users have had for years and is not addressed anywhere else. 

Every different toolkit/app seems to implement touchpad scrolling very differently (e.g firefox vs cosmic-files), not just inertial feel but also scroll speed, this makes using touchpad scroll on Linux an nightmare. In the x11/synaptics driver days we had a buch of config knobs to fix this, but since the move to wayland and libinput pretty much all of these are gone. Libinput recently released lua scripting support which will enable users to configure hardware level behavior but theres still a bunch of inconsistencies at the toolkit/compositor level. 

The inertial scrolling in this PR + some per app config (https://github.com/damianvander/cosmic-comp/tree/kinetic-overrides) makes a huge improvement to the UX and lets users tune the scroll per-app to get a consistent speed. I think this is a huge feature for laptop users and I would personally choose a desktop environment based on a feature like this. If you have a laptop, I would encourage you to give it a try for yourself. 

With regard to side-effects, this is definitely something I ran into whilst developing. Particularly Firefox and java apps seem to be really sensitive to scroll start/end events and behave weirdly (sticking, layering extra inertia on top). I have found that by changing scroll source to AxisSource::Continuous and blocking scroll start/stop events when the setting is enabled, all of these side effects dissapear compltely.

I know this is not a perfect solution, but I cant see another way to solve this problem. If you have any ideas or suggestions, Id love to hear them. If this is something that ends up being added, I would be happy to help maintain it. 

Kind regards
Damian

lgandecki:
@damianvander thanks for the work. This is one thing that keeps me going back to MacOS everytime I try to go back to Linux. I used to be a developer of PLD Linux Distribution two decades ago, and a huge Linux fanboy. But I do use touchpad, I use laptops. It's hard for me to believe how bad touchpad experience is. Input is, literally, one of the, if not THE most important element of using an operating system. And it's just wrong. I also keep thinking that maybe I should just fix it, but then I see threads like this, where someone tries and gets ignored. And yeah, I could fork, and maintain a fork.. but then I start interpolating from that and get the feel that my OS will no longer be a work machine that I can just use, but something I have to constantly tinker with. It's great to be able to tinker. It's not great to HAVE to, especially with such basic things like input.. 

To get to the question, when you say "this is not a perfect solution" - why not? What cons you discovered?

damianvander:
@lgandecki 
I have published https://github.com/damianvander/cosmic-scroll which patches the cosmic-comp and cosmic-settings binary to enable this. Only tested on cachyos so far but feel free to open a PR with whatever distro you use

PR #2397, opened 2026-05-21
input: support libinput custom acceleration curves

algesten:
AI-generated: This code was co-authored with Claude.

I got a 5K monitor and Magic Trackpad and no matter what I do, I don't get the feeling I want for moving the cursor.

Libinput 1.30 added a `Custom` accel profile that takes piecewise-linear multiplier tables instead of the single scalar that `Flat` and `Adaptive` use. Wire it through cosmic-comp's config so users can supply their own curves (e.g. from cosmic-settings).

Ultimately I'm aiming for this, which I got going on my desktop.

image

Gated behind a new flag `custom-accel` so it's possible to compile with libinput < 1.30. With the feature off the input crate stays at its default `libinput_1_21` floor, the `Custom` variant on libinput's `AccelProfile` isn't compiled in, and configs that specify the `Custom` profile deserialize to `None` (libinput falls back to its default). With the feature on the `input` crate is built with `libinput_1_30`, the apply path becomes available, and a runtime libinput >= 1.30 is required.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -116,6 +116,7 @@ features = ["svg"]
 optional = true
 
 [features]
+custom-accel = ["cosmic-comp-config/custom-accel"]
 debug = ["egui", "egui_plot", "smithay-egui", "anyhow/backtrace"]
 default = ["systemd"]
 systemd = ["libsystemd", "logind-zbus"]
--- a/cosmic-comp-config/Cargo.toml
+++ b/cosmic-comp-config/Cargo.toml
@@ -17,5 +17,6 @@ tracing = { version = "0.1.44", features = [
 
 [features]
 default = []
+custom-accel = ["input/libinput_1_30"]
 output = ["ron", "tracing"]
 randr = ["cosmic-randr-shell", "output"]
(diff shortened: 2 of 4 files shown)


Drakulix:
Ping @pop-os/ux before review on this proposed feature.

algesten:
@Drakulix sorry, I just realised a mistake I made. The `libinput` requirement is 1.23, not 1.30. The current floor is 1.21 by `cosmic-config` not setting any requirement, which falls back on defaults to 1.21.

* 1.21 was released Jun 2022
* 1.23 was released Mar 2023

A simpler version of this PR would ditch the feature flag and bump the floor for libinput. Would be preferable and acceptable? A quick investigation into current PopOS 24.04, Fedora 41, NixOS 25, seem to indicate they are all 1.25 and above.

Vixea:
Honestly I would love to see this as currently no Desktop Environments that I know do this and allows for some crazy things when it comes to gaming.

git-f0x:
FWIW, GNOME seems to be adding support for this in mutter. Won't be exposed in the UI at this point, but might be nice to have the config in COSMIC as well: https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/4296

orowith2os:
Person adding that support to GNOME here: if someone wants to open an issue somewhere discussing the design and accessibility of such a UI (either in cosmic to reference, or in gnome as a tracker issue, or both), feel free, and it would be appreciated. I will not be doing such work myself.

PR #2414, opened 2026-05-26
fix: use configured blur strength for lock surface and its subsurfaces

wash2:

### DIFF
--- a/src/backend/render/mod.rs
+++ b/src/backend/render/mod.rs
@@ -821,7 +821,7 @@ where
                 });
             }
             Stage::SessionLock(lock_surface) => {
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
I guess this is relevant for subsurfaces?

We don't render anything below, so this cannot accidentally reveal anything or show any blur effect for that matter, so there is no reason for lock-surfaces themselves to use it.

wash2:
> I guess this is relevant for subsurfaces?

Yes, it is really just for the subsurfaces on the lock surface, like for the lock screen menu that is on top of the lock surface with the wallpaper.

PR #2432, opened 2026-06-01
wayland: implement pointer-warp-v1

### LINKED ISSUE #2405: [Feature Request] Implement wp_pointer_warp_v1 Wayland protocol for Wine/Proton gaming compatibility
## Problem

Games running via Wine/Proton with the native Wayland driver  (PROTON_ENABLE_WAYLAND=1) have broken mouse input on COSMIC: the cursor moves visually but clicks do not register correctly.

I am completely alient to these kind of things, but with a bit of AI troubleshooting the hypothesis I ended up with is that those games make use of `SetCursorPos` calls, which get ignored / land at wrong coordinates / something like that. And apparently Wine's Wayland driver uses the `wp_pointer_warp_v1` protocol to handle `SetCursorPos` (which is the Windows API call games use to reposition the 
cursor, used heavily in menu navigation, FPS camera control,...) and cosmic-comp does not currently implement this protocol.

This makes those games effectively unplayable on COSMIC with Wine Wayland.

## Impact

Without this protocol, the only workarounds are:
- Disable Wine Wayland entirely (falling back to XWayland) but it causes 
  fullscreen/borderless window regression on COSMIC with NVIDIA (already reported in #3434, 
  #3495)
- Use a different compositor (KDE/KWin already supports wp_pointer_warp_v1 
  since 6.4. *(Indeed, installing KDE alongside COSMIC on my system and using that session to launch the game solved the issue, but it is an annoying workaround)*

## Tested configuration

- Distro: Fedora 44 COSMIC Spin
- Hardware: Tuxedo InfinityBook Pro 16 Gen 8
  - GPU: NVIDIA GeForce RTX 4060 Max-Q / Mobile (proprietary drivers from RPMFusion)
- Games affected: Civilization VI (Heroic Launcher Flatpak, from Epic Games Store, Using `Proton-GE-latest`)
- cosmic-comp version: `cosmic-comp-1.0.13-1.fc44.x86_64`

## Request

If someone more expert than me can confirm this diagnosis, the request would be to implement `wp_pointer_warp_v1` in cosmic-comp.

If this is the wrong place for this issue, or if a similar one has already been opene, I apologize for the mistake and please let me know in the comments.

skygrango:
A few weeks ago, the winewayland driver [merged](https://gitlab.winehq.org/wine/wine/-/merge_requests/10830) support for pointer-warp-v1, and given that we already have pointer-constraint-v1, enabling this protocol is simple.

I also add keyboard focus check to prevent the pointer from being warped just by hovering over the window.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -2393,7 +2393,7 @@ impl State {
             });
 
             if let Some((output, geometry, surface_offset)) = found {
-                let mut pos_in_element = location + surface_offset.to_f64();
+                let pos_in_element = location + surface_offset.to_f64();
                 let window_size = geometry.size.to_f64();
 
                 let is_legal = |p: Point<f64, Logical>| {
@@ -2417,29 +2417,13 @@ impl State {
                 let workspace_origin = output.geometry().loc.to_f64();
                 let origin = geometry.loc.to_f64();
 
-                if !is_legal(pos_in_element) {
-                    let original_global = pointer.current_location();
-
-                    let original_pos_in_element = Point::new(
-                        original_global.x - workspace_origin.x - origin.x,
-                        original_global.y - workspace_origin.y - origin.y,
-                    );
-
-                    let y_only_pos = Point::new(original_pos_in_element.x, pos_in_element.y);
-                    let x_only_pos = Point::new(pos_in_element.x, original_pos_in_element.y);
-
-                    if is_legal(y_only_pos) {
-                        pos_in_element = y_only_pos;
-                    } else if is_legal(x_only_pos) {
-                        pos_in_element = x_only_pos;
-                    } else {
-                        pos_in_element = original_pos_in_element;
-                    }
+                if is_legal(pos_in_element) {
+                    let x = workspace_origin.x + origin.x + pos_in_element.x;
+                    let y = workspace_origin.y + origin.y + pos_in_element.y;
+                    Some((Point::new(x, y), output.clone()))
+                } else {
+                    None
                 }
-
-                let x = workspace_origin.x + origin.x + pos_in_element.x;
-                let y = workspace_origin.y + origin.y + pos_in_element.y;
-                Some((Point::new(x, y), output.clone()))
             } else {
                 None
             }
--- a/src/state.rs
+++ b/src/state.rs
@@ -86,6 +86,7 @@ use smithay::{
         output::OutputManagerState,
         pointer_constraints::PointerConstraintsState,
         pointer_gestures::PointerGesturesState,
+        pointer_warp::PointerWarpManager,
         presentation::PresentationState,
         seat::WaylandFocus,
         security_context::{SecurityContext, SecurityContextState},
@@ -671,6 +672,7 @@ impl State {
         XWaylandKeyboardGrabState::new::<Self>(dh);
         let xwayland_shell_state = XWaylandShellState::new::<Self>(dh);
         PointerConstraintsState::new::<Self>(dh);
+        PointerWarpManager::new::<Self>(dh);
         PointerGesturesState::new::<Self>(dh);
         TabletManagerState::new::<Self>(dh);
         SecurityContextState::new::<Self, _>(dh, client_has_no_security_context);
--- a/src/wayland/handlers/mod.rs
+++ b/src/wayland/handlers/mod.rs
@@ -25,6 +25,7 @@ pub mod output_configuration;
 pub mod output_power;
 pub mod overlap_notify;
 pub mod pointer_constraints;
+pub mod pointer_warp;
 pub mod primary_selection;
 pub mod seat;
 pub mod security_context;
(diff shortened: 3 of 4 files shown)


ids1024:
Other client implementations include https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/2031 for XWayland (which hasn't been merged) and the support in SDL. Though I'm not sure what is best to test this.

I don't know if there are any binary distributions of Wine with a recent enough version, though https://github.com/valvesoftware/proton seems to be good as a way to build Wine (it has scripting to help build within Podman). Though I'm still figuring out how to run something in Proton manually with `WAYLAND_DEBUG` output...

skygrango:
> and the support in SDL. Though I'm not sure what is best to test this.

for SDL3, I modified its demo for testing

```C
diff --git a/examples/demo/02-woodeneye-008/woodeneye-008.c b/examples/demo/02-woodeneye-008/woodeneye-008.c
index 9a53323d9..1373ea1b0 100644
--- a/examples/demo/02-woodeneye-008/woodeneye-008.c
+++ b/examples/demo/02-woodeneye-008/woodeneye-008.c
@@ -357,7 +357,7 @@ SDL_AppResult SDL_AppInit(void **appstate, int argc, char *argv[])
     debug_string[0] = 0;

     SDL_SetRenderVSync(as->renderer, false);
-    SDL_SetWindowRelativeMouseMode(as->window, true);
+    SDL_SetWindowRelativeMouseMode(as->window, false);
     SDL_SetHintWithPriority(SDL_HINT_WINDOWS_RAW_KEYBOARD, "1", SDL_HINT_OVERRIDE);
     return SDL_APP_CONTINUE;
 }
@@ -403,6 +403,14 @@ SDL_AppResult SDL_AppEvent(void *appstate, SDL_Event *event)
             break;
         }
         case SDL_EVENT_MOUSE_BUTTON_DOWN: {
+        switch (event->button.button) {
+        case SDL_BUTTON_LEFT:
+        SDL_WarpMouseInWindow(as->window, 400, 300);
+        break;
+        case SDL_BUTTON_RIGHT:
+        SDL_WarpMouseInWindow(as->window, 1000, 1000);
+        break;
+        }
             SDL_MouseID id = event->button.which;
             int index = whoseMouse(id, players, player_count);
             if (index >= 0) {
```

During this process, I discovered that coordinates outside the surface require change. Currently, simply swapping the x or y axes  to the original position is not sufficient; we should clamp the position to the window bounds. The current approach is only suitable for pointer motion, not for pointer warps.

However, this could very well behave differently from other compositors, because, as far as I know, both [KWin](https://invent.kde.org/plasma/kwin/-/merge_requests/6460) and [Mutter](https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/4525) simply reject the pointer warp request when the target position is out of bounds.

Do you think we should stay consistent with other compositor behaviors, or use clamping when performing pointer warps for potential compatibility issues?

skygrango:
> Though I'm still figuring out how to run something in Proton manually with `WAYLAND_DEBUG` output...

run `script -f steam.log -c "WAYLAND_DEBUG=1 steam" | tail -f steam.log | grep "warp"` in shell, and add `PROTON_ENABLE_WAYLAND=1` to the game's launch options, this is how I record wayland logs.

ids1024:
> Do you think we should stay consistent with other compositor behaviors, or use clamping when performing pointer warps for potential compatibility issues?

According to https://wayland.app/protocols/pointer-warp-v1:

> Request the compositor to move the pointer to a surface-local position. Whether or not the compositor honors the request is implementation defined, but it should
>
> * honor it if the surface has pointer focus, including when it has an implicit pointer grab
> * reject it if the enter serial is incorrect
> * reject it if the requested position is outside of the surface

Rejecting warps that are out of bounds seems to be specified in the protocol (the last point here) so we should do that like other compositors, if I understand correctly.

skygrango:
> Rejecting warps that are out of bounds seems to be specified in the protocol (the last point here) so we should do that like other compositors, if I understand correctly.

This will also affect the behavior of `set_cursor_position_hint`

[zwp_locked_pointer_v1::set_cursor_position_hint](https://wayland.app/protocols/pointer-constraints-unstable-v1#zwp_locked_pointer_v1:request:set_cursor_position_hint) does not mention rejecting request

Do you think we should make a distinction here, or should we use the same implementation ?

skygrango:
I think we can safely follow the implementation of kwin, they don't use clamping.

skygrango:
I initially noticed that SDL clamp the position on its own, so I considered doing the same for other clients. However, this isn't really the compositor's responsibility.

ids1024:
`zwp_locked_pointer_v1::set_cursor_position_hint` is much more vaguely defined. Simply ignoring the "hint" was a valid behavior, and technically clients were wrong to rely on it. (Which I guess is part of why `warp` is now a dedicated protocol).

So clamping or ignoring both seem valid, but it's probably best for compatibility to match what other compositors like Gnome are doing.

Drakulix (approved):
LGTM

jacobgkau (approved):
Support for the new protocol is being advertised, and works to warp the pointer in a test app using the protocol.

PR #2441, opened 2026-06-05
fix xwayland windows with CSD shrinking on any event

### LINKED ISSUE #2440: X11/XWayland windows progressively shrink on each interaction due to bug in 0.1~1780466969~24.04~2f7c34f caused by `smithay` breaking change
**Affects versions:** `0.1~1780466969~24.04~2f7c34f` and later (regression introduced in commit 9f28a764)
**Not present in:** `0.1~1779893019~24.04~22fe419` and earlier

# Symptoms

Any X11/XWayland window that sets `_GTK_FRAME_EXTENTS` (e.g. Firefox launched with `GDK_BACKEND=x11 MOZ_ENABLE_WAYLAND=0`) shrinks slightly each time it is interacted with. The shrink amount is fixed per interaction and corresponds to the application's declared frame extents — for Firefox with a default GTK theme this is approximately the width of a scrollbar (~15–17 px).

The effect is **cumulative**: the window shrinks by one step on every trigger, so over time it becomes noticeably smaller.

Here's a video of the problem happening : https://www.youtube.com/watch?v=d9I7jA0CSl8

## Reproduction steps

1. Launch Firefox (or any GTK app) as an X11/XWayland client:
   `GDK_BACKEND=x11 MOZ_ENABLE_WAYLAND=0` firefox
2. Resize the window to a non-fullscreen size and note its dimensions.
3. Perform **any** of the following actions:
   - Click and drag the title bar to move the window
   - Alt-Tab away to another window
   - Alt-Tab back to Firefox
   - Click and drag a corner/edge to resize the window
4. Observe that the window is now slightly smaller than before step 3.
5. Repeat step 3 — the window shrinks again by the same amount each time.

## Root cause

The `smithay` dependency was updated in commit 9f28a764 ( rev [774f2ab](https://github.com/Smithay/smithay/commit/df73da7183c365b00eb4558690a48061462053f2) → [85f83ab](https://github.com/Smithay/smithay/commit/85f83ab61711e725aa1f3481df4f4dc5370e925b) with [this diff](https://github.com/Smithay/smithay/compare/774f2ab...85f83ab)). That update introduced a **breaking semantic change** to `X11Surface::geometry()` that cosmic-comp did not fully adapt to.

**Before the `smithay` update:**
`X11Surface::geometry()` returned the full window rectangle — position and size exactly as sent to the client via the last `configure_with_sync` call.

**After the `smithay` update** (per the `smithay` changelog):
> `X11Surface::geometry` has been renamed to `X11Surface::last_configure`. `X11Surface::geometry` now returns the bounding box **minus the surface's frame extents**. For clients
that set `_GTK_FRAME_EXTENTS`, the shadow extents are subtracted from the bbox to give the geometry.

Firefox/GTK sets `_GTK_FRAME_EXTENTS` to declare invisible shadow pixels around its window. The new `X11Surface::geometry()` therefore returns a rectangle that is smaller than the
actual window by those extents. `X11Surface::last_configure()` is now the correct way to retrieve the actual configured bounds.

This smaller geometry propagates up the entire call chain:

```
X11Surface::geometry()   ← now returns bbox − _GTK_FRAME_EXTENTS
  └─ CosmicSurface::geometry()
       └─ CosmicWindow::geometry()
            └─ CosmicMapped::geometry()  /  space.element_geometry()
```

Cosmic-comp uses `mapped.geometry().size` in multiple places to compute the size to send in `configure_with_sync`. Because that size is now smaller by `frame_extents`, each
configure shrinks the window by one step, which produces a new (smaller) bbox, which produces an even smaller `geometry()`, and so on — a progressive shrink loop.

Every user-visible trigger forces one iteration of this loop:

| Trigger | Code path |
|---|---|
| Drag to move | `MoveGrab::drop` → `grab_state.window.geometry().size` → `set_geometry` → `configure_with_sync` |
| Alt-Tab | `Shell::update_focus` → `unmap` saves `space.element_geometry().size` (now smaller) as `last_geometry`; next `map_internal` configures with that shrunken size |
| Corner drag to resize | `apply_resize_to_location` → `surface.geometry()` → `surface.configure(geometry)` (`resize.rs:546`) |

# Suggested fix

Two specific call sites need to use `last_configure()` instead of `geometry()` for X11 surfaces:

**`src/shell/element/surface.rs` — `pending_size()`:**
```rust
// Before (broken):
WindowSurface::X11(surface) => Some(surface.geometry().size),
// After:
WindowSurface::X11(surface) => Some(surface.last_configure().size),

src/shell/layout/floating/grabs/resize.rs — apply_resize_to_location (~line 546):
// Before (broken):
let mut geometry = surface.geometry();
// After:
let mut geometry = surface.last_configure();
```

The broader issue is that anywhere `mapped.geometry().size` is used as input to `set_geometry` / `configure_with_sync`, it will now return the content-area size rather than the full
configured size for X11 windows. A `configure_size()` helper on `CosmicSurface` that returns `last_configure().size` for X11 and `geometry().size` for Wayland would make the distinction
explicit and prevent similar regressions at other call sites.

cc @ids1024

hojjatabdollahi:
This fixes the resizing issue in xwayland windows with CSD. We add the `frame_extents` to the size and pass it to xwayland and remove it from the size we receive from xwayland.

Also use `last_configure`, instead of `geometry` in resize. This was missed in https://github.com/pop-os/cosmic-comp/pull/2357/changes/9c3f50410a0e17aa7b9ed4276dc8b114263b37ab.~~

Fixes #2440

~~Draft until I can confirm it fixes the issue, ~~I'm having trouble reproducing the issue~~, and there are other call sites that seem to need to be updated too.~~
___

### DIFF
--- a/src/shell/element/surface.rs
+++ b/src/shell/element/surface.rs
@@ -238,7 +238,8 @@ impl CosmicSurface {
                 toplevel.with_pending_state(|state| state.size = Some(geo.size.as_logical()))
             }
             WindowSurface::X11(surface) => {
-                let _ = surface.configure_with_sync(geo.as_logical(), None);
+                let _ =
+                    surface.configure_with_sync(geo.as_logical() + surface.frame_extents(), None);
             }
         }
     }
--- a/src/shell/layout/floating/grabs/resize.rs
+++ b/src/shell/layout/floating/grabs/resize.rs
@@ -543,7 +543,7 @@ impl ResizeSurfaceGrab {
                             );
                         }
                         WindowSurface::X11(surface) => {
-                            let mut geometry = surface.geometry();
+                            let mut geometry = surface.last_configure();
                             geometry.loc += (location - new_location).as_logical();
                             let _ = surface.configure(geometry);
                         }


ids1024:
Hm, I thought I tested it with the `geometry` function removed to catch any remaining uses...

hojjatabdollahi:
Yeah, this doesn't seem to be the proper fix.

ids1024:
Or actually, this may be right. The difference from the ones I changed here is that `SpaceElement` is in scope in these files, so the `SpaceElement::geometry` function is available. (If both are available, the one implemented on the type rather than the trait takes precedence. Before https://github.com/Smithay/smithay/commit/14ba35559079faea7953a113242a71686802f923, there were confusingly then two `geometry()` methods on `X11Surface` with different meanings).

hojjatabdollahi:
@ids1024 can you check out my latest changes? It adds the frame extent size when sending it to xwayland, and subtract it when receiving it from xwayland. In my testing this works.

src/xwayland.rs
```diff
@@ -946,11 +946,13 @@ impl XwmHandler for State {
...
                 let ssd_height = mapped.ssd_height(false).unwrap_or(0);
+                let extents = window.frame_extents();
                 mapped.set_geometry(Rectangle::new(
                     current_geo.loc,
                     (
-                        w.map(|w| w as i32).unwrap_or(current_geo.size.w),
-                        h.map(|h| h as i32 + ssd_height)
+                        w.map(|w| w as i32 - extents.left - extents.right)
+                            .unwrap_or(current_geo.size.w),
+                        h.map(|h| h as i32 + ssd_height - extents.top - extents.bottom)
```
ids1024: It seems suspicious to me that this modifies `size` but not `loc`, particularly given `Add<FrameExtents<N, Kind>>` when invoked in `set_geometry` seems like it will change both?

ids1024: I'm not sure I understand why we need to subtract it here anyway. `CosmicMapped::set_geometry` passes through to `set_geometry` on `CosmicStack` or `CosmicWindow`.`CosmicWindow::set_geometry` passes through to `CosmicSurface::set_geometry` except adjusting for `ssd_height`.

What exactly is the aim here, and does it really need a change both here in this invocation and in `CosmicSurface::set_geometry`, rather than just one?

hojjatabdollahi: I just tested it and it works without this too. In my head `set_geometry` was cosmic-to-x and `request_configure` is x-to-cosmic and I was just undoing what I did in `set_geometry` (while ignoring `loc` since the client shouldn't set its location). But I was wrong.

ids1024 (approved):

jklgrasso (approved):
Looks good

gene1wood:
This was released in [`epoch-1.0.16`](https://github.com/pop-os/cosmic-comp/releases/tag/epoch-1.0.16)

PR #2442, opened 2026-06-05
Ei protocol support

hojjatabdollahi:
Based on https://github.com/pop-os/cosmic-comp/pull/465

Requires:

- [x] https://github.com/Smithay/smithay/pull/2051
- [x] https://github.com/Smithay/smithay/pull/2073
- [x] For pointer emulation in Steam Input: It also requires xwayland to be compiled with `-Dxwayland_ei=portal` or `auto` with `liboeffis-dev` present https://github.com/pop-os/xwayland/issues/2
- [x] https://github.com/Smithay/smithay/pull/2115

Remote Desktop Portal using this PR: https://github.com/pop-os/xdg-desktop-portal-cosmic/pull/317

Clients I've tested with:

- [x] Steam Input
  - [x] Mouse emulation works
- [x] Lan Mouse (PopOS host, Mac client)
  - [x] Mouse and keyboard emulation works
- [x] Rustdesk (Flatpak) from Android
  - [x] Flatpak version doesn't support Mutlimonitor
  - [x] The client's pointer doesn't match 1 to 1 with the host
  - [x] ASCII text and modifiers work
  - [x] Non-latin text works
- [x] Rustdesk (Flatpak) from Mac
  - [x] Flatpak version doesn't support Mutlimonitor
  - [x] The pointers are not 1 to 1 (scale is not taken into account)
  - [x] The pointer is not mapped to the shared output, when you move the cursor on the client, it may be on another output and it's moving somewhere you're not seeing. It should jump to the shared screen.
  - [x] ASCII text and modifiers work
  - [x] Non-latin isn't relevant, looks like it sends key-code, so changing the layout on the client does nothing.
- [x] Rustdesk (system) from Mac
  - [x] Single output works
  - [Not planned] It allows you to choose your output, or show all outputs at once, the pointer is restricted to one output. (This is a rustdesk bug, `RdpInputMouse` is not display-aware).
  - [Not planned] In native mode Rustdesk doesn't use Remote Desktop portal, it uses the Screen Sharing portal and for input goes directly to `uinput`. :upside_down_face: 

TODO:
- [Not planned] Multi monitor support
- [x] Proper EiRegion support
- [x] Absolute pointer position support
- [x] ei_text.utf8 support

Relevant mattermost thread: https://chat.pop-os.org/pop-os/pl/po79ozukbjrj3c1jps4g4u57ry
___

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: GPL-3.0-only
 
 use crate::{
+    input::InputBackendId,
     shell::Shell,
     state::{BackendData, State},
     utils::prelude::OutputExt,
@@ -819,6 +820,27 @@ fn config_changed(config: cosmic_config::Config, keys: Vec<String>, state: &mut 
                             change_modifier_state(&keyboard, CAPSLOCK_SCANCODE, state);
                         }
                     }
+                }
+                let ei_connections = state
+                    .common
+                    .ei_keyboard_source
+                    .keys()
+                    .cloned()
+                    .collect::<Vec<_>>();
+                for conn in &ei_connections {
+                    state.release_ei_keyboard(conn);
+                    state.clear_input_source_state(&InputBackendId::Ei(conn.clone()));
+                }
+                for ei_seat in state.common.ei_seats.values() {
+                    if let Err(err) =
+                        ei_seat.add_keyboard("virtual keyboard", xkb_config_to_wl(&value))
+                    {
+                        warn!(?err, "Failed to update libei keyboard keymap");
+                    }
+                }
+                if !state.common.ei_seats.is_empty() {
+                    let seat = state.common.shell.read().seats.last_active().clone();
+                    state.broadcast_ei_keyboard_modifiers(&seat);
                 }
                 state.common.config.cosmic_conf.xkb_config = value;
             }
--- a/src/dbus/ei.rs
+++ b/src/dbus/ei.rs
@@ -1,0 +1,79 @@
+use std::{
+    os::unix::net::UnixStream,
+    sync::{Arc, Mutex},
+};
+
+use smithay::reexports::calloop;
+use zbus::names::{UniqueName, WellKnownName};
+
+use super::name_owners::NameOwners;
+
+static ALLOWED_NAMES: &[WellKnownName] = &[WellKnownName::from_static_str_unchecked(
+    "org.freedesktop.impl.portal.desktop.cosmic",
+)];
+
+/// Channel for handing the EI socketpair (and requested device types)
+/// It's `None` until the EI sender side has been set up
+type EiSender = Arc<Mutex<Option<calloop::channel::Sender<crate::libei::EiRequest>>>>;
+
+struct Ei {
+    ei_sender: EiSender,
+    name_owners: NameOwners,
+}
+
+impl Ei {
+    async fn check_sender_allowed(&self, sender: &UniqueName<'_>) -> zbus::fdo::Result<()> {
+        if self.name_owners.check_owner(sender, ALLOWED_NAMES).await {
+            Ok(())
+        } else {
+            Err(zbus::fdo::Error::AccessDenied("Access denied".to_string()))
+        }
+    }
+}
+
+#[zbus::interface(name = "com.system76.CosmicComp.Ei")]
+impl Ei {
+    /// Create a new EI sender context
+    async fn get_sender_socket(
+        &self,
+        device_types: u32,
+        #[zbus(header)] header: zbus::message::Header<'_>,
+    ) -> zbus::fdo::Result<zbus::zvariant::OwnedFd> {
+        if let Some(sender) = header.sender() {
+            self.check_sender_allowed(sender).await?;
+        }
+
+        let (comp_stream, client_stream) = UnixStream::pair().map_err(|err| {
+            zbus::fdo::Error::Failed(format!("Failed to create socket pair: {err}"))
+        })?;
+
+        {
+            let guard = self.ei_sender.lock().unwrap();
+            let sender = guard
+                .as_ref()
+                .ok_or_else(|| zbus::fdo::Error::Failed("EI sender not available".to_string()))?;
+            sender.send((comp_stream, device_types)).map_err(|err| {
+                zbus::fdo::Error::Failed(format!("Failed to hand off EI socket: {err}"))
+            })?;
+        }
+
+        Ok(std::os::fd::OwnedFd::from(client_stream).into())
+    }
+}
+
+/// Register the `com.system76.CosmicComp.Ei` interface on the shared session connection.
+pub async fn init(
+    conn: &zbus::Connection,
+    name_owners: &NameOwners,
+    ei_sender: EiSender,
+) -> zbus::Result<()> {
+    let ei = Ei {
+        ei_sender,
+        name_owners: name_owners.clone(),
+    };
+    conn.object_server()
+        .at("/com/system76/CosmicComp/Ei", ei)
+        .await?;
+    conn.request_name("com.system76.CosmicComp").await?;
+    Ok(())
+}
--- a/src/dbus/ei.rs
+++ b/src/dbus/ei.rs
@@ -8,9 +8,10 @@ use zbus::names::{UniqueName, WellKnownName};
 
 use super::name_owners::NameOwners;
 
-static ALLOWED_NAMES: &[WellKnownName] = &[WellKnownName::from_static_str_unchecked(
-    "org.freedesktop.impl.portal.desktop.cosmic",
-)];
+static ALLOWED_NAMES: &[WellKnownName] = &[
+    WellKnownName::from_static_str_unchecked("org.freedesktop.impl.portal.desktop.cosmic"),
+    WellKnownName::from_static_str_unchecked("com.system76.CosmicOSK"),
+];
 
 /// Channel for handing the EI socketpair (and requested device types)
 /// It's `None` until the EI sender side has been set up
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -11,7 +11,7 @@ use crate::{
     },
     input::gestures::{GestureState, SwipeAction},
     shell::{
-        LastModifierChange, SeatExt, Trigger,
+        SeatExt, Trigger,
         focus::{
             Stage, render_input_order,
             target::{KeyboardFocusTarget, PointerFocusTarget},
@@ -42,14 +42,14 @@ use smithay::{
     backend::input::{
         AbsolutePositionEvent, Axis, AxisRelativeDirection, AxisSource, Device, DeviceCapability,
         GestureBeginEvent, GestureEndEvent, GesturePinchUpdateEvent as _,
-        GestureSwipeUpdateEvent as _, InputBackend, InputEvent, KeyState, KeyboardKeyEvent,
-        PointerAxisEvent, ProximityState, TabletToolButtonEvent, TabletToolEvent,
-        TabletToolProximityEvent, TabletToolTipEvent, TabletToolTipState, TouchEvent,
+        GestureSwipeUpdateEvent as _, InputBackend, InputEvent, KeyState, PointerAxisEvent,
+        ProximityState, TabletToolButtonEvent, TabletToolEvent, TabletToolProximityEvent,
+        TabletToolTipEvent, TabletToolTipState, TouchEvent,
     },
     desktop::{PopupKeyboardGrab, WindowSurfaceType, utils::under_from_surface_tree},
     input::{
         Seat,
-        keyboard::{FilterResult, KeysymHandle, ModifiersState},
+        keyboard::{FilterResult, KeyboardSource, KeysymHandle, ModifiersState},
         pointer::{
             AxisFrame, ButtonEvent, GestureHoldBeginEvent, GestureHoldEndEvent,
             GesturePinchBeginEvent, GesturePinchEndEvent, GesturePinchUpdateEvent,
@@ -64,7 +64,7 @@ use smithay::{
         input::Device as InputDevice,
         wayland_server::{Resource as _, protocol::wl_surface::WlSurface},
     },
-    utils::{Clock, Logical, Monotonic, Point, Rectangle, SERIAL_COUNTER, Serial},
+    utils::{Clock, Logical, Monotonic, Point, Rectangle, SERIAL_COUNTER, Serial, Size},
     wayland::{
         compositor::CompositorHandler,
         image_copy_capture::CursorSessionRef,
@@ -80,13 +80,26 @@ use std::{
     any::Any,
     borrow::Cow,
     cell::RefCell,
-    collections::HashSet,
+    collections::{HashMap, HashSet},
     ops::ControlFlow,
     time::{Duration, Instant},
 };
 
 pub mod actions;
 pub mod gestures;
+
+/// Identifies the input backend instance an event came from, used to disambiguate device ids
+/// (which are only unique within a single backend instance, see
+/// [`smithay::backend::input::Device::id`]).
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub enum InputBackendId {
+    /// The session input backend (libinput / winit / x11) these are mutually exclusive
+    Normal,
+    /// A specific Ei client connection
+    Ei(smithay::reexports::reis::eis::Connection),
+    /// The `zwp_virtual_keyboard_v1` protocol (all virtual keyboards share this source)
+    VirtualKeyboard,
+}
 
 /// Used for debouncing focus updates due to pointer motion, if after the focus change is
 /// triggered the event will cancel if the pointer moves to the original target
@@ -100,19 +113,35 @@ pub struct PointerFocusState {
 }
 
 #[derive(Default)]
-pub struct SupressedKeys(RefCell<Vec<(Keycode, Option<RegistrationToken>)>>);
+pub struct SupressedKeys(
+    RefCell<HashMap<InputBackendId, Vec<(Keycode, Option<RegistrationToken>)>>>,
+);
 #[derive(Default)]
-pub struct SupressedButtons(RefCell<HashSet<u32>>);
+pub struct SupressedButtons(RefCell<HashMap<InputBackendId, HashSet<u32>>>);
 #[derive(Default, Debug)]
-pub struct ModifiersShortcutQueue(RefCell<Option<shortcuts::Binding>>);
+pub struct ModifiersShortcutQueue(RefCell<HashMap<InputBackendId, shortcuts::Binding>>);
 
 impl SupressedKeys {
-    fn add(&self, keysym: &KeysymHandle, token: impl Into<Option<RegistrationToken>>) {
-        self.0.borrow_mut().push((keysym.raw_code(), token.into()));
+    fn add(
+        &self,
+        backend_id: &InputBackendId,
+        keysym: &KeysymHandle,
+        token: impl Into<Option<RegistrationToken>>,
+    ) {
+        self.0
+            .borrow_mut()
+            .entry(backend_id.clone())
+            .or_default()
+            .push((keysym.raw_code(), token.into()));
     }
 
-    fn filter(&self, keysym: &KeysymHandle) -> Option<Vec<RegistrationToken>> {
-        let mut keys = self.0.borrow_mut();
+    fn filter(
+        &self,
+        backend_id: &InputBackendId,
+        keysym: &KeysymHandle,
+    ) -> Option<Vec<RegistrationToken>> {
+        let mut by_source = self.0.borrow_mut();
+        let keys = by_source.get_mut(backend_id)?;
         let (removed, remaining) = keys
             .drain(..)
             .partition(|(key, _)| *key == keysym.raw_code());
@@ -129,44 +158,60 @@ impl SupressedKeys {
                 .collect::<Vec<_>>(),
         )
     }
+
+    fn clear_source(&self, backend_id: &InputBackendId) {
+        self.0.borrow_mut().remove(backend_id);
+    }
 }
 
 impl SupressedButtons {
-    fn add(&self, button: u32) {
-        self.0.borrow_mut().insert(button);
+    fn add(&self, backend_id: &InputBackendId, button: u32) {
+        self.0
+            .borrow_mut()
+            .entry(backend_id.clone())
+            .or_default()
+            .insert(button);
     }
 
-    fn remove(&self, button: u32) -> bool {
-        self.0.borrow_mut().remove(&button)
+    fn remove(&self, backend_id: &InputBackendId, button: u32) -> bool {
+        self.0
+            .borrow_mut()
+            .get_mut(backend_id)
+            .is_some_and(|buttons| buttons.remove(&button))
+    }
+
+    fn clear_source(&self, backend_id: &InputBackendId) {
+        self.0.borrow_mut().remove(backend_id);
     }
 }
 
 impl ModifiersShortcutQueue {
-    pub fn set(&self, binding: shortcuts::Binding) {
-        let mut set = self.0.borrow_mut();
-        *set = Some(binding);
+    pub fn set(&self, backend_id: &InputBackendId, binding: shortcuts::Binding) {
+        self.0.borrow_mut().insert(backend_id.clone(), binding);
     }
 
-    pub fn take(&self, binding: &shortcuts::Binding) -> bool {
+    pub fn take(&self, backend_id: &InputBackendId, binding: &shortcuts::Binding) -> bool {
         let mut set = self.0.borrow_mut();
-        if set.is_some() && set.as_ref().unwrap() == binding {
-            *set = None;
+        if set.get(backend_id).is_some_and(|queued| queued == binding) {
+            set.remove(backend_id);
             true
         } else {
             false
         }
     }
 
-    pub fn clear(&self) {
-        let mut set = self.0.borrow_mut();
-        *set = None;
+    pub fn clear(&self, backend_id: &InputBackendId) {
+        self.0.borrow_mut().remove(backend_id);
     }
 }
 
 impl State {
     #[profiling::function]
-    pub fn process_input_event<B: InputBackend>(&mut self, event: InputEvent<B>)
-    where
+    pub fn process_input_event<B: InputBackend>(
+        &mut self,
+        event: InputEvent<B>,
+        backend_id: InputBackendId,
+    ) where
         <B as InputBackend>::Device: 'static,
     {
         crate::wayland::handlers::output_power::set_all_surfaces_dpms_on(self);
@@ -177,19 +222,25 @@ impl State {
                 let shell = self.common.shell.read();
                 let seat = shell.seats.last_active();
                 let led_state = seat.get_keyboard().unwrap().led_state();
-                seat.devices().add_device(&device, led_state);
+                seat.devices().add_device(&device, led_state, &backend_id);
                 if device.has_capability(DeviceCapability::TabletTool) {
                     seat.tablet_seat().add_wp_tablet(
                         &self.common.display_handle,
                         &TabletDescriptor::from(&device),
                     );
                 }
+                let has_keyboard = device.has_capability(DeviceCapability::Keyboard);
+                std::mem::drop(shell);
+                // send the seat's current modifier state to the new ei keyboard
+                if has_keyboard && let InputBackendId::Ei(conn) = &backend_id {
+                    self.send_ei_keyboard_modifiers(conn);
+                }
             }
             InputEvent::DeviceRemoved { device } => {
                 for seat in &mut self.common.shell.read().seats.iter() {
                     let devices = seat.devices();
-                    if devices.has_device(&device) {
-                        devices.remove_device(&device);
+                    if devices.has_device(&device, &backend_id) {
+                        devices.remove_device(&device, &backend_id);
                         if device.has_capability(DeviceCapability::TabletTool) {
                             seat.tablet_seat()
                                 .remove_tablet(&TabletDescriptor::from(&device));
@@ -210,7 +261,7 @@ impl State {
                     .shell
                     .read()
                     .seats
-                    .for_device(&event.device())
+                    .for_device(&event.device(), &backend_id)
                     .cloned();
                 if let Some(seat) = maybe_seat {
                     self.common.idle_notifier_state.notify_activity(&seat);
@@ -231,51 +282,17 @@ impl State {
                             serial,
                             time,
                             |data, modifiers, handle| {
-                                if previous_modifiers != *modifiers {
-                                    *seat
-                                        .user_data()
-                                        .get::<LastModifierChange>()
-                                        .unwrap()
-                                        .0
-                                        .lock()
-                                        .unwrap() = Some(serial);
-                                }
-
-                                let current_focus = seat.get_keyboard().unwrap().current_focus();
-                                let shortcuts_inhibited = current_focus.as_ref().is_some_and(|f| {
-                                    f.wl_surface()
-                                        .map(|surface| {
-                                            seat.keyboard_shortcuts_inhibitor_for_surface(&surface)
-                                                .map(|inhibitor| inhibitor.is_active())
-                                                .unwrap_or(false)
-                                                || seat.has_active_xwayland_grab(&surface)
-                                        })
-                                        .unwrap_or(false)
-                                });
-                                let sym = handle.modified_sym();
-
-                                let result = Self::filter_keyboard_input(
-                                    data, &event, &seat, modifiers, handle, serial,
-                                );
-
-                                if (matches!(result, FilterResult::Forward)
-                                    && !seat.get_keyboard().unwrap().is_grabbed()
-                                    && !shortcuts_inhibited
-                                    && !matches!(
-                                        current_focus,
-                                        Some(KeyboardFocusTarget::LockSurface(_))
-                                    ))
-                                // we don't want to accidentally leave any keys pressed
-                                // and do more filtering in `xwayland_notify_key_event`
-                                // for released keys
-                                    || state == KeyState::Released
-                                {
-                                    data.common.xwayland_notify_key_event(
-                                        sym, keycode, state, serial, time,
-                                    );
-                                }
-
-                                result
+                                data.process_keyboard_filter(
+                                    &backend_id,
+                                    &seat,
+                                    modifiers,
+                                    handle,
+                                    serial,
+                                    time,
+                                    keycode,
+                                    state,
+                                    previous_modifiers,
+                                )
                             },
                         )
                         .flatten()
@@ -286,7 +303,7 @@ impl State {
                                 FilterResult::<()>::Forward
                             });
                         }
-                        self.handle_action(action, &seat, serial, time, pattern, None)
+                        self.handle_action(action, &backend_id, &seat, serial, time, pattern, None)
                     }
 
                     // If we want to track numlock state so it can be reused on the next boot...
@@ -311,7 +328,11 @@ impl State {
                 use smithay::backend::input::PointerMotionEvent;
 
                 let shell = self.common.shell.write();
-                if let Some(seat) = shell.seats.for_device(&event.device()).cloned() {
+                if let Some(seat) = shell
+                    .seats
+                    .for_device(&event.device(), &backend_id)
+                    .cloned()
+                {
                     self.common.idle_notifier_state.notify_activity(&seat);
                     notify_cursor_activity(self, &seat);
                     let current_output = seat.active_output();
@@ -665,19 +686,47 @@ impl State {
                     .shell
                     .read()
                     .seats
-                    .for_device(&event.device())
+                    .for_device(&event.device(), &backend_id)
                     .cloned();
                 if let Some(seat) = maybe_seat {
                     self.common.idle_notifier_state.notify_activity(&seat);
                     notify_cursor_activity(self, &seat);
-                    let output = seat.active_output();
-                    let output_geometry = output.geometry();
-                    let position = output_geometry.loc.to_f64()
-                        + smithay::backend::input::AbsolutePositionEvent::position_transformed(
-                            &event,
-                            output_geometry.size.as_logical(),
-                        )
-                        .as_global();
+                    let (output, position) = if matches!(&backend_id, InputBackendId::Ei(_)) {
+                        // EI absolute coordinates are in the compositor's *global*
+                        // logical space: each advertised region carries its output's
+                        // global offset, so the client sends a global position. Use
+                        // the coordinate directly and find the output it lands in,
+                        // rather than mapping relative to the focused output (which
+                        // cannot address other monitors). This is the KWin/mutter
+                        // model.
+                        let position =
+                            smithay::backend::input::AbsolutePositionEvent::position_transformed(
+                                &event,
+                                // smithay's EI impl ignores the size and returns the
+                                // raw coordinate.
+                                Size::from((0, 0)),
+                            )
+                            .as_global();
+                        let output = self
+                            .common
+                            .shell
+                            .read()
+                            .outputs()
+                            .find(|o| o.geometry().to_f64().contains(position))
+                            .cloned()
+                            .unwrap_or_else(|| seat.active_output());
+                        (output, position)
+                    } else {
+                        let output = seat.active_output();
+                        let output_geometry = output.geometry();
+                        let position = output_geometry.loc.to_f64()
+                            + smithay::backend::input::AbsolutePositionEvent::position_transformed(
+                                &event,
+                                output_geometry.size.as_logical(),
+                            )
+                            .as_global();
+                        (output, position)
+                    };
                     let serial = SERIAL_COUNTER.next_serial();
                     let under = State::surface_under(position, &output, &self.common.shell.write())
                         .map(|(target, pos)| (target, pos.as_logical()));
@@ -693,6 +742,14 @@ impl State {
                         },
                     );
                     ptr.frame(self);
+
+                    // Keep the seat's active output following the pointer. Click-to-
+                    // focus (PointerButton) resolves its target via
+                    // `seat.active_output()`
+                    let previous_output = seat.active_output();
+                    if previous_output != output {
+                        seat.set_active_output(&output);
+                    }
 
                     let shell = self.common.shell.read();
                     update_output_image_copy_cursor_position(
@@ -713,7 +770,7 @@ impl State {
                     .shell
                     .read()
                     .seats
-                    .for_device(&event.device())
+                    .for_device(&event.device(), &backend_id)
                     .cloned()
                 else {
                     return;
@@ -736,7 +793,26 @@ impl State {
                 let serial = SERIAL_COUNTER.next_serial();
                 let button = event.button_code();
 
-                let mut pass_event = !seat.supressed_buttons().remove(button);
+                // Track buttons held by a libei source so they can be released if the connection
+                // drops mid-press
+                if let InputBackendId::Ei(conn) = &backend_id {
+                    match event.state() {
+                        ButtonState::Pressed => {
+                            self.common
+                                .ei_pointer_buttons
+                                .entry(conn.clone())
+                                .or_default()
+                                .insert(button);
+                        }
+                        ButtonState::Released => {
+                            if let Some(held) = self.common.ei_pointer_buttons.get_mut(conn) {
+                                held.remove(&button);
+                            }
+                        }
+                    }
+                }
+
+                let mut pass_event = !seat.supressed_buttons().remove(&backend_id, button);
                 if event.state() == ButtonState::Pressed {
                     // change the keyboard focus unless the pointer is grabbed
                     // We test for any matching surface type here but always use the root
@@ -758,7 +834,7 @@ impl State {
                         );
                         if let Some(target) = under.filter(|_| !on_resize_fork) {
                             if let Some(surface) = target.toplevel().map(Cow::into_owned)
-                                && seat.get_keyboard().unwrap().modifier_state().logo
+                                && self.source_modifiers(&backend_id, &seat).logo
                                 && !shortcuts_inhibited
                             {
                                 let seat_clone = seat.clone();
@@ -769,17 +845,18 @@ impl State {
                                     // aimed at the compositor and shouldn't be passed
                                     // to the application.
                                     pass_event = false;
-                                    seat.supressed_buttons().add(button);
+                                    seat.supressed_buttons().add(&backend_id, button);
                                 };
 
                                 fn dispatch_grab<G: PointerGrab<State> + 'static>(
                                     grab: Option<(G, smithay::input::pointer::Focus)>,
                                     seat: Seat<State>,
+                                    backend_id: &InputBackendId,
                                     serial: Serial,
                                     state: &mut State,
                                 ) {
                                     if let Some((target, focus)) = grab {
-                                        seat.modifiers_shortcut_queue().clear();
+                                        seat.modifiers_shortcut_queue().clear(backend_id);
 
                                         seat.get_pointer()
                                             .unwrap()
@@ -791,6 +868,7 @@ impl State {
                                     match mouse_button {
                                         smithay::backend::input::MouseButton::Left => {
                                             supress_button();
+                                            let backend_id = backend_id.clone();
                                             self.common.event_loop_handle.insert_idle(
                                                 move |state| {
                                                     let mut shell = state.common.shell.write();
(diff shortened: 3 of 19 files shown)


samrocketman:
Just chiming to say as a PopOS system76 user (on laptop) who enjoys synergy from my Ubuntu desktop computer, I am happy to see some movement toward this. I started reviewing all of the background material regarding InputCapture and RemoteDesktop Wayland portals on cosmic and happened upon this today. I am still reviewing but wanted to at least chime in my positive support for it.

src/xwayland.rs
```diff
@@ -115,6 +115,7 @@ impl State {
             &self.common.display_handle,
             None,
             std::iter::empty::<(OsString, OsString)>(),
+            ["-enable-ei-portal"],
```
Drakulix, line 118: Can we re-try without the argument, in case we find an Xwayland version not supporting this? Or would Xwayland silently ignore the argument anyway?

Drakulix (commented):
Please don't upgrade libcosmic as part of this PR. This creates a lot of unnecessary churn with https://github.com/pop-os/cosmic-comp/pull/2179

src/xwayland.rs
hojjatabdollahi: XWayland has support for this arg since 23.2. And I think PopOS 24.04 ships with 24.1. And it knows the arg even if it's not compiled with ei_portal support. So it won't error out.

But I can add a retry logic on error if you want.

Drakulix: Even debian stable seems to ship 24.*. So yeah we can go ahead with depending on this argument.

hojjatabdollahi: Relevant discussion here: https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/3303

src/dbus/ei.rs
```diff
@@ -0,0 +1,57 @@
...
+        }
+    }
+}
+
+#[zbus::interface(name = "com.system76.CosmicComp.Ei")]
+impl Ei {
+    async fn get_socket_path(
+        &self,
+        #[zbus(header)] header: zbus::message::Header<'_>,
+    ) -> zbus::fdo::Result<String> {
```
ids1024: If we're exposing this in a DBus protocol, it doesn't really make sense to return a path to a listener socket. We should create a `UnixStream::pair`. Create and insert an ` EiInput` event source with one, and return an `OwnedFd` from the other.

It can also be called something like `get_sender_socket` since we'll need another method later with receiver contexts (and some additional information) if we want to support https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.impl.portal.InputCapture.html.

src/libei.rs
```diff
@@ -0,0 +1,48 @@
...
+    let handle_clone = handle.clone();
+    handle
+        .insert_source(listener_source, move |context, _, _| {
+            let source = EiInput::new(context);
+            handle_clone
+                .insert_source(source, |event, connection, data| match event {
+                    EiInputEvent::Connected => {
+                        let seat = connection.add_seat("default");
+                        let conf = data.common.config.xkb_config();
+                        let _ = seat.add_keyboard("virtual keyboard", xkb_config_to_wl(&conf));
```
ids1024: We need to call `add_keyboard()` again for each EI source when the cosmic-comp keyboard configuration is changed during runtime.

ids1024:
Since https://github.com/Smithay/smithay/pull/2051 defines device IDs to be specific to a backend instance, we'll need to change `Devices` in `src/shell/seats.rs` to index by both the device ID and something that disambiguates different backends. I was wondering if we might want input backends in `smithay` to have their own IDs of some kind too, but I don't think I got around to experimenting with that idea.

We want to make sure we reliably remove all devices associated with an EI backend when the EI connection ends. The smithay ei backend is supposed to make sure that happens (with `DeviceDropWrapper`, etc.), though it may be something that needs testing.

hojjatabdollahi:
I was having a regression with my game not going fullscreen in this branch until I reverted the update to libcosmic. Which tells me probably something in iced/libcosmic recently broke fullscreen games. I will look closer later to find the culprit but thought to write it here in case it's broken on the frosted glass branch.

EDIT: Actually I had updated all dependencies. So it could be anything. I'll look into it later.

hojjatabdollahi:
> We want to make sure we reliably remove all devices associated with an EI backend when the EI connection ends. The smithay ei backend is supposed to make sure that happens (with `DeviceDropWrapper`, etc.), though it may be something that needs testing.

What do you suggest for this? 
- Add a counter or something to cosmic-comp for each backend?
- Or update Device type in smithay to store the id? (breaking change)

ids1024:
I thought `smithay` could have an `id()` method of some kind on the `InputBackend` trait (though that might not really be needed). Then `Devices` could index by a tuple of the backend id, and the backend-specific device ID.

Alternately, perhaps cosmic-comp could define such an ID itself with something like

```rust
#[derive(Hash)]
enum InputBackendId {
    /// X11, Wayland, or libinput input backend
    Normal,
    Ei(Backend),
}
```

`reis` would need to provide `Hash` for `Backend`, but that would be easy enough to add. Or maybe it should use `BackendWeak`, though then `reis` would need to make that public.

Using something that holds a reference to the reis backend would guarantee a unique match without some kind of incrementing counter, though that can work too.

ids1024:
Incorporating additional information in `Device::id()` is hard because `Device` is implemented for `reis::request::Device`, and `Event` (which has a `device()` method) for various other `reis` types.

`reis` doesn't allow adding extra user data to objects, so to add addition information we'd either need to add that to reis, or wrap both `reis::request::Device` and all those events in some struct with the extra information, or do something wild with global state, or re-design the smithay input API so these functions have a reference to the backend.

... But we shouldn't need that since our `EiInput` handler already can know what backend instance the events came from, and maybe pass that as an argument to `process_input_event()`.

ids1024:
Or I guess `reis::eis::Connection` already has `Hash`/`Eq`, and could be used.

hojjatabdollahi:
@ids1024 the connection is a good idea https://github.com/Smithay/smithay/pull/2072

hojjatabdollahi:
I added the socketpair and updated https://github.com/pop-os/xdg-desktop-portal-cosmic/pull/317 to use it. 

Added the Enum for Normal/EI backend using the connection to distinguish them and also `add_keyboard` is being called on keymap/layout change.

src/dbus/ei.rs
```diff
@@ -0,0 +1,78 @@
...
+    }
+}
+
+#[zbus::interface(name = "com.system76.CosmicComp.Ei")]
+impl Ei {
+    /// Create a new EI sender context
+    async fn get_sender_socket(
+        &self,
+        #[zbus(header)] header: zbus::message::Header<'_>,
+    ) -> zbus::fdo::Result<zbus::zvariant::OwnedFd> {
```
ids1024, line 41: As mentioned in https://github.com/pop-os/xdg-desktop-portal-cosmic/pull/317#discussion_r3422936224, do we need to add a bitflag argument here for which device types to expose on the EI socket? Then decide which `add_*` methods to call based on that.

src/dbus/ei.rs
```diff
@@ -0,0 +1,78 @@
...
+        })?;
+
+        {
+            let guard = self.ei_sender.lock().unwrap();
+            let sender = guard
+                .as_ref()
+                .ok_or_else(|| zbus::fdo::Error::Failed("EI sender not available".to_string()))?;
+            sender.send(comp_stream).map_err(|err| {
+                zbus::fdo::Error::Failed(format!("Failed to hand off EI socket: {err}"))
+            })?;
```
ids1024: If we store a `LoopHandle` inside of `Ei`, couldn't we just insert the event source here, instead of sending a message to a channel source that does so?

src/dbus/ei.rs
hojjatabdollahi: Ah... I didn't see your last comment.

src/dbus/ei.rs
hojjatabdollahi: Can we store LoopHandle in Ei? Isn't it `RC`-based? Also I assumed zbus's executor and calloop's event-loop are on different threads.

ids1024: Ah yeah, I was assuming the handle was `Send + Sync`. A channel is reasonable then.

hojjatabdollahi:
I took a stab at `ei_text`. I'm handling `TextKeysym` the old-fashioned way, mapping to keycode and dropping any keysym that doesn't exist in the current layout. For `TextUtf8`, I'm injecting the text into the current textbox.

Since there is no prior art for `ei_text`, I assumed that we handle `Keysym` the same way Gnome/KDE handle it for their dbus remote desktop interface. And I imagine my use of `TextUtf8` is how it was designed to be used?!

src/libei.rs
```diff
@@ -0,0 +1,191 @@
...
+                keyboard.input_forward(self, keycode, KeyState::Pressed, serial, time, needs_mods);
+            }
+            KeyState::Released => {
+                keyboard.input_forward(self, keycode, KeyState::Released, serial, time, false);
+                if needs_mods {
+                    keyboard.set_modifier_state(ModifiersState::default());
+                    keyboard.advertise_modifier_state(self);
+                }
+            }
+        }
```
Drakulix, line 191: So how does this interact with any real attached keyboard.

E.g. if the user holds Shift, while some emulated input happens, will the modifier state have unset `Shift` afterwards?

Also this assumes that if `needs_mods` is empty, that currently there isn't any pressed modifier. So e.g. an injected "a" press, while "Shift" is held on a physical keyboard would inject "A", no?

I am pretty sure need to backup the current modifier state and set/reset it in any case.

src/libei.rs
```diff
@@ -0,0 +1,191 @@
...
+        let serial = SERIAL_COUNTER.next_serial();
+        let time = self.common.clock.now().as_millis();
+
+        match key_state {
+            KeyState::Pressed => {
+                if needs_mods {
+                    keyboard.set_modifier_state(mods);
+                    keyboard.advertise_modifier_state(self);
+                }
+                keyboard.input_forward(self, keycode, KeyState::Pressed, serial, time, needs_mods);
```
Drakulix: So emulated inputs cannot cause any keybinds? That is a bit unfortunate and definitely not wanted for use-cases like `input-leap`.

src/libei.rs
hojjatabdollahi: You're right. We do need to backup the current modifier state.

src/libei.rs
hojjatabdollahi: hmmm. That would depend on which path `input-leap` calls. We have both `DeviceCapability::Keyboard` and `DeviceCapability::Text`.

But either way we probably want to process keybind for keysym too. I'll probably need to change `keyboard.input()` to make it work. I'll look into it.

hojjatabdollahi: Or to avoid changing `keyboard.input()` I can create a thin synthetic event struct that implements `KeyboardKeyEvent`.

Drakulix: I think we want to keep using input_forward specifically for the fallback code path for text-input.

That is specifically meant to be pasted verbatim into an application and not to be changed by some compositor handling.

hojjatabdollahi: I'm testing with rustdesk. It sends keysym. I think input-leap would send keysym too. I'm having trouble with shortcut keys. For example I get ctrl+c as ctrl and a c. And the way I'm handling modifiers, it presses control (since it doesn't need a modifier) and then it presses c but it doesn't do it with a modifier. So sending ctrl+c from Rustdesk on my phone, would just type c in the terminal.

I think the issue is I'm directing `notify_keyboard_keysym` to `ei_text::Keysym` in `xdg-desktop-portal-cosmic`. I should probably turn that to keycode and send it to `ei_keyboard` and leave `ei_text::Keysym` to more modern clients who connect directly to Ei. In that case then `input_forward` works fine.

Drakulix: > So sending ctrl+c from Rustdesk on my phone, would just type c in the terminal.

Yeah I think the ei-keyboard needs to keep it's own XkbState. This is essentially the same problem implementations of the virtual-keyboard in various compositors have. Their implementations can be taken for inspiration.

> I think the issue is I'm directing notify_keyboard_keysym to ei_text::Keysym in xdg-desktop-portal-cosmic. I should probably turn that to keycode and send it to ei_keyboard and leave ei_text::Keysym to more modern clients who connect directly to Ei. In that case then input_forward works fine.

No, please don't do the translation outside of the compositor. This makes solving issues affecting combinations of different input systems harder.

> I'm testing with rustdesk. It sends keysym. I think input-leap would send keysym too.

There might not be an application sending Text-events now (though you could try pasting in rustdesk), but we still need to make sure we don't handle any synthetic key events we create from text.

hojjatabdollahi: Other compositors have a dedicated method for `notify_keyboard_keysym`. For example mutter has `clutter_virtual_input_device_notify_keyval` and kwin has `org_kde_kwin_fake_input.keyboard_keysym`. Which does the keysym to keycode mapping. They treat keysym as a layout independent character and not "text". So keycode and keysym both end up at the same place.

Rustdesk sends "ctrl+c" as:
```
[2026-06-17T17:34:53Z INFO  xdg_desktop_portal_cosmic::remote_desktop] NotifyKeyboardKeycode: keycode=29 (0x1d), state=1 (pressed)
[2026-06-17T17:34:53Z INFO  xdg_desktop_portal_cosmic::remote_desktop] NotifyKeyboardKeysym: keysym=99 (0x63), state=1 (pressed)
[2026-06-17T17:34:53Z INFO  xdg_desktop_portal_cosmic::remote_desktop] NotifyKeyboardKeycode: keycode=29 (0x1d), state=0 (released)
[2026-06-17T17:34:53Z INFO  xdg_desktop_portal_cosmic::remote_desktop] NotifyKeyboardKeysym: keysym=99 (0x63), state=0 (released)
```

Gnome and KDE can handle that because they're handling keysym the same way they're doing keycode. But if we're treating keysym as "logical text" and forwarding it to the focused client. But we're handling keycode as a keyboard event. Then super+t would just type t in the current window and hold down super in the compositor. 

EDIT: here's what happens for super+t:
I receive this:
```
Jun 17 11:52:17 Cosmic xdg-desktop-portal-cosmic[104300]: [2026-06-17T17:52:17Z WARN  xdg_desktop_portal_cosmic::remote_desktop] NotifyKeyboardKeycode: keycode=125 (0x7d), state=1 (pressed)
Jun 17 11:52:17 Cosmic xdg-desktop-portal-cosmic[104300]: [2026-06-17T17:52:17Z WARN  xdg_desktop_portal_cosmic::remote_desktop] NotifyKeyboardKeysym: keysym=116 (0x74), state=1 (pressed)
Jun 17 11:52:17 Cosmic xdg-desktop-portal-cosmic[104300]: [2026-06-17T17:52:17Z WARN  xdg_desktop_portal_cosmic::remote_desktop] NotifyKeyboardKeycode: keycode=125 (0x7d), state=0 (released)
Jun 17 11:52:17 Cosmic xdg-desktop-portal-cosmic[104300]: [2026-06-17T17:52:17Z WARN  xdg_desktop_portal_cosmic::remote_desktop] NotifyKeyboardKeysym: keysym=116 (0x74), state=0 (released)
```
so `super pressed` goes to the compositor, then `t pressed` goes to the focused window, `super released` goes to the compositor so it shows the cosmic launcher. And `t released` goes to the cosmic launcher which doesn't do anything with it.

hojjatabdollahi: I'm working on adding `XkbState` per ei connection.

hojjatabdollahi:
@Drakulix @ids1024 Here's the design that I came up with, I'd appreciate your feedback here:

The problem:
- the RD keyboard input are being affected by the physical keyboard modifiers
- `ei_text::Keysym` semantically is "text", but in practice it's mixed with other keycode events and should be handled with them (Rustdesk sends super+t as keycode+keysym)
- seat has one place for modifiers and pressed keys so different inputs will step on eachother's foot
- The way we currently detect shortcuts only looks at the seat and not the source

Here's my final design:
Each ei connection owns an isolated keyboard state. Actions that are based on modifiers use the modifiers of the event's source. focus and cursor stay shared.

- Per source state: `ei_isolated_kbd` owns xkb state + modifiers. 
- `ei_keyboard` and `ei_text::Keysym` for a connection are stored in the connection's state, so super+t reassembels there
- update shortcut matcher: `filter_keyboard_input` uses the connection's modifiers
- for actions with the pointer (super+drag, super+scroll) it reads the `source_modifiers` so a physically-held super never affects a remote drag and vice versa.
- forwarded keys goes to the shared focus with the source's keymap/mods.
- if no modifier is being held, the keysym is not modifier or non-printable, and the input text box can handle zwp_text_input, then `ei_text::Keysym` is forwarded to the input (just like `ei_text::TextUtf8`) otherwise it goes through the process above turning into a keycode.


This is the best I could come up with.

hojjatabdollahi:
Here's me connecting using Rustdesk, moving the cursor typing, pressing super+g, super+drag and ctrl+c:


https://github.com/user-attachments/assets/f51da4ac-a5e4-47ea-8bcd-af1cfee5702d

hojjatabdollahi:
I used the same method to fix shortcuts in virtual keyboard too. Here's cosmic-osk:

https://github.com/user-attachments/assets/3e7660a3-35be-4203-ad29-ba8fb7700701

I understand that this PR is getting too big to properly review, and I'm ok with removing all the `ei_text` related stuff into its own PR. It would at least enable pointer emulation in Steam Input, until we figure out all the kinks of Remote Desktop.

Drakulix:
>     * if no modifier is being held, the keysym is not modifier or non-printable, and the input text box can handle zwp_text_input, then `ei_text::Keysym` is forwarded to the input (just like `ei_text::TextUtf8`) otherwise it goes through the process above turning into a keycode.
> 

I like the overall design, but I think you misunderstood me. I fully agree, that we need to convert the `Keysym` to keycodes and process them like any other keyboard input. Which also means they will trigger shortcuts.

**Only** for the path of explicit `TextUtf8` events, which contain valid utf8, I don't want to handle any shortcuts, **if** zwp_text_input ends up being not available and we synthesize keysyms for the passed text.

In general I like the design.

hojjatabdollahi:
> **Only** for the path of explicit `TextUtf8` events, which contain valid utf8, I don't want to handle any shortcuts, **if** zwp_text_input ends up being not available and we synthesize keysyms for the passed text.

Got it. So under no circumstance `ei_text.utf8` will cause a shortcut. Previously the fallback branch of `ei_text.utf8` would potentially cause shortcuts if a modifier was held down. I'll make sure that doesn't happen. `TextUtf8` is just a string and as a fallback will become a bunch of keycodes, but won't cause any shortcuts. I'll add a commit soon for that.

But we still want `ei_text.keysym` to cause shortcuts (under certain conditions), because we're redirecting the legacy `notify_keyboard_keysym` to it. 

However, if there are no modifiers and a text input is focused and the keysym is printable we just forward the keysym to the input, that would allow characters that are [not in our layout](https://libinput.pages.freedesktop.org/libei/api/group__libei.html#ggad07c425e3144e8751f7a2ea72f6c8a24ab6e65d7812dbc3458097a4780083007d) to not be silently dropped. (for example if I connect using my phone and type a chinese character).

Gnome drops those but KDE has a fallback layout to handle these. If an input is focused we forward these, but if it's not, then we drop them too.

src/input/mod.rs
```diff
@@ -1639,15 +1688,262 @@ impl State {
...
+            .get(conn)
+            .and_then(|iso| iso.keycode_for_keysym(keysym))
+        else {
+            tracing::warn!(
+                "EI text keysym {:?} is not in the active layout; ignoring",
+                keysym
+            );
+            return;
+        };
+        self.inject_ei_key_internal(conn, keycode, key_state, handle_shortcuts);
```
ids1024: This seems to just be sending the keycode without activating the right shift level/group first? Which doesn't actually send the correct keysym.

src/config/mod.rs
```diff
@@ -820,6 +821,36 @@ fn config_changed(config: cosmic_config::Config, keys: Vec<String>, state: &mut
...
                     }
                 }
+                // Re-create the virtual keyboard on each libei sender seat with the new keymap.
+                for seat in state.common.ei_seats.values() {
+                    if let Err(err) =
+                        seat.add_keyboard("virtual keyboard", xkb_config_to_wl(&value))
+                    {
+                        warn!(?err, "Failed to update libei virtual keyboard config");
+                    }
+                }
```
hojjatabdollahi: Testing with freeRDP, if I press super+space to change the keyboard layout, it does change, but because we remove and recreate the keyboard, we also lose focus, I have to click in the textbox to be able to type.

src/libei.rs
```diff
@@ -0,0 +1,155 @@
...
+}
+
+impl State {
+    /// Inject UTF-8 text (from an EI `ei_text` device) into the focused client.
+    pub fn inject_ei_text(&mut self, conn: &smithay::reexports::reis::eis::Connection, text: &str) {
+        let seat = self.common.shell.read().seats.last_active().clone();
+        let text_input = seat.text_input();
+        let mut injected = false;
+        text_input.with_active_text_input(|ti, _surface| {
+            ti.commit_string(Some(text.to_owned()));
```
hojjatabdollahi: Turns out you can't just commit text to an input. The compositor needs to send `enter` for the app to enable the input for us to commit. And we only send `enter` if an IME is enabled. So, utf8 character that are not in the current layout are just dropped.

hojjatabdollahi: So either we have to register as IME, relax the condition in smithay and send "enter" to every, or send `ei_text.utf8` strings through the fallback method. :shrug:

src/config/mod.rs
hojjatabdollahi: for now I removed this bit, everything is still working :shrug:

src/input/mod.rs
hojjatabdollahi: Should be fixed, also won't ignore missing ones anymore either.

src/libei.rs
```diff
@@ -0,0 +1,155 @@
...
+                                data.inject_ei_key(
+                                    connection.eis_connection(),
+                                    event.key_code(),
+                                    event.state(),
+                                );
+                            }
+                            other => {
+                                let backend_id =
+                                    InputBackendId::Ei(connection.eis_connection().clone());
+                                data.process_input_event(other, backend_id);
```
ids1024, line 142: Testing with `lan-mouse`, it looks cursor movement is unresponsive unless we add a `.schedule_render()` call after this, as is used with libinput. (We'll need to do this only on the kms backend).

hojjatabdollahi: Feel free to push any fixes/changes you want to this (I think I did enable that option in my PR).

I was planning to test Deskflow, but lan-mouse looks pretty good too. I'll test that too.

hojjatabdollahi: I have been testing with Rustdesk, Windows App, and freeRDP.

Here's a bunch of issues that I noticed (some may not be real issues, I haven't dug deep):
1- when you share one screen the focus/pointer is not confined to that, so it may jump to another output, but you can't bring it back because Rustdesk for example is mapping your touches on the phone to mouse movements and they're limited, so you can't move the cursor back.
2- No absolute pointer position
3- if you switch to touch mode in Windows App, the cursor gets hidden, but you can't really touch anything (I know you're working on touch mode for cosmic)
4- I think there is some offset in Rustdesk's cursor vs the compostior's cursor and looked like it's related to my fractional scaling.

I also noticed Rustdesk uses shift+insert to copy paste utf8 characters instead of sending keysym and libcosmic apps don't handle it (except for cosmic-term) so I can type Persian via Rustdesk on most programs except libcosmic ones.

ids1024: I guess https://github.com/enigo-rs/enigo/blob/main/src/linux/libei.rs needs to be updated to use text input with the latest `reis` version, but `rustdesk` seems to vendor an older version in https://github.com/rustdesk/rustdesk/tree/master/libs/enigo. I guess that is using the `xdo` backend? So it's going via `XWayland` maybe?

ids1024: I've added a commit to call `schedule_render()`, and one to improve scroll event handling.

hojjatabdollahi:
Noticed an issue with absolute pointer position when remoting to "pop os" using "windows app" on "Mac os". 😅

ids1024:
Having used it for a while, receiving keyboard and pointer input from another system with `lan-mouse` seems to work pretty well now, though that's not a client that's doing anything fancy like `ei_text`.

src/shell/seats.rs
```diff
@@ -185,7 +198,7 @@ struct FocusedOutput(pub Mutex<Option<Output>>);
 pub struct PointerConstraintHint(pub Mutex<Option<(WlSurface, Point<f64, Logical>)>>);
 
 #[derive(Default)]
-pub struct LastModifierChange(pub Mutex<Option<Serial>>);
+pub struct LastModifierChange(pub Mutex<(HashMap<InputBackendId, Serial>, Option<Serial>)>);
```
Drakulix, line 210: Is this cleaned up at some point or will this just continue to accumulate entries, the more ei-connections we handled?

src/libei.rs
Drakulix: Apps use the enter event to detect an IME present. We should only send this if we have an EI connection at the very least. I'd argue we probably should register an IME or provide an escape hatch in smithay.

hojjatabdollahi: I've never used IMEs and I don't know if apps would change their behavior if they think an IME is present.
Also what if an actual IME exists on the system, I don't think you can have two IMEs.

Maybe we add the escape hatch to smithay, that the compositor will act as an input method (send `enter` + `commit`) only if: the call is coming from an active ei connection and no real IMEs are bound.

What do you think?

src/libei.rs
hojjatabdollahi: > I guess that is using the `xdo` backend? So it's going via `XWayland` maybe?

It's just simply pressing `shift+insert`: https://github.com/rustdesk/rustdesk/blob/16570ee34f5eaaf83fd9e76258e5675b4645f33d/src/server/input_service.rs#L1837

src/shell/seats.rs
hojjatabdollahi: It was not!

hojjatabdollahi:
I tested adding the compositor as an input method if there are no other IME bound, and there is a zwp_text_input_v3 field present and an active ei_connection.

But

Looks like we're doing this: 
```
GTK_IM_MODULE=ibus
QT_IM_MODULE=ibus
CLUTTER_IM_MODULE=ibus
XMODIFIERS=@im=ibus
```
So GTK based apps, won't use `im_wayland` they'll be using `ibus` which means they don't even create `zwp_text_input_v3` object at all.

And I don't think libcosmic/iced has any im-module either, so they're not creating zwp_text_input_v3 either and ei_text.utf8 always falls back to character based method. 

I did add a batch mode to it though, otherwise the many layout switches was causing the system to freeze and the panel to crash. (I'll push that soon)

ids1024:
> And I don't think libcosmic/iced has any im-module either, so they're not creating zwp_text_input_v3 either and ei_text.utf8 always falls back to character based method.

It looks like I get a `zwp_text_input_v3` at least being being created by cosmic-term, cosmic-edit, etc.:

```
> WAYLAND_DEBUG=1 cosmic-term 2>&1 | grep zwp_text_input_v3
[3630173.187]  -> zwp_text_input_manager_v3#17.get_text_input(new id zwp_text_input_v3#23, wl_seat#9)
[3630186.023]  -> zwp_text_input_manager_v3#60.get_text_input(new id zwp_text_input_v3#79, wl_seat#43)
```

hojjatabdollahi:
Yeah, I had a bug in smithay, it normally ignores text-input requests if there are no IME, so it was dropping the responses from the app. We'd send `enter` never get `enable`.


We probably should also drop those env vars so GTK and QT apps also properly create text input v3. As Jacob mentions in this PR they're still being set, so all non libcosmic apps will go through the fallback method.
https://github.com/pop-os/default-settings/pull/183

ids1024:
`GTK_IM_MODULE=ibus` may still work better than `GTK_IM_MODULE=wayland`.

The smithay/cosmic-comp IME support could probably be better, but also the Wayland protocols have known design issues that are expect to need breaking changes to fix, which is why https://gitlab.freedesktop.org/wayland/wayland-protocols/-/tree/main/experimental/xx-text-input and https://gitlab.freedesktop.org/wayland/wayland-protocols/-/tree/main/experimental/xx-input-method exist.

hojjatabdollahi:
We serve the video stream at physical size, but our cursor is in physical space. As far as I know Gnome streams at logical size, so the cursor coords match. What I did was kept streaming at physical size, and reported physical coords as the logical coords to the client (with a scale of 1) but then convert the coords coming from the user back to logical coords using the correct scale. [here](https://github.com/pop-os/xdg-desktop-portal-cosmic/pull/317/changes/a8cf3193df9e4f6debd863083e867e67fd3ec3a8).

Now the cursor movements in Rustdesk are 1:1 in the viewer and the host.

hojjatabdollahi:
I don't know how to test multi monitor. What client would be using multi monitor?

Native Rustdesk allows for streaming multiple outputs, but then it uses uinput for input instead of Remote Desktop portal, and in cosmic comp we map uinput to a single output (I assume for tablet support).

Tried TeamViewer but its Wayland support is abysmal.

The only thing I can think of is if we create a RDP server for COSMIC, but that's out of scope here.


So as far as this PR is concerned I have tested with what I can:
- XWayland (via XTestFakeMotionEvent)
- Steam Input
- Lan Mouse
- Rustdesk Flatpak (with Android, and PC guests).

So, I'd say it's ready for review. Let me know what you think.

src/config/mod.rs
```diff
@@ -820,6 +821,32 @@ fn config_changed(config: cosmic_config::Config, keys: Vec<String>, state: &mut
                         }
                     }
                 }
+                // TODO: we deliberately do NOT recreate the EIS virtual keyboard device
+                // here. It was causing loss of focus when layout changes. I'll have to come back to
+                // this issue.
+
+                // Rebuild the per-connection isolated keyboard state with the new keymap
+                // after first releasing any keys that source still has held.
```
ids1024: If this isn't calling `.add_keyboard` on the `EiInputSeat`, but **does** create a new `IsolatedKeyboardState`, then doesn't that mean the compositor is interpreting keycodes from the client using the new keymap, but without sending that keymap to the client? If so, that presumably could lead to non-nonsensical results in clients that try to map keysyms/text to keycodes.

hojjatabdollahi: > non-nonsensical

Sounds good to me! :sunglasses: 

IIRC this was because of the way we were applying keymap. I think the changes you recently made to just switch the keymap in the group would affect this. I have to look closer, double check and will get back to you.

ids1024: Yeah, once https://github.com/pop-os/cosmic-settings-daemon/pull/153 and https://github.com/pop-os/cosmic-applets/pull/1441 are merged, the XKB configuration should only change when the keyboard configuration is changed in cosmic-settings.

hojjatabdollahi: I suppose it will also fix the issue that my repeat/delay settings kept getting reset too, since we won't change the config.

hojjatabdollahi:
We should migrate the ci in cosmic-comp and smithay to something that runs clippy for the correct toolchain version (like https://github.com/pop-os/cosmic-comp/pull/2306).

I think the clippy complains are Rust 1.97 linter complaining.

ids1024:
> I think the clippy complains are Rust 1.97 linter complaining.

I believe this is intentional. Clippy should be `rust-version` aware, so all the lints it produces should be valid on the version of Rust we target.

hojjatabdollahi:
> > I think the clippy complains are Rust 1.97 linter complaining.
> 
> I believe this is intentional. Clippy should be `rust-version` aware, so all the lints it produces should be valid on the version of Rust we target.

But cosmic-comp is on 1.93. The CI pulls in the latest stable toolchain (1.97) and clippy creates complaints that if you run locally (without `+nightly` that I saw you did in Smithay) you won't see those complaints.

https://github.com/pop-os/cosmic-comp/pull/2306/changes#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR16

hojjatabdollahi:
But more importantly, our clippy test starts failing on code that was fine and nobody touched just because a new stable rust is released.

ceast3:
The `clippy-check` job here is red on current stable, which I think is the last thing blocking this from a clean CI run. The failures are all new stable-clippy lints, not anything wrong with the Ei work itself:

- CI runs `cargo +stable clippy --all-features -- -D warnings` (stable is 1.97.1), while the tree pins 1.93 via `rust-toolchain.toml`. Between those, clippy extended `question_mark` to cover `else if let Some(x) = … else { return None }`, extended `collapsible_if` to let-chains, and added `useless_borrows_in_formatting`.

Fixed all 10 errors from the failing run, following clippy's own machine-applicable suggestions, plus `cargo fmt`:

- 6× `question_mark` in `src/shell/mod.rs`
- 2× `collapsible_if` collapsed into let-chains (`src/input/mod.rs`, `src/wayland/protocols/keyboard_layout.rs`) — the crate is edition 2024, so this is the idiomatic form
- 1× `useless_borrows_in_formatting` (`src/backend/kms/socket.rs`)
- 1× needless `&` (`src/state.rs`)

No behavior change intended. Verified locally on stable 1.97.1: `cargo +stable clippy --all-features -- -D warnings` is clean and `cargo +stable test --all-features` passes. I also smoke-tested the compositor nested (winit) against the #317 portal branch — `org.freedesktop.portal.RemoteDesktop` → `ConnectToEIS` → `GetSenderSocket` handed over a working EIS socket, and a libei sender injected both pointer motion and keyboard input successfully.

Happy to open this as a PR against your branch instead if that's easier to merge — just say the word.

<details>
<summary>Diff (git diff against this PR's head; +32 −39)</summary>

```diff
diff --git a/src/backend/kms/socket.rs b/src/backend/kms/socket.rs
index 3babd35a..f1c2b0eb 100644
--- a/src/backend/kms/socket.rs
+++ b/src/backend/kms/socket.rs
@@ -36,7 +36,7 @@ impl Common {
     ) -> Result<Socket> {
         let socket_name = format!(
             "{}-{}",
-            &self.socket.to_string_lossy(),
+            self.socket.to_string_lossy(),
             render_node
                 .dev_path()
                 .unwrap()
diff --git a/src/input/mod.rs b/src/input/mod.rs
index e87c2c3e..759f6fe1 100644
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -1965,10 +1965,10 @@ impl State {
             let seat = self.common.shell.read().seats.last_active().clone();
             self.seat_active_layout(&seat)
         };
-        if let Some(layout) = active_layout {
-            if let Some(iso) = self.common.ei_isolated_kbd.get_mut(conn) {
-                iso.set_layout(layout);
-            }
+        if let Some(layout) = active_layout
+            && let Some(iso) = self.common.ei_isolated_kbd.get_mut(conn)
+        {
+            iso.set_layout(layout);
         }
 
         // If this keysym was injected via the keycode path on press, its release must go the
diff --git a/src/shell/mod.rs b/src/shell/mod.rs
index 326b064b..39d692e2 100644
--- a/src/shell/mod.rs
+++ b/src/shell/mod.rs
@@ -3878,13 +3878,13 @@ impl Shell {
                 }
 
                 (initial_window_location, layer, workspace.handle)
-            } else if let Some(sticky_layer) = self
-                .workspaces
-                .sets
-                .get_mut(&cursor_output)
-                .filter(|set| set.sticky_layer.mapped().any(|m| m == &old_mapped))
-                .map(|set| &mut set.sticky_layer)
-            {
+            } else {
+                let sticky_layer = self
+                    .workspaces
+                    .sets
+                    .get_mut(&cursor_output)
+                    .filter(|set| set.sticky_layer.mapped().any(|m| m == &old_mapped))
+                    .map(|set| &mut set.sticky_layer)?;
                 let elem_geo = sticky_layer.element_geometry(&old_mapped).unwrap();
                 let mut initial_window_location = elem_geo.loc.to_global(&cursor_output);
 
@@ -3927,8 +3927,6 @@ impl Shell {
                     ManagedLayer::Sticky,
                     self.active_space(&cursor_output).unwrap().handle,
                 )
-            } else {
-                return None;
             };
 
         toplevel_leave_workspace(&window, &workspace_handle);
@@ -4222,14 +4220,13 @@ impl Shell {
                 .unwrap()
                 .to_global(&set.output);
             (&mut set.sticky_layer, geometry)
-        } else if let Some(workspace) = self.space_for_mut(mapped) {
+        } else {
+            let workspace = self.space_for_mut(mapped)?;
             let geometry = workspace
                 .element_geometry(mapped)
                 .unwrap()
                 .to_global(workspace.output());
             (&mut workspace.floating_layer, geometry)
-        } else {
-            return None;
         };
 
         let new_loc = if edge.contains(ResizeEdge::LEFT) {
@@ -4266,7 +4263,8 @@ impl Shell {
             ReleaseMode::Click,
         ) {
             grab.into()
-        } else if let Some(ws) = self.space_for_mut(mapped) {
+        } else {
+            let ws = self.space_for_mut(mapped)?;
             let node_id = mapped.tiling_node_id.lock().unwrap().clone()?;
             let (node, left_up_idx, orientation) = ws.tiling_layer.resize_request(node_id, edge)?;
             ResizeForkGrab::new(
@@ -4279,8 +4277,6 @@ impl Shell {
                 ReleaseMode::Click,
             )
             .into()
-        } else {
-            return None;
         };
 
         Some(((focus, new_loc), (grab, Focus::Keep)))
@@ -4517,10 +4513,9 @@ impl Shell {
             .find(|set| set.sticky_layer.mapped().any(|m| m == &mapped))
         {
             &mut set.sticky_layer
-        } else if let Some(workspace) = self.space_for_mut(&mapped) {
-            &mut workspace.floating_layer
         } else {
-            return None;
+            let workspace = self.space_for_mut(&mapped)?;
+            &mut workspace.floating_layer
         };
 
         let grab: ResizeGrab = if let Some(grab) = floating_layer.resize_request(
@@ -4532,7 +4527,8 @@ impl Shell {
             ReleaseMode::NoMouseButtons,
         ) {
             grab.into()
-        } else if let Some(ws) = self.space_for_mut(&mapped) {
+        } else {
+            let ws = self.space_for_mut(&mapped)?;
             let node_id = mapped.tiling_node_id.lock().unwrap().clone()?;
             let (node, left_up_idx, orientation) =
                 ws.tiling_layer.resize_request(node_id, edges)?;
@@ -4546,8 +4542,6 @@ impl Shell {
                 ReleaseMode::NoMouseButtons,
             )
             .into()
-        } else {
-            return None;
         };
 
         Some((grab, Focus::Clear))
@@ -4904,7 +4898,8 @@ impl Shell {
                 })),
                 Some(from),
             );
-        } else if let Some(workspace) = self.space_for_mut(&mapped) {
+        } else {
+            let workspace = self.space_for_mut(&mapped)?;
             if mapped.is_minimized() {
                 // TODO: Rewrite the `MinimizedWindow` to restore to fullscreen
                 return None;
@@ -4952,8 +4947,6 @@ impl Shell {
                 },
                 Some(from),
             );
-        } else {
-            return None;
         };
 
         Some(KeyboardFocusTarget::Fullscreen(window))
diff --git a/src/state.rs b/src/state.rs
index 2a39a35d..4bf99259 100644
--- a/src/state.rs
+++ b/src/state.rs
@@ -707,7 +707,7 @@ impl State {
         AlphaModifierState::new::<Self>(dh);
         SinglePixelBufferState::new::<Self>(dh);
         FixesState::new::<Self>(dh);
-        let keyboard_layout_state = KeyboardLayoutState::new::<State, _>(&dh, client_not_sandboxed);
+        let keyboard_layout_state = KeyboardLayoutState::new::<State, _>(dh, client_not_sandboxed);
 
         let background_effect_state = BackgroundEffectState::new::<Self>(dh);
 
diff --git a/src/wayland/protocols/keyboard_layout.rs b/src/wayland/protocols/keyboard_layout.rs
index 0d674f88..f3a44f0a 100644
--- a/src/wayland/protocols/keyboard_layout.rs
+++ b/src/wayland/protocols/keyboard_layout.rs
@@ -51,15 +51,15 @@ impl KeyboardLayoutState {
     {
         let mut keyboard_layouts = mem::take(&mut state.keyboard_layout_state().keyboard_layouts);
         for (keyboard_layout, last_layout) in &mut keyboard_layouts {
-            if let Some(data) = keyboard_layout.data::<LayoutUserData<D>>() {
-                if let Some(handle) = &data.handle {
-                    let active_layout = handle.with_xkb_state(state, |context| {
-                        context.xkb().lock().unwrap().active_layout()
-                    });
-                    if *last_layout != Some(active_layout) {
-                        keyboard_layout.group(active_layout.0);
-                        *last_layout = Some(active_layout);
-                    }
+            if let Some(data) = keyboard_layout.data::<LayoutUserData<D>>()
+                && let Some(handle) = &data.handle
+            {
+                let active_layout = handle.with_xkb_state(state, |context| {
+                    context.xkb().lock().unwrap().active_layout()
+                });
+                if *last_layout != Some(active_layout) {
+                    keyboard_layout.group(active_layout.0);
+                    *last_layout = Some(active_layout);
                 }
             }
         }
```
</details>

Drakulix:
Looking at the code now, but please just rebase this against https://github.com/pop-os/cosmic-comp/pull/2513 instead for the dmabuf changes. I am hoping to be able to merge that very soon.

src/config/mod.rs
```diff
@@ -820,6 +821,22 @@ fn config_changed(config: cosmic_config::Config, keys: Vec<String>, state: &mut
                         }
                     }
                 }
+                // TODO: we deliberately do NOT recreate the EIS virtual keyboard device
+                // here. It was causing loss of focus when layout changes. I'll have to come back to
+                // this issue.
```
ids1024: I still think this needs to have an `EiInputSeat::add_keyboard` call with to handle changes to the keyboard configuration.

hojjatabdollahi: Oh, I thought I fixed that, maybe I didn't remove the comment. I'll check soon.

hojjatabdollahi:
There are a bunch of questions that I suppose at some point we have to get to:
- Will we have `cosmic-remote-desktop-server`?
  - if yes, will it be a remote-assistant mode? (only support an already logged in user session?)
  - or will it support headless remote desktop too?
  - What if it's not headless, what if a monitor is connected to the remote machine, does it mirror? So, if I remote to my computer at work people will see what I'm doing and my computer is unlocked?
  - Or does it take-over and lock the local user out (mentioning on the lock screen that the machine is being controlled remotely, like Windows)?
  - Each one of these decisions changes the scope (I think gnome-remote-desktop switches to headless if no monitor is connected, and switches to remote-assistant/mirroring if one is connected, not super sure)
- How do we handle privacy?
  - Other OSs have a mandatory panel that they can inject a privacy indicator telling the user their screen is being shared, or their input is being controlled remotely, but cosmic doesn't really. Nothing mandates me to use any of the applets. So how would you make sure to notify the user their screen and input is being shared?
  - If a program uses Remote Desktop portal (Rustdesk Flatpak), then we can detect that they're streaming/controlling and notify the user. But cosmic-comp doesn't have a standard dbus interface for remote desktop (like mutter does) so if someone runs Rustdesk Native version, it can just run in the background, stream the display or control the input without the user's knowledge. (The same for the future cosmic-remote-desktop if it's running on the login screen, how do you tell a user?!)
  - We can't tap into screen copy api either, because we use that for everything (thumbnails, workspaces...)
- How do we distinguish between different xwayland apps when they ask for controlling the input via libei? One of the annoyances for people using Steam Input is the portals permission dialog that pops up everytime (imagine trying to use your controller as a mouse, but then need a real mouse to click "allow"), you can't just blanket accept it (then what's the point), but also it's xwayland asking, it could be steam, it could be some other x11 app. TBH I haven't looked deep into this, but from what I know it's just saying "xwayland is asking to control your keyboard and mouse" which is no good. So we may wanna look into that.
- Also https://gitlab.freedesktop.org/xdg/xdg-specs/-/merge_requests/115

Alright, rant over. Just leaving this here to be buried by my next force push! :)

src/config/mod.rs
hojjatabdollahi: I had not done that. Just added a new commit.

isFakeAccount:
I'd imagine the session is done like Gnome where you have both Desktop Sharing and Remote Login.

Remote Login locks the session and Desktop sharing does not. 

Gnome remote desktop struggles with headless for me. Idk if it is a bug or intended behavior. 

I hope a remote desktop server is added because my work won't let me use rust desk.

hojjatabdollahi:
I don't have a touch screen and I'm not sure how good the touch support is in cosmic-comp. So, for the most part I've ignored the "touch" part of libei. For example when an ei connection is made, we send the ei_region for the pointer, but it's empty for touch.

I'm not aware of any client that would use it. Pretty sure Rustdesk in touch mode just sends absolute pointer events.

There is a touch rework in upstream Smithay too. So, I'm just gonna leave this message here, in case we wanna tackle touch support in cosmic-comp.

ids1024:
> I don't have a touch screen and I'm not sure how good the touch support is in cosmic-comp.

https://github.com/pop-os/cosmic-epoch/issues/241 probably needs to be updated, but in general the compositor generally does support passing libinput touch events to clients fine, and then for the most part the client can handle the touch events normally, though there are various things to improve still.

I do have a touch device I can test, though it's hard to test without some client that actually uses it, yeah.

> For example when an ei connection is made, we send the ei_region for the pointer, but it's empty for touch.

https://libinput.pages.freedesktop.org/libei/interfaces/ei_touchscreen/index.html says that "The x/y coordinates must be within the device’s regions or the event is silently discarded" for `motion`/`down` requests.

I guess the ei touchscreen interface is really usable without a defined region? In which case we should probably either send an appropriate region of just not use `add_touch()` on the EI seat until we're actually ready to support it.

hojjatabdollahi:
Turns out we were not too far off.
I tested this using a script that send ei touch events, and a client that visualizes it. Looks good to me. But I can't test regressions with physical mapped touchscreens.

https://github.com/user-attachments/assets/ad4f5cc0-6850-47e7-881f-d63d454a2970

However this depends on:
- an API change in Smithay https://github.com/Smithay/smithay/pull/2115

In addition to the Dmabuf v6:
- [x] https://github.com/pop-os/cosmic-comp/pull/2513

But that means we have to wait for cosmic-comp to be updated to the latest commit in Smithay, which means we have to wait for the touch rework to land in cosmic-comp.

I can remove my last commit, and we do the ei touchscreen stuff in a followup PR, so this PR is not blocked.

ids1024:
Rebased on https://github.com/pop-os/cosmic-comp/pull/2696 for the Smithay update and squashed.

If we don't see any regressions or other major issues with this (or the tablet API change), I think this should be good to merge. Smaller issues can be improved later.

This can be tested using https://github.com/pop-os/xdg-desktop-portal-cosmic/pull/317 with Ei clients like `lan-mouse`.

ids1024:
https://github.com/pop-os/xdg-desktop-portal-cosmic/pull/317 is probably blocked on some improvement to the portal dialog, but this part should be okay to merge before that.

I've added a fix to make `clippy` pass as well.

Drakulix (dismissed):
Would've loved to get https://github.com/pop-os/cosmic-comp/pull/2725 ready in time to land before this, but that didn't happen. This is good code-wise and ready for merging imo.

ids1024:
Rebased and added a commit to whitelist `com.system76.CosmicOSK` on the DBus protocol.

Drakulix (approved):

leviport (approved):

ids1024:
Okay, that should be everything to finally get this merged.

Thanks for working on this!

samrocketman:
When can this be expected to be in a release?  Part of existing LTS or only a future LTS?

hojjatabdollahi:
COSMIC is rolling release, this is already available on PopOS if you have the staging repo enabled, and today or tomorrow it will be available in stable channel.

You can't use this yet though, you have to wait for https://github.com/pop-os/xdg-desktop-portal-cosmic/pull/317.

Liquidmasl:
this landed for me on pop 24.04 and steam input mouse emulation works now, the cursor finally moves. thanks for doing it 🙂

one thing it surfaced though: the RemoteDesktop consent dialog now shows up over and over. once for steam itself, then again for every game i launch, and again every time i start a Remote Play stream. for game streaming that's rough — the prompt appears on the host, so i have to physically walk to the pc and click allow while the stream sits dead.

dbus-monitor on the RemoteDesktop interface suggests xwayland/oeffis never asks for persistence:

```
member=SelectDevices
   array [
      dict entry( string "handle_token"  variant string "oeffis_2_1681692777" )
      dict entry( string "types"         variant uint32 3 )
   ]

member=Start
   array [
      dict entry( string "handle_token"  variant string "oeffis_3_1714636915" )
   ]
```

no `persist_mode` in `SelectDevices`, no `restore_token` in `Start` (0 occurrences in the entire capture), and a fresh `CreateSession → SelectDevices → Start → ConnectToEIS` cycle from a new bus connection each time rather than reusing a session. `app_id` comes through empty too, so the dialog just says "unknown application".

i don't think this is a cosmic bug — the portal is right to prompt when the client doesn't request persistence, and ScreenCast persistence works fine here (restore tokens show up in `flatpak permission-list`). it's specifically the RemoteDesktop/ei path that never asks.

wrote it up in pop-os/xwayland#5, but that repo looks like packaging only so posting here too since you've been in this path recently. do you know if oeffis/xwayland is meant to request `persist_mode` here, or is that an upstream gap?

PR #2450, opened 2026-06-08
Fix login screen not showing up

hojjatabdollahi:
This avoids a race that causes a deadlock on laptops connected to an external monitor.

This is a hack, the latest changes in cosmic-comp makes creating these connections lazy. But this makes it sync and creates it so no-one can call block_on on an inflight init.
___

### DIFF
--- a/src/dbus/mod.rs
+++ b/src/dbus/mod.rs
@@ -42,6 +42,14 @@ impl DBusState {
             a11y_keyboard_monitor: RefCell::new(None),
         }));
         evlh.insert_source(source, |_, _, _| {}).unwrap();
+
+        if let Err(err) = futures_executor::block_on(state.session_conn()) {
+            tracing::error!("Failed to connect to session bus: {}", err);
+        }
+        if let Err(err) = futures_executor::block_on(state.system_conn()) {
+            tracing::error!("Failed to connect to system bus: {}", err);
+        }
+
         let state_clone = state.clone();
         state.spawn(async move {
             if let Err(err) = init_session(&state_clone).await {


ids1024:
Hm, yeah. I guess using `block_on` for proxy calls is fine, but if we use it for `session_conn()`/`system_conn()`, but elsewhere invoke those with the calloop executor on the main thread, we can end up blocking the main thread on a future that will never be woken since the other future doesn't get the chance to progress.

I think we really should avoid blocking the main thread on DBus calls at all (which is the goal here, though more is needed), but in the meantime we need to either block initially to create the connections, or we'd need to use an async executor in a different thread.

ids1024:
Or actually, maybe simply using https://docs.rs/async-once-cell/latest/async_once_cell/struct.Lazy.html instead of `async_once_cell::OnceCell` would work here. Looks like it's designed so that it would?

hojjatabdollahi:
Interesting. That would work.

hojjatabdollahi:
`Pin<Box<dyn Future<Result<...` incoming...

hojjatabdollahi:
I tried Lazy. But the issue persists. The type is too complicated for me to know what I'm doing wrong. Also it seems like Lazy caches Err, so we won't retry on the subsequent calls if the first time it errors out.

ids1024:
Yeah, the fact in theory it would allow re-trying is why I used `OnceCell` instead of `Lazy` in the first place. Though it probably is expected that these things won't work properly if connecting to the DBus daemon fails; and we (and most software) don't really handle daemon restarts anyway. The reliance on the central server is a limitation of DBus.

This workaround would also be potentially racy; if the first blocking call failed, the race could still happen. So if `Lazy`  doesn't work either it would probably be better to change `session_conn: async_once_cell::OnceCell<zbus::Connection>` to just `session_conn: zbus::Connection` for now; and eliminate all the blocking calls some time in the future.

PR #2451, opened 2026-06-09
fix(dbus): Block on creation of zbus connection to avoid deadlock

ids1024:
A fix for the issue reported in https://github.com/pop-os/cosmic-comp/pull/2450.

Using `block_on` in some places for zbus calls is fine (that's what zbus's blocking API does; it has its own executor for background tasks), but this is problematic with `async_once_cell`. If we also use that in the calloop async executor.

I think ideally we should avoid blocking the main thread on any async tasks, but for now we can just block in initialization here.

I still need to try to reproduce the issue, and test this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -243,12 +243,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "async-once-cell"
-version = "0.5.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4288f83726785267c6f2ef073a3d83dc3f9b81464e9f99898240cced85fce35a"
-
-[[package]]
 name = "async-process"
 version = "2.5.0"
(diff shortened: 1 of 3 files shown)


hojjatabdollahi:
The way I'm reproducing the issue is by testing on a laptop (so there's a lid) and connecting my two external monitors and logging out, or restarting. Does that not deadlock on your system?

[I'm testing this]

hojjatabdollahi (approved):
Fixes my issue.

ids1024:
Yeah, it seems to be fairly reproducible on a laptop with one external monitor, when testing a release build.

This does seem to fix the issue, as expected, so this should be good enough for now. It would be good to eliminate all main-thread blocking dbus calls in the future.

leviport (approved):

wash2 (approved):

Issue #930, opened 2024-10-18
[BUG] Moving focus between monitors with shortcut puts cursor in the center of the screen


BrodieRobertson:
Current Behavior
When moving window focus with the shortcuts the cursor is placed in the center of a window, with the exception of moving the focus between 2 monitors, where upon switching monitors it is first placed in the center of the screen

Expected Behavior
Like when moving focus between windows on one monitor, the same should apply when moving between windows on two monitors.

AviiNL:
Addition: With Cursor follows focus **disabled** the cursor still jumps between monitors when switching focus to a window on that specific monitor, although it does **not** move the mouse if focus switches to a window on the same monitor the cursor is already on.

Maleya:
I have been looking everywhere to turn this off, only to realize this is a bug.

PR #2410, opened 2026-05-25
fix: follow the focus after alt+tab to another output

hojjatabdollahi:
If you use alt+tab to switch to a window on another output, the cursor should follow you there regardless of "cursor follows focus" config. Just like it does when you move the focus to a window on another window using super+hjkl. 

Right now, you can use alt+tab to switch focus to a window on another output, but if you use super+jk to change workspace, the workspace on your old output will change not where your focus is.

___
- [x] I have disclosed use of any AI generated code in my commit messages.
  - If you are using an LLM, and do not fully understand the changes it is making to the code base, do not create a PR.
  - In our experience, AI generated code often results in overly complex code that lacks enough context for a proper fix or feature inclusion. This results in considerably longer code reviews. Due to this, AI authored or partially authored PRs may be closed without comment.
- [x] I understand these changes in full and will be able to respond to review comments.
- [x] My change is accurately described in the commit message.
- [x] My contribution is tested and working as described.
- [x] I have read the [Developer Certificate of Origin](https://developercertificate.org/) and certify my contribution under its conditions.

PR #2457, opened 2026-06-11
Respect cursor_follows_focus when activating a window on another output

severine-akki:
## Problem

When clicking an app in the dock/panel while its window is on a **different output**, the pointer is warped to the center of that window — even when **Cursor follows focus** is disabled in Settings. This is surprising and disruptive on multi-monitor setups (you click on monitor A, the window focuses on monitor B, and your cursor is yanked to B).

This is the root cause behind several reports:
- #930 (moving focus between monitors centers the cursor)
- pop-os/cosmic-settings#1269
- pop-os/cosmic-epoch#3121

## Cause

In `src/wayland/handlers/toplevel_management.rs`, the `activate` handler warps the pointer whenever `switching_output` is true, without checking the `cursor_follows_focus` config. Every other focus path in the compositor (`shell::focus::update_focus_state` at `focus/mod.rs`, and `input::actions`) already gates the warp on `cursor_follows_focus`; this path was missed.

## Fix

Gate the pointer warp on `self.common.config.cosmic_conf.cursor_follows_focus`. Keyboard focus and `set_active_output` are unchanged — only the pointer warp becomes conditional, consistent with the rest of the codebase.

When `cursor_follows_focus` is **enabled**, behavior is unchanged (cursor still follows to the window on the other output). When **disabled**, the cursor stays where it is.

## Testing

- Two monitors, `cursor_follows_focus` disabled: click an app in the dock whose window is on the other monitor → cursor no longer jumps. ✅
- `cursor_follows_focus` enabled: cursor still follows to the activated window. ✅

### DIFF
--- a/src/wayland/handlers/toplevel_management.rs
+++ b/src/wayland/handlers/toplevel_management.rs
@@ -110,11 +110,14 @@ impl ToplevelManagementHandler for State {
 
             std::mem::drop(shell);
 
-            // move pointer to window if it’s on a different monitor/output
+            // move pointer to window if it’s on a different monitor/output,
+            // but only when the user opted into cursor_follows_focus.
             let switching_output = seat.active_output() != *output;
             if switching_output && let Some(new_pos) = new_pos {
                 seat.set_active_output(output);
-                if let Some(ptr) = seat.get_pointer() {
+                if self.common.config.cosmic_conf.cursor_follows_focus
+                    && let Some(ptr) = seat.get_pointer()
+                {
                     let serial = SERIAL_COUNTER.next_serial();
                     ptr.motion(
                         self,


Drakulix (approved):
Thanks!

hojjatabdollahi:
@Drakulix 

This reverts https://github.com/pop-os/cosmic-comp/pull/2410. 

There is some discussion related to this in https://github.com/pop-os/cosmic-epoch/issues/2233 and https://chat.pop-os.org/pop-os/pl/ykifswdoyfyydcwsy4dfyticac. 

And Maria said they may need to talk to the team to see what's the best approach here.

We can revert my PR since clicking on a window in the taskbar shouldn't warp the cursor to another output. But I think we should just change the way cosmic-comp works and assume output with keyboard focus (regardless of pointer location) is the active output (just like other OSs) the only thing is a way to indicate an empty workspace has keyboard focus (right now the only indication is the mouse cursor jumping to the middle of that output when you change output, if we find a better way we can get rid of the pointer warping for this purpose).

jklgrasso:
I assigned myself before I read the comment. It looks like some discussion needs to happen before this gets to QA. I'll unassign myself for now.

gordan-ramsey:
This is a broken feature again. can we please get it refixed?

hojjatabdollahi:
@h-n-a what do you mean? what's broken?

gordan-ramsey:
@hojjatabdollahi why would the cursor still follow when the option is disabled in cosmic? its a regression.

hojjatabdollahi:
@h-n-a under what condition is it following? With alt+tab?

henrique-eba:
@hojjatabdollahi For me at least, it is whenever the window gets focused. I have all the options disabled, but the cursor still jumps.

Even windows within the open software make the cursor jump, which is extremely annoying. (like dialogs and such).

Common scenario for me: I have 2 monitors and 2 Firefox windows, 1 on each monitor. Sometimes I want to bring the one on the second monitor to the foreground but still keep working on my first monitor, but the cursor jumps to the other monitor, making me drag the mouse all over again.

Issue #2446, opened 2026-06-07
Add support for `xdg-dialog-v1` protocol


dkonigsberg:
The Wayland xdg-dialog-v1 protocol is described here:
https://wayland.app/protocols/xdg-dialog-v1

It appears to be a way for applications to register surfaces as "dialogs," and thus indicate things like modal state to the compositor.  It may end up being the Wayland-correct way of fixing the issues Cosmic currently has with not handling modal dialogs correctly.

It is currently implemented in both Mutter (Gnome) and KWin (KDE), both of which handle such dialogs correctly.

Most of the compositors listed as not currently supporting it, appear to be tiling-only or special-purpose compositors and not general purpose floating window desktop compositors.

Issue #2259, opened 2026-04-07
X11: Child windows of application can't stay on top of parent window


SotoFidel:
Splintered off of https://github.com/pop-os/cosmic-comp/issues/2242. 

I've noticed this issue mainly in Bitwig, but this issue might be related to https://github.com/pop-os/cosmic-comp/issues/1805

In general, when running a DAW (Bitwig, Reaper, etc), it's expected that plugin windows (like Surge XT), should remain on top of the DAW, even when the window/mouse focus is on the DAW rather than on the plugin. When a workspace is in floating mode,COSMIC fails at keeping plugin windows on top of the DAW when the focus moves to the DAW.

Initially I though that this was a Wayland type of issue rather than a COSMIC one, so for the sake of testing I installed Gnome 46 by running 
`sudo apt install gnome-session gnome-shell-extension-manager`
and launching the Gnome Wayland session. On Gnome, I'm able to work inside of Bitwig without plugin windows moving below the DAW. 
So I think this is a COSMIC issue.

**To Reproduce:**
1. Open up a DAW like Bitwig or Reaper
2. Open a plugin in the DAW like Surge XT
3. With the plugin window open, click anywhere on the parent DAW window so it gets focus
4. See the plugin window go behind the DAW

**Expected behavior:**
The menu item should be clicked, and focus should still be on Surge XT, not the window behind it.

**Extra Context**
These applications still run as X11 applications through XWayland

odd-git:
I have similar issue. When I start Ardour/reaper/bitwig everything seem t work. But when i open any plugin  DAW freeze and the plugin appear after a while with a blank screen or completely dark.  The plugin seems to process the audio  properly but i can't tweak with the plugin control unless it's natively without a GUI.

Issue #1989, opened 2026-01-14
Cosmic does not respect modal dialog windows


dkonigsberg:
I've noticed very inconsistent behavior in the case of modal windows on Cosmic.

On some applications, when an application shows a modal dialog box, clicking on the base window causes focus to move from the dialog box to the base window.  The dialog remains on top, but is no longer focused.  I can make this happen with the majority of apps on my system.

On some some other applications, however, doing this causes the base window to become raised above the supposedly-modal dialog box. I've been able to make this happen with  Beyond Compare, hterm, CLion, apps using older Java UIs, and probably others.

This bug is part of the reason focus-follows-mouse basically breaks the system, and why I'm so eager for #863 before I can comfortably daily-drive Cosmic.

dkonigsberg:
Just wanted to add a few screenshots to illustrate the issue I'm talking about.

First, here's a picture of "gedit" showing the base window focused when there's a modal dialog on top of it.  You can't type in the base window in this state, but it shouldn't be focus-able in the first place:
<img width="1198" height="802" alt="Image" src="https://github.com/user-attachments/assets/78648d1d-dae0-46f6-b332-a28b22974ffd" />

Second, here's a picture of "HTerm" where I can actually raise the base window above the modal dialog. This also should not be possible, and isn't under any other window manager:
<img width="1265" height="757" alt="Image" src="https://github.com/user-attachments/assets/48372a12-edcf-482d-ad6f-07e1e30239fc" />

RafaelKr:
In JetBrains IDEs (the ones based on IntelliJ IDEA) there's a `File > Recent projects > Manage projects...` (default shortcut `Alt+P` on Linux) modal dialog. The modal disappears when it loses focus. So when I try to open it, it shows up for a split-second and then instantly closes again. I can imagine this behavior is also related to this issue.

Update 2026-03-27: This works now, not sure if it's related to the IntelliJ IDEA 2026.1 release or the cosmic updates, which I both made yesterday. cc @loffa

dkonigsberg:
From a bit of further testing, I think modal windows being covered mostly happens with X11 applications.  However, modal windows being capable of losing focus (when this shouldn't be possible) still happens for both Wayland and X11 applications.

rodraah:
I’ve noticed the same issue. 
Not sure if it’s the same bug, but when an app has multiple child windows and you close one, the main window jumps in front of the remaining child windows:
[...]

Issue #1805, opened 2025-11-12
X11 transient parent not working


abique:
## Problem

X11 transient parent windows don't stay on top of the parent window and are moved behind by clicking on the parent window. They must stay on top!

## Repro Steps

- install bitwig studio, eg: `flatpak install com.bitwig.BitwigStudio`
- run bitwig studio
- click on the "cpu icon" in the transport area, this will open the dsp performance graph
- click on the parent window, the performance graph will move behind, which shouldn't happen and doesn't with GNOME

This problem happens will all child windows, including plugin's window which is really bad for the application's workflow.

PR #2471, opened 2026-06-14
feat: implement xdg-dialog-v1, modal dialog handling and fix issues with overlapping children windows

### LINKED ISSUE #2776: xdg-dialog-v1
From the Poland summit

* Position on top of the window
* Movable
* Floating
* Don’t dim window beneath

For COSMIC apps
  * Use this protocol
  * Close when un-focused

mayakwd:
Implements [xdg-dialog-v1](https://wayland.app/protocols/xdg-dialog-v1) and makes modal dialogs behave consistently for both Wayland and X11 clients. Also, it fixes issues with parent windows overlapping transient children (huge pain, simple solution) 

Fixes:
- #2446 (the modal-dialog behavior part: focus, stacking, and dimming for a window blocked by a modal child).
- Transient children overlapped by parent window issues fixes: 
  - #1805,
  - #1989,
  - #2259, 
  - pop-os/cosmic-epoch#1944, 
  - pop-os/cosmic-epoch#3171
  - probably more (some of them partially)

https://github.com/user-attachments/assets/55db8d6a-0404-40a6-a8ed-82d278b79379

## Important scope note
#1989's thread mixes modal and *non-modal* transient windows, which end up behind their parent toplevel. For example:
- gedit's "About" window: GTK3 apps on Wayland can't signal modality (only via gtk-shell protocol, which is not supported). The GTK4 gedit build displays the "About" window as a modal one.
- JetBrains products are now on Wayland, and they implemented their own client-side focus-managing system, intentionally moving focus to pseudo-modal windows like Settings)

Non-modal transient windows ending up behind their parent is a separate mechanism from modality — and this PR addresses it too: the parent-child check is unified in CosmicSurface::is_parent_of (xdg parent for Wayland, WM_TRANSIENT_FOR for X11), and raise_with_children now uses it. Previously, only Wayland children were raised with their parents, which is why mostly X11 apps (Reaper, Bitwig, JetBrains) were hit the hardest.

## Note on smithay bump
The X11 side relies on the smithay change I landed in Smithay/smithay#2062 (`_NET_WM_STATE_MODAL` client messages -> `XwmHandler::modal_request` / `unmodal_request`, `X11Surface::set_modal` / `is_modal`). **UPD:** no smithay bump needed anymore since cosmic-comp's master already includes everything needed.

## What it does

A toplevel that has a visible modal child dialog is now treated as *blocked*:

- **Focus**
  Keyboard focus on the blocked parent is redirected to the modal dialog, so the parent can't regain focus while the dialog is open. The redirect lives in a single choke point in `Shell::set_focus`, so every focus path is covered. Closing the dialog switches focus back to its parent with an intentional focus override to provide a good UX and avoid focusing on an unrelated window in the focus stack.
- **Stacking**
  The dialog always stays above its parent. Focusing a modal dialog raises its entire parent chain, so the parent can't end up buried behind unrelated windows.
- **Refused actions**
  Move/resize/maximize/minimize/fullscreen of a blocked parent is refused, and the dialog shortly shakes to point the user to it. Clicking the blocked parent also shakes it (watch the video).
- ~~**Dimming**
   The blocked parent's content is dimmed with a short fade.~~
- **Transient stacking (modal and non-modal)**
  Transient children are kept above their parents for both Wayland and X11. The parent-child relation is unified in `is_parent_of`. Previously, the `raise_with_children` logic only recognized Wayland parents, so X11 transient children could end up buried under their parent.
- **Placement**
   Modal dialog opens centered on top of its parent. This is initial placement only. The dialog stays floating and movable.
  

## What I've tested
- Wayland apps.
- X11: a GTK app under `GDK_BACKEND=x11`, including toggling `set_modal(true)` *after* the window is mapped (checking the new smithay callback from Smithay/smithay#2062).
- Some applications from the mentioned issues: Reaper, Bitwig, Godot
- Scenarios
   - focus redirect
   - shake on each refused action and on click
   - ~~dim fade in/out~~
   - dialog close -> focus returns to parent
   - modal-of-modal chains
   - tiled/floating/stacking

## What I didn't do
There are pre-existing tiling/stacking/transient gaps that the modal support just makes more visible. I left them out on purpose - they touch a lot of things, and it's better to discuss things and land them separately:

- A transient family isn't moved between workspaces as a whole, so a modal dialog can be sent to a different workspace from its parent. In that case, focus restoration then falls back to the focus stack, not a parent in a different space.
- A dialog of an inactive stack tab stays visible over the active tab.
- ~~A fullscreen parent isn't dimmed.~~

*Not relevant for this PR but still valuable:*
- While building the dim, I found an issue related to geometry presented by `CosmicMapped::active_window_geometry()`. For a non-stacked X11 window, its `Window` branch counts the SSD header twice (the `loc` already skips the header, and the `size` adds `SSD_HEIGHT` again), while the `Stack` branch is content-only. It also affects the screencopy cursor hit-test and the tiling group overlay. For now, I've made a workaround (with an explanation in the code comments and `TODO` markers). 

### DIFF
--- a/src/shell/layout/floating/mod.rs
+++ b/src/shell/layout/floating/mod.rs
@@ -8,7 +8,10 @@ use std::{
 
 use cosmic_comp_config::AppearanceConfig;
 use cosmic_settings_config::shortcuts::action::ResizeDirection;
-use keyframe::{ease, functions::EaseInOutCubic};
+use keyframe::{
+    ease,
+    functions::{EaseInOutCubic, EaseOutCubic},
+};
 use smithay::{
     backend::renderer::element::{
         AsRenderElements, RenderElement,
@@ -47,6 +50,8 @@ pub use self::grabs::*;
 
 pub const ANIMATION_DURATION: Duration = Duration::from_millis(200);
 pub const MINIMIZE_ANIMATION_DURATION: Duration = Duration::from_millis(320);
+const SHAKE_ANIMATION_DURATION: Duration = Duration::from_millis(320);
+const SHAKE_AMPLITUDE: f32 = 10.0;
 
 #[derive(Debug, Default)]
 pub struct FloatingLayout {
@@ -76,6 +81,10 @@ enum Animation {
         previous_geometry: Rectangle<i32, Local>,
         target_geometry: Rectangle<i32, Local>,
     },
+    Shake {
+        start: Instant,
+        geometry: Rectangle<i32, Local>,
+    },
 }
 
 impl Animation {
@@ -84,12 +93,13 @@ impl Animation {
             Animation::Tiled { start, .. } => start,
             Animation::Minimize { start, .. } => start,
             Animation::Unminimize { start, .. } => start,
+            Animation::Shake { start, .. } => start,
         }
     }
 
     fn alpha(&self) -> f32 {
         match self {
-            Animation::Tiled { .. } => 1.0,
+            Animation::Tiled { .. } | Animation::Shake { .. } => 1.0,
             Animation::Minimize { start, .. } => {
                 let percentage = Instant::now()
                     .duration_since(*start)
@@ -120,6 +130,7 @@ impl Animation {
             Animation::Unminimize {
                 previous_geometry, ..
             } => previous_geometry,
+            Animation::Shake { geometry, .. } => geometry,
         }
     }
 
@@ -131,6 +142,21 @@ impl Animation {
         gaps: (i32, i32),
     ) -> Rectangle<i32, Local> {
         let (duration, target_rect) = match self {
+            Animation::Shake { start, geometry } => {
+                let progress = Instant::now()
+                    .duration_since(*start)
+                    .min(SHAKE_ANIMATION_DURATION)
+                    .as_secs_f32()
+                    / SHAKE_ANIMATION_DURATION.as_secs_f32();
+                let envelope: f32 = ease(EaseOutCubic, 1.0, 0.0, progress);
+                let offset =
+                    (SHAKE_AMPLITUDE * envelope * (progress * 3.0 * std::f32::consts::TAU).sin())
+                        .round() as i32;
+
+                let mut geometry = *geometry;
+                geometry.loc.x += offset;
+                return geometry;
+            }
             Animation::Minimize {
                 target_geometry, ..
             }
@@ -375,7 +401,9 @@ impl FloatingLayout {
                     } => {
                         *target_geometry = geometry;
                     }
-                    Animation::Minimize { .. } | Animation::Tiled { .. } => {}
+                    Animation::Minimize { .. }
+                    | Animation::Tiled { .. }
+                    | Animation::Shake { .. } => {}
                 }
             } else {
                 self.animations.insert(
@@ -1413,6 +1441,7 @@ impl FloatingLayout {
         self.animations.retain(|_, anim| {
             let duration = match anim {
                 Animation::Tiled { .. } => ANIMATION_DURATION,
+                Animation::Shake { .. } => SHAKE_ANIMATION_DURATION,
                 _ => MINIMIZE_ANIMATION_DURATION,
             };
             Instant::now().duration_since(*anim.start()) < duration
@@ -1420,6 +1449,25 @@ impl FloatingLayout {
         if self.animations.is_empty() != was_empty {
             self.dirty.store(true, Ordering::SeqCst);
         }
+    }
+
+    /// Plays a nudge animation on `mapped` if it lives in this layer; returns
+    /// whether it does, so callers can stop searching other layers.
+    pub fn shake(&mut self, mapped: &CosmicMapped) -> bool {
+        let Some(geometry) = self.space.element_geometry(mapped) else {
+            return false;
+        };
+        if !self.animations.contains_key(mapped) {
+            self.animations.insert(
+                mapped.clone(),
+                Animation::Shake {
+                    start: Instant::now(),
+                    geometry: geometry.as_local(),
+                },
+            );
+            self.dirty.store(true, Ordering::SeqCst);
+        }
+        true
     }
 
     pub fn merge(&mut self, other: FloatingLayout) {
@@ -1653,6 +1701,16 @@ impl FloatingLayout {
                     );
                     window_elements.insert(0, element.into());
                 }
+            }
+
+            let shade = theme.shade;
+            if let Some(dim) = elem.modal_dim_element(
+                renderer,
+                geometry,
+                alpha,
+                [shade.red, shade.green, shade.blue],
+            ) {
+                window_elements.insert(0, dim);
             }
 
             elements.extend(window_elements);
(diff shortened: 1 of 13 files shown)


dkonigsberg:
Just one question about this:
> gedit's "About" window is not modal (it shouldn't block gedit's main window)

gedit's "About" window absolutely does block gedit's main window on Gnome and KDE, so how exactly is it not modal?  (Unless there's another definition of window state that looks modal to a user, but is technically something else.)

Also, I have to thank you immensely for taking on this task.

mayakwd:
> gedit's "About" window absolutely does block gedit's main window on Gnome and KDE, so how exactly is it not modal? (Unless there's another definition of window state that looks modal to a user, but is technically something else.)

Hey @dkonigsberg, thanks for a great question! 

It made me think that I haven't done deep enough research. :-) Yet I've checked gedit once again. So your question is totally valid. 

gedit's "About" window is modal by design, but the installation I have and tested uses GTK3. GTK3 does not support a modality over the Wayland protocol, only X11. Modality support for Wayland was introduced in GTK4. 

GTK uses the `GDK_BACKEND` environment variable to determine which backend to use.  Since COSMIC exports `GDK_BACKEND=wayland,x11` - `wayland` always wins by priority. 

If we enforce X11 usage `GDK_BACKEND=x11 gedit`, then we'll see that the "About" window is modal and blocking the parent, as expected :-)

Drakulix:
> gedit's "About" window is modal by design, but the installation I have and tested uses GTK3. GTK3 does not support a modality over the Wayland protocol, only X11. Modality support for Wayland was introduced in GTK4.

Nitpick just for completeness sake: It does support modality, but only via the private gtk-shell protocol: https://wayland.app/protocols/gtk-shell#gtk_surface1:request:set_modal

Which we will not implement for obvious reasons.

HigherOrderLogic:
Any chance this can be partially extracted to Smithay? I'm also looking into implement this protocol for Niri.

mayakwd:
@HigherOrderLogic, surprisingly, there is nothing to extract to Smithay. Smithay already has everything required for implementation - it already provides [XdgDialogState](https://docs.rs/smithay/latest/smithay/wayland/shell/xdg/dialog/struct.XdgDialogState.html),  https://github.com/Smithay/smithay/pull/2062, added modality requests handling.

The entire implementation in this PR only implements the modality in terms of cosmic-comp window handling and doesn't include anything to extract to Smithay.

HigherOrderLogic:
> surprisingly, there is nothing to extract to Smithay

Thanks. Seems like the infra have been merged into Smithay some time ago, but I just missed it.

And by the way, the app you showcased was Gnome Calculator right? Did you do any special setting to the Calculator app's preference dialog to get it as a separate window? Somehow mine still show up as embedded into the main window instead.

mayakwd:
> The app you showcased was Gnome Calculator right? Did you do any special setting to the Calculator app's preference dialog to get it as a separate window?

It is Gnome Calculator (pre-installed), I didn't do anything special.

mayakwd:
Dropped smithay bump commit, rebased on master.
Slightly updated description.

mayakwd:
@Drakulix, hi! 
Does it make sense to update this PR to resolve the conflicts, or should I just close it?

Drakulix:
> @Drakulix, hi! Does it make sense to update this PR to resolve the conflicts, or should I just close it?

Hi! Sorry, we certainly want this feature, however I still need to discuss exact behavior with our UX team (for things like dimming the window).

So please don't close it, but maybe just wait for me to come back to this in a couple of weeks? (Things are a little busy atm.)

mayakwd:
> Hi! Sorry, we certainly want this feature, however I still need to discuss exact behavior with our UX team (for things like dimming the window).
> 
> So please don't close it, but maybe just wait for me to come back to this in a couple of weeks? (Things are a little busy atm.)

Of course! I apologize if my message gave the impression that I was rushing you. My original intention was to find out whether it makes sense to support this PR, whether a more suitable alternative is in preparation or already available, or maybe it's worth splitting this PR in some way.

Thank you, I'll just update it then to align with the new changes.

WatchMkr:
The behavior we decided on at the Poland summit:

- Position on top of the window
- Movable
- Floating
- Don’t dim window beneath

mayakwd:
@WatchMkr,

Ah, sad that I failed to join you in Warsaw.
So it is movable and floating (I hope I understand it correctly), and I have to remove the dimmer. 

What about shaking? Is it relevant, or should I remove it as well?

WatchMkr:
Sorry we missed you!

The shake is good to bring the users attention to the blocking modal. Since it's already movable and floating, I think it's just removing dimming and positioning the model on top of its parent.

Thanks for working on this!

mayakwd:
> Sorry we missed you!
> 
> The shake is good to bring the users attention to the blocking modal. Since it's already movable and floating, I think it's just removing dimming and positioning the model on top of its parent.
> 
> Thanks for working on this!

Thanks! Updated PR to match the summit decision:
- dimming is gone
- modal now opens centered on top of parent
- it was already floating and movable

Also, rebased on top of master, so conflicts were resolved (CI failures are inherited from master).

src/shell/layout/floating/mod.rs
```diff
@@ -406,6 +434,16 @@ impl FloatingLayout {
...
 
+    /// Geometry of the window `mapped` is transient for, if that parent is mapped here.
+    fn parent_geometry(&self, mapped: &CosmicMapped) -> Option<Rectangle<i32, Local>> {
+        let child = mapped.active_window();
+        self.space
+            .elements()
+            .find(|elem| elem.active_window().is_parent_of(&child))
+            .and_then(|parent| self.space.element_geometry(parent))
+            .map(RectExt::as_local)
+    }
```
Drakulix, line 445: So modals will only be centered for floating and sticky windows? What about a model for a tiled window?

Drakulix (commented):
Looks great code wise, just a single question.

src/shell/layout/floating/mod.rs
mayakwd: Uh, that's right, I just forgot about tiled windows. Working on that. I believe that we need a special branch for sticky windows as well.

mayakwd: Sorry for the delay, during more thorough testing I found several edge cases I'd like to fix, before bringing complete solution to the project. I hope to finish this week (busy days).

mayakwd: Update - fixed several edge-cases and rewrote some paths (PR description also updated).
What changed:

- Dialog used to open on whatever output was active, so on a multi-monitor setup it regularly appeared away from the window it blocks. It now opens on its parent's output and workspace, with modal dialogs additionally centered on the parent. There is more info in PR description (Notes on placement design)
- Locating the parent and its rectangle goes through one unified path now, so centering is correct for a floating, tiled, stacked, sticky or fullscreen parent, on any output or workspace. A fullscreen parent previously yielded no anchor at all and the dialog fell back to cascade placement.
- Dialog's own child window was buried when the dialog was clicked, because the focused window was raised above its own descendants. The raise walk also restacked layers that didn't hold the focused window at all.
- The window menu's Resize entries bypassed the modal block that every other path honored.
- Shake now is  skipped for maximized/fullscreen dialogs.
- `WM_TRANSIENT_FOR` isn't validated by the server, so walk could hang on self-transient windows or cyclic chains.

Issue #2062, opened 2026-02-04
[Bug] high-cpu usage leads to mouse stuttering


lincolnthalles:
### Problem
After a while, `cosmic-comp` starts to use more and more CPU with simple mouse pointer movement.

It drops to under 1% with no movement, but constant movement makes it use a whole CPU core, and the mouse pointer stutters like hell after it gets in this state.

Closing all user apps doesn't fix the behavior.

This bug was getting triggered much more frequently before this week's updates. Since the last version, I have had a full day of work without issues, but I just experienced this again.

Not a hardware issue, since nothing like this happens under GNOME X11. Also, I haven't connected or disconnected a single USB device the whole day, so it shouldn't be some controller or driver weirdness.

Not sure if it's related, but this goes over and over on the log:
```log
# journalctl /usr/bin/cosmic-comp
fev 04 18:05:58 pop-os cosmic-comp[3646]: surface missing from known popups
fev 04 18:06:03 pop-os cosmic-comp[3646]: Client bug: Unable to re-configure repositioned popup.
fev 04 18:06:03 pop-os cosmic-comp[3646]: Client bug: Unable to re-configure repositioned popup.
fev 04 18:06:11 pop-os cosmic-comp[3646]: Client bug: Unable to re-configure repositioned popup.
fev 04 18:06:11 pop-os cosmic-comp[3646]: Client bug: Unable to re-configure repositioned popup.
```

### Reproducing
I haven't found a specific trigger. It will happen after multitasking with browsers, messaging apps, and editors for a while. Gaming isn't necessarily involved to trigger.

### Workaround
The only solution I found is to kill `cosmic-comp`. It will behave normally for a while after this. The issue with this workaround is that, due to Wayland's nature, all applications are closed, and the session is lost when killing the compositor.

### System Information
OS: Pop!_OS 24.04 LTS
Kernel: 6.17.9-76061709-generic
CPU: Intel(R) Core(TM) i7-9700 (8) @ 4.70 GHz
GPU 1: NVIDIA GeForce RTX 2060 Rev. A [driver 590.48.01, main output]
GPU 2: Intel UHD Graphics 630 @ 1.20 GHz [Integrated, not in use]
Display: Single monitor, HDMI, 1920x1080@60Hz
Mouse: Razer Viper Mini@1000 Hz polling rate
cosmic-comp: 0.1\~1769425738\~24.04\~1dfc948

mayakwd:
[...]

Issue #972, opened 2024-11-01
High CPU usage when idle


wiiznokes:
Both cosmic-comp and cosmic-panel are at `2%` of CPU usage when idle (only system monitor opened and the mouse not moving for some time).
This cause my laptop fans to make some noise (this is not the case in KDE or Gnome).
I don't think this was the case in previous versions of cosmic (i could try to bisect but my laptop is really not that great).

`cosmic-comp` commit: 90883c6
system: Fedora 41

Let me know if you need more information

![screenshot-2024-11-01-23-03-33](https://github.com/user-attachments/assets/5df66a96-6f82-48ec-8f97-89c1bc4a9da9)

romanstingler:
you lucky one. here it jumps between 4% and 60% `Intel(R) Core(TM) i7-10750H`
Archlinux/CachyOS

CodedNil:
I'm getting 6% consistent cpu usage, often jumping to 10%, fedora 41 with the latest git version, intel i7-1260P

Alfatango12:
Same here. cosmic-comp hits around 20% with just a terminal and even a bit higher with system settings open. AMD Ryzen 7 3700X

TerminalHash:
on alpha 3 i has high CPU usage only when moving cursor on screen.
after updating to alpha 4, i have constant high CPU usage from 60% to 89%.

on alpha 2 and below i dont expected high CPU usage.

Arch Linux/EndeavourOS, CPU: AMD Athlon II X2 280 (2) @ 3.600GHz

jacobgkau:
![screenshot-2024-12-10-18-07-11](https://github.com/user-attachments/assets/6444cd7e-26aa-4760-8627-7eae7f1d9e8c)

One core seems to be pegged near 100% with Slack & Mattermost on one workspace, some Firefox windows on a couple others, and the system sitting on an empty workspace doing nothing. This causes the system's fans to remain spun up indefinitely. (This is on 2c01c94.)

matilde-t:
I also have a very high CPU usage when moving the mouse

Completely still:
![image](https://github.com/user-attachments/assets/10f6484a-9dee-495a-a51c-e39fbfa296f5)

Moving the mouse in circles for a couple of seconds (inside the terminal window):
![image](https://github.com/user-attachments/assets/0b4e5f1d-3e2f-460f-832d-4d47ae4c004c)

`apt list` says 
[...]

PR #2485, opened 2026-06-17
fix: prevent gradual slowdown and rising CPU usage in long-running sessions

mayakwd:
**The problem**

A new comment on the already-closed issue https://github.com/pop-os/cosmic-comp/issues/2062#issuecomment-4720637087 noted that the original leak is still occurring. Yet it's a totally different leak.

So, from the user's perspective, the problem is that `cosmic-comp` slowly consumes more and more CPU over the course of the session, and as a consequence, the UI starts to feel laggy after some time.  Restarting the compositor fixes it (no reboot is needed), but it slowly builds up again.

**Root cause**

In the original message, we can find clues that everything starts with `Common::refresh`, and that "more and more CPU the longer session lasts" suggests that some collection is growing larger and taking longer to process. 

After some diagnostic time, I've found that in the `SetRectangle` request handler inside `zcosmic_toplevel_manager_v1`, we store a rectangle per surface, and in the set branch, it only pushes, never removing the previous rectangle for the same surface. 

So long-lasting apps push new rectangles on almost every relayout (including workspace changes, min, max, etc), so `ToplevelStateInner::rectangles` grows without any limit.

The only cleanup happens in `ToplevelInfoState::refresh` - it drops rectangles of the surfaces that are already dead.
But long-lasting surfaces are not dead, so we iterate over a growing vector of rectangles and call `surface.upgrade()` on every element on every refresh.

I measured locally and got **~23000** rectangles after **~2-3h** of work, with a small number of windows (**up to 5** at different times) open (it should be exactly 1 per window).

**Fix**
Make `set_rectangle` idempotent per surface: drop any existing rectangle for that surface before pushing the new one (the same thing the unset branch already does).

I checked with temporary debug counters that rectangles now exactly equal the number of windows and reset when windows are closed, instead of only growing.

### DIFF
--- a/src/wayland/protocols/toplevel_management.rs
+++ b/src/wayland/protocols/toplevel_management.rs
@@ -237,11 +237,10 @@ where
                     window_from_handle::<<D as ToplevelInfoHandler>::Window>(toplevel).unwrap();
                 if let Some(toplevel_state) = window.user_data().get::<ToplevelState>() {
                     let mut toplevel_state = toplevel_state.lock().unwrap();
-                    if width == 0 && height == 0 {
-                        toplevel_state
-                            .rectangles
-                            .retain(|(s, _)| s.id() != surface.id());
-                    } else {
+                    toplevel_state
+                        .rectangles
+                        .retain(|(s, _)| s.id() != surface.id());
+                    if width != 0 || height != 0 {
                         toplevel_state.rectangles.push((
                             surface.downgrade(),
                             Rectangle::new((x, y).into(), (width, height).into()),


src/wayland/protocols/toplevel_management.rs
```diff
@@ -242,6 +242,9 @@ where
                             .rectangles
                             .retain(|(s, _)| s.id() != surface.id());
                     } else {
+                        toplevel_state
+                            .rectangles
+                            .retain(|(s, _)| s.id() != surface.id());
```
Drakulix: Thanks for identifying this issue!

The solution looks correct and keeps rectangles of other clients around.
(However technically we could be more strict, we only want to keep the latest rectangle around per client. So we could compare the client_id of both surfaces.)

If we do the same in both cases of the condition though, please just move the retain out of the `if`-statement and invert the condition.

Drakulix (approved):
Thanks!

src/wayland/protocols/toplevel_management.rs
mayakwd: Thanks for the review. I was so happy when I finally figured out the root cause that I didn't realize I'd duplicated the code - I got a little carried away :-)

mmstick:
May fix #972

jklgrasso (approved):

Issue #649, opened 2024-07-23
Handle panic in DRM threads in some way


ids1024:
With https://github.com/pop-os/cosmic-comp/pull/546, panics that would previously have killed cosmic-comp now only kill the DRM thread for that surface.

This is potentially good, but maybe a bit confusing? The logs will indicate a panic, but otherwise it's unclear what happened.

Should we make sure a panic in a thread kills the whole compositor, again? Or can we do something to try to re-spawn a thread for rendering an output (and hope it doesn't repeatedly panic)? Not sure what other options there might be.

Drakulix:
I agree that the current behaviour isn't optimal.

> Should we make sure a panic in a thread kills the whole compositor, again? Or can we do something to try to re-spawn a thread for rendering an output (and hope it doesn't repeatedly panic)? Not sure what other options there might be.

Panic'ing everything seems to be a solution to make it at least visible, but that is still much more disruptive and potentially breaks applications.

Restarting might be an option, but I don't want to end up in a busy loop like situation, where we re-try continuously either. If we do that we should have some kind of counter and reset that again, once the thread has been running for some time.

Alternatively or after hitting some max value for the retry counter, we could just set the output to disabled at least in multi-output situations. If we only have one output, we might want to fall back to panic'ing, even though we explicitly added support for removing all outputs... *sigh*

danik292:
> If we do that we should have some kind of counter and reset that again, once the thread has been running for some time.

Yeah like restart it 5 times, if it won't work then panick and reset the counter after 2 mins

Issue #2149, opened 2026-03-03
cosmic-comp triggers amdgpu gfxhub page fault (UTCL2 permission fault) on Radeon 890M, causing GPU reset and full session restart


J450n-4-W:
## Summary

`cosmic-comp` repeatedly triggers GPU page faults in the amdgpu kernel driver on a Radeon 890M (Strix Halo / RDNA 3.5) iGPU. The page faults escalate into a GFX ring timeout, a failed ring reset, and ultimately a full GPU MODE2 reset. This kills `cosmic-comp` with SIGKILL (exit code 137) and takes down the entire desktop session — all running applications are lost.

This is a recurring issue. I have observed the same crash pattern at least 3 times across 2 boots:
- 2026-02-08 08:13 (boot -1)
- 2026-02-09 22:13 (boot -1)
- 2026-03-03 08:14 (current boot)

## System information

| Component | Value |
|---|---|
| **OS** | Pop!_OS 24.04 LTS |
| **Kernel** | 6.18.7-76061807-generic |
| **CPU** | AMD Ryzen 9 9955HX 16-Core (Strix Halo) |
| **RAM** | 92 GiB DDR5 |
| **GPU** | AMD Radeon 890M iGPU [1002:13c0] rev d8 (RDNA 3.5) |
| **PCI address** | 0000:01:00.0 |
| **Kernel driver** | amdgpu |
| **Mesa** | 25.1.5-1pop0~1753463422~24.04~8af185e |
| **cosmic-comp** | 0.1~1771344098~24.04~ef14b7e |
| **cosmic-session** | 1.0.0~1771011327~24.04~da592cc |
| **cosmic-panel** | 0.1.0~1771011327~24.04~6a783a1 |

## Steps to reproduce

No reliable reproducer — but it usually happens to me while using Flatpak Chromium extensions. 

### Phase 1: Page faults begin (08:14:12)

`cosmic-comp` (pid 3463, thread `cosmic-com:cs0` pid 3505) starts generating repeated gfxhub page faults from UTCL2 client TCP (0x8):

```
amdgpu 0000:01:00.0: amdgpu: [gfxhub] page fault (src_id:0 ring:24 vmid:6 pasid:32770)
amdgpu 0000:01:00.0: amdgpu:  Process cosmic-comp pid 3463 thread cosmic-com:cs0 pid 3505
[...]

PR #2487, opened 2026-06-18
fix(kms): don't panic when renderer creation fails after a GPU reset

V0UP3R:
**The problem**

On a hybrid laptop whose internal panel is driven by an AMD iGPU (Radeon 680M / Rembrandt), the amdgpu driver intermittently performs a GPU reset while COSMIC is running. The kernel log shows:

```
amdgpu …: [gfxhub] page fault … Faulty UTCL2 client ID: TCP (0x8)
amdgpu …: ring gfx_0.1.0 timeout, signaled seq=…, emitted seq=…
amdgpu …: [drm] device wedged, but recovered through reset
cosmic-session[…]: process ' … cosmic-comp ' terminated with signal 6
```

The GPU itself recovers ("recovered through reset"), but `cosmic-comp` is killed with `SIGABRT` and the whole session is dropped back to the greeter, closing every running application.

**Root cause**

In `SurfaceThreadState::redraw` (`src/backend/kms/surface/mod.rs`) the renderer is acquired with `.unwrap()`:

```rust
let mut renderer = if render_node != self.target_node {
    self.api.renderer(&render_node, &self.target_node, compositor.format()).unwrap()
} else {
    self.api.single_renderer(&self.target_node).unwrap()
};
// …
renderer = self.api.single_renderer(&self.target_node).unwrap();
```

While the DRM device is briefly lost during a GPU reset these calls fail, and the `.unwrap()` panics the surface's render thread. Because that panic unwinds across the EGL/GBM FFI boundary it aborts the entire process rather than only ending the thread — which is what turns a recoverable GPU reset into a full session loss.

**Fix**

`redraw` already returns `anyhow::Result<()>`; its caller already logs the error and calls `queue_redraw(true)`, and the frame-submission path (`render_frame` / `queue_frame`) already propagates its errors the same way. This change just propagates the renderer-acquisition errors too (`map_err(…)?`) instead of unwrapping, so a transient device loss is retried on the next frame — once the device is back — rather than taking down the compositor. It does not change behaviour when renderer acquisition succeeds.

**Testing**

- Builds cleanly against `master` (Rust 1.90).
- The change is limited to error propagation along an already-fallible path.
- I have reproduced the underlying amdgpu GPU reset on real hardware (Radeon 680M). Runtime confirmation that the surface recovers on the next frame with a locally-built compositor is still in progress; I will update the PR once verified.

Relates to #649.

---

### DIFF
--- a/src/backend/kms/surface/mod.rs
+++ b/src/backend/kms/surface/mod.rs
@@ -1006,12 +1006,16 @@ impl SurfaceThreadState {
             &self.shell.read(),
         );
 
+        // Acquiring a renderer can fail transiently when the underlying DRM
+        // device is lost (e.g. after a GPU reset).
         let mut renderer = if render_node != self.target_node {
             self.api
                 .renderer(&render_node, &self.target_node, compositor.format())
-                .unwrap()
+                .map_err(|err| anyhow::format_err!("Failed to create renderer: {:?}", err))?
         } else {
-            self.api.single_renderer(&self.target_node).unwrap()
+            self.api
+                .single_renderer(&self.target_node)
+                .map_err(|err| anyhow::format_err!("Failed to create renderer: {:?}", err))?
         };
 
         self.timings.start_render(&self.clock);
@@ -1267,7 +1271,10 @@ impl SurfaceThreadState {
                 })
                 .context("Failed to draw to offscreen render target")?;
 
-            renderer = self.api.single_renderer(&self.target_node).unwrap();
+            renderer = self
+                .api
+                .single_renderer(&self.target_node)
+                .map_err(|err| anyhow::format_err!("Failed to create renderer: {:?}", err))?;
 
             elements = postprocess_elements(
                 &mut renderer,


turidduu:
Confirming this fixes a real, recurring failure mode in the wild: on Strix Halo (gfx1151, issue #2149 signature) my kernel now recovers the amdgpu gfx ring cleanly on every hang —

```
amdgpu 0000:07:00.0: Ring gfx_0.1.0 reset succeeded
amdgpu 0000:07:00.0: [drm] device wedged, but recovered through reset
```

— yet cosmic-comp still aborts with SIGABRT one second later, taking the whole session down:

```
cosmic-session[3583]: process ' COSMIC_SESSION_SOCK=12 cosmic-comp ' terminated with signal 6
```

That's ~22 full session losses in five weeks on this machine that this PR would presumably have turned into momentary glitches. Would love to see this reviewed/merged.

src/backend/kms/surface/mod.rs
```diff
@@ -1006,12 +1006,20 @@ impl SurfaceThreadState {
             &self.shell.read(),
         );
 
+        // Acquiring a renderer can fail transiently when the underlying DRM
+        // device is lost (e.g. after a GPU reset). Propagate the error instead
+        // of unwrapping: `redraw` is fallible and its caller already logs the
+        // failure and reschedules a redraw, so we recover on the next frame
+        // once the device comes back, rather than aborting the whole
+        // compositor. See https://github.com/pop-os/cosmic-comp/issues/649
```
Drakulix: Please shorten this comment to the first sentence.

Drakulix (dismissed):
Code-wise this looks good. thanks!

jacobktm:
I've verified that this does work, I'm ready to approve this once the change @Drakulix requested is made unless that's not considered blocking, in which case let me know and I will approve.

V0UP3R:
@Drakulix shortened the comment to the first sentence as requested. Thanks for the review!

Drakulix (approved):
Thanks!

jacobktm (dismissed):

jacobktm:
whoops, I just noticed more changes were pushed after the comment change, I just want to do another quick sanity check before I approve

V0UP3R:
@jacobktm the second push was just a rebase onto current master to clear the "behind master" state, no code changes. The net diff vs master is still only the three `unwrap()` to `map_err(...)?` conversions in `surface/mod.rs` plus the shortened comment.

The failing `clippy-check` is unrelated to this PR. It's failing tree-wide on master and the other open PRs after the stable clippy bump (CI runs `cargo +stable clippy -- -D warnings`); the lints are in `input/mod.rs` and `xdg_shell/mod.rs`, not in this diff.

jacobktm (approved):

jacobktm:
I see, thanks.

Drakulix (approved):
still lgtm

turidduu:
Follow-up now running the merged version (`cosmic-comp 0.1~1785355703~24.04~091583a`): kernel-side recovery works on every hang, but the session is still lost each time. The failure mode changed shape with this patch — previously cosmic-comp SIGABRTed ~1 s after the reset; now it dies silently after `device wedged, but recovered through reset`, with no panic output in the journal, so some other path in the post-reset recovery still gives up. Six occurrences on 2026-07-30/31.

I've armed `RUST_BACKTRACE=1` and raised `kernel.core_pipe_limit` so the next occurrence should finally produce a usable backtrace/core — will post it here. (Machine details and full timeline tracked in #2149.)

PR #2489, opened 2026-06-19
Restore fullscreen surface back to stack it was part of before fullscreening

ids1024:
Seems to be working, but there are a couple things I still need to test and adjust.

### DIFF
--- a/src/shell/workspace.rs
+++ b/src/shell/workspace.rs
@@ -61,7 +61,7 @@ use wayland_backend::server::ClientId;
 use super::{
     CosmicMappedRenderElement, CosmicSurface, ResizeDirection, ResizeMode,
     element::{
-        CosmicMapped, MaximizedState, resize_indicator::ResizeIndicator,
+        CosmicMapped, CosmicMappedKey, MaximizedState, resize_indicator::ResizeIndicator,
         stack::CosmicStackRenderElement, swap_indicator::SwapIndicator,
         window::CosmicWindowRenderElement,
     },
@@ -262,6 +262,9 @@ pub enum FullscreenRestoreState {
         output: WeakOutput,
         state: FloatingRestoreData,
     },
+    Stack {
+        state: StackRestoreData,
+    },
 }
 
 impl FullscreenRestoreState {
@@ -270,6 +273,7 @@ impl FullscreenRestoreState {
             FullscreenRestoreState::Floating { state, .. }
             | FullscreenRestoreState::Sticky { state, .. } => state.was_maximized,
             FullscreenRestoreState::Tiling { state, .. } => state.was_maximized,
+            FullscreenRestoreState::Stack { .. } => false,
         }
     }
 }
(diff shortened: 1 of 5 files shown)


ids1024:
Okay, I think I've found a fairly clean way to do this that covers various different cases:

- Stacks in floating and tiling layers
- Stacks in the sticky layer
- Stacks with multiple elements, or only one
- Restoring to the index the window previously occupied in the stack

I often use Firefox in a stack, so this has been annoying me.

jklgrasso (dismissed):
Looks good

src/shell/workspace.rs
```diff
@@ -320,6 +315,12 @@ pub struct TilingRestoreData {
     pub was_maximized: bool,
 }
 
+#[derive(Debug, Clone)]
+pub struct StackRestoreData {
+    pub stack: CosmicMapped,
```
Drakulix: Should this be a weak reference? Also have you tested what happens, if you unstack the original stack? That tends to modify `CosmicMapped` somewhat in place, if I remember correctly.

Drakulix (commented):
Looks mostly good apart from a small question.

src/shell/workspace.rs
ids1024: Good point, I guess `remap_unfullscreened_window` should avoid unwrapping `stack_ref()`, at least. It can treat a `CosmicMapped` that isn't a stack the same way it treats a stack that isn't `alive()`.

I would have used a weak reference if we had a `WeakCosmicMapped`, but there isn't one. This shouldn't cause a reference cycle since it's been removed from the stack, and the `alive()` hopefully would be equivalent to checking if a weak ref upgrades. Though a weak ref would be preferable if we had one regardless.

Adding a `WeakCosmicMapped` looks a bit non-trivial with how `CosmicMapped` defined as a bunch of `Arc`s and `CosmicMappedInternal`. Maybe there could be a way to lookup a mapped from `Shell` with a `CosmicMappedKey`.

(More generally, I wonder if things like `CosmicMapped`/`CosmicWindow`/`CosmicStack`/etc. shouldn't be clonable and we should distinguish between the owned version of these things, stored in exactly one place, and an unowned handle...)

jklgrasso: I didn't check unstacking...

Only happens with tiling on:
stack, fullscreen a youtube video (used firefox), un-fullscreen, un-stack, tiling seems to break. I moved the window(s) around after the unstack failure to show how it broke.

https://github.com/user-attachments/assets/b934aae3-f539-4244-8c99-16c65462b01e

Drakulix: > Adding a WeakCosmicMapped looks a bit non-trivial with how CosmicMapped defined as a bunch of Arcs and CosmicMappedInternal. Maybe there could be a way to lookup a mapped from Shell with a CosmicMappedKey.
> 
> (More generally, I wonder if things like CosmicMapped/CosmicWindow/CosmicStack/etc. shouldn't be clonable and we should distinguish between the owned version of these things, stored in exactly one place, and an unowned handle...)

Yeah that is pretty much a reason why I added the key at some point. This would probably be a very worthwhile refactor.

ids1024:
This now stores `CosmicMappedKey` instead of the `CosmicMapped`, and searches through mapped surfaces for a matching one on restore.

Further refactoring to use `CosmicMappedKey` in more places instead of cloning `CosmicMapped` would probably be good, but isn't done here.

There seem to be some odd behaviors around the geometry of a stack when unstacked with the fullscreen window open, but that appears to occur before this change as well, and isn't related to the changes here. So it would be good to fix that, but this fix should be mergeable now.

Drakulix (approved):
LGTM!

jklgrasso (approved):

PR #2491, opened 2026-06-21
fix: unminimized window at wrong size after animation ends

mayakwd:
**Fixes**: pop-os/cosmic-epoch#2743

**Problem**
Screenshot_2026-06-19_20-25-03
After the unminimizing animation window receives the wrong size.

**Root cause**
`FloatingLayout::recalculate` reconfigures every floating window. When selecting the window geometry, it checks `mapped.is_maximized(false)`, which reflects the **committed** maximized state. Right after unminimization, the client hasn't acked the maximize yet, so the committed state is still **not maximized**. As a result, `recalculate` recomputes the geometry from the old, unmaximized size. 

The `configure` it sends still has the pending-maximized flag set, so the client receives **an unmaximized size with the maximized state** and renders it small, thinking it's maximized.

**Fix**
Use the geometry from the **pending** maximized state (`is_maximized(true)`) instead of the committed one - intended state, not the client has acked yet.

### DIFF
--- a/src/shell/layout/floating/mod.rs
+++ b/src/shell/layout/floating/mod.rs
@@ -1346,7 +1346,7 @@ impl FloatingLayout {
             mapped.set_bounds(geometry.size.as_logical());
             let prev = self.space.element_geometry(&mapped).map(RectExt::as_local);
 
-            let window_geometry = if mapped.is_maximized(false) {
+            let window_geometry = if mapped.is_maximized(true) {
                 geometry
             } else {
                 prev.map(|mut rect| {


jacobgkau (approved):
I was able to recreate https://github.com/pop-os/cosmic-epoch/issues/2743 by performing the repro steps on top of another window that's maximized. Confirmed this fixes it. Not seeing any regressions regarding maximized windows keeping their state during other times this recalculate function's used, like when the display scale changes.

Drakulix (approved):
Thanks!

PR #2500, opened 2026-06-25
fix: VRAM and memory leaks in window, session, output, and multi-GPU paths

minyek:
## Summary

Ten independent fixes to compositor-owned leaks. Each keeps GPU buffers or
compositor objects alive after the window, session, or output that needed them is
gone, so memory climbs during minimize, zoom, monitor-hotplug, multi-GPU, and
activation workflows and does not return to the idle baseline.

This is the cosmic-comp half of a two-part investigation into a steady VRAM climb
on the proprietary NVIDIA driver; the companion
[Smithay PR](https://github.com/Smithay/smithay/pull/2080) fixes library-level leaks in
the allocator and GLES renderer. The
[cache-management APIs](https://github.com/Smithay/smithay/pull/2139) are now available
in cosmic-comp's pinned Smithay, so this branch is rebased onto `master`
(`7d1daab7`) with no dependency changes. The separate Smithay update commit has
been dropped as requested.

A residual VRAM climb on NVIDIA — memory not reclaimed for sampled, cross-process
dmabuf imports — reproduces with no compositor code and is an NVIDIA driver
issue, not a bug in either PR. The driver issue has been reported
[here](https://forums.developer.nvidia.com/t/bug-report-vram-not-reclaimed-for-sampled-cross-process-dma-buf-imports/374816).

## What's fixed

1. **Minimized windows** — a client that disconnects while minimized left its
   `MinimizedWindow` holding strong references, trapping GPU textures; now dropped
   via an `IsAlive` retain in `Workspace::refresh` / `WorkspaceSet::refresh`.
2. **Cursor image cache** — cached frames accumulated across cursor-shape changes;
   now cleared when the shape changes, preserving upstream's SVG rendering and
   shake-to-find behavior. Tests cover changed and unchanged shapes.
3. **Stale Wayland activations** — `pending_activations` entries for surfaces that
   died before completing were never removed; now pruned in `Shell::refresh`.
4. **Stale X11 activations** — the retain filter always returned `true`, so entries
   for X11 windows that died before mapping accumulated; now kept only while a
   matching pending window exists.
5. **Session-lock surfaces** — removing an output left its `LockSurface` in
   `SessionLock.surfaces`; configuration disable and physical removal now share
   output teardown that removes it.
6. **Postprocess textures** — `postprocess_textures` kept a DRM node's
   `GlesTexture`s after the node was removed; now freed on GPU hot-unplug.
7. **Zoom reference cycle** — `Output → OutputZoomState → IcedElement → Output`
   kept the output alive after disconnect; broken by calling `output_leave()` on
   the zoom element before removing the output.
8. **Main-thread renderer caches** — these renderers draw only on demand, leaving
   destruction queues and cached imports holding GPU resources between draws.
   Buffer and surface destruction now schedule a batched drain through the
   existing refresh loop, replacing the periodic cleanup. Inactive sessions
   defer the drain, and failed drains remain pending for retry. Output
   configuration, scanout-flag rendering, and screenshots invalidate caches
   afterwards, including rendering-error paths. Image-copy captures drain
   destruction queues while retaining imports reused by the stream. Manager-level
   APIs also release resources cached between render and target nodes.
9. **Multi-GPU client disconnect** — the disconnect handler stopped after the
   first device it freed, leaving the client's id on other devices it used; now
   removed from every device before refreshing device usage.
10. **Surface threads** — connector removal dropped the surface without joining
    its thread, leaving its renderer, swapchain, and postprocess buffers alive;
    now joined before teardown completes.

## Testing

Both binaries were instrumented for this hunt — this compositor and its Smithay
dependency — with a GPU-resource census on a dual-output NVIDIA system. The
September 8 normal and injected-fault desktop suites passed, exercising output
disable/restore, client churn, captures, cursor and pointer behavior, cleanup
retry, and rendering-failure recovery. The tested build also contains separate
capture and sticky-window fixes, so these runs provide supporting coverage rather
than hardware validation of this exact PR tree.

Physical hot-unplug while session-locked and assisted multi-GPU validation remain
outstanding. Earlier soak results do not establish coverage of those scenarios.
The residual driver-side VRAM growth is tracked separately with the standalone
reproducer above.

At `01a3aafe`, formatting, all-feature Clippy with warnings denied, all four CI
feature configurations, and all-feature tests pass (6 passed; none failed or
ignored). The focused tests cover cleanup scheduling and cursor-cache transitions;
they are not full Wayland/KMS integration tests. The pointer-constraint fix and
its four tests are maintained on a separate branch. The branch uses master's Smithay
revision `e3d461a`. Cargo reports the existing transitive `proc-macro-error2`
future-compatibility warning.

## AI assistance

This work was developed with AI assistance (Claude Code and OpenAI Codex); use of
AI-generated code is disclosed in the commit messages per the contribution
guidelines. All changes have been reviewed and are understood by the author.

## Checklist

### DIFF
--- a/src/backend/kms/mod.rs
+++ b/src/backend/kms/mod.rs
@@ -79,6 +79,7 @@ pub struct KmsState {
 
     pub syncobj_state: Option<DrmSyncobjState>,
     pub dmabuf_global: Option<DmabufGlobal>,
+    pending_renderer_cleanup: bool,
 }
 
 pub struct KmsGuard<'a> {
@@ -137,6 +138,7 @@ pub fn init_backend(
 
         syncobj_state: None,
         dmabuf_global: None,
+        pending_renderer_cleanup: false,
     });
 
     // manually add already present gpus
@@ -698,6 +700,27 @@ impl KmsState {
         }
 
         Ok(node)
+    }
+
+    /// Request a drain of the main-thread renderers' destruction queues on the next refresh.
+    ///
+    /// Destroying a surface or buffer drops the textures these renderers imported from it,
+    /// which only *queues* the GL deletions on their contexts; the queues are flushed by
+    /// rendering or an explicit drain. Those renderers may not draw again for a long time,
+    /// so until the drain runs the dead client's buffers stay pinned in VRAM.
+    pub fn schedule_renderer_cleanup(&mut self) {
+        self.pending_renderer_cleanup = true;
+    }
+
+    /// Drain the GL destruction queues of the main-thread renderers, if scheduled.
+    pub fn run_scheduled_renderer_cleanup(&mut self) {
+        if !self.pending_renderer_cleanup || !self.session.is_active() {
+            return;
+        }
+        self.pending_renderer_cleanup = false;
+        if let Err(err) = self.api.cleanup_texture_cache() {
+            debug!(?err, "Failed to drain main-thread renderer cleanup queue");
+        }
     }
 
     pub fn schedule_render(&mut self, output: &Output) {
@@ -1285,6 +1308,25 @@ impl KmsGuard<'_> {
             }
         }
 
+        self.invalidate_renderer_caches();
+
         Ok(())
+    }
+
+    /// Drop all cached imports held by the main-thread renderers, along with the
+    /// framebuffers cached for copying between a render and a target node.
+    ///
+    /// Unlike the per-output render threads (which draw every frame), these renderers
+    /// only draw during output (re-)configuration. Between those infrequent draws their
+    /// import caches provide no benefit yet keep live clients' buffers pinned in VRAM;
+    /// the next render re-imports what it needs. Imports belonging to clients that have
+    /// already exited are released by the destruction-scheduled drain.
+    fn invalidate_renderer_caches(&mut self) {
+        if !self.session.is_active() {
+            return;
+        }
+        if let Err(err) = self.api.invalidate_caches() {
+            debug!(?err, "Failed to invalidate main-thread renderer caches");
+        }
     }
 }
--- a/src/backend/kms/mod.rs
+++ b/src/backend/kms/mod.rs
@@ -83,6 +83,34 @@ pub struct KmsState {
 
     pub syncobj_state: Option<DrmSyncobjState>,
     pub dmabuf_global: Option<DmabufGlobal>,
+    renderer_cleanup: RendererCleanupSchedule,
+}
+
+#[derive(Debug, Default)]
+struct RendererCleanupSchedule {
+    pending: bool,
+}
+
+impl RendererCleanupSchedule {
+    fn schedule(&mut self) {
+        self.pending = true;
+    }
+
+    fn run_if_active<E>(
+        &mut self,
+        session_active: bool,
+        cleanup: impl FnOnce() -> Result<(), E>,
+    ) -> Result<(), E> {
+        if !self.pending || !session_active {
+            return Ok(());
+        }
+        self.pending = false;
+        if let Err(err) = cleanup() {
+            self.pending = true;
+            return Err(err);
+        }
+        Ok(())
+    }
 }
 
 pub struct KmsGuard<'a> {
@@ -141,6 +169,7 @@ pub fn init_backend(
 
         syncobj_state: None,
         dmabuf_global: None,
+        renderer_cleanup: RendererCleanupSchedule::default(),
     });
 
     // manually add already present gpus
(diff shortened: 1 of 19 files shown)


jklgrasso (dismissed):

minyek:
Resolved the conflict with master (upstream's new KeyboardLayoutState::refresh landed at the same spot in refresh() as the KMS cache cleanup — kept both); the approval was auto-dismissed by the push, mind re-approving?

src/backend/kms/mod.rs
```diff
@@ -588,6 +595,31 @@ impl KmsState {
...
+                    if let Err(err) = device.renderer_mut().cleanup_texture_cache() {
+                        debug!(?err, "Failed to drain main-thread renderer cleanup queue");
+                    }
+                }
+            }
+            Err(err) => debug!(?err, "Failed to enumerate render devices for cleanup"),
+        }
+        self.last_renderer_cleanup = Instant::now();
+    }
+
```
Drakulix, line 622: Instead of adding a new periodic cleanup function, should we be able to simply do a cleanup (or schedule a cleanup with `loop_handle.insert_idle` after output (re-)configuration?

Drakulix (changes requested):
Looks mostly good. Just one nitpick.

src/backend/kms/mod.rs
minyek: I don't think that'd fix the issue as the queue is populated at drop time, not import time — surfaces imported during (re-)configuration only become garbage later, when the window closes or the client exits. A cleanup right after configuration would run before that garbage exists and then never again, so e.g. a client exiting an hour after the last hotplug would stay pinned in VRAM until the next mode change. Since the resources are queued at arbitrary points (window close, client exit, cache eviction inside smithay), there's no single event that covers them all — hence the throttled drain (a make_current + empty try_iter at most every 2s).

We could make this event-based with a smithay change (e.g. a notifier/ping when a cleanup queue gets work), but for this PR I decided to keep the change contained to cosmic-comp.

Drakulix: Right. I think ideally we would simply have a method on the `GraphicsApi`/`MultiRenderer` and `GlesRenderer` to invalidate all caches. We know we don't need them and cleaning them up periodically still seems quite unnecessary to me.

Would you consider making a smithay PR to add such methods?

minyek: Yes, I'm happy to make the smithay change and I agree it's a better solution.  I'm away on holiday, so when I'm back next week I'll complete my testing and submit the smithay PR. Once that's approved, I'll update this PR.

minyek:
Smithay/smithay#2139 has landed, so this branch is rebased onto current `master` (`f97a852e`) and force-pushed. The 2 s drain timer you asked me to replace is gone; three new commits take its place. The other eight commits are as reviewed, apart from formatting-only rustfmt/clippy touch-ups in two.

**1. `chore: update smithay`** bumps the pin to `347b2b3` (the revision with `Renderer::invalidate_caches` and the `GpuManager` cache methods) and adapts to the API changes in between: the `input::tablet` rewrite, and the new pointer-constraints contract — `remove_constraint` now runs with the constraint mutex held inside pointer dispatch, where applying the cursor hint deadlocks, so the hint is applied from an idle callback, and on deactivation as well as destruction per the `set_cursor_position_hint` spec.

**2. `fix(kms): drain and invalidate the main-thread renderer caches`** is the event-driven cache management you asked for: buffer/surface destruction sets a flag that `refresh` drains via `GpuManager::cleanup_texture_cache` (one batched drain even when a client exit destroys hundreds of surfaces), and infrequent draws — output reconfiguration, screenshots, image-copy captures — call `invalidate_caches` when done (captures only drain, since a screencast reuses its imports every frame). Going through the `GpuManager` methods rather than a `devices_mut` loop matters: only the manager can reach the buffers cached per render/target node pair, which screencopy's `offscreen_renderer` creates — a per-device loop left a full output-sized dmabuf per pair pinned in VRAM. That is also why the loop `f86cd933` added is dropped.

**3. `fix(kms): join the surface thread when its connector is removed`** is a one-liner: connector removal dropped the `Surface` without joining its thread, and a thread that missed the shutdown signal blocked on the compositor read lock forever, stranding its renderer, swapchain and postprocess offscreens — one full set of output-sized targets leaked per unplug.

Validation: a multi-day soak plus a scripted regression pass on an instrumented build with a GPU-resource census (27 censuses over 2.5 h, dual-output NVIDIA), covering output power-cycles, client churn, screencopy, workspace overview, and popups/zoom/minimize/cross-output moves. Every compositor-side counter returns to baseline or saturates flat, with no panic, GL error or DRM commit failure; the only residual growth is the known NVIDIA driver pooling, tracked separately with a standalone reproducer. One caveat: the cursor-hint deferral postdates the soak and is covered by manual testing of pointer-locking clients only.

Drakulix:
>  chore: update smithay bumps the pin to 347b2b3 (the revision with Renderer::invalidate_caches and the GpuManager cache methods) and adapts to the API changes in between: the input::tablet rewrite, and the new pointer-constraints contract — remove_constraint now runs with the constraint mutex held inside pointer dispatch, where applying the cursor hint deadlocks, so the hint is applied from an idle callback, and on deactivation as well as destruction per the set_cursor_position_hint spec.

We already have an update in progress that also implements tablet grabs and other features coming with the smithay update here: https://github.com/pop-os/cosmic-comp/pull/2725

I'll ping you once that is merged and the smithay version used by cosmic-comp is new enough. I'd prefer to do that myself and then have you simply drop this commit and rebase the fixes on top of it. Thanks for validating all the fixes though!

Drakulix:
Smithay has been updated on master. Please drop the update commits and rebase.

minyek:
I've updated to use the new smithay event driven cleanup, rebased and retested.

jklgrasso (approved):

PR #2501, opened 2026-06-25
backend/kms: Use udmabuf for multi-gpu buffer sharing

Drakulix:
Draft because I am not sure we want this and due to very limited testing.

But I think this is useful for debugging and maybe also generally at some point, so I'll throw it into a PR for visibility and later use.

Uses a udmabuf to share contents between gpus, forcing the buffer to be linear (unfortunately), but avoids any weird migrations by immediately allocating in RAM. Needs a very recent nvidia driver to render correctly (595) and some Intel cards also have some rendering issues, but also works (without crashing) with llvmpipe and pixman! Not scanout-able, but otherwise just a plain better version of dumb buffers.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,7 +74,7 @@ xdg-user = "0.2.1"
 xkbcommon = "0.9"
 zbus = "5.14.0"
 profiling = { version = "1.0" }
-rustix = { version = "1.1.4", features = ["process"] }
+rustix = { version = "1.1.4", features = ["process", "param"] }
 rand = "0.10"
 # CLI arguments
 clap_lex = "1.0"
--- a/src/backend/kms/render/gles.rs
+++ b/src/backend/kms/render/gles.rs
@@ -3,13 +3,13 @@
 use smithay::backend::{
     SwapBuffersError,
     allocator::{
-        Allocator,
+        Allocator, Format, Fourcc, Modifier,
         dmabuf::{AnyError, Dmabuf, DmabufAllocator},
         gbm::GbmAllocator,
     },
     drm::{CreateDrmNodeError, DrmNode},
     renderer::{
-        RendererSuper,
+        Bind, RendererSuper,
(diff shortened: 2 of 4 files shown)


ids1024:
It would be good to add the `UdmabufAllocator` to smithay, even if we're not sure if any compositors want to use it by default currently. Udmabuf may also be useful for `wl_shm` (https://github.com/Smithay/smithay/issues/2044).

So this behaves differently from a linear iGPU buffer? Superficially it seems like it should be the same given that both would be the same layout in system memory. Though maybe it has some impact of cache coherence behavior or such; and may ends up being handled different by the driver in certain ways.

Drakulix:
> It would be good to add the `UdmabufAllocator` to smithay, even if we're not sure if any compositors want to use it by default currently. Udmabuf may also be useful for `wl_shm` ([Smithay/smithay#2044](https://github.com/Smithay/smithay/issues/2044)).

Done here: https://github.com/Smithay/smithay/pull/2084

> So this behaves differently from a linear iGPU buffer? Superficially it seems like it should be the same given that both would be the same layout in system memory. Though maybe it has some impact of cache coherence behavior or such; and may ends up being handled different by the driver in certain ways.

Yeah I think the main different is, that this will also go through the dmabuf-import code of any iGPU driver. So while it *should* be the same, this can use different code paths.

E.g. it also doesn't depend on the iGPU driver playing nice on dmabuf-export and e.g. let us "map" the buffer to cpu-memory, so this is also guaranteed to be importable by (recent) versions of llvmpipe.

ids1024:
It looks like 26.04 has a line in `/usr/lib/udev/rules.d/70-uaccess.rules` with a rule to provide access to `/dev/udmabuf` that 24.04 doesn't.

Easy enough to change the permissions for testing, but if we want to actually use udmabuf for anything, we'll want to update https://github.com/pop-os/systemd to have that on 24.04.

Issue #2384, opened 2026-05-16
Rapid hover over dock icons renders garbage in tooltips, occasionally causes hard display freeze on amdgpu (Renoir)


diamondminer857:
Yesterday i did an apt upgrade on Pop!_OS 24.04 (which bumped cosmic-applets from 1.0.2 to 1.0.12 along with cosmic-panel, cosmic-app-library, cosmic-comp and others) I started getting display freezes that cannot be recovered from. TTY switch (Ctrl+Alt+F2-F6) doesn't respond during the freeze, audio keeps playing, caps lock LED toggles, but the screen is stuck.
The trigger seems to be hovering over dock icons. Specifically:

Hovering a single icon slowly is fine
Quickly moving the cursor across 3+ icons (left-to-right or right-to-left) makes the dock flicker and I sometimes see what looks like uninitialized memory rendered into the app tray area (small squares of random colors, very brief, few ms)
Doing the hover over some icons (specifically COSMIC text editor) triggers a full freeze.

When the freeze happens, the kernel logs amdgpu [drm] *ERROR* [CRTC:86:crtc-0] flip_done timed out followed by commit wait timeouts on multiple planes, and a WARNING in amdgpu_dm_commit_planes. The driver then attempts a recovery (SMU resume, ring re-init) but it doesn't restore display output.
The first time (while I was typing up another issue on GitHub, ironic), it froze on hover over the CosmicEdit icon, which made me think it was specific to that app, but the flickering shows it's not the CosmicEdit itself. If I try pinning new apps, they also do the same thing.

Setup:
Pop!_OS 24.04 LTS, COSMIC alpha, Wayland session
Kernel 6.18.7-76061807-generic (System76)
ThinkPad T14 Gen 1, Ryzen 5 4650U (Renoir iGPU, DCN 2.1), eDP-1 panel
Mesa 25.2.8
cosmic-applets 1.0.12, cosmic-panel 0.1.0-1778807001, cosmic-comp 0.1-1778668982, cosmic-app-library 1.0.12

Relevant journalctl log from the first freeze:

> amdgpu 0000:07:00.0: [drm] *ERROR* [CRTC:86:crtc-0] flip_done timed out
amdgpu 0000:07:00.0: [drm] *ERROR* flip_done timed out
amdgpu 0000:07:00.0: [drm] *ERROR* [PLANE:59:plane-3] commit wait timed out
[...]

PR #2504, opened 2026-06-26
kms: gate overlay scanout per-element by buffer GPU node

Ericky14:

This is aim to fix a similar issue as this one: https://github.com/pop-os/cosmic-comp/issues/2384

On multi-GPU / hybrid-graphics setups I was getting texture corruption and a flicker-on-hover that only shows on the physical display. It comes down to surfaces being promoted to a DRM overlay plane for direct scanout. On multi-GPU the client buffer lives on the render GPU but the plane is on the display GPU, so it ends up scanning out a foreign buffer instead of going through the cross-GPU blit.

This PR decides scanout candidacy per surface instead of per output: a surface is only flagged as a DRM scanout candidate (Kind::ScanoutCandidate) when its buffer actually lives on the output's scan-out GPU (target_node). Single-GPU is unaffected, and surfaces that legitimately render on the display GPU keep their overlay scanout even when we composite on another GPU.

### DIFF
--- a/src/backend/kms/surface/mod.rs
+++ b/src/backend/kms/surface/mod.rs
@@ -8,7 +8,7 @@ use crate::{
         init_shaders, output_elements,
     },
     config::ScreenFilter,
-    shell::Shell,
+    shell::{Shell, element::surface::ScanoutTargetNodeGuard},
     state::SurfaceDmabufFeedback,
     utils::prelude::*,
     wayland::handlers::{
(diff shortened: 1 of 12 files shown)


src/backend/kms/surface/mod.rs
```diff
@@ -1046,11 +1046,18 @@ impl SurfaceThreadState {
             }
         };
 
-        if has_active_fullscreen || animations_going {
+        if has_active_fullscreen || animations_going || render_node != self.target_node {
```
Drakulix: Hmm, these checks aren't entirely correct.

We can have clients originating from different gpus on a single output. Just because we composite on one gpu, doesn't mean we don't have buffers from the other, that we couldn't put on an overlay plane.

The correct solution would be to make sure render-elements generated from clients on different gpus don't get the `ScanoutCandidate`-flag.

Ericky14: You're correct, I didn't account for that.. 

I updated the PR with a different fix which does fix the issue on my end.
I used thread local to avoid having to pass DrmNode through a bunch of functions.

Drakulix: Thanks, this already looks much better.

> I used thread local to avoid having to pass DrmNode through a bunch of functions.

But I personally don't like this. This introduces more not immediately obvious side-effects and we already pass the drm-node to `output_elements`. The sane thing to do here in my opinion is to just pass it through the whole element-creation.

Ericky14: Updated it to remove thread local

Drakulix (approved):
Thanks!

jacobgkau (approved):
I wasn't able to recreate the issue (tried on an AMDGPU integrated system, and two Intel/NVIDIA hybrid systems since this PR talked about hybrid, but I don't think we have an AMD/NVIDIA or Intel/AMD hybrid system handy for testing).

I took a look at this on the two hybrid (Intel + NVIDIA) systems, the AMD-only system, and an NVIDIA-only system, and am not seeing any problems on any of them.

PR #2508, opened 2026-06-28
fix: internal laptop screen stuck on vendor logo with external monitor

AlexDaichendt:

This PR fixes this issue https://github.com/pop-os/cosmic-epoch/issues/2933. In summary, when you disable a laptops screen in the display configuration menu due to having external monitors connected, on boot, the laptop screen will retain the last frame buffer. This effectively means, that the laptop screen, although disabled in the COSMIC settings, is active and holding a frame forever. This is especially bad for OLED screens due to burn in risk. 
The workaround so far was, on every login open the settings and turn the laptop screen on and off. 

LLM disclosure: Claude Opus came up with the initial idea, but I had to refactor it due to unnecessary code duplication and the usual excessive comments. I can reason about the changes. 

What is this MR changing:
- extracted the clean up logic for disabling CRTs that was already present in the `display_configuration` method into a new function `disable_crts` (that's the largest part of the diff really) 
- use the `disable_crts` function in `apply_config_for_outputs` - after finding out which monitors should be powered down, actually power them down (this part missing caused the bug).
- Regarding the change `if !is_primary && !cleanup.contains(&crtc) {`: clean up is done now in the separate function. to keep responsabilities where they should, this no longer cleans up

I tested: 
- the issue is 100% reproducible on main 
- my MR fixes the issue 
- my MR does not break turning on/off screens from the settings menu or starting with a different permutation of display config.

Awesome work btw, I'm daily driving this DE for almost 2 years now and its shaped up so nice :)

### DIFF
--- a/src/backend/kms/drm_helpers.rs
+++ b/src/backend/kms/drm_helpers.rs
@@ -3,6 +3,7 @@
 use anyhow::{Context, Result, anyhow};
 use libdisplay_info::{edid::DisplayDescriptorTag, info::Info};
 use smithay::{
+    backend::drm::DrmDevice,
     reexports::drm::control::{
         AtomicCommitFlags, Device as ControlDevice, Mode, ModeFlags, PlaneType, ResourceHandle,
         atomic::AtomicModeReq,
@@ -16,8 +17,7 @@ use smithay::{
 use std::{collections::HashMap, ops::Range};
 
 pub fn display_configuration(
-    device: &mut impl ControlDevice,
-    supports_atomic: bool,
+    device: &mut DrmDevice,
 ) -> Result<HashMap<connector::Handle, Option<crtc::Handle>>> {
     let res_handles = device.resource_handles()?;
     let connectors = res_handles.connectors();
@@ -73,7 +73,7 @@ pub fn display_configuration(
     }
 
     // And then cleanup
-    if supports_atomic {
+    if device.is_atomic() {
         let mut req = AtomicModeReq::new();
         let plane_handles = device.plane_handles()?;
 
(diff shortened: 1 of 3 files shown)


src/backend/kms/mod.rs
```diff
@@ -852,11 +852,26 @@ impl KmsGuard<'_> {
...
+                        .any(|o| !o.is_enabled() && surface.output == *o)
+                    {
+                        disabled_crtcs.push(*crtc);
+                        false
+                    } else {
+                        true
+                    }
+                });
+
+                let supports_atomic = device.drm.device().is_atomic();
```
Drakulix: You are passing the device to `disable_crtc` anyway. Why not query this inside the function and remove the argument?

src/backend/kms/drm_helpers.rs
```diff
@@ -77,24 +77,6 @@ pub fn display_configuration(
         let mut req = AtomicModeReq::new();
```
Drakulix, line 77: We need to be careful here, why I like the code-deduplication with the new `disable_crtcs` function, this also means we split the cleanup into two atomic commits.

This means the intermediate state between the two requests, needs to be valid. However if I am seeing this correctly, all this request does now is disable some overlay/cursor planes. correct? That should always be valid.

Drakulix (commented):
Thanks, looks mostly good to me. Just a couple of questions.

src/backend/kms/drm_helpers.rs
AlexDaichendt: The first commit removes overlay/cursor planes, but keeps the primary plane - for the screens that remain active. 
The 2nd commit tears down all resources - for screens that should be off

This should be fine. there is a performance tradeoff with 2 commits, but I have no idea (or experience with kms) to say how large that is. considering its not in a hot path, its probably fine?

src/backend/kms/mod.rs
AlexDaichendt: I went with the same type as `display_configuration` does (`&mut impl ControlDevice`) - that does not offer a `is_atomic()` fn. We could change the signature of `disable_crts` to `&mut DrmDevice` instead. 

It seems like once we do that and remove the now unused variable `supports_atomic` from `device.rs`, we can shrink the diff by 10 lines. I've pushed that commit now. If you are unhappy with it, I'll gladly revert.

Drakulix (approved):
Looks good to me now. Thanks!

jklgrasso:
This needs to be rebased. Otherwise, this looks good

AlexDaichendt:
> This needs to be rebased. Otherwise, this looks good

Done! Thank you both for the swift review! Really happy to get this merged. Will you close this issue then? https://github.com/pop-os/cosmic-epoch/issues/2933

Drakulix (approved):

jklgrasso (approved):

PR #2510, opened 2026-06-29
quirk for newer Intel iGPUs

Drakulix:
Same fix as https://github.com/pop-os/cosmic-workspaces-epoch/pull/218, but on the compositor level.

Can fix flickering issues, when buffers are copied from an Intel iGPU to any other GPU. (nVidia and AMD are likewise affected.)

Based on https://github.com/pop-os/cosmic-comp/pull/2509 and draft while waiting for https://github.com/Smithay/smithay/pull/2083.

### DIFF
--- a/src/backend/kms/render/gles.rs
+++ b/src/backend/kms/render/gles.rs
@@ -1,21 +1,30 @@
 // SPDX-License-Identifier: GPL-3.0-only
 
-use smithay::backend::{
-    SwapBuffersError,
-    allocator::{
-        Allocator,
-        dmabuf::{AnyError, Dmabuf, DmabufAllocator},
-        gbm::GbmAllocator,
+use clap_lex::OsStrExt;
+use rustix::path::Arg;
+use smithay::{
+    backend::{
+        SwapBuffersError,
+        allocator::{
+            Allocator,
+            dmabuf::{AnyError, Dmabuf, DmabufAllocator},
+            gbm::GbmAllocator,
+        },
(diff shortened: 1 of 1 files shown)


src/backend/kms/render/gles.rs
```diff
@@ -123,11 +132,26 @@ impl<A: AsFd + Clone + 'static> GraphicsApi for GbmGlowBackend<A> {
                     .any(|renderer| renderer.node.dev_id() == node.dev_id())
             })
             .flat_map(|(node, (allocator, renderer))| {
-                let renderer = renderer.replace(None)?;
+                let mut renderer = renderer.replace(None)?;
+                let is_intel = allocator
+                    .as_ref()
+                    .get_driver()
+                    .is_ok_and(|drv| drv.name().contains("i915"));
```
ids1024, line 139: So we don't need this workaround with the `xe` kernel module?

ids1024 (dismissed):
The implementation seems good if we need this workaround. The only issue would be the performance impact of the change (obviously), though if we need this to avoid visual corruption, that takes priority.

src/backend/kms/render/gles.rs
Drakulix: Not in my testing, but it causes different corruption (like in the cosmic-apps context menus) independently of the gpu that is doing the composition. So we should probably still consider the driver to be experimental.

ids1024 (approved):

leviport (approved):

PR #2513, opened 2026-06-30
Dmabuf v6

Drakulix:
Draft for https://github.com/Smithay/smithay/pull/1430

### DIFF
--- a/src/backend/kms/device.rs
+++ b/src/backend/kms/device.rs
@@ -59,7 +59,7 @@ use std::{
     time::Duration,
 };
 
-use super::{drm_helpers, socket::Socket, surface::Surface};
+use super::{drm_helpers, surface::Surface};
 
 #[derive(Debug)]
 pub struct EGLInternals {
@@ -98,14 +98,12 @@ pub struct Device {
 
     pub texture_formats: FormatSet,
     event_token: Option<RegistrationToken>,
-    pub socket: Option<Socket>,
 }
 
 #[derive(Debug)]
 struct ReusableDevice {
     leasing_global: Option<DrmLeaseState>,
     active_clients: HashSet<ClientId>,
-    socket: Option<Socket>,
 }
 
 #[derive(Debug)]
@@ -593,7 +591,7 @@ impl State {
         }
     }
 
-    pub fn device_removed(&mut self, dev: dev_t, dh: &DisplayHandle) -> Result<()> {
+    pub fn device_removed(&mut self, dev: dev_t, _dh: &DisplayHandle) -> Result<()> {
         let backend = self.backend.kms();
         // we can't use DrmNode::from_node_id, because that assumes the node is still on sysfs
         let drm_node = backend
@@ -615,13 +613,6 @@ impl State {
             }
             if let Some(token) = device.event_token.take() {
                 self.common.event_loop_handle.remove(token);
-            }
-            if let Some(socket) = device.socket.take() {
-                self.common.event_loop_handle.remove(socket.token);
-                self.common
-                    .dmabuf_state
-                    .destroy_global::<State>(dh, socket.dmabuf_global);
-                dh.remove_global::<State>(socket.drm_global);
             }
             backend.api.as_mut().remove_node(&device.inner.render_node);
             backend
@@ -755,23 +746,7 @@ impl Device {
         let ReusableDevice {
             leasing_global,
             active_clients,
-            socket,
         } = reuse.unwrap_or_else(|| {
-            let socket = match (!is_software)
-                .then(|| common.create_socket(dh, render_node, texture_formats.clone()))
-                .transpose()
-            {
-                Ok(socket) => socket,
-                Err(err) => {
-                    warn!(
-                        ?err,
-                        "Failed to initialize hardware-acceleration for clients on {}.",
-                        render_node,
-                    );
-                    None
-                }
-            };
-
             let leasing_global = match (!is_software)
                 .then(|| DrmLeaseState::new::<State>(dh, &dev_node))
                 .transpose()
@@ -790,7 +765,6 @@ impl Device {
             ReusableDevice {
                 leasing_global,
                 active_clients: HashSet::new(),
-                socket,
             }
         });
 
@@ -831,7 +805,6 @@ impl Device {
 
             texture_formats,
             event_token: Some(token),
-            socket,
         })
     }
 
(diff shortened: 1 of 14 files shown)


src/backend/kms/device.rs
```diff
@@ -99,14 +99,12 @@ pub struct Device {
     supports_atomic: bool,
     pub texture_formats: FormatSet,
     event_token: Option<RegistrationToken>,
-    pub socket: Option<Socket>,
```
Drakulix, line 101: Since there is no advantage with dmabuf-v6 anymore, we can drop our special render-node-wayland-sockets.

src/backend/kms/mod.rs
```diff
@@ -465,87 +474,189 @@ impl State {
...
-            if let Some(old_state) = self.syncobj_state.take() {
+            if let Some(old_state) = kms.syncobj_state.take() {
                 dh.remove_global::<State>(old_state.into_global());
             }
         }
 
         Ok(())
     }
 
+    fn update_default_feedback(&mut self) {
```
Drakulix, line 596: Since we only have one global for all devices now, we need to update the feedback from time to time. This is usually not disruptive, except if a device goes away.

Most apps relying on WSI will simply die from either the GPU disappearing or trying to import their next dmabuf, since mesa does not and cannot update the used gpu at runtime. But that isn't any different from before, we just need to make sure to keep the global valid so all apps using different gpus stay functional.

(If we change the primary gpu, we also need to update the device `wl_drm` uses... can we just drop that global at this point? Due to this being a private-protocol, no consumer - mesa - expected to ever receive a second `device`-event. As such we need to re-create the global.)

src/backend/kms/mod.rs
```diff
@@ -465,87 +474,189 @@ impl State {
...
+            for dev in kms
+                .drm_devices
+                .values()
+                .filter(|dev| dev.inner.render_node != primary_node)
+            {
+                feedback = feedback.add_preference_tranche(
+                    dev.inner.render_node.dev_id(),
+                    TrancheFlags::Sampling,
+                    dev.texture_formats.iter().cloned(),
+                );
```
Drakulix, line 620: TODO: the current smithay code would send these tranches for v5 clients as well. We didn't used to do this, so we probably should skip these instead to not trigger any bugs?

src/backend/kms/surface/mod.rs
```diff
@@ -1539,60 +1539,37 @@ fn get_surface_dmabuf_feedback(
...
-    // as e.g. the nvidia driver might then send us dmabufs, that makes e.g. the iris hangs on import...
-    if target_node != render_node.dev_id() && !combined_formats.is_empty() {
+    if target_node != render_node {
         builder = builder.add_preference_tranche(
             target_node.dev_id(),
-            Some(zwp_linux_dmabuf_feedback_v1::TrancheFlags::Scanout),
-            combined_formats,
+            zwp_linux_dmabuf_feedback_v1::TrancheFlags::Sampling,
+            target_formats,
         );
```
Drakulix, line 1552: This is to make sure we don't stop advertising the sampling tranche in the surface-feedback, that the client likely choose based on the default-feedback. That would be kinda weird.

Actually we might want to advertise the target formats first.

Drakulix (commented):
Some comments/open questions on the implementation so far.

ids1024:
On my AMD+Nvidia laptop, for **some** reason this seems to cause a SIGSEGV in `libnvidia-egl-wayland2` when calling `eglGetPlatformDisplay`; for instance in a minimal wgpu test client, or in cosmic-panel:

```
Thread 1 "fragment" received signal SIGSEGV, Segmentation fault.
0x00007ffff7cadd75 in __GI___libc_free (mem=0x4) at ./malloc/malloc.c:3375
warning: 3375	./malloc/malloc.c: No such file or directory
(gdb) bt
#0  0x00007ffff7cadd75 in __GI___libc_free (mem=0x4) at ./malloc/malloc.c:3375
#1  0x00007ffff7faeaf9 in wl_array_release () from /lib/x86_64-linux-gnu/libwayland-client.so.0
#2  0x00007ffff687b732 in ?? () from /lib/x86_64-linux-gnu/libnvidia-egl-wayland2.so.1
#3  0x00007ffff6879f8a in ?? () from /lib/x86_64-linux-gnu/libnvidia-egl-wayland2.so.1
#4  0x00007ffff687b08b in ?? () from /lib/x86_64-linux-gnu/libnvidia-egl-wayland2.so.1
#5  0x00007ffff6878261 in ?? () from /lib/x86_64-linux-gnu/libnvidia-egl-wayland2.so.1
#6  0x00007fffd9cefbc9 in ?? () from /lib/x86_64-linux-gnu/libEGL_nvidia.so.0
#7  0x00007fffd9c8a50f in ?? () from /lib/x86_64-linux-gnu/libEGL_nvidia.so.0
#8  0x00007ffff7e48145 in ?? () from /lib/x86_64-linux-gnu/libEGL.so.1
#9  0x00005555566755ce in khronos_egl::{impl#140}::eglGetPlatformDisplay<libloading::safe::Library> (
    self=0x55555711fab0, platform=12760, native_display=0x555556cdb3c0, attrib_list=0x7fffffff8c78)
    at /home/ian/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/khronos-egl-6.0.0/src/lib.rs:2333
```

src/backend/kms/mod.rs
Drakulix: Skipped with the new version range api for older clients.

src/backend/kms/surface/mod.rs
Drakulix: Skipped for old clients now.

Drakulix:
> On my AMD+Nvidia laptop, for **some** reason this seems to cause a SIGSEGV in `libnvidia-egl-wayland2` when calling `eglGetPlatformDisplay`

The last version only sends the current tranches anymore for old clients. I verified that the output of wayland-info regarding `wp_linux_dmabuf_v1` is the same and that the crash observed doesn't happen anymore.

Also tested https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/33344 against this and it behaves as expected.

I think this is safe to merge. If any client misuses v6 we can still stop advertising it in the meantime, but I think this is a good testing ground.

ids1024 (approved):
As stated, I see this advertise an additional tranche on v6 but not v5. The changes here look good.

There's probably more we can do with deciding what tranches to advertise, in what order, when to update, etc. but the changes here seem like a safe start. We can experiment more with these things alongside testing client code (Mesa and Nvidia) that uses v6.

leviport (approved):

PR #2527, opened 2026-07-06
fix: fall back to legacy X11 names when loading pointer themes

maskedsyntax:
## Summary
- Add an alias table mapping modern shape-v1 names to legacy X11 names used by themes such as DMZ-White
- Update icon loading to try each alias (then the original name) before failing

## Test plan
- [ ] Set `XCURSOR_THEME=DMZ-White` and verify default, resize, and pointer shapes display correctly
- [ ] Confirm themes that already ship modern names (e.g. Adwaita, Bibata) are unaffected
- [ ] Exercise edge and diagonal resize shapes over window borders

### DIFF
--- a/src/backend/render/cursor.rs
+++ b/src/backend/render/cursor.rs
@@ -114,18 +114,57 @@ enum Error {
     NoDefaultCursor,
     #[error("Error opening xcursor file: {0}")]
     File(#[from] std::io::Error),
-    #[error("Failed to parse XCursor file")]
-    Parse,
+}
+
+fn cursor_aliases(name: &str) -> &[&str] {
+    match name {
+        "default" => &["default", "left_ptr", "arrow"],
+        "pointer" => &["pointer", "hand2", "hand"],
+        "text" => &["text", "xterm"],
+        "wait" => &["wait", "watch"],
+        "progress" => &["progress", "left_ptr_watch"],
+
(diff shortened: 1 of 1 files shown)


Drakulix:
So this is mostly to help old potentially unmaintained cursor-themes? Not sure if I am particularly in favor of that.

maskedsyntax:
That’s fair. The main motivation isn’t to preserve abandoned themes for their own sake, but to improve compatibility with themes that are still widely used (e.g. Adwaita, Bibata and other X11-era themes) without requiring users to patch or rename them manually.

The aliases are only used as a fallback when the modern shape name isn’t found, so themes that already ship the current names remain completely unaffected. This keeps existing behavior unchanged while making older themes “just work.”

Drakulix (approved):
Well code-wise this looks good and it's not much more difficult to maintain. So from an engineering perspective this is fine. Leaving it up to @pop-os/quality-assurance to find and test any cursor-themes this helps with.

jacobgkau (approved):
Confirmed that when `XCURSOR_THEME` is set to `DMZ-White`, this gets the cursor theme to load and display all of its variants as expected, where it used to be stuck on a black fallback cursor with no variants.

No problems with our default cursor theme. (Side note, it appears installing the `dmz-cursor-theme` package overrides our cursor theme with our default config/no environment variables, so simply installing it broke the cursor before and takes effect with that cursor now.)

Issue #1420, opened 2025-05-17
[suggestions] thoughts on better accessibility features


newsgrep:
I really much appreciate that you introduced accessibility functions in the alpha version of the desktop environment. I would like to share a few thoughts about farther developing accessibility for COSMIC. I used *** to mark those that IMHO are specaly important.

1. *** Improve the zoom to be less shaky as I described here: https://github.com/pop-os/cosmic-epoch/issues/1883 and here: https://github.com/pop-os/cosmic-comp/issues/1237#issuecomment-2844794568 
2. *** COSMIC is the only desktop I know where zooming into one screen does not disturb the content on the other screen. The zoom provided by other desktops does not work well with dual display setups. Given that background I would like to suggest adding a magnifier function additionally to the zoom. Meaning when we have a setup with two or more displays the whole screen of the primary display would be used to show the magnification of the display that contains the mouse pointer. Meaning that if you have 3 Displays and switch to magnifier mode the system behaves like you have only two screens to place your windows because the primary screen is only used for zooming the content of the other two screens. That way it would be possible for a visually impaired person and a healthy person to work together on one workstation. If you have only one screen the magnifier mode would  a) not work b) introduce a magnifying glassbar on the top of your screen c) produce a magnifying glass surrounding the mouse pointer. 
3. Invert - Inverts the color of the desktop and windows; activated with a keyboard shortcut, similar to here: https://support.microsoft.com/images/en-us/131609e1-de2b-54c4-3d58-da784235b635
4. Track Mouse - Display a mouse curser locating effect; activated with a keyboard shortcut
5. *** Mouse Mark - Allows you to draw lines on the desktop; activated with a keyboard shortcut: very use full for presentations or temporal highlights of screen content
6. *** Invert Mouse Pointer: turn the mouse pointer white when moving over a dark surface, see picture 3 here: https://support.microsoft.com/images/en-us/27886b44-6dbf-457c-b455-733af070dfa0 
7. Focus highlight: draw a red rectangle around the elements that are in focus, see the pictures here for an example: https://support.freedomscientific.com/Downloads/ZoomText/ZoomTextWhatsNew
8. Text cursor highlight: Add some highlighting to make the text cursor more visible, especially when it is on the edge of a Fullscreen window, maybe by rendering a small triangle on the top and bottom of the curser
9. adjust size of mouse pointer
10. adjust color of mouse pointer
[...]

PR #2533, opened 2026-07-08
fix: a11y: magnifier: correct focal-point math and update focal point on zoom-in

### LINKED ISSUE #1503: Magnifier breaks at >2x zoom when Cursor Tracking is set to Center — cursor no longer centered, screen reframes incorrectly

### LINKED ISSUE #2511: Accessibility: Magnifier uses stale cursor position (focal point) when zooming in after zooming out
## Summary

When zooming in after zooming out completely, the magnifier reuses the cursor position ("focal point") from the last active zoom instead of the current pointer location.

## Expected behavior

As a low vision user, after fully zooming out, when I zoom in again with the magnifier, I expect it to zoom in to the area of the screen containing the mouse cursor (varies slightly with the **Zoomed view moves** setting). (This is the behavior on macOS, Windows, and other Linux desktops.)

## Actual behavior

After fully zooming out, when I zoom in again, the magnifier zooms in to the **previous cursor position** when I was last zoomed in.

Please see attached screen recording.

This affects all three "Zoomed view moves" modes:
- Continuously with pointer
- When pointer reaches edge
- To keep pointer centered

## Accessibility UX impact

1. The magnifier zooms in to a completely different area of the screen than expected.
2. After each initial zoom in from fully zoomed out, the screen "jumps" to the new cursor location as soon as it is moved.

## Steps to reproduce

1. Enable the magnifier and zoom in to corner A of the screen.
2. Without moving the cursor, zoom out completely.
3. Move the cursor to corner B of the screen.
4. Being careful not to move the cursor yet, use either the keyboard shortcut or the modifier scroll gesture to zoom in.
5. Note the magnifier zooms in to corner A, not corner B.

## Root cause

`update_focal_point` is only called when zoom level > 1.0, so the focal point becomes stale while zoomed out.

https://github.com/user-attachments/assets/37e0d0c2-4a6f-40c6-a537-68d7d9847cae

### LINKED ISSUE #2683: Accessibility - Magnifier | When zooming in dock disapears
### Issue
When zooming in at above 300+% the dock disapears and becomes inaccessible

## Steps to reproduce
1. Accessibility -> Magnifier
2. Zoom at above 300/400%  levels
3. pull mouse cursor down to where the dock should be
4. notice the dock disapears 

Related issue:
1. Move mouse to top panel to top right
2. applets would disapear into the right edge of the screen out of view outside what  the monitor can display
3. mouse cursor could move outside but the applets even opened do not show up fuly on view

System configuration
POP_OS: 24.04 Latest 
No external GPU

tgrushka:
Happy [Disability Pride Month](https://thearc.org/blog/why-and-how-to-celebrate-disability-pride-month/)!

I am celebrating this occasion as a legally blind contributor to help improve the magnifier for the COSMIC user community.

Feedback and suggestions for revisions welcome.

## Summary

This PR corrects the focal-point math in `ZoomMovement::Centered` and `OnEdge` (which broke and caused jumping at high zoom levels) and refreshes the stale focal point on zoom in.

## Use of AI / references

- I used AI to assist in research of the focal point computation formulae for OnEdge and Centered modes.

- I co-wrote the final formulas with AI assistance and my corrections. I tested the formulas.

- I used the formulas found in [KDE's zoom magnifier - link directly to zoom.cpp](https://invent.kde.org/plasma/kwin/-/blob/master/src/plugins/zoom/zoom.cpp) and converted them to Rust and the COSMIC focal point system.

## Changes

- Remove unnecessary focal point precomputation in `OutputZoomState::new`, because it is now updated on first zoom in anyway.

- Do focal point computation in `Local` coordinates in `update_focal_point` to eliminate local-global conversion confusion / issues.

- `ZoomState::update_focal_point`

  - `ZoomMovement::OnEdge` branch - single code path rewrite - old two-branch logic caused jumping and unpredictability at high zoom. Fixes [#2683](https://github.com/pop-os/cosmic-epoch/issues/2683), partially addresses [#2688](https://github.com/pop-os/cosmic-epoch/issues/2688).

  - `ZoomMovement::Centered` branch - translation/clamp/invert - prevents "blow up" of calculation at zoom level > 2.0 that caused cursor to disappear off bottom right of screen. Fixes [#1503](https://github.com/pop-os/cosmic-comp/issues/1503).

  - `is_level_change` - sets `previous_point` so that `animating_focal_point` interpolates instead of jumping. Still room for improvement (needs focal point recomputation per-frame during animation to prevent "boomerang" feel), but this is still an improvement over jumping.

- In `Shell::trigger_zoom`, call `ZoomState::update_focal_point` when `level > 1.0` - overlaps proposal in [#1859](https://github.com/pop-os/cosmic-comp/pull/1859) + the level gate to prevent stale focal point on zoom in. Fixes [#2511](https://github.com/pop-os/cosmic-comp/issues/2511) and [#1883](https://github.com/pop-os/cosmic-epoch/issues/1883). Partially addresses [#1420](https://github.com/pop-os/cosmic-comp/issues/1420) (item 1 only).

### DIFF
--- a/src/shell/mod.rs
+++ b/src/shell/mod.rs
@@ -2471,6 +2471,14 @@ impl Shell {
             increment: zoom_config.increment,
             movement: zoom_config.view_moves,
         });
+
+        if level > 1.0 {
+            self.update_focal_point(
+                seat,
+                seat.get_pointer().unwrap().current_location().as_global(),
+                zoom_config.view_moves,
+            );
+        }
     }
 
     pub fn update_focal_point(
--- a/src/shell/mod.rs
+++ b/src/shell/mod.rs
@@ -2472,13 +2472,11 @@ impl Shell {
             movement: zoom_config.view_moves,
         });
 
-        if level > 1.0 {
-            self.update_focal_point(
-                seat,
-                seat.get_pointer().unwrap().current_location().as_global(),
-                zoom_config.view_moves,
-            );
-        }
+        self.update_focal_point(
+            seat,
+            seat.get_pointer().unwrap().current_location().as_global(),
+            zoom_config.view_moves,
+        );
     }
 
     pub fn update_focal_point(
--- a/src/shell/zoom.rs
+++ b/src/shell/zoom.rs
@@ -257,11 +257,13 @@ impl ZoomState {
             }
             ZoomMovement::OnEdge => {
                 if is_level_change {
+                    // Ensure the cursor doesn't disappear off screen
                     output_state_ref.focal_point = cursor_position;
                     return;
                 }
 
                 // Compute small margin relative to zoomed output to keep cursor within
+                // (can be user-configurable in the future)
                 let margin_size = zoomed_output_geometry.size.h * 0.02;
                 let margins = FrameExtents::new(margin_size, margin_size, margin_size, margin_size);
                 let inner_rect = zoomed_output_geometry - margins;
@@ -305,7 +307,7 @@ impl ZoomState {
                 let center = (output_geometry.size / 2.).to_point();
 
                 if level <= 1.0 + f64::EPSILON {
-                    // Without this break, focal point will jump to (0, 0) on zoom out
+                    // Prevent focal point jumping to top-left corner (0, 0) on zoom out
                     output_state_ref.focal_point = center;
                     return;
                 }
--- a/src/shell/zoom.rs
+++ b/src/shell/zoom.rs
@@ -27,7 +27,7 @@ use smithay::{
         },
     },
     output::Output,
-    utils::{IsAlive, Point, Rectangle, Serial, Size},
+    utils::{FrameExtents, IsAlive, Point, Rectangle, Serial, Size},
 };
 use tracing::error;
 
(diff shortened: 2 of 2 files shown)


src/shell/mod.rs
```diff
@@ -2468,6 +2468,14 @@ impl Shell {
             increment: zoom_config.increment,
             movement: zoom_config.view_moves,
         });
+
+        if level > 1.0 {
```
Drakulix: Please just drop the condition, we already return early, if the level equals 1.0 (and lower/negative levels aren't supported anyway).

src/shell/zoom.rs
```diff
@@ -256,86 +228,101 @@ impl ZoomState {
...
+                    output_geometry.loc.y + output_geometry.size.h - 1.0,
                 );
-                output_state_ref.focal_point = focal_point.to_local(output);
+
+                output_state_ref.focal_point = focal_point;
+            }
+            ZoomMovement::Centered => {
+                let center = (output_geometry.size / 2.).to_point();
+
+                if level <= 1.0 + f64::EPSILON {
```
Drakulix: Isn't this the same as `level == 1.0`?

Drakulix (commented):
Code wise this looks good, just a couple of nitpicks. Obviously this will need to go through QA first, but thanks a lot for working on this!

leviport (approved):
Seems to be working nicely. Thanks!

tgrushka:
Thanks for the reviews, @Drakulix  and @leviport . Have I resolved the concerns and do you need anything else on this one? I just rebased and the clippy errors are in another module and unrelated to my changes. If the failing check is blocking, I can run clippy, if you want.

Drakulix (approved):

Drakulix:
> Have I resolved the concerns and do you need anything else on this one? I just rebased and the clippy errors are in another module and unrelated to my changes.

All good! Just needed to find the time to look into the code again. Thanks again :)

Issue #863, opened 2024-09-14
[FEATURE REQUEST] Grab focus without raising the window


BrodieRobertson:
Continuing on from a discussion in https://github.com/pop-os/cosmic-settings/issues/566

Currently when a window is given focus that window is placed above all of the other windows, this makes sense for dialogs and other such windows that demand focus but for many others can be quite disruptive if you tend to overlap windows, especially so when using focus follows cursor. KDE Plasma and a lot of window managers addresses this by allowing the user to separate focus and window raising into 2 separate actions if they desire.

As an example, in this video the window with the red border currently has focus, I have it so window raise only occurs after you click on a window.  

https://github.com/user-attachments/assets/51084c23-ec55-48a0-b776-735781bf4ba4

WatchMkr:
If you have to click, how is this different than Focus follow cursor being off? Is there more to it than an outline showing what will have focus when clicking?

This suggestion might be related. https://github.com/pop-os/cosmic-comp/issues/876

My suggestion there is to raise the window when cursor stops so as to not erroneously raise other windows.

BrodieRobertson:
If a window is in focus but not raised you should still be able to type on that window

https://github.com/user-attachments/assets/1b22ac0b-a104-4b1a-8da3-05c93d7df64a

skewballfox:
Just adding a bit of context. Right now, focus follows mouse, causes things windows like steam settings (x11) and chrome captions (wayland) to be hidden behind the parent window. 

Not sure what the approach should be. 

> My suggestion there is to raise the window when cursor stops so as to not erroneously raise other windows.

That's the approach gnome takes. Though there are situations where you want a window to remain on top even if it doesn't have keyboard focus.

brendanrankin:
I would like to see something along the lines of what Brodie proposes. I call it more "traditional" UNIX window manager...I guess.

Basically, I think raising windows and cursor focus are separate features and should be treated so. For me, it would be good enough to have a distinct checkbox option for "Raise Window when clicked" which would mandate clicking to raise any window. IMHO, focus shouldn't be attached to window raising.

I frequently have a large terminal (or Zellij session) in the background with other windows that I'd like to select text from or some-such on the foreground. This auto-raise whatever the cursor is over when "Focus follows cursor" is enabled really harshes my vibe.

Make sense? It does to me!

Hyperchaotic:
[...]

PR #2535, opened 2026-07-08
Add option to disable auto-raise for focus-follows-cursor

perpetualbits:

Addresses #863.

## AI assistance disclosure

This change was developed with AI assistance (Claude Code); the commits carry an
`Assisted-by:` trailer. Per your policy I want to be upfront about it rather than
hide it. The design decisions here are mine and I understand the change in full:
the intent-enum approach, the reason a persisted mark is needed (see
Implementation), and the trade-offs in the scope note are all things I can defend
in review. It was validated by me on real hardware (see Testing). I'm happy to
walk through any part of it or make changes on request.

## Summary

Adds a new compositor config option, `focus_follows_cursor_raise` (default `true`,
preserving current behaviour), that decouples raising from focusing when focus
follows the cursor.

With `focus_follows_cursor` enabled and `focus_follows_cursor_raise` set to
`false`, moving the pointer over a floating window gives it keyboard focus
without raising it to the front ("sloppy focus"). Explicit focus — a click, a
keyboard focus action, or an application activating itself — still raises the
window as before.

## Motivation

Today focus and raising are coupled: any focus change runs `update_active()`,
which raises the focused floating window. With focus-follows-cursor that means a
window jumps to the front merely because the pointer crossed it. This option lets
the two behaviours be chosen independently, which is the long-standing "focus
follows mouse, no auto-raise" model many users prefer.

## Implementation

- New field `focus_follows_cursor_raise` in `cosmic-comp-config` (default `true`),
  with live reload wired into the existing config watcher.
- A small `Raise { Yes, No }` intent enum threaded through `Shell::set_focus`. A
  named enum is used rather than a second `bool` next to the existing
  `update_cursor` argument, so call sites stay self-documenting and the compiler
  enforces completeness — a missed call site cannot build. Every explicit-focus
  call site passes `Raise::Yes`; only the focus-follows-cursor timer passes the
  configured choice.
- A persisted `no_raise_window` mark on `Shell`. This is needed because focus is
  reconciled periodically: `update_active()` is re-run and would re-raise the
  focused floating window on the next pass, so suppressing the raise once is not
  enough. Instead the hovered window is recorded and `update_active()` skips
  raising it every time it runs, under the same write lock that runs
  `update_active()` so the two never disagree. The mark is set on passive focus
  with raising disabled and cleared on any explicit focus.

## Testing

- `cargo build` and `cargo fmt --check` are clean.
- The default (`focus_follows_cursor_raise = true`) is covered by a unit test in
  `cosmic-comp-config`.
- Validated on real hardware: run as the session compositor. With
  `focus_follows_cursor` on and `focus_follows_cursor_raise` off, hovering a
  partly-covered floating window focuses it without raising it, and clicking still
  raises it. Toggling `focus_follows_cursor_raise` at runtime switches the
  behaviour immediately (live reload).

Note: focus-follows-cursor only triggers under the KMS backend, since the
follow-cursor logic lives in the relative-motion (`PointerMotion`) handler; the
nested `winit`/`x11` backends deliver absolute motion, so this must be exercised on
a real session rather than nested.

## Note on scope

The no-raise mark is a single value on `Shell` rather than per-seat. In the common
single-seat case this is exact. In a multi-seat setup, one seat's explicit focus
clears another seat's hover mark, so the other seat's hovered window can be
re-raised on the next reconciliation. This keeps the change small; happy to make
it per-seat if preferred.

### DIFF
--- a/cosmic-comp-config/src/lib.rs
+++ b/cosmic-comp-config/src/lib.rs
@@ -91,6 +91,15 @@ pub struct CosmicCompConfig {
     pub cursor_follows_focus: bool,
     /// The delay in milliseconds before focus follows mouse (if enabled)
     pub focus_follows_cursor_delay: u64,
+    /// When `focus_follows_cursor` is enabled, controls whether the window
+    /// that gains keyboard focus from the pointer passing over it is also
+    /// raised to the front of the stack.
+    ///
+    /// `true`  = historical behaviour: hovering focuses AND raises.
+    /// `false` = "sloppy focus": hovering focuses but leaves stacking order
+    ///           alone; a window only rises when explicitly acted upon
+    ///           (clicked, keyboard-focused, or activated by an app).
+    pub focus_follows_cursor_raise: bool,
     /// Let X11 applications scale themselves
     pub descale_xwayland: XwaylandDescaling,
     /// Let X11 applications snoop on certain key-presses to allow for global shortcuts
@@ -133,6 +142,9 @@ impl Default for CosmicCompConfig {
             focus_follows_cursor: false,
             cursor_follows_focus: false,
             focus_follows_cursor_delay: 250,
+            // Default true preserves the historical focus-follows-cursor
+            // behaviour (raise on hover) for everyone who has not opted out.
+            focus_follows_cursor_raise: true,
             descale_xwayland: XwaylandDescaling::Fractional,
             xwayland_eavesdropping: XwaylandEavesdropping::default(),
             edge_snap_threshold: 0,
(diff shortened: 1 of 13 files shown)


leviport:
The PR template is required.

perpetualbits:
Superseded by #2612.

While addressing the checklist request I re-tested this change on real hardware
and found a bug: with raise disabled, moving the pointer off a hovered floating
window onto the empty desktop still raised the window. Rather than patch a stalled
PR, I've opened a fresh, fully-documented replacement — #2612 — containing the
same feature plus the fix (root cause and fix explained in the description and
commit message), with the required checklist completed and AI assistance
disclosed.

Closing this in favour of #2612 to keep review in one place. Thanks!

Issue #2153, opened 2026-03-03
High GPU utilization (40%) at idle with mixed refresh rate multi-monitor setup (NVIDIA)


judecale:
## System Information

- **OS:** Pop!_OS 24.04 LTS
- **Kernel:** 6.18.7-76061807-generic
- **cosmic-comp:** 0.1~1771890949~24.04~3f68eee
- **GPU:** NVIDIA GeForce RTX 3060 12GB
- **Driver:** 580.119.02
- **CUDA:** 13.0

## Display Configuration

- **DP-1:** ASUS MX34V 3440x1440 @ 100Hz (center, normal orientation)
- **HDMI-A-1:** Dell S2421H 1920x1080 @ 60Hz (left, rotate90)
- **HDMI-A-2:** AOC 22E1W 1920x1080 @ 60Hz (right, rotate90)
- **Scale:** 125% on all three

## Problem

With all three monitors enabled, cosmic-comp drives GPU utilization to ~40% at idle (no user interaction, no animations, nothing moving on screen). The GPU stays in P5 power state at 19W despite the high utilization, suggesting the compositor is generating high-frequency work that doesn't trigger the GPU to clock up.

This causes visible choppiness and occasional flickering on the HDMI-A-1 (Dell, portrait rotation) display.

### Single monitor vs. three monitors

```
# All three monitors enabled:
utilization.gpu [%], power.draw [W], pstate
40 %, 18.97 W, P5

# one side monitors disabled:
utilization.gpu [%], power.draw [W], pstate
19 %, 29.02 W, P3

# Both side monitors disabled (ultrawide only, after settling):
utilization.gpu [%], power.draw [W], pstate
0 %, 29.95 W, P5
```

### Syscall trace (5-second sample at idle, three monitors enabled)

```
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
 35.20    0.052128           8      6119           epoll_pwait
 12.67    0.018761           1     10933           epoll_ctl
 12.60    0.018653           1     10744      6361 read
  9.37    0.013874          51       271         7 futex
  8.37    0.012389           2      5020           timerfd_settime
  7.94    0.011761           4      2682           sendto
  7.51    0.011125           2      5543           write
  1.67    0.002466           2      1149       568 recvmsg
  1.51    0.002229           2       926           ioctl
  1.30    0.001918           1      1350           epoll_wait
  0.73    0.001075           2       452           eventfd2
[...]

PR #2536, opened 2026-07-08
kms/timings: fix idle GPU busy-loop from render_start collapsing to zero

Guihal:
## Summary

Related to the ~20-25% idle `gpu_busy_percent` reported in #2153, reproduced on a non-Nvidia (AMD Phoenix1 iGPU) machine with one VRR output (144Hz eDP) and one non-VRR output (60Hz HDMI) — so the existing Nvidia-specific hack in `Timings::next_render_time()` was not the cause here.

Note: `next_render_time()` has an early `if self.vendor == Some(0x10de) { return Duration::ZERO; }` return for Nvidia, which runs before the `next_render_time()` fix below — so that half of this PR does not affect Nvidia hardware. The `next_presentation_time()` VRR-threshold fix also ends up masked on Nvidia by that same early return once it reaches `next_render_time()`. This PR is unlikely to be a full fix for #2153 on Nvidia; it targets the busy-loop reproduced on the AMD machine described above.

Root-caused via live `gdb`/`perf` instrumentation of a running KMS-backend session (no code changes to the live process, non-stopping `dprintf` breakpoints only). Confirmed with direct memory reads and call-rate probes that `queue_redraw`'s `render_start` was landing on `Duration::ZERO` (-> `Timer::immediate()`) on nearly every frame, via two independent branches in `timings.rs`:

- `Timings::next_presentation_time()`'s VRR-immediate check (`self.vrr && to_next_ns > refresh_interval_ns`) triggers whenever more than one refresh interval has elapsed since the last presentation — which is true on essentially every steady-state frame at higher refresh rates, not just on a genuine missed vblank. Confirmed firing on the 144Hz VRR output.
- `Timings::next_render_time()`'s final `saturating_sub(margin)` silently clamps to zero whenever the computed render deadline has already passed by even a few nanoseconds, instead of scheduling shortly before the *next* vblank. Confirmed firing independently (~60% of calls) on the non-VRR 60Hz output.

Both feed `queue_redraw` (`surface/mod.rs`) arming `Timer::immediate()` instead of a real pre-vblank deadline, so the compositor re-renders immediately after every vblank instead of idling between frames — keeping the GPU busy for nearly the entire refresh interval on an otherwise fully idle desktop.

## Changes

- `next_presentation_time()`: require the VRR-immediate branch to represent more than two full refresh intervals having elapsed since the last presentation, instead of firing on the normal one-interval-ahead computation that was true on essentially every steady-state frame. This trades off some presentation latency for VRR content that legitimately becomes ready between one and two intervals after the last presentation; a more precise fix would need damage tracking to distinguish that case, which is out of scope here.
- `next_render_time()`: instead of collapsing to `Duration::ZERO` (or an arbitrary fixed floor) when the computed deadline has already passed, roll the target forward one refresh interval at a time until it again has the full submission margin ahead of it, and schedule against that. This is not a latency optimization — on some inputs it defers the render further than a short fixed delay would — but it avoids repeatedly re-triggering immediate-render behavior right after the previous vblank, letting the GPU idle over more of the interval instead.

## Test plan

- [x] `cargo check --lib` passes
- [x] Reproduced the original ~1200+ `queue_redraw`/`Timer::immediate()` calls-per-second behavior live via gdb `dprintf` probes on both outputs before the fix
- [x] Built and installed on the actual reporting machine in place of the stock package binary, logged back into a real KMS session running it live: idle `gpu_busy_percent` dropped from ~20-25% to a steady 6-8%, confirmed via `/sys/class/drm/card*/device/gpu_busy_percent`

## Checklist

### DIFF
--- a/src/backend/kms/surface/timings.rs
+++ b/src/backend/kms/surface/timings.rs
@@ -316,9 +316,15 @@ impl Timings {
             since_last.as_secs() * 1_000_000_000 + u64::from(since_last.subsec_nanos());
         let to_next_ns = (since_last_ns / refresh_interval_ns + 1) * refresh_interval_ns;
 
-        // If VRR is enabled and we are already a full frame or more behind schedule (i.e. this is
-        // a genuine missed-vblank recovery, not just the normal one-interval-ahead computation),
-        // assume that we can present immediately.
+        // If VRR is enabled and more than two full intervals have passed since last presentation,
+        // assume that we can present immediately. This is deliberately more conservative than
+        // "more than one interval" (the original condition): at higher refresh rates the original
+        // threshold was met on essentially every steady-state frame, not just on a genuine
+        // late/missed vblank, which kept re-triggering Timer::immediate() and starving the GPU of
+        // idle time between frames. This does trade off some presentation latency for content that
+        // legitimately becomes ready between one and two intervals after the last presentation;
+        // a more precise fix would need to distinguish that case (e.g. via damage tracking) rather
+        // than widening this threshold.
         if self.vrr && to_next_ns > 2 * refresh_interval_ns {
             Duration::ZERO
         } else {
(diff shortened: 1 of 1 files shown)


leviport:
The PR template is required

Guihal:
Built and installed this on the actual reporting machine (dual-output AMD iGPU setup, 144Hz VRR eDP + 60Hz HDMI) in place of the stock package binary, then logged out/in to actually run it as the live compositor (not nested/winit — real KMS session).

Confirmed empirically via `/sys/class/drm/card*/device/gpu_busy_percent`:

- Before (stock `1:1.2.0-1.1` binary): ~20-25% idle GPU usage
- After (this patch): steady **6-8%** idle GPU usage

Solid improvement, though not fully down to near-0 — there's likely still headroom from unrelated contributors surfaced during root-causing (e.g. an unconditional `schedule_render()` on every `wl_surface.commit` regardless of damage, `src/wayland/handlers/compositor.rs:274` — not addressed by this patch, out of scope here since it's a separate, more invasive change). But this patch alone gets a ~3-4x reduction in idle GPU usage on real hardware, not just in theory.

Guihal:
> The PR template is required

Added

skygrango:
I think this change may be mixing two different responsibilities.

`next_presentation_time()` is already responsible for selecting the target presentation time by predicting the next suitable vblank/presentation opportunity. `next_render_time()` should only translate that presentation target into a render deadline by subtracting the required submission margin.

By advancing `target` inside `next_render_time()` when the deadline has already passed:

```rust
while target <= margin {
    target += refresh_interval;
}
```

we are effectively making `next_render_time()` choose a new presentation target. This duplicates part of the scheduling logic that already exists in `next_presentation_time()` and makes the separation between presentation scheduling and render scheduling less clear.

There is also a potential latency concern, especially with VRR. An expired render deadline does not necessarily mean that the compositor should wait for the next refresh interval. In a VRR scenario, starting rendering immediately can be intentional because rendering itself has latency, and delaying the render may increase input latency.

The original `Duration::ZERO` behavior may have been problematic because it could cause an unintended immediate-render loop, but the fix may be better handled by preventing repeated immediate scheduling rather than by making `next_render_time()` select a later presentation opportunity.

skygrango:
When modifying vblank scheduling to improve GPU utilization, we should evaluate more than just GPU idle time. It is also necessary to verify that frame rendering and presentation remain efficient under sustained GPU load, and that the changes do not introduce additional latency or missed presentation opportunities.

Drakulix:
>     * `next_presentation_time()`: require the VRR-immediate branch to represent more than two full refresh intervals having elapsed since the last presentation, instead of firing on the normal one-interval-ahead computation that was true on essentially every steady-state frame. This trades off some presentation latency for VRR content that legitimately becomes ready between one and two intervals after the last presentation; a more precise fix would need damage tracking to distinguish that case, which is out of scope here.

We already have damage-tracking, the next_presentation_time just gives us an estimate. The actual render step should determine, if we have any changes in the first place. If that isn't the case the render is skipped and no immediate new render is scheduled, thus breaking the loop.

If that doesn't happen on your system, then this isn't the right fix and the logic bug is elsewhere.

>     * `next_render_time()`: instead of collapsing to `Duration::ZERO` (or an arbitrary fixed floor) when the computed deadline has already passed, roll the target forward one refresh interval at a time until it again has the full submission margin ahead of it, and schedule against that. This is not a latency optimization — on some inputs it defers the render further than a short fixed delay would — but it avoids repeatedly re-triggering immediate-render behavior right after the previous vblank, letting the GPU idle over more of the interval instead.

This has the chance of repeatedly missing vblanks. We don't have a good story for when rendering takes more time than a single interval, we'd need something akin to triple buffering to handle that better.

We are already behind in this case, imo it makes sense to start rendering immediately. The important part again is, that if there is nothing happening on an idle desktop, we should be able to break the cycle. If there is damage then this condition makes sure we at least render as quickly as we can trying to keep up, even if that might cause uneven frame timings instead of missing additional frames by accident.

PR #2541, opened 2026-07-10
wayland: avoid redraws for empty surface commits

Guihal:
## What this changes

`CompositorHandler::commit()` currently calls `schedule_render()` for every commit on a visible surface. Smithay may later find that the output damage is empty and skip drawing, but by then the backend scheduling and render path has already been entered.

This adds a conservative check before scheduling the output. For each committed surface tree, the compositor keeps a value-only signature containing:

- surface identity and tree order;
- attached buffer identity;
- Smithay's damage commit counter;
- surface view, scale, transform, and opaque region;
- the alpha modifier.

The handler also compares the mapped window geometry before and after shell commit processing. Geometry is not part of the persisted surface-tree signature.

The commit handler now has three outcomes:

- schedule a render when that state changed or the commit is otherwise ambiguous;
- schedule frame callbacks without a redraw when the visual state is unchanged but `wl_surface.frame` callbacks are pending;
- do nothing when an already mapped toplevel has neither a visual change nor pending callbacks.

The last case is deliberately narrow. Initial mapping, buffer removal, geometry changes, popups, layer surfaces, lock surfaces, cursor and drag surfaces, fullscreen surfaces, subsurface commits, and other cases that cannot be proven unchanged continue through the existing render path.

A callback-only request uses the existing estimated-vblank timer on KMS. If the output is idle, it arms the timer for the next estimated presentation time and sends frame callbacks without calling `redraw()`. If a render or vblank is already pending, that work delivers the callbacks instead. X11 retains its existing scheduled output pass, while winit continues to use its fixed render loop.

XDG toplevel titles need separate handling because they affect server-side decorations without changing Smithay's renderer surface state. `title_changed()` now refreshes the mapped decoration and schedules its output directly.

This does not pass surface damage rectangles into KMS or replace Smithay's output damage tracking. Smithay still decides which output regions need to be redrawn after the compositor has decided that the renderer needs to wake up at all.

This is independent of the render-timing change and is kept in a separate PR so that the scheduling policy can be reviewed and tested on its own.

## Testing

- `cargo check`
- `cargo test --lib` (4 passed)
- `rustfmt --check` on the touched Rust files
- `git diff --check`
- `make` (release build)
- installed the release binary and verified that its checksum matched the built artifact

The installed build is ready for real-session testing. I have left the runtime-testing checkbox open until that is complete.

### DIFF
--- a/src/backend/kms/mod.rs
+++ b/src/backend/kms/mod.rs
@@ -599,6 +599,17 @@ impl KmsState {
         }
     }
 
+    pub fn schedule_frame_callbacks(&mut self, output: &Output) {
+        for surface in self
+            .drm_devices
+            .values()
+            .flat_map(|d| d.inner.surfaces.values())
+            .filter(|s| s.output == *output || s.output.mirroring().is_some_and(|o| &o == output))
+        {
+            surface.schedule_frame_callbacks();
+        }
+    }
+
     pub fn target_node_for_output(&self, output: &Output) -> Option<DrmNode> {
         self.drm_devices
             .values()
--- a/src/backend/kms/surface/mod.rs
+++ b/src/backend/kms/surface/mod.rs
@@ -216,6 +216,7 @@ pub enum ThreadCommand {
     UpdateScreenFilter(ScreenFilter),
     VBlank(Option<DrmEventMetadata>),
     ScheduleRender,
+    ScheduleFrameCallbacks,
     AdaptiveSyncAvailable(SyncSender<Result<VrrSupport>>),
     UseAdaptiveSync(AdaptiveSync),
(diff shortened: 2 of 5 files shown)


Guihal:
Runtime testing looks good so far. With no visible content changes, idle GPU usage dropped from roughly 20% to 2%, and I have not noticed any regressions from this change during normal use. I will add more testing details here if anything else turns up.

skygrango:
That’s an interesting idea, but my concern is that `queue_estimated_vblank` gets called when the surface hasn't changed; this triggers unnecessary calls to `send_frame_callbacks`, which in turn affects other surfaces.

PR #2543, opened 2026-07-10
cleanup: remove attempts to force redraw

wash2:
They seem to always fail anyways and it has no negative effect as far as I can tell


Drakulix (approved):

Drakulix:
> They seem to always fail anyways and it has no negative effect as far as I can tell

I think the issue simply is, that libcosmic sets this value before mapping the window / committing the first buffer. So obviously `visible_output_for_surface` returns None. I think we simply should remove the log message, but this works as well.

jklgrasso (approved):

PR #2556, opened 2026-07-11
feat: per-application workspace assignment rules

AldenDana:
Implements pop-os/cosmic-epoch#1395 / pop-os/cosmic-workspaces-epoch#186: opening an app always places it on a chosen workspace.

Adds a `workspace_assignments` config to `com.system76.CosmicComp` — a list of `(app_id, workspace)` rules (workspace numbers 1-based, matching the UI). When a new window's `app_id` matches, it maps to that workspace instead of the active one, regardless of launch origin; the target workspace is flagged urgent rather than stealing focus (same behavior as other windows mapped to inactive workspaces). Rules hot-reload like other compositor settings.

Rules can be managed from the **window context menu**: right-click a titlebar → *Always open on this workspace* toggles a rule for that app on the window's current workspace (the entry then shows *Always opens on workspace N* with a checkmark). The menu writes through cosmic-config, so the config file stays the single source of truth:

```ron
// ~/.config/cosmic/com.system76.CosmicComp/v1/workspace_assignments
[
    (app_id: "code", workspace: 5),
]
```

**Alt+Space opens the window menu.** Not every window has a compositor-drawn titlebar to right-click — client-side-decorated windows (most GTK4/Qt6/Electron/libcosmic apps, including COSMIC's own Files) have none, so there was previously no way to reach the context menu (and therefore the workspace-assignment toggle) for them at all. Alt+Space is added as a hardcoded, non-configurable compositor shortcut (`PrivateAction::OpenWindowMenu`) that opens the menu for the currently focused window regardless of decoration mode — chosen because it's unused by default and is the traditional "window system menu" shortcut on other desktops (Super+M was the first attempt, but that's already bound to Maximize by default, so it's Alt+Space instead).

**Rename support.** The `zcosmic_workspace_handle_v2::rename` request was defined in the protocol but never handled compositor-side; this wires it up (advertises the `Rename` capability, applies the name, and persists it for pinned workspaces) so clients can let users rename workspaces.

Happy to follow up with a cosmic-settings page for bulk management if the config shape is acceptable.


AldenDana:
**AI disclosure**

This implementation was developed with assistance from Claude (Anthropic's AI coding assistant), directed and reviewed by me throughout — I specified the requirements, reviewed the generated diffs, and tested every feature live on my own machine. I'm continuing to work with Claude to address review feedback, so turnaround should be quick either way.

  - (Disclosing here in the PR rather than retroactively rewriting already-pushed commit messages.)

PR #2562, opened 2026-07-12
Skip stale frame extents when maximizing/fullscreening

### LINKED ISSUE #2503: PR #2441 Introduces Stale Frame Extents on Maximize
In #2440 I reported that the update of the `smithay` dependency in 9f28a764a7cac5f9acd16a6cff46d72d0656c87d introduced a breaking semantic change to `X11Surface::geometry()` that cosmic-comp did not fully adapt to.

@hojjatabdollahi kindly fixed this in #2441

The fix however introduces a different problem.

Now, in `comsic-comp` `0.1~1781804234~24.04~aac1e19` ([`epoch-1.0.16`](https://github.com/pop-os/cosmic-comp/releases/tag/epoch-1.0.16)) when you maximize a window (like Firefox launched with `GDK_BACKEND=x11 MOZ_ENABLE_WAYLAND=0 firefox`), the window extends past the right side of the screen, and then when you Alt-Tab switch to another app, it fixes itself to fill the screen.

Here's a video of it happening : https://youtu.be/uFbaV5YAgiI

## LLM Asserted Root Cause: PR #2441 Introduces Stale Frame Extents on Maximize

Here's a possible explanation generated by an LLM which is beyond my understanding of how the compositor works, feel free to either use this for reference or ignore it. I don't know if it's true

### Background

PR #2441 fixed XWayland CSD windows shrinking on events by changing `src/shell/element/surface.rs:242` to add `surface.frame_extents()` to the geometry sent to `configure_with_sync()`:

```rust
// Before
surface.configure_with_sync(geo.as_logical(), None);

// After
surface.configure_with_sync(geo.as_logical() + surface.frame_extents(), None);
```

`frame_extents` comes from the `_GTK_FRAME_EXTENTS` X11 property that GTK apps (like Firefox) set to advertise the pixel widths of their CSD drop shadows. Adding it expands the configured bounding box so the X11 window's shadow area is accounted for — the content area inside the shadow still matches the intended `geo`.

### The New Bug: Stale Frame Extents During Maximize

The problem is a **stale state / ordering race** during maximize. The maximize path in `src/shell/layout/floating/mod.rs:360` calls these two lines back-to-back:

```rust
mapped.set_maximized(true);                       // line 360
mapped.set_geometry(geometry.to_global(&output)); // line 361
```

Here is what happens step by step:

1. **Normal state**: Firefox has drop shadows. `_GTK_FRAME_EXTENTS` = `{left:10, right:10, top:10, bottom:10}` (example). `surface.frame_extents()` returns these values.

2. **`set_maximized(true)`** is called. Smithay's `change_net_state()` updates its internal `net_state` synchronously, so `is_maximized()` immediately returns `true`. The `_NET_WM_STATE_MAXIMIZED` atoms are written to the X11 property.

3. **`set_geometry(screen_geo)`** is called immediately after. It calls:
   ```
   configure_with_sync(screen_geo + frame_extents_OLD)
   ```
   For a 1920×1080 screen with 10px shadows on all sides, this sends:
   ```
   loc: (-10, -10), size: (1940, 1100)  ← 20px too wide and tall
   ```
   At this point `frame_extents()` still returns the **old** non-zero shadow values — Firefox hasn't had a chance to process the maximize and update `_GTK_FRAME_EXTENTS` yet.

4. **Firefox processes the configure**, sees it is being maximized, removes its drop shadows, and sets `_GTK_FRAME_EXTENTS` → `{0,0,0,0}`. Smithay's `update_toolkit_frame_extents()` stores the new zero values. But Firefox renders at the already-configured `1940×1100`, causing the right (and bottom) edge to extend past the screen boundary.

5. **Alt-Tab** triggers another geometry update for the focus change. By now `frame_extents()` returns `{0,0,0,0}`. The configure becomes `screen_geo + {0,0,0,0}` = `screen_geo`, and the window correctly fills the screen.

### Why Alt-Tab Fixes It

The focus change triggers a re-configure. By that point Firefox has already processed the maximize and zeroed out `_GTK_FRAME_EXTENTS`, so no extra size is added and the window snaps into place.

### Fix Direction

Since `is_maximized()` already returns `true` at the time `set_geometry` is called (smithay updates its internal state synchronously in `change_net_state`), the fix can simply skip adding frame extents for maximized or fullscreen windows:

```rust
WindowSurface::X11(surface) => {
    let frame_extents = if surface.is_maximized() || surface.is_fullscreen() {
        Default::default()
    } else {
        surface.frame_extents()
    };
    let _ = surface.configure_with_sync(geo.as_logical() + frame_extents, None);
}
```

This is correct because GTK windows always remove their CSD shadows when maximized or fullscreen — their eventual `_GTK_FRAME_EXTENTS` will be `{0,0,0,0}`, which is exactly what `Default::default()` supplies here, just without waiting for the client to confirm it.

gene1wood:
PR #2441 added the client's advertised _GTK_FRAME_EXTENTS to the geometry sent in configure_with_sync() so XWayland CSD windows (e.g. Firefox) wouldn't shrink. But set_maximized()/set_geometry() are called back-to-back, and Smithay's internal maximized/fullscreen state updates synchronously while the client's _GTK_FRAME_EXTENTS does not update until it processes the configure. This meant the window was briefly configured larger than the output using its old, pre-maximize shadow extents, leaving it jutting past the screen edge until an unrelated reconfigure (e.g. focus change on alt-tab) caught up with the client's now-zeroed extents.

GTK clients always drop their CSD shadows when maximized or fullscreen, so once is_maximized()/is_fullscreen() report true, skip frame_extents() and use zero instead of waiting for the client to confirm it.

Fixes #2503

### DIFF
--- a/src/shell/element/surface.rs
+++ b/src/shell/element/surface.rs
@@ -270,8 +270,12 @@ impl CosmicSurface {
                 toplevel.with_pending_state(|state| state.size = Some(geo.size.as_logical()))
             }
             WindowSurface::X11(surface) => {
-                let _ =
-                    surface.configure_with_sync(geo.as_logical() + surface.frame_extents(), None);
+                let frame_extents = if surface.is_maximized() || surface.is_fullscreen() {
+                    Default::default()
+                } else {
+                    surface.frame_extents()
+                };
+                let _ = surface.configure_with_sync(geo.as_logical() + frame_extents, None);
             }
         }
     }


gene1wood:
Could this be reviewed and merged, the issue in #2441 continues to occur on affected windows, it's pretty irritating. @hojjatabdollahi do you have time to take a look at this?

hojjatabdollahi:
@gene1wood my review holds no wait since I'm just a contributor like yourself.


But I did take a look and test your PR. This solves maximize, but unmaximize has the same issue (tile 2 windows and maximize/unmaximize firefox).

My guess is it's better if we implement `XwmHandler::property_notify` and on `WmWindowProperty::FrameExtents` re-call `set-geometry`, so we don't need to guess where and when shadows are being added/removed. :shrug:

gene1wood:
Thanks for the feedback! I'll work on reproducing the unmaximize issue and fixing this PR.

PR #2568, opened 2026-07-14
fix: improve screen edge mouse pointer accuracy

RipleyTom:
This fixes scrolling by moving to the edge in games when using scaling.
For testing purposes I encountered the initial issue in XCOM 2 and The Guild - Europa 1410 Demo in 3840*2160 at 175% scale but I think a lot of games were affected.

The current implementation removed 1 from the logical size as the upper margin of the clamp, which could easily remove several physical pixels and the mouse pointer would never reach the edges which resulted in games not scrolling down and right properly.

We now use 2 extra steps, first we use a margin of (last logical pixel + edge) / 2.
Then we limit the size to the floating value of the logical size. This is necessary as the values we get from Wayland are rounded and we can easily overflow into the upper bound otherwise.

I made a simple rust test:
<details>
<summary>Test code</summary>

```rust
const RESOLUTIONS: [(f64, f64); 5] = [(1280.0, 720.0), (1920.0, 1080.0), (2560.0, 1440.0), (3840.0, 2160.0), (7680.0, 4320.0)];
const SCALINGS: [f64; 11] = [0.5, 0.75, 1.0, 1.25, 1.50, 1.75, 2.0, 2.25, 2.50, 2.75, 3.0];

fn ret_result(b: bool) -> &'static str {
	if b {
		return "\x1b[92mSuccess\x1b[0m";
	}

	return "\x1b[91mFailure\x1b[0m";
}

struct Counters {
	success_old: u64,
	success_new: u64,
	total: u64,
}

impl Counters {
	fn new() -> Counters {
		Counters {
			success_old: 0,
			success_new: 0,
			total: 0,
		}
	}

	fn test_old(&mut self, b: bool) -> &'static str {
		if b {
			self.success_old += 1;
		}

		return ret_result(b);
	}

	fn test_new(&mut self, b: bool) -> &'static str {
		if b {
			self.success_new += 1;
		}

		return ret_result(b);
	}
}

fn main() {
	let mut count = Counters::new();
	for scaling in SCALINGS {
		println!("At {}% scaling:", scaling * 100.0);
		for resolution in RESOLUTIONS {
			println!("{}x{}", resolution.0, resolution.1);
			let logical_size = (resolution.0 / scaling, resolution.1 / scaling);
			let logical_geometry_size = (logical_size.0.round(), logical_size.1.round());
			let physical_last_pixels = (resolution.0 - 1.0, resolution.1 - 1.0);
			let current_actual_last_pixels = (((logical_geometry_size.0 - 1.0) * scaling).floor(), ((logical_geometry_size.1 - 1.0) * scaling).floor());

			let logical_last_pixels_with_scaled_edge = (
				((resolution.0 - 1.0) / scaling + (logical_geometry_size.0.min(logical_size.0))) / 2.0,
				((resolution.1 - 1.0) / scaling + (logical_geometry_size.1.min(logical_size.1))) / 2.0,
			);
			let physical_last_pixel_with_scaled_edge = ((logical_last_pixels_with_scaled_edge.0 * scaling).floor(), (logical_last_pixels_with_scaled_edge.1 * scaling).floor());

			let old = count.test_old(current_actual_last_pixels == physical_last_pixels);
			let success_new = physical_last_pixel_with_scaled_edge == physical_last_pixels;
			let new = count.test_new(success_new);
			count.total += 1;
			println!("old: {}, logical_edge: {}", old, new);

			if !success_new {
				println!("logical_size: {:?}", logical_size);
				println!("geometry_size: {:?}", logical_geometry_size);
				println!("real_last_pixels: {:?}", physical_last_pixels);
				println!("current_actual_last_pixels: {:?}", current_actual_last_pixels);
				println!("physical_last_pixel_with_scaled_edge: {:?}", physical_last_pixel_with_scaled_edge);
			}
		}
		println!("=====================================================")
	}

	println!("old: {}/{}", count.success_old, count.total);
	println!("new: {}/{}", count.success_new, count.total);
}
```
</details>

<details>
<summary>Results</summary>

```
At 50% scaling:
1280x720
old: Success, logical_edge: Success
1920x1080
old: Success, logical_edge: Success
2560x1440
old: Success, logical_edge: Success
3840x2160
old: Success, logical_edge: Success
7680x4320
old: Success, logical_edge: Success
=====================================================
At 75% scaling:
1280x720
old: Success, logical_edge: Success
1920x1080
old: Success, logical_edge: Success
2560x1440
old: Success, logical_edge: Success
3840x2160
old: Success, logical_edge: Success
7680x4320
old: Success, logical_edge: Success
=====================================================
At 100% scaling:
1280x720
old: Success, logical_edge: Success
1920x1080
old: Success, logical_edge: Success
2560x1440
old: Success, logical_edge: Success
3840x2160
old: Success, logical_edge: Success
7680x4320
old: Success, logical_edge: Success
=====================================================
At 125% scaling:
1280x720
old: Failure, logical_edge: Success
1920x1080
old: Failure, logical_edge: Success
2560x1440
old: Failure, logical_edge: Success
3840x2160
old: Failure, logical_edge: Success
7680x4320
old: Failure, logical_edge: Success
=====================================================
At 150% scaling:
1280x720
old: Failure, logical_edge: Success
1920x1080
old: Failure, logical_edge: Success
2560x1440
old: Failure, logical_edge: Success
3840x2160
old: Failure, logical_edge: Success
7680x4320
old: Failure, logical_edge: Success
=====================================================
At 175% scaling:
1280x720
old: Failure, logical_edge: Success
1920x1080
old: Failure, logical_edge: Success
2560x1440
old: Failure, logical_edge: Success
3840x2160
old: Failure, logical_edge: Success
7680x4320
old: Success, logical_edge: Success
=====================================================
At 200% scaling:
1280x720
old: Failure, logical_edge: Success
1920x1080
old: Failure, logical_edge: Success
2560x1440
old: Failure, logical_edge: Success
3840x2160
old: Failure, logical_edge: Success
7680x4320
old: Failure, logical_edge: Success
=====================================================
At 225% scaling:
1280x720
old: Failure, logical_edge: Success
1920x1080
old: Failure, logical_edge: Success
2560x1440
old: Failure, logical_edge: Success
3840x2160
old: Failure, logical_edge: Success
7680x4320
old: Failure, logical_edge: Success
=====================================================
At 250% scaling:
1280x720
old: Failure, logical_edge: Success
1920x1080
old: Failure, logical_edge: Success
2560x1440
old: Failure, logical_edge: Success
3840x2160
old: Failure, logical_edge: Success
7680x4320
old: Failure, logical_edge: Success
=====================================================
At 275% scaling:
1280x720
old: Failure, logical_edge: Failure
logical_size: (465.45454545454544, 261.8181818181818)
geometry_size: (465.0, 262.0)
real_last_pixels: (1279.0, 719.0)
current_actual_last_pixels: (1276.0, 717.0)
physical_last_pixel_with_scaled_edge: (1278.0, 719.0)
1920x1080
old: Failure, logical_edge: Success
2560x1440
old: Failure, logical_edge: Success
3840x2160
old: Failure, logical_edge: Failure
logical_size: (1396.3636363636363, 785.4545454545455)
geometry_size: (1396.0, 785.0)
real_last_pixels: (3839.0, 2159.0)
current_actual_last_pixels: (3836.0, 2156.0)
physical_last_pixel_with_scaled_edge: (3839.0, 2158.0)
7680x4320
old: Failure, logical_edge: Success
=====================================================
At 300% scaling:
1280x720
old: Failure, logical_edge: Success
1920x1080
old: Failure, logical_edge: Success
2560x1440
old: Failure, logical_edge: Success
3840x2160
old: Failure, logical_edge: Success
7680x4320
old: Failure, logical_edge: Success
=====================================================
old: 16/55
new: 53/55
```
</details>

So we go from 16/55 tests passed to 53/55, the 2 failures are at very high scaling(275%) and are the result of the rounding of the geometry_size where the shortfall can be >= 1 pixels at those scales. I have no easy solution for this.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -337,7 +337,7 @@ impl State {
                                 // Constraint does not apply if not within region
                                 if !constraint.region().is_none_or(|x| {
                                     x.contains(
-                                        (ptr.current_location() - *surface_loc).to_i32_round(),
+                                        (ptr.current_location() - *surface_loc).to_i32_floor(),
                                     )
                                 }) {
                                     return;
@@ -382,14 +382,25 @@ impl State {
                         .unwrap_or(current_output.clone());
                     drop(shell);
                     let output_geometry = output.geometry();
-                    position.x = position.x.clamp(
-                        output_geometry.loc.x as f64,
-                        (output_geometry.loc.x + output_geometry.size.w - 1) as f64,
-                    );
-                    position.y = position.y.clamp(
-                        output_geometry.loc.y as f64,
-                        (output_geometry.loc.y + output_geometry.size.h - 1) as f64,
-                    );
+
+                    let scale = output.current_scale().fractional_scale();
+                    let physical = output
+                        .current_mode()
+                        .map(|mode| output.current_transform().transform_size(mode.size))
+                        .unwrap_or_default();
+                    let logical = physical.to_f64().to_logical(scale);
+                    let output_geometry_loc = output_geometry.loc.to_f64();
+                    // output_geometry.size is a rounded value and may undershoot/overshoot the accurate logical size
+                    // We constrain the position with:
+                    // - output_geometry.size so that we don't send leave events to a fullscreen app
+                    // - logical size so that the position doesn't end up outside the actual size of the output
+                    // See https://github.com/pop-os/cosmic-comp/pull/2568
+                    let max_x = output_geometry_loc.x
+                        + logical.w.min(output_geometry.size.w as f64).next_down();
+                    let max_y = output_geometry_loc.y
+                        + logical.h.min(output_geometry.size.h as f64).next_down();
+                    position.x = position.x.clamp(output_geometry_loc.x, max_x);
+                    position.y = position.y.clamp(output_geometry_loc.y, max_y);
 
                     if ptr.is_grabbed() {
                         if seat
@@ -543,7 +554,7 @@ impl State {
 
                             if let Some(region) = &confine_region
                                 && !region
-                                    .contains((pos.as_logical() - *surface_loc).to_i32_round())
+                                    .contains((pos.as_logical() - *surface_loc).to_i32_floor())
                             {
                                 return (false, None);
                             }
@@ -614,7 +625,7 @@ impl State {
                                         PointerConstraint::Confined(confined) => confined.region(),
                                     };
                                     let point =
-                                        (ptr.current_location() - surface_location).to_i32_round();
+                                        (ptr.current_location() - surface_location).to_i32_floor();
                                     if region.is_none_or(|region| region.contains(point)) {
                                         constraint.activate();
                                     }
@@ -2210,7 +2221,7 @@ impl State {
                         if Rectangle::new(offset.as_local(), output_geo.size)
                             .intersection(output_geo)
                             .is_some_and(|geometry| {
-                                geometry.contains(global_pos.to_local(output).to_i32_round())
+                                geometry.contains(global_pos.to_local(output).to_i32_floor())
                             })
                             && let Some(element) = workspace.popup_element_under(location, seat)
                         {
@@ -2224,7 +2235,7 @@ impl State {
                         if Rectangle::new(offset.as_local(), output_geo.size)
                             .intersection(output_geo)
                             .is_some_and(|geometry| {
-                                geometry.contains(global_pos.to_local(output).to_i32_round())
+                                geometry.contains(global_pos.to_local(output).to_i32_floor())
                             })
                             && let Some(element) = workspace.toplevel_element_under(location, seat)
                         {
@@ -2427,7 +2438,7 @@ impl State {
                         if let Some(constraint) = constraint
                             && let Some(region) = constraint.region()
                         {
-                            let point_in_surface = (p - surface_offset.to_f64()).to_i32_round();
+                            let point_in_surface = (p - surface_offset.to_f64()).to_i32_floor();
                             return region.contains(point_in_surface);
                         }
                         true
--- a/src/shell/element/stack.rs
+++ b/src/shell/element/stack.rs
@@ -530,7 +530,7 @@ impl CosmicStack {
             let geo = p.windows.lock().unwrap()[p.active.load(Ordering::SeqCst)].geometry();
 
             if surface_type.contains(WindowSurfaceType::TOPLEVEL) {
-                let point_i32 = relative_pos.to_i32_round::<i32>();
+                let point_i32 = relative_pos.to_i32_floor::<i32>();
                 if (point_i32.x - geo.loc.x >= -RESIZE_BORDER && point_i32.x - geo.loc.x < 0)
                     || (point_i32.y - geo.loc.y >= -RESIZE_BORDER && point_i32.y - geo.loc.y < 0)
                     || (point_i32.x - geo.loc.x >= geo.size.w
--- a/src/shell/element/window.rs
+++ b/src/shell/element/window.rs
@@ -113,7 +113,7 @@ impl Focus {
         location: Point<f64, Logical>,
     ) -> Option<Focus> {
         let geo = surface.geometry();
-        let loc = location.to_i32_round::<i32>() - geo.loc;
+        let loc = location.to_i32_floor::<i32>() - geo.loc;
         if loc.y < 0 && loc.x < 0 {
             Some(Focus::ResizeTopLeft)
         } else if loc.y < 0 && loc.x >= geo.size.w {
@@ -292,7 +292,7 @@ impl CosmicWindow {
             {
                 let geo = p.window.geometry();
 
-                let point_i32 = relative_pos.to_i32_round::<i32>();
+                let point_i32 = relative_pos.to_i32_floor::<i32>();
                 let ssd_height = if has_ssd { SSD_HEIGHT } else { 0 };
 
                 if (point_i32.x - geo.loc.x >= -RESIZE_BORDER && point_i32.x - geo.loc.x < 0)
--- a/src/shell/grabs/menu/mod.rs
+++ b/src/shell/grabs/menu/mod.rs
@@ -548,7 +548,7 @@ impl PointerGrab<State> for MenuGrab {
                 let mut bbox = elem.iced.bbox();
                 bbox.loc = elem.position.as_logical();
 
-                bbox.contains(event_location.to_i32_round())
+                bbox.contains(event_location.to_i32_floor())
             }) {
                 let element = &mut elements[i];
 
@@ -753,7 +753,7 @@ impl TouchGrab<State> for MenuGrab {
                 let mut bbox = elem.iced.bbox();
                 bbox.loc = elem.position.as_logical();
 
-                bbox.contains(event_location.to_i32_round())
+                bbox.contains(event_location.to_i32_floor())
             }) {
                 let element = &mut elements[i];
 
--- a/src/shell/layout/tiling/mod.rs
+++ b/src/shell/layout/tiling/mod.rs
@@ -3150,7 +3150,7 @@ impl TilingLayout {
         location_f64: Point<f64, Local>,
         seat: &Seat<State>,
     ) -> Option<KeyboardFocusTarget> {
-        let location = location_f64.to_i32_round();
+        let location = location_f64.to_i32_floor();
 
         for (mapped, geo) in self.mapped() {
             if !mapped.bbox().contains((location - geo.loc).as_logical()) {
@@ -3177,7 +3177,7 @@ impl TilingLayout {
         location_f64: Point<f64, Local>,
         seat: &Seat<State>,
     ) -> Option<KeyboardFocusTarget> {
-        let location = location_f64.to_i32_round();
+        let location = location_f64.to_i32_floor();
 
         for (mapped, geo) in self.mapped() {
             // Tiled windows are rendered cropped to their tile (`geo`), so input must be bound to the tile as well
@@ -3209,7 +3209,7 @@ impl TilingLayout {
         overview: OverviewMode,
         seat: &Seat<State>,
     ) -> Option<(PointerFocusTarget, Point<f64, Local>)> {
-        let location = location_f64.to_i32_round();
+        let location = location_f64.to_i32_floor();
 
         if matches!(overview, OverviewMode::None) {
             for (mapped, geo) in self.mapped() {
@@ -3244,7 +3244,7 @@ impl TilingLayout {
     ) -> Option<(PointerFocusTarget, Point<f64, Local>)> {
         let tree = &self.queue.trees.back().unwrap().0;
         let root = tree.root_node_id()?;
-        let location = location_f64.to_i32_round();
+        let location = location_f64.to_i32_floor();
 
         if matches!(overview, OverviewMode::None) {
             for (mapped, geo) in self.mapped() {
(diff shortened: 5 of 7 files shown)


src/input/mod.rs
```diff
@@ -382,13 +382,27 @@ impl State {
...
+                        .unwrap_or_default()
+                        .to_f64()
+                        .to_logical(scale);
+
                     position.x = position.x.clamp(
                         output_geometry.loc.x as f64,
-                        (output_geometry.loc.x + output_geometry.size.w - 1) as f64,
+                        output_geometry.loc.x as f64
+                            + (output_geometry.size.w as f64).min(logical_size.w) as f64
+                            - edge_inset,
```
Drakulix: I am sorry, but "value determined by experimentation" won't fly here.

Could you try this instead to see if it fixes your issue?

```suggestion
                        ((output_geometry.loc.x + output_geometry.size.w) as f64).next_down()
```

RipleyTom: I already considered it before and it is worse with 41/55 tests passed.
I updated with a better version, that 0.25 was indeed a bit lazy. inset is now a midway point between the edge and the last pixel.
And before someone suggest it, yes I tried physical-1.0/scale before too.

Result is actually now 53/55.
If someone has a better suggestion, feel free to edit the test and see if you can come up with something better.
Also updated the test.

Drakulix: This seems much more sane, but I am not sure I 100% understand what's going on here yet.

1. ~~If the output_geometry.size is too inaccurate because it rounds to i32 again (see src/utils/prelude.rs) why don't we just always use logical here?~~ just using logical seems to push the test results to 55/55.
2. Why the half-way point between edge and last-pixel? Either one has to be better/more accurate, no?
3. So without knowing any better about 1. and 2. I would assume something like `let max_x = (output_geometry.loc.x as f64 + logical.w).next_down();` would work or does that simply fail due to floating point inaccuracies?

RipleyTom: 1. I think it matters for surface detection, which is why we min between the output geometry and the logical. The geometry is the underlying value used by Wayland and it expects something within those bounds. At least that's my assumption but I'm no Wayland expert(and those values are the result of round() so it's not necessarily bigger or smaller than the accurate logical size, hence the min).
2. Because of floating point accuracy, I initially went with  -(1/scale) which sounds logical as we want the cap to be the last pixel in the clamp but I ended up with a 2158.9999999999~ value that got floored to 2158. And we don't want edge included so that's out. Half-way point seems the sanest.
3. see 1.

RipleyTom: For 2. there is also the concern that wayland uses wl_fixed(24_8 fixed floating point) to transmit position to clients.

Drakulix: 1. It doesn't. Wayland clients no nothing about their position relative to outputs, the resolution is purely informative. We can do what we want here.
2. Cursor coordinates are floating-point as well though. So I am not really sure, why we cannot simply use 2158.9999~, if that is the largest possible floating point value included in the output geometry.

RipleyTom: 1. Yes sorry, I meant more the compositor, something like Workspace::toplevel_surface_under(https://github.com/pop-os/cosmic-comp/blob/master/src/shell/workspace.rs#L931) uses output_geometry to see which surface is under a position.

2. I made a simple example in C(`clang test_wl_fixed.c -o test_wl_fixed -lwayland-client -lm`):
```c
#include <stdio.h>
#include <math.h>
#include <wayland-util.h>

int main()
{
	double val = 2160.0;
	double nd = nextafter(val, 0.0);
	wl_fixed_t wl_value = wl_fixed_from_double(nd);
	double val_final = wl_fixed_to_double(wl_value);
	printf("Val: %.15f, NextAfter: %.15f, Final value: %.15f\n", val, nd, val_final);
}
```
```
$ ./test_wl_fixed
Val: 2160.000000000000000, NextAfter: 2159.999999999999545, Final value: 2160.000000000000000
```

RipleyTom: Ideally we should maybe even put a guard so that it doesn't round up actually.

Drakulix: > 1. Yes sorry, I meant more the compositor, something like Workspace::toplevel_surface_under(https://github.com/pop-os/cosmic-comp/blob/master/src/shell/workspace.rs#L931) uses output_geometry to see which surface is under a position.

> $ ./test_wl_fixed
> Val: 2160.000000000000000, NextAfter: 2159.999999999999545, Final value: 2160.000000000000000
> ```

Okay now this starts making sense.

So in short: If we pick too high a value `toplevel_surface_under` might not register our window as clicked any more as we are falling out of the geometry. And if we pick to low of a value due to scaling *and* rounding when converting to wl_fixed, we might end up with missing a few pixels.

So I see why the previous code path in particular caused issues, when the geometry was reduced by a whole logical pixel.

However with your `test_wl_fixed` example, a value of 2159.999 seems like something a sensible client should imo round to 2160. So the question still remains for me, whether taking the logical resolution as a f64 and using `next_down` really results in real-world client issues?

If it does, I think we should investigate how other compositors handle this, because we need to define some value where the cursor jumps from one compositor to the next. And if we choose the include 2160.0 and only move to the next output on the next larger value, I would assume we hit the same issue with left and top edges.

src/input/mod.rs
```diff
@@ -382,14 +382,24 @@ impl State {
...
+                    let lower_x = (physical.w as f64 - 1.0) / scale;
+                    let upper_x = (output_geometry.size.w as f64).min(logical.w);
+                    let max_x = output_geometry.loc.x as f64 + (lower_x + upper_x) / 2.0;
+
+                    let lower_y = (physical.h as f64 - 1.0) / scale;
+                    let upper_y = (output_geometry.size.h as f64).min(logical.h);
+                    let max_y = output_geometry.loc.x as f64 + (lower_y + upper_y) / 2.0;
+
+                    position.x = position.x.clamp(output_geometry.loc.x as f64, max_x);
+                    position.y = position.y.clamp(output_geometry.loc.y as f64, max_y);
```
Drakulix: So too sum up my last comment. Does this fix the games in question? If so that is the implementation we should arguably use.

```suggestion
                    let scale = output.current_scale().fractional_scale();
                    let physical = output
                        .current_mode()
                        .map(|mode| output.current_transform().transform_size(mode.size))
                        .unwrap_or_default();
                    let logical = physical.to_f64().to_logical(scale);
                    let output_loc = output_geometry.loc.to_f64();

                    let max_x = output_loc.x + logical.w.next_down();
                    let max_y = output_loc.y + logical.h.next_down();

                    position.x = position.x.clamp(output_loc.x, max_x);
                    position.y = position.y.clamp(output_loc.y, max_y);
```

RipleyTom: Sorry I took a bit long as I was looking into stuff as it is a complicated issue.
First I found there is actually a difference in implementation between libwayland and wayland-rs, opened an issue for it(https://github.com/Smithay/wayland-rs/issues/947). But the short of it is wayland-rs doesn't round so there is no need to worry about the value ending up rounded up and on the edge!

The version you proposed is still incorrect(even if it works in games) as the pointer will leave the surface and emit wl_pointer::Event::Leave events when the pointer is pushed into the right and bottom. Because the pointer location is outside the geometry surface. This, I hope you'll agree, doesn't make any sense in a fullscreen application.

I also found a second issue where the location tested was rounded up and always ended up outside the geometry(with both your version and mine) and emited leave events even when it shouldn't. I added it to the commit as it's very much relevant.

Drakulix: > The version you proposed is still incorrect(even if it works in games) as the pointer will leave the surface and emit wl_pointer::Event::Leave events when the pointer is pushed into the right and bottom. Because the pointer location is outside the geometry surface. This, I hope you'll agree, doesn't make any sense in a fullscreen application.

Yeah I absolutely agree. I would still love to get rid of the `.min(output_geometry.*)` calls and just use `logical`, but I guess to do that we would need to side-step `under_from_surface_tree` and always consider fullscreen surfaces to be in focus as long as the pointer is within the logical output-geometry (or copy `under_from_surface_tree` from smithay to use these values / adjust smithays api).

All of this is a bigger refactoring task, than I am comfortable to ask from any contributor and your solution is now quite clean and improves the situation a lot. So I think we can just continue with this approach for now.

(Should you be willing to continue working on this, it would obviously be very welcome though.)

Thanks for sticking with it for this long!

src/input/mod.rs
```diff
@@ -382,14 +382,18 @@ impl State {
...
                     let output_geometry = output.geometry();
-                    position.x = position.x.clamp(
-                        output_geometry.loc.x as f64,
-                        (output_geometry.loc.x + output_geometry.size.w - 1) as f64,
-                    );
-                    position.y = position.y.clamp(
-                        output_geometry.loc.y as f64,
-                        (output_geometry.loc.y + output_geometry.size.h - 1) as f64,
-                    );
+
```
Drakulix, line 385: Could you add a short comment here, why simply clamping to the output_geometry is not sufficient? (Feel free to add a link to this PR for anyone wanting to read more about it.)

src/input/mod.rs
```diff
@@ -382,14 +382,18 @@ impl State {
...
+                    let scale = output.current_scale().fractional_scale();
+                    let physical = output
+                        .current_mode()
+                        .map(|mode| output.current_transform().transform_size(mode.size))
+                        .unwrap_or_default();
+                    let logical = physical.to_f64().to_logical(scale);
+                    let output_geometry_loc = output_geometry.loc.to_f64();
+                    let max_x = output_geometry_loc.x + logical.w.min(output_geometry.size.w as f64).next_down();
+                    let max_y = output_geometry_loc.y + logical.h.min(output_geometry.size.h as f64).next_down();
```
Drakulix: Please make sure the formatting check is happy.

Drakulix (dismissed):
Looks good apart from two small nitpicks.

git-f0x:
This has a regression in that you can't grab the scrollbar of a maximized window at the right screen edge, with the resize indicator showing up (which was the reason I originally added the -1, since the cursor goes off-screen)  :(

The same is true for the bottom edge.

RipleyTom:
> This has a regression in that you can't grab the scrollbar of a maximized window at the right screen edge, with the resize indicator showing up (which was the reason I originally added the -1, since the cursor goes off-screen) :(
> 
> The same is true for the bottom edge.

I have a separate PR for that issue, https://github.com/pop-os/iced/pull/367 .

git-f0x:
Not iced related. This is with all apps (e.g. Firefox).

RipleyTom:
I see, added a separate commit for it as it is a separate issue: rounding was used in calculations which could result in it rounding up and putting the relative position on edge while it actually being below.

RipleyTom:
I added one more commit to address possible issues brought to life by the new pointer accuracy in tiling/mod.rs. The issue is still the same, `let location = location_f64.to_i32_round();` can round up location and put it on the edge. It wasn't an issue before as pointer was constrained by -1 logical so could never end up close enough to the edge to round up to it.

I think that will be the last commit unless @git-f0x finds another issue.

RipleyTom:
Actually relooking at it, I can probably fix most of those with just a to_i32_floor() without changing everything to f64 :facepalm: . Will rework it.

RipleyTom:
I did a full sweep where location and to_i32_round() was used and fixed them.
I think it's pretty much done now.

Drakulix (approved):
Thanks!

git-f0x:
I think the only very minor regression is that a 1 pixel line of the cursor isn't visible anymore when it's all the way at the right or bottom edges (like on other compositors). Might be related to the behavior in maximized Firefox when the cursor is all the way to the right and clicking the scrollbar with the **stationary** cursor to go to a point, the scrollbar proceeds to shrink after the click, instead of staying fully revealed like before. It gets revealed again if the cursor is moved (including pushing further into the edge).

But that's very minor, and this fixes bigger issues.

jklgrasso (approved):

PR #2570, opened 2026-07-14
fix: no ceil for scaled outline thickness

wash2:

I believe this will fix the border width discrepancy for SSD vs CSD windows on scaled outputs. 125% and zoom makes the issue easier to check. `ceil` was done intentionally and mentioned in the commit that it was introduced, but I don't think removing it would cause a panic or anything so it may be OK to just remove it? I imagine that outlines might just be less crisp.

### DIFF
--- a/src/backend/render/mod.rs
+++ b/src/backend/render/mod.rs
@@ -267,7 +267,7 @@ impl IndicatorShader {
             .filter(|(old_settings, _)| &settings == old_settings)
             .is_none()
         {
-            let thickness: f32 = ((thickness as f64 * scale).ceil() / scale) as f32;
+            let thickness: f32 = ((thickness as f64 * scale) / scale) as f32;
             let shader = Self::get(renderer);
 
             let elem = PixelShaderElement::new(


Drakulix:
> I believe this will fix the border width discrepancy for SSD vs CSD windows on scaled outputs. 125% and zoom makes the issue easier to check. `ceil` was done intentionally and mentioned in the commit that it was introduced, but I don't think removing it would cause a panic or anything so it may be OK to just remove it? I imagine that outlines might just be less crisp.

Yeah it was added to make sure we always draw a pixel-perfect outline.

But consistency is more important and I am not sure we want to change this in libcosmic.

Drakulix (approved):

jacobgkau (approved):
Not seeing a significant difference between the outline on a CSD or SSD app even before this change. I see a slight difference in how thick they both are at certain display scale values (with e.g. 125% being thicker with the old version than the new version). The change seems fine-- comparing with other scale values, it seems like the 125% thickness is more in-line with where it should be after this change.

PR #2571, opened 2026-07-15
Various cursor image-copy fixes

ids1024:
There are more edge cases around image-copy to address, as well as inconsistency between handling of `PointerMotion`/`PointerMotionAbsolute`, along with more code deduplication, workspace capture, and https://github.com/pop-os/cosmic-comp/pull/2248. But the changes here should be a good incremental improvement.

The changes related to metadata cursor capture won't have an effect without https://github.com/pop-os/xdg-desktop-portal-cosmic/pull/175.

This also has a change to hold a references to `Buffer`, that should prevent some potential synchronization issues (though in most cases it should have no effect). And it also changes the pointer warp behavior to send `wl_pointer::motion`, which seems to match Gnome.

### DIFF
--- a/src/backend/kms/surface/mod.rs
+++ b/src/backend/kms/surface/mod.rs
@@ -1674,7 +1674,7 @@ fn send_screencopy_result<'a>(
     pre_postprocess_data: &mut PrePostprocessData,
     tx: &std::sync::mpsc::Sender<PendingImageCopyData>,
     frame_result: &RenderFrameResult<GbmBuffer, GbmFramebuffer, CosmicElement<GlMultiRenderer<'a>>>,
-    elements: &[CosmicElement<GlMultiRenderer>],
+    elements: &[CosmicElement<GlMultiRenderer<'a>>],
     (session, frame, res): (
         &ScreencopySessionRef,
         ScreencopyFrame,
@@ -1855,6 +1855,8 @@ fn send_screencopy_result<'a>(
         transform,
         damage.as_deref(),
         sync,
+        // Don't reference `Buffer`s since we blit from framebuffer/postprocess buffer
+        vec![],
     )? {
         if frame_result.is_empty {
             data.frame
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -2426,7 +2426,7 @@ impl State {
                 if is_legal(pos_in_element) {
                     let x = workspace_origin.x + origin.x + pos_in_element.x;
                     let y = workspace_origin.y + origin.y + pos_in_element.y;
-                    Some((Point::new(x, y), output.clone()))
+                    Some((Point::<_, Global>::new(x, y), output.clone()))
                 } else {
                     None
                 }
@@ -2437,10 +2437,10 @@ impl State {
 
         if let Some((point, output)) = point_and_output {
             let original_position = pointer.current_location();
-            pointer.set_location(point);
+            pointer.set_location(point.as_logical());
 
             let mut shell = self.common.shell.write();
-            shell.update_pointer_position(point.as_global().to_local(&output), &output);
+            shell.update_pointer_position(point.to_local(&output), &output);
 
             let seat = shell
                 .seats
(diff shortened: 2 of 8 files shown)


src/backend/kms/surface/mod.rs
```diff
@@ -1848,13 +1850,16 @@ fn send_screencopy_result<'a>(
...
+    let buffers = render_element_buffers(renderer, elements);
+
     if let Some(data) = submit_buffer(
         frame,
         renderer,
         shm_buffer.then_some(fb.as_mut().unwrap()),
         transform,
         damage.as_deref(),
         sync,
+        buffers,
```
Drakulix: Is this really necessary in the kms case, given we just blit from either the pre_postprocessing texture or the framebuffer?

src/input/mod.rs
```diff
@@ -2432,7 +2432,19 @@ impl State {
...
+                .map(|(target, pos)| (target, pos.as_logical()));
+            pointer.motion(
+                self,
+                under,
+                &MotionEvent {
+                    location: point.as_logical(),
+                    serial,
+                    time: self.common.clock.now().as_millis(),
+                },
+            );
```
Drakulix, line 2426: If we do this then we need to make sure for Xwayland to send a relative_motion event of (0,0) in the same frame. (source Xaver from KDE)

Drakulix (commented):
Looks mostly good. Lots of neat little changes, I like it :)

src/input/mod.rs
ids1024: Is that basically just a bug in `Xwayland` we need to work around?

https://wayland.app/protocols/relative-pointer-unstable-v1#zwp_relative_pointer_v1:event:relative_motion doesn't mention anything about `(0, 0)` events but does say:

> Relative motions are not coupled to wl_pointer.motion events, and can be sent in combination with such events, but also independently. There may also be scenarios where wl_pointer.motion is sent, but there is no relative motion. The order of an absolute and relative motion event originating from the same physical motion is not guaranteed.

Which certainly makes it sound like not sending `relative_motion` is the correct thing to do here, or at least a correct thing to do.

Drakulix: > Is that basically just a bug in Xwayland we need to work around?

Less a bug and more a "feature". Xwayland needs to know if a given even is a wrap or not apparently and uses relative motion to differentiate. I'll try to find an actual reference for this.

ids1024: If it's intended that a warp can be recognized this way, the protocol spec should say that, but I guess this will be superseded by `wl_pointer.warp` in the upcoming wayland release: https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/340

Drakulix: - https://invent.kde.org/plasma/kwin/-/merge_requests/5422
- https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/1391

> If it's intended that a warp can be recognized this way, the protocol spec should say that, but I guess this will be superseded by wl_pointer.warp in the upcoming wayland release.

Yeah I am pretty sure that is what motivated this in the first place.

ids1024: Or actually it looks like https://gitlab.freedesktop.org/wayland/wayland/-/tags/1.26.0 was tagged 16 minutes ago.

I guess what we want is to have a `PointerHandle::warp` that uses that all, and I maybe sends `motion`/`relative_motion` to clients binding an older version.

src/backend/kms/surface/mod.rs
ids1024: Good catch. I guess we can just use an empty `buffers` here.

src/input/mod.rs
ids1024: Updated to also send `relative_motion` for now, with a `TODO` comment to use `wl_pointer.warp` here later. Once we get Smithay updated for that.

Drakulix (approved):
LGTM!

ids1024:
Rebased on latest `master`.

jklgrasso (approved):

PR #2572, opened 2026-07-15
Fix frozen/detached active window hint

graham-walker:
This PR reorders the active window hint rendering so it happens after the window’s geometry is calculated.

This fixes a case where the active window hint can appear frozen/detached when unmaximizing a window when the setting keep_style_on_maximize is enabled pop-os/cosmic-panel#601.

This used to be the rendering order before push based rendering. The previous ordering can be seen in [031fdc3](https://github.com/pop-os/cosmic-comp/commit/031fdc389aaa4dac8546e0fc429bbea2295cc00d), before it changed in [e4c0716](https://github.com/pop-os/cosmic-comp/commit/e4c0716951fd69984c582720c4c4572c327b2493).

I searched for issues, PRs, etc. but did not find a reason why the order was changed for push based rendering.

https://github.com/user-attachments/assets/083dc0c9-128a-4a9e-995a-e50ca0c4ec3f

https://github.com/user-attachments/assets/ccb67dde-b583-48ab-aae0-5b9836596dda

___

### DIFF
--- a/src/shell/layout/floating/mod.rs
+++ b/src/shell/layout/floating/mod.rs
@@ -1522,49 +1522,6 @@ impl FloatingLayout {
                 .unwrap_or_else(|| (self.space.element_geometry(elem).unwrap().as_local(), alpha));
             let render_location = geometry.loc - elem.geometry().loc.as_local();
 
-            if focused == Some(elem) && !elem.is_maximized(false) {
-                let active_window_hint = crate::theme::active_window_hint(theme);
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
Thanks!

jacobgkau (approved):
This seems to make the active hint track windows more smoothly while they're being unmaximized (the hint now follows the window, instead of being stuck at the maximized size until after the window's done unmaximizing). Not seeing any problems with the hint or the resize indicators (which also appear to be part of this code block). Active hint customization still works as expected.

PR #2575, opened 2026-07-15
chore: clippy fixes

Drakulix:

Only non-functional code changes.

### DIFF
--- a/src/backend/kms/socket.rs
+++ b/src/backend/kms/socket.rs
@@ -36,7 +36,7 @@ impl Common {
     ) -> Result<Socket> {
         let socket_name = format!(
             "{}-{}",
-            &self.socket.to_string_lossy(),
+            self.socket.to_string_lossy(),
             render_node
                 .dev_path()
                 .unwrap()
--- a/src/shell/mod.rs
+++ b/src/shell/mod.rs
@@ -3880,11 +3880,11 @@ impl Shell {
                 (initial_window_location, layer, workspace.handle)
             } else {
                 let sticky_layer = self
-                .workspaces
-                .sets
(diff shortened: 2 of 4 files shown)


git-f0x:
#2306 would be needed to fully fix the CI, since it currently runs lints on 1.97 (similar is true for some other repos, both in COSMIC and Smithay), instead of 1.93 (MSRV).

Drakulix:
Hmm, this shouldn't be necessary. Clippy understands the MSRV and won't suggest fixes, that aren't compatible with that version and we want to have the newest set of lints available, because some do apply to older rust versions.

git-f0x:
Ah, ok. Though that PR should still be good if `+stable` is added back, since that action does seem to noticeably reduce CI time (it reduced it by about a minute in libcosmic).

mmstick (approved):

hojjatabdollahi:
But running 1.97 lints by default just creates churn, new contributions are failing ci, people need to figure out if it's their fault or a new lint somewhere else in the codebase, because a new rust version is released. When rust 1.98 is released all PRs probably fail ci until someone clears up the codebase again.

Iced also periodically does apply [newer lint](https://github.com/iced-rs/iced/commit/56b706fcd35aa881baede77a696bbb842773fdc6) but their [ci](https://github.com/iced-rs/iced/blob/master/.github/workflows/lint.yml) runs the same clippy as the current toolchain version.

Of course you're the maintainer and it makes sense to stay on top of the latest lints, just as a contributor I hate seeing that failing ci! :fearful:

wash2 (approved):

jklgrasso (approved):

PR #2584, opened 2026-07-16
fix: release active popup grabs on toplevel destruction. If a window …

littlezabi:
### The Problem

When a client window is destroyed, any active Wayland popup grabs associated with the seat are not automatically cleared. The compositor remains in a grab state, redirecting all keyboard and mouse inputs to a non-existent surface. The user's desktop becomes unresponsive (mouse cursor still moves, but clicks are ignored and keyboard shortcuts stop working).

### How to Reproduce
1. Open a terminal and launch several background processes (e.g., local development servers/workers).
2. Abruptly close the terminal or kill the terminal processes all at once.
3. This triggers a client process crash, causing the system's crash reporter dialog (`apport-gtk` or similar) to spawn a modal popup.
4. Because the grab state was orphaned during the abrupt client close, the input focus gets permanently locked, rendering the desktop session frozen.

I submitted a bug report three weeks ago via the Pop! OS github repository.
check [https://github.com/pop-os/pop/issues/4030](https://github.com/pop-os/pop/issues/4030)

### Video Demonstration on testing device 

https://github.com/user-attachments/assets/fbea36b3-1147-44d9-95ba-760723e69500

### System Specifications Summary

Device Model: Lenovo ThinkPad (Modal 20S00043UK)
Processor: Intel(R) Core(TM) i5-10210U 
RAM: 16 GB System Memory
GPU: Intel CometLake-U GT2 [UHD Graphics]
Display Interface: `/dev/fb0`
Storage: 512GB SanDisk SSD

Thank you guys! :) I really like Pop os and appreciate your works

### DIFF
--- a/src/wayland/handlers/xdg_shell/mod.rs
+++ b/src/wayland/handlers/xdg_shell/mod.rs
@@ -317,8 +317,45 @@ impl XdgShellHandler for State {
     }
 
     fn toplevel_destroyed(&mut self, surface: ToplevelSurface) {
+        for (popup, _) in smithay::desktop::PopupManager::popups_for_surface(surface.wl_surface()) {
+            if let smithay::desktop::PopupKind::Xdg(ref xdg_popup) = popup {
+                xdg_popup.send_popup_done();
+            }
+        }
+
         let (output, clients) = {
             let mut shell = self.common.shell.write();
+
+            for seat in shell.seats.iter() {
+                if let Some(data) = seat.user_data().get::<PopupGrabData>() {
+                    let mut should_ungrab = false;
+                    let grab = data.take();
+                    if let Some(ref grab) = grab {
+                        if grab.has_ended() {
+                            should_ungrab = true;
+                        } else if let Some(target) = grab.current_grab() {
+                            if let Some(wl_surface) = target.wl_surface() {
+                                if wl_surface.as_ref() == surface.wl_surface()
+                                    || smithay::desktop::PopupManager::popups_for_surface(
+                                        surface.wl_surface(),
+                                    )
+                                    .any(|(p, _)| p.wl_surface() == wl_surface.as_ref())
+                                {
+                                    should_ungrab = true;
+                                }
+                            }
+                        }
+                    }
+                    if should_ungrab {
+                        if let Some(mut grab) = grab {
+                            grab.ungrab(PopupUngrabStrategy::All);
+                        }
+                    } else {
+                        data.set(grab);
+                    }
+                }
+            }
+
             let seat = shell.seats.last_active().clone();
 
             // Clean up pending_windows for surfaces that were never mapped.


src/wayland/handlers/xdg_shell/mod.rs
```diff
@@ -317,6 +317,23 @@ impl XdgShellHandler for State {
...
+        }
+
+        // Release any active popup grabs on the last active seat to prevent
+        // input focus from getting permanently stuck if the window was closed/crashed.
+        let seat = self.common.shell.read().seats.last_active().clone();
+        if let Some(data) = seat.user_data().get::<PopupGrabData>()
+            && let Some(mut grab) = data.take()
+        {
+            grab.ungrab(PopupUngrabStrategy::All);
+        }
```
Drakulix: Hmm, pretty sure this would remove any popup grab on any toplevel description, which isn't really what we want here.

Can you please change this code to:
1. Iterate through all seat, we might want to add multi-seat support at some point.
2. Compare `grab.current_grab().wl_surface()` with `surface.wl_surface()` before removing the GrabData and calling ungrab?

Drakulix (commented):
Thanks for tracking this down, couple of remarks:

src/wayland/handlers/xdg_shell/mod.rs
littlezabi: Thanks for the suggestion. I had experimented with a few different approaches, but they were causing some unexpected behavior. I'll implement your recommended changes, test them thoroughly, and update the PR.

littlezabi: I've updated
* Iterates over all seats.
* Checks if the active grab's surface matches either the destroyed toplevel window itself or one of its active child popups before releasing the grab.

src/wayland/handlers/xdg_shell/mod.rs
```diff
@@ -317,8 +317,42 @@ impl XdgShellHandler for State {
...
+
+        
         let (output, clients) = {
             let mut shell = self.common.shell.write();
+            
+            for seat in shell.seats.iter() {
+                if let Some(data) = seat.user_data().get::<PopupGrabData>() {
+                    let mut should_ungrab = false;
+                    data.set(data.take().map(|grab| {
```
Drakulix: This is a little ugly. Please use `data.update` instead of `data.set(data.take())`

Drakulix (commented):
Just one little nitpick. Looks good otherwise! Thanks :)

src/wayland/handlers/xdg_shell/mod.rs
littlezabi: Ok i'm on it.

littlezabi: Thanks for the suggestion! I tried using data.update, but it fails to compile because PopupGrab does not implement Copy.
```
error[E0599]: the method `update` exists for reference `&Cell<Option<PopupGrab<State>>>`, but its trait bounds were not satisfied
  |
  = note: the following trait bounds were not satisfied:
          PopupGrab<state::State>: std::marker::Copy
          which is required by `std::option::Option<PopupGrab<state::State>>: std::marker::Copy`
```
          
 Instead, I used data.take() and data.set(), which is the standard way in Rust to safely update non-Copy data inside a Cell.

Drakulix (dismissed):

jklgrasso (dismissed):

Drakulix:
@littlezabi Can you please run `cargo fmt` on your PR?

littlezabi:
> @littlezabi Can you please run `cargo fmt` on your PR?
All formatted now.

Drakulix (approved):

jklgrasso (approved):

PR #2585, opened 2026-07-16
Fix flicker when switching between menus libcosmic apps in tiling mode

hojjatabdollahi:
Tiling mode uses keyboard focus to draw focus hint, while floating mode uses focus stack.

Sometimes when you open a pop up (for example menus in cosmic-files menu bar) when you're going from one menu to the other menu, for a moment, the system is destroying the old popup and creating a new one, and the keyboard focus is on nothing. That causes the focus hint to disappear for a moment, creating a flicker effect.

This took me a while to figure out. And I'm not sure if the solution below is the best solution. But it does fix the problem.

To reproduce the issue:
- open a cosmic-term
- open a cosmic-files
- make sure you're in tiling mode
- click on one of them and open a menu and try to move the cursor to a neighboring menu
- then click on the other one and do the same
(The issue normally happens the first time that you focus a window and between the first two menus that you open).

You can see it looks like the window is losing focus when I go from one menu to another (Sort to View in cosmic-files):

https://github.com/user-attachments/assets/8f9efc1b-7c56-4c38-bfe4-81d9855cbe72

___

### DIFF
--- a/src/shell/layout/tiling/mod.rs
+++ b/src/shell/layout/tiling/mod.rs
@@ -4013,6 +4013,7 @@ impl TilingLayout {
         &self,
         renderer: &mut R,
         seat: Option<&Seat<State>>,
+        focused: Option<&CosmicMapped>,
         non_exclusive_zone: Rectangle<i32, Local>,
         overview: (OverviewMode, Option<(SwapIndicator, Option<&Tree<Data>>)>),
         resize_indicator: Option<(ResizeMode, ResizeIndicator)>,
@@ -4143,6 +4144,7 @@ impl TilingLayout {
             old_geometries,
             is_overview,
             seat,
+            focused,
             &self.output,
             percentage,
             draw_groups,
@@ -5276,6 +5278,7 @@ fn render_new_tree_windows<R>(
     old_geometries: Option<HashMap<NodeId, Rectangle<i32, Local>>>,
     is_overview: bool,
     seat: Option<&Seat<State>>,
+    focused: Option<&CosmicMapped>,
     output: &Output,
     percentage: f32,
     transition: Option<f32>,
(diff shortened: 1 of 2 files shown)


Drakulix:
No the focus fallback isn't super nice.

I think we should simply also use the focus stack in the tiling code.

hojjatabdollahi:
@Drakulix I changed it to be mirroring floating mode (except for focused groups which floating mode doesn't have).

While testing to make sure my changes didn't break focus logic, I noticed that opening workspaces changes focused window. I opened an [issue](https://github.com/pop-os/cosmic-workspaces-epoch/issues/320). I also added a [commit](https://github.com/pop-os/cosmic-comp/pull/2585/commits/0402100c98b8cb52708ba3861f9556721dcf5794) that fixes it. But I don't like it, it feels hacky. I can remove it if you want, but I don't know what's a better solution to that bug?

PS. I asked in [mattermost](https://chat.pop-os.org/pop-os/pl/wbjyujyi8fds9dm7n3grq83kgy) about system76's plan for workspaces, would appreciate if you leave your feedback so I don't waste time on hacky solution.

Drakulix:
> @Drakulix I changed it to be mirroring floating mode (except for focused groups which floating mode doesn't have).

That first commit looks good!

> While testing to make sure my changes didn't break focus logic, I noticed that opening workspaces changes focused window. I opened an [issue](https://github.com/pop-os/cosmic-workspaces-epoch/issues/320). I also added a [commit](https://github.com/pop-os/cosmic-comp/pull/2585/commits/0402100c98b8cb52708ba3861f9556721dcf5794) that fixes it. But I don't like it, it feels hacky. I can remove it if you want, but I don't know what's a better solution to that bug?

I would drop that commit for now, while we think about a better solution. Thanks for the report though!

> PS. I asked in [mattermost](https://chat.pop-os.org/pop-os/pl/wbjyujyi8fds9dm7n3grq83kgy) about system76's plan for workspaces, would appreciate if you leave your feedback so I don't waste time on hacky solution.

I would like to come back to this in a couple of weeks. We have an engineering meetup planned, where the workspace overview might be a discussion topic of it's own. At the moment I cannot give you a clear answer.

hojjatabdollahi:
I removed that commit.

Drakulix (approved):

jklgrasso (approved):

Issue #248, opened 2023-12-16
Wayland tearing protocol support


Jontique:
Hey.

I was wondering if COSMIC will support the Wayland tearing protocol, at very least for fullscreen applications.

Allowing screen tearing makes a huge difference in reducing input latency in FPS games and games like Osu!

https://wayland.app/protocols/tearing-control-v1

Drakulix:
Yes, it is on the TODO list and we will eventually support it. **But** it's not a very high priority.

LucasGGamerM:
Where can one see the TODO list? Is it implemented on the alpha?

Drakulix:
If it would be implemented this issue wouldn't be open any more.
You can take a look at the current prioritization here: https://github.com/orgs/pop-os/projects/23
Any open issue is part of the hypothetical "TODO" list, but those are the issues we actively working towards at the moment.

FinGu:
Bumping this.. ( would also love for an option to always enable tearing on fullscreen apps )

AHuman-Person:
Chiming in as someone trying to switch to popos from win11. Having like 300 quadrillion years of input latency really makes gaming not worth it on pop os(geometry dash specifically, ever single frame counts, especially on 60hz)

gabriele2000:
> Chiming in as someone trying to switch to popos from win11. Having like 300 quadrillion years of input latency really makes gaming not worth it on pop os(geometry dash specifically, ever single frame counts, especially on 60hz)

Funny: I've never had issues with latency, could be because I have a 120Hz screen, even though I play at 60FPS?
Now I'm really wondering if that's the case, since I've never had latency issues.

EmanuelKroell:
Hi, missing support for tearing is currently (and has been for almost 2 years now) the biggest issue keepinge from using Cosmic or another smithay based compositor like Niri.
I hope that especially with the influx of people who primarily want to play games coming over from windows, tearing would be a higher priority/at least be added to the roadmap at some point.

Jontique:
Hello!
Seeing this had gained some reactions and comments over the years, I am requesting an update on whether if this is still planned. Or perhaps prioritizing it higher, as it is a quite vital for low input latency-dependent games. 

Also adding the Gaming tag on the issue would be great, so it is grouped with the rest of the gaming-related improvements at: https://github.com/pop-os/cosmic-comp/issues?q=is%3Aissue%20state%3Aopen%20label%3Agaming

PR #2588, opened 2026-07-17
Implement wp_tearing_control_v1 (composited tearing)

s0kil:
## What

Implements `wp_tearing_control_v1` with **composited tearing** — async page flips of the compositor's own composited buffer. Addresses #248.

- Protocol: `wp_tearing_control_manager_v1` (v1) global; per-surface double-buffered presentation hint (promoted on commit, `tearing_control_exists` error on duplicate controls, destroy resets pending to vsync).
- Policy: an output enters tearing mode iff an opt-in toggle is set (`COSMIC_ALLOW_TEARING=1` env for now), the output is not mirroring, and the focused fullscreen toplevel's committed hint is `async`.
- Frame clock: in tearing mode the estimated-vblank throttle is bypassed — composite is driven by client commits (rate-ceiled ~500fps) and submitted with `FrameFlags::ALLOW_ASYNC_PAGE_FLIP`; frame callbacks are sent immediately, so clients run uncapped.
- Cursor plane is disabled in tearing mode (cursor composites instead): kernel atomic async flips may only change the primary plane FB, so this keeps every async commit valid.
- Depends on smithay async-flip support: Smithay/smithay#2099 (Cargo patch currently points at that branch).

## Why composited tearing

On muxless hybrid laptops (game renders on the dGPU, iGPU scans out the internal panel) direct scanout of client buffers is impossible — the classic "tearing only for direct-scanout fullscreen" design delivers nothing on such machines. Async-flipping the composited buffer does: validated end-to-end on an i915-scanout + NVIDIA-render laptop, CS2 (native Wayland, fullscreen) runs 240fps on a 144Hz eDP with visible tearing, where it was previously FIFO-capped at 144. Compositor cost went *down* (composite-per-commit at ~240/s measured cheaper than the previous 144Hz vsync path in CPU time). Zero rejected async flips observed on i915.

## Current limitations (draft status — seeking direction)

- Composited-tearing only; the direct-scanout fullscreen path does not async-flip yet (planned follow-up — needed for single-GPU desktops).
- Single-output policy; no VRR interaction handling yet.
- Toggle is an env var; should become cosmic-config + Settings UI.
- Xwayland (24.1.12) did not forward tearing hints for an X11 Vulkan IMMEDIATE client on the NVIDIA stack during testing — under investigation whether that's driver- or Xwayland-side; native Wayland clients work.
- No `DRM_CAP_ATOMIC_ASYNC_PAGE_FLIP` gating yet (tracked in the smithay PR).

Implementation was developed with AI assistance and human review/testing on real hardware; happy to iterate per maintainer direction and split/reshape as preferred.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4813,7 +4813,6 @@ checksum = "8ed6a63f02c8539c91a8685a86f4099661ba3da017932f6ebbea6de3f0fa7c90"
 [[package]]
 name = "smithay"
 version = "0.7.0"
-source = "git+https://github.com/smithay/smithay.git?rev=efeb597#efeb59768eb9a2d8dae9f71300a56aef06b0c1c4"
 dependencies = [
  "aliasable",
  "appendlist",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1194,7 +1194,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users",
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -1508,7 +1508,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


Drakulix:
Three things:

- First off all the PR template is not optional
- Secondly this feature should be implemented in smithay and is already being worked on
- Thirdly, I am not sure if the approach taken for composited tearing is what we want, since we still end up synchronizing on copying the client buffer from one gpu to the other. This requires further work in smithay to allow copying a potentially not-ready buffer *and* then allowing to submit the composited result asynchronously. Otherwise you aren't gaining anything in terms of latency, which is the whole goal for allowing teared presentation.

s0kil:
@Drakulix thank you for the review.

PR #2589, opened 2026-07-17
render: `cgmath` -> `glam`

git-f0x:

Updates smithay to include https://github.com/Smithay/smithay/pull/2095.
Edit: Hmm, maybe glam doesn't need the `float-types` feature, since that's only for f64 (DVec, DMat,...)? Smithay might not need it either. Or should maybe f64 types be used as long as possible and then cast into their f32 counterparts (with e.g. `as_mat3()`)? Or maybe using affine types where possible (e.g. Affine2 is recommended over Mat3 for 2D affine transformations)?

Should I fix clippy in a separate commit (it's a one line change)?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -792,7 +792,6 @@ dependencies = [
  "anyhow",
  "bitflags 2.11.0",
  "calloop",
- "cgmath",
  "clap_lex",
  "cosmic-comp-config",
  "cosmic-config",
@@ -804,6 +803,7 @@ dependencies = [
  "egui_plot",
  "futures-executor",
  "futures-util",
+ "glam 0.33.2",
  "i18n-embed",
  "i18n-embed-fl",
  "iced_graphics",
@@ -1194,7 +1194,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users",
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


Drakulix (dismissed):

Drakulix:
> Edit: Hmm, maybe glam doesn't need the `float-types` feature, since that's only for f64 (DVec, DMat,...)? Smithay might not need it either. Or should maybe f64 types be used as long as possible and then cast into their f32 counterparts (with e.g. `as_mat3()`)? Or maybe using affine types where possible (e.g. Affine2 is recommended over Mat3 for 2D affine transformations)?

Feel free to remove the feature then. I am not aware of any bugs that would be fixed by the higher precision and I thus would rather merge a 1-to-1 match of our current code given this is quite critical.

> Should I fix clippy in a separate commit (it's a one line change)?

Please feel free to.

git-f0x:
We do lose some precision with the port in `shadow.rs`, since there's casting to f32 that wasn't there before. Is that fine?

I think `Affine2` could be used in some places in Smithay as well, since it's more efficient for 2D (and compatible with `Mat3`). But that's probably out of scope for now.

Drakulix:
> We do lose some precision with the port in `shadow.rs`, since there's casting to f32 that wasn't there before. Is that fine?

Right. Well the only way to know for sure is to run it, but my gut feeling is that this is fine. If it actually causes problems on some highly specific resolution+scaling, I am sure we'll have a bug report soon. :sweat_smile: 

> I think `Affine2` could be used in some places in Smithay as well, since it's more efficient for 2D (and compatible with `Mat3`). But that's probably out of scope for now.

Yeah I think it makes sense to adopt glam quickly to not introduce bit-rot and then incrementally improve it. Feel free to work on that though, I definitely agree.

Drakulix (dismissed):

git-f0x:
Sorry!

Drakulix (approved):

jklgrasso (approved):

Issue #2407, opened 2026-05-25
[Bug] Edge barrier between multiple monitors works only in one direction


SammyMcFly:
On a multi-monitor setup the edge barrier or sticky edges that is supposed to keep the mouse cursor from changing monitor too easily works only in one direction.

**Expected behavior**
Either there should be no edge barrier or it should work in both directions.

**What actually happens**
The cursor is prevented from moving from monitor one to monitor two as long as the mouse speed is low, but not from monitor two to monitor one.

**Comments**
- I use one 4K and one 1440p monitor alongside each other with 150% and 100% scaling, respectively.

Solarunit:
>The cursor is prevented from moving from monitor one to monitor two as long as the mouse speed is low, but not from monitor two to monitor one.

Interesting, I haven't noticed this at all. 
I made a related feature request recently to disable cursor transition between monitors https://github.com/pop-os/cosmic-comp/issues/2404

szpolny:
Same issue with 1080p monitor on the left and 2k on right. Each on 100% scaling.

Software and hardware:
Cosmic:  1.0.16
NixOS: 26.11.20260612.49a4bd0 (Zokor)
Kernel: 7.0.12-zen1
Nvidia Driver: 595.80
GPU: GeForce RTX 4070 Ti SUPER

SammyMcFly:
I just found out that the issue might be even trickier to reproduce. The issue occurred with the following setup:
- 4K monitor: connected with HDMI cable
- 1440p monitor: connected with DP cable
It did not occur when both were connected with DP cables. The graphics card is an Intel Arc B580.

PR #2590, opened 2026-07-17
fix(input): stop cursor sticking to screen edge during cross monitor movement

### LINKED ISSUE #2102: Cursor getting stuck at display edge when moving from left to right

k-start:
This PR fixes an issue where the cursor gets stuck on the edge of a display when moving across monitors. This occurs when the mouse is moved slowly from a monitor on the left to a monitor on the right, causing the cursor to get stuck on the edge of the left monitor. The same occurs from top to bottom.

The bug is caused when the coordinates (`f64`) land on the last pixel of a display (e.g. `1919.5`). Since the coordinate is still on Screen A and not yet crossing over to Screen B, it gets clamped back to `1919.0`. The cursor gets stuck in this zone between `size - 1.0` and `size` until a large enough movement occurs to break it out.

This PR fixes this by only clamping the cursor coordinates when the cursor is fully off-screen. If `matching_output` is not `None`, we skip the clamping since the cursor is still in bounds of our monitors. If it is `None`, we clamp it to the edge of the screen.

Fixes #2102 #2407

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -375,21 +375,24 @@ impl State {
                     let original_position = position;
                     position += event.delta().as_global();
                     let shell = self.common.shell.read();
-                    let output = shell
+                    let matching_output = shell
                         .outputs()
                         .find(|output| output.geometry().to_f64().contains(position))
-                        .cloned()
-                        .unwrap_or(current_output.clone());
+                        .cloned();
                     drop(shell);
+                    let needs_clamping = matching_output.is_none();
+                    let output = matching_output.unwrap_or_else(|| current_output.clone());
                     let output_geometry = output.geometry();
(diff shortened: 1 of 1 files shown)


Drakulix:
See https://github.com/pop-os/cosmic-comp/pull/2568

k-start:
Yep fair, I only saw a stale PR fix when I was looking earlier. Will close this in favour of that fix!

PR #2595, opened 2026-07-18
Disable blur for cursor elements

git-f0x:

Blur strength was set to non-zero values for cursor elements in most places, and I can't think of a reason why they would need blur, so I removed it.
I can't notice any performance difference or anything, so maybe it was already handled somewhere, but less function arguments is also nice. :)

### DIFF
--- a/src/backend/render/cursor.rs
+++ b/src/backend/render/cursor.rs
@@ -185,7 +185,6 @@ pub fn draw_surface_cursor<R>(
     surface: &wl_surface::WlSurface,
     location: Point<f64, Logical>,
     scale: impl Into<Scale<f64>>,
-    blur_strength: usize,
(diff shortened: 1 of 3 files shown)


Drakulix:
> Blur strength was set to non-zero values for cursor elements in most places, and I can't think of a reason why they would need blur, so I removed it. I can't notice any performance difference or anything, so maybe it was already handled somewhere, but less function arguments is also nice. :)

Just because a blur-strength is set, doesn't mean the surface gets blur, if it doesn't request it. However technically the wayland protocol allows cursor surfaces to have background-effects (which includes blur), so I don't really see a reason to explicitly disable that?

git-f0x:
Ah, yeah, makes sense. Closing. :)

PR #2620, opened 2026-07-23
fix: ensure X window focus on unmap/map

RipleyTom:
This fixes some games where you have to alt-tab out and back in to get input focus.

There is an issue caused by basically two things:
- a race on whether refresh_focus runs between unmap/map
- the fact that we suppress(rightfully so) Element(W) to Fullscreen(W) focus changes

On unmap, the X server input is lost by the window, in cosmic-comp it removes it from the focus_stack.

Now 2 things can happen:
1) refresh_focus runs after the unmap but before the map, it finds that keyboard.current_focus() refers to something not in the focus_stack, it corrects the focus(ie another window gets it), then when the window maps, KeyboardTarget::enter is called => conn.set_input_focus corrects the X server internal focus. The game correctly has input.
2) refresh_focus doesn't run between the unmap and the map, Fullscreen(W) is put in the stack, when it maps and update focus, all it sees is a transition between Element(W) and Fullscreen(W) which is suppressed in KeyboardTarget::replace. From cosmic-comp's perspective the game has focus, from the X server's pov it lost focus when unmapping and never regained it.

So to fix it we force a transition(only in the case of map and if the mapped window had focus in the first place) by transitioning focus to None then back to the window.

### DIFF
--- a/src/shell/focus/mod.rs
+++ b/src/shell/focus/mod.rs
@@ -212,6 +212,29 @@ impl Shell {
         state.common.shell.write().update_active();
     }
 
+    // We suppress Element(X) to Fullscreen(X) transition to avoid
+    // loss of focus by the X window when having a transition to fullscreen
+    // but in the case of X11 unmap/map the leave/enter needs to happen for the X11
+    // internal state to be focused on the window
+    pub fn set_focus_on_x11_map(
+        state: &mut State,
+        target: &KeyboardFocusTarget,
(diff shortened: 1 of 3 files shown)


Drakulix (approved):
Thanks for tracking that down. Focus behavior especially for games has been surprisingly tricky.

jacobktm (approved):

PR #2622, opened 2026-07-24
fix: dbus: deadlock in a11y keyboard monitor

tgrushka:
## Problem

`A11yKeyboardMonitorState::key_event` calls `self.has_key_grab()` inside `self.clients` guard loop.

A "well-behaved" a11y client calls `WatchKeyboard` before `SetKeyGrabs`.

In the existing code, with such "well-behaved" clients,`!client.watched` short-circuits the call to `self.has_key_grab()` and thus prevents a double-lock.

If `WatchKeyboard` is not called first (poorly written client, no error or warning to the client developer other than this freeze, and **if** they test it on COSMIC), the `!client.watched` check does not short-circuit, and `self.has_key_grab()` is called **inside the existing guard**, which then tries to lock `self.clients` again, causing the deadlock.

This causes the compositor to freeze, requiring a hard reboot.

## Fix

This commit moves the `self.has_key_grab()` before the guard to prevent a double lock.

## Background

I discovered this bug/freeze while writing my `say-selection` screen reader to work with the new a11y KeyboardMonitor DBus API (**thanks COSMIC!**).

While prototyping, I failed to call `WatchKeyboard` before `SetKeyGrabs` and discovered this deadlock when COSMIC froze and I had to hard-reset.

I used AI to research the root cause, but chose the fix approach and wrote the code myself.

### DIFF
--- a/src/dbus/a11y_keyboard_monitor.rs
+++ b/src/dbus/a11y_keyboard_monitor.rs
@@ -150,9 +150,10 @@ impl A11yKeyboardMonitorState {
     }
 
     pub fn key_event(&self, modifiers: &ModifiersState, keysym: &KeysymHandle, state: KeyState) {
+        let has_key_grab = self.has_key_grab(modifiers, keysym.modified_sym());
         let clients = self.clients.lock().unwrap();
         for (unique_name, client) in clients.0.iter() {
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
Nice catch! Thanks :)

jacobktm (approved):
I was able to recreate this freeze and verify that this PR does correct it.

PR #2623, opened 2026-07-24
fix: workaround for Chromium apps breaking below 1.0 scaling

RipleyTom:
This is a fix for chromium apps running under Wayland breaking below 1.0 scaling(ie at 50% and 75%).
Note that I called the commit a fix but it is really a workaround for an issue that is not ours.
Chromium apps under Wayland seem to break when under 100%.
Other applications like Firefox using Wayland seem to work fine under those scales.

I suspect this has something to do with this clamp in Chromium but I couldn't find the exact logic chain that breaks:
https://chromium.googlesource.com/chromium/src/+/refs/heads/main/ui/ozone/platform/wayland/common/wayland_util.cc#329

For scales >= 100% this doesn't change anything.
For scales < 100%, scaling will still works but the surface buffer will be bigger(up to 4x for 50% compared to what it should be) and the surface will be downscaled when rendered(extra GPU work). This is obviously not ideal but I think the tradeoff of running Chromium apps until it gets fixed is worth it.

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -956,8 +956,11 @@ impl Common {
                 );
                 if let Some(output) = primary_scanout_output {
                     with_fractional_scale(states, |fraction_scale| {
-                        fraction_scale
-                            .set_preferred_scale(output.current_scale().fractional_scale());
+                        // The 1.0 clamp is a workaround for Chromium
+                        // TODO: remove if Chromium ever gets fixed
+                        fraction_scale.set_preferred_scale(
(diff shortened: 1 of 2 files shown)


Drakulix (approved):
yeah I wanted to look into this anyway, because I don't think any app was ever really tested for < 1.0 scaling.

So imo we should always clamp to 1 anyway.

jacobktm (approved):
Chromium apps behave correctly after applying this PR.

PR #2638, opened 2026-07-26
fix(corner_radius): only error on radius values the client submitted

zyads:
### The Problem

`cosmic-comp` kills clients with a fatal Wayland protocol error over a purely cosmetic corner-radius hint, in two situations the client cannot avoid.

The radius and padding hints are validated in a **pre-commit hook**, which runs on *every* commit of the surface — not only on the commit that carries a new `set_radius`/`set_padding` request. As a result:

1. **A window that is resized smaller is killed.** A client sets a legal radius (say 20px on a 400x400 window). Later the compositor resizes it — tiling, maximising, snapping, a resize animation. The client acks the configure and commits its new, smaller window geometry. The hook re-validates the *old* radius against the *new* geometry, decides it is too large, and posts `radius_too_large`. The client never sent a new request and had no way to prevent this.

2. **A layer surface that is not mapped yet is killed.** `layer_radius_hook` validates against `bbox_from_surface_tree`, which reports an empty rectangle for a surface with no buffer. `half_min_dim` is then `0`, so *every* non-zero radius counts as too large. The xdg hook does not have this problem because it only validates when window geometry is set.

Because `post_error` tears down the whole connection, a single stale hint takes down everything that client owns.

### Impact

Not theoretical. On my machine (Pop!_OS 24.04, 4K display, `cosmic-comp` 693d1ad) these two errors account for **349,861 of the 392,958** Wayland protocol errors in the last 7 days of my journal — 89% of all of them.

**`cosmic-panel` stops drawing and never recovers (case 1).** The panel's Wayland connection is killed, but the process does not exit, so `cosmic-session` never respawns it. The panel and every applet stay alive and healthy and simply never render again. 345,220 lines of

```
Protocol error 1 on object cosmic_corner_radius_toplevel_v1@375
```

at a sustained ~20,000/minute for 15 minutes, stopping when `cosmic-panel` was killed by hand and respawned by the session. The panel-side "alive but not rendering, so it is never restarted" behaviour is pop-os/cosmic-panel#635; this is one of the triggers for it.

**`xdg-desktop-portal-cosmic` dies, taking the clipboard with it (case 2).** The portal's screenshot overlay is a layer surface built with `size_limits: Limits::NONE.min_height(1.0).min_width(1.0)` (`src/screenshot.rs`), so it can legitimately be 1x1 before it is configured. From my journal:

```
xdg-desktop-portal-cosmic[852701]: cosmic_corner_radius_layer_v1#1087: error 1: ... corner radius too large
xdg-desktop-portal-cosmic[852701]: SCTK failed to send Control::AboutToWait. TrySendError { kind: Disconnected }
xdg-desktop-portal-cosmic[852701]: Error: EventLoop(ExitFailure(1))
systemd[1638]: dbus-...portal.desktop.cosmic@1.service: Main process exited, code=exited, status=1/FAILURE
```

The portal owns the `wl_data_offer` for a screenshot, so when it dies the screenshot can no longer be pasted. This looks like a contributor to the recurring "screenshot sometimes can't be pasted" reports (e.g. pop-os/cosmic-screenshot#224).

### How to Reproduce

Minimal client, no special setup — full source: https://gist.github.com/zyads/d07ee40c32de6a162dbaae96fe5cbb28

1. Map a 400x400 `xdg_toplevel` with `set_window_geometry(0, 0, 400, 400)`.
2. `cosmic_corner_radius_toplevel_v1.set_radius(20, 20, 20, 20)` and commit. Accepted, correctly — 20 <= 400/2.
3. **Without touching the radius object again**, `set_window_geometry(0, 0, 400, 20)` and commit.

Result on current `main`:

```
step 1: mapped a 400x400 toplevel
step 2: set_radius(20,20,20,20) on a 400x400 window - accepted, as it should be
step 3: committing a smaller window geometry (400x20). Radius NOT touched.
RESULT: KILLED by the compositor
  protocol error: code=1 interface=cosmic_corner_radius_toplevel_v1 object=12
                  message="... corner radius too large"
```

Step 3 is exactly what every client does when it acks a compositor-initiated resize.

### The Fix

Only raise a protocol error for state the **client actually submitted in the commit being processed**.

- A small `Submitted { corners, padding }` marker is stored in the surface's `data_map`, set by `set_radius`/`set_padding`, and read-and-cleared once per commit by the hooks. A value is therefore only ever validated against the geometry it was submitted alongside.
- `layer_radius_hook` additionally skips surfaces whose padded box has no area, mirroring the xdg hook's existing "no window geometry yet" behaviour.

Genuinely bad clients — ones that submit an over-large radius for the geometry they are committing — still get the protocol error, so protocol conformance is unchanged for the case the error exists for.

Nothing is left unprotected: `surface_corners()` and `surface_padding()` in `src/wayland/handlers/corner_radius.rs` already clamp to `half_min_dim` when rendering, with the comment *"guard against corner radius being too large, potentially disconnecting the outline"*. The rendering path was always safe; the fatal error was redundant.

### Verification

Built and checked on Rust 1.93 / Ubuntu 24.04: `cargo fmt --all -- --check` clean and `cargo build --release` clean. `cargo clippy --all-features -- -D warnings` reports no lint from the changed file — the four it does report are pre-existing on `master` under 1.93's clippy (`needless_borrow` in `src/input/mod.rs`, `collapsible_if` in `src/wayland/handlers/xdg_shell/mod.rs`), both untouched here.

Behaviour, checked by running the reproducer against a **patched `cosmic-comp` running nested** (`COSMIC_BACKEND=winit`) and against an unpatched one:

| | unpatched | patched |
|---|---|---|
| shrink geometry, radius untouched | **killed**, `radius_too_large` | survives |
| `set_radius(300,…)` on a 400x400 window (`--conformance`) | rejected | **still rejected** |

So the false positive is gone and the error still fires for the case it exists for.

I did not run a patched compositor as my real session, so I have not directly watched the panel survive over a long session. The evidence for that half is the journal correlation above plus the reproducer.

### Note

I also noticed, but deliberately did **not** touch in this PR, that `GetCornerRadiusLayer` checks `ToplevelHookId` rather than `LayerHookId` when deciding whether a corner-radius object already exists for a layer surface (`corner_radius.rs`, `GetCornerRadiusLayer` arm). That looks like a copy-paste slip, but it is a separate issue and I did not want to mix it in. Happy to file it separately.

___

### DIFF
--- a/src/wayland/protocols/corner_radius.rs
+++ b/src/wayland/protocols/corner_radius.rs
@@ -12,6 +12,7 @@ use smithay::reexports::wayland_server::New;
 use smithay::reexports::wayland_server::protocol::wl_surface::WlSurface;
 use smithay::utils::{HookId, Logical, Point, Rectangle};
 use smithay::wayland::compositor::Cacheable;
+use smithay::wayland::compositor::SurfaceData;
 use smithay::wayland::compositor::add_pre_commit_hook;
 use smithay::wayland::compositor::with_states;
 use smithay::wayland::shell::wlr_layer::WlrLayerShellHandler;
@@ -28,6 +29,49 @@ use wayland_backend::server::GlobalId;
 
 type ToplevelHookId = Mutex<Option<(HookId, Weak<CosmicCornerRadiusToplevelV1>)>>;
 type LayerHookId = Mutex<Option<(HookId, Weak<CosmicCornerRadiusLayerV1>)>>;
+
+/// Which of the double-buffered corner-radius values the client itself
+/// submitted since the surface was last committed.
+///
+/// The radius and padding hints are validated against the surface geometry in
+/// a pre-commit hook, which runs on *every* commit. Without this marker a
+/// client would be killed whenever a compositor-initiated resize (tiling,
+/// maximise, snapping, a resize animation) made a previously legal radius too
+/// large for the new geometry, even though the client never sent a new
+/// request and had no way to avoid it. Only state the client actually
+/// submitted may raise a protocol error.
+#[derive(Debug, Default, Clone, Copy)]
+struct Submitted {
+    corners: bool,
+    padding: bool,
+}
+
+type SubmittedState = Mutex<Submitted>;
+
+fn mark_corners_submitted(surface_data: &SurfaceData) {
+    let submitted = surface_data
+        .data_map
+        .get_or_insert_threadsafe(SubmittedState::default);
+    submitted.lock().unwrap().corners = true;
+}
+
+fn mark_padding_submitted(surface_data: &SurfaceData) {
+    let submitted = surface_data
+        .data_map
+        .get_or_insert_threadsafe(SubmittedState::default);
+    submitted.lock().unwrap().padding = true;
+}
+
+/// Reads and clears the markers. Called once per commit from the pre-commit
+/// hooks, so a value is only ever validated against the geometry it was
+/// submitted with.
+fn take_submitted(surface_data: &SurfaceData) -> Submitted {
+    let submitted = surface_data
+        .data_map
+        .get_or_insert_threadsafe(SubmittedState::default);
+    let mut guard = submitted.lock().unwrap();
+    std::mem::take(&mut *guard)
+}
 
 #[derive(Debug)]
 pub struct CornerRadiusState {
@@ -374,6 +418,7 @@ where
                     let mut cached = s.cached_state.get::<CacheableCorners>();
                     let pending = cached.pending();
                     *pending = CacheableCorners(guard.corners);
+                    mark_corners_submitted(s);
                 });
                 drop(guard);
 
@@ -505,6 +550,7 @@ where
                     let mut cached = s.cached_state.get::<CacheableCorners>();
                     let pending = cached.pending();
                     *pending = CacheableCorners(guard.corners);
+                    mark_corners_submitted(s);
                 });
                 drop(guard);
 
@@ -568,6 +614,7 @@ where
                     let mut cached = s.cached_state.get::<CacheablePadding>();
                     let pending = cached.pending();
                     *pending = CacheablePadding(guard.padding);
+                    mark_padding_submitted(s);
                 });
                 drop(guard);
 
@@ -609,6 +656,14 @@ where
 
 fn xdg_radius_hook<D: 'static>(_state: &mut D, _dh: &DisplayHandle, surface: &WlSurface) {
     with_states(surface, |surface_data| {
+        // Only a radius the client submitted itself may be fatal. A radius
+        // that only became too large because the surface geometry shrank is
+        // clamped when rendering (see `surface_corners`), so there is nothing
+        // to protect against here and killing the client would be a bug.
+        if !take_submitted(surface_data).corners {
+            return;
+        }
+
         let corners = *surface_data
             .cached_state
             .get::<CacheableCorners>()
@@ -654,9 +709,32 @@ fn pad_rect(
     Some(rect)
 }
 
+fn post_layer_error(
+    surface_data: &SurfaceData,
+    error: cosmic_corner_radius_layer_v1::Error,
+    reason: &str,
+) {
+    if let Some(hook) = surface_data.data_map.get::<LayerHookId>() {
+        let hook_ref = hook.lock().unwrap();
+        if let Some((_, obj)) = hook_ref.as_ref()
+            && let Ok(obj) = obj.upgrade()
+        {
+            obj.post_error(error as u32, format!("{obj:?} {reason}"));
+        }
+    }
+}
+
 fn layer_radius_hook<D: 'static>(_state: &mut D, _dh: &DisplayHandle, surface: &WlSurface) {
     let bbox = bbox_from_surface_tree(surface, Point::default());
     with_states(surface, |surface_data| {
+        // Only values the client submitted itself may be fatal. Values that
+        // only became too large because the surface was resized are clamped
+        // when rendering (see `surface_corners` / `surface_padding`).
+        let submitted = take_submitted(surface_data);
+        if !submitted.corners && !submitted.padding {
+            return;
+        }
+
         let corners = *surface_data
             .cached_state
             .get::<CacheableCorners>()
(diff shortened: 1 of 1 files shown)


Drakulix:
> `cosmic-comp` kills clients with a fatal Wayland protocol error over a purely cosmetic corner-radius hint, in two situations the client cannot avoid.

It absolutely can in both cases.

> 
>     1. **A window that is resized smaller is killed.** A client sets a legal radius (say 20px on a 400x400 window). Later the compositor resizes it — tiling, maximising, snapping, a resize animation. The client acks the configure and commits its new, smaller window geometry. The hook re-validates the _old_ radius against the _new_ geometry, decides it is too large, and posts `radius_too_large`. The client never sent a new request and had no way to prevent this.

`ack_configure` is acknowledged on commit, so in the same commit the client can submit a new radius. That is what acknowledging the new size brings.

>     2. **A layer surface that is not mapped yet is killed.** `layer_radius_hook` validates against `bbox_from_surface_tree`, which reports an empty rectangle for a surface with no buffer. `half_min_dim` is then `0`, so _every_ non-zero radius counts as too large. The xdg hook does not have this problem because it only validates when window geometry is set.

This is a again a client bug. An unmapped surface should not have a corner size set, but the corner size should only be set, once a buffer is attached. On any commit that removes the buffer again (unmapping the surface), the client needs to unset the corner radius as well.

This is inline with other protocols and is why the atomic commit machinery exists. To prevent having any in-progress states visible to the user but only submit valid consistent visuals.

zyads:
Ah ok thanks for explaining. Since the compositor side is behaving as designed I'll close this.

But the two errors account for 349,861 of the 392,958 Wayland protocol errors in 7 days of my journal (89%), and the client is cosmic-panel. When it's killed the process doesn't exit, so cosmic-session never respawns it.. the panel and its applets stay alive and simply never render again until they're killed by hand. I logged 345,220 of

    Protocol error 1 on object cosmic_corner_radius_toplevel_v1@375

at ~20,000/minute for 15 minutes.

If that reads as a cosmic-panel bug to you, I'll file it there with the journal data. Happy to be pointed elsewhere if it belongs somewhere else.

Separately, I'll open a small issue for the GetCornerRadiusLayer arm looking up ToplevelHookId instead of LayerHookId, which I mentioned above and left out of this PR.

Thanks again! :)

Drakulix:
> Ah ok thanks for explaining. Since the compositor side is behaving as designed I'll close this.
> 
> But the two errors account for 349,861 of the 392,958 Wayland protocol errors in 7 days of my journal (89%), and the client is cosmic-panel. When it's killed the process doesn't exit, so cosmic-session never respawns it.. the panel and its applets stay alive and simply never render again until they're killed by hand. I logged 345,220 of
> 
> ```
> Protocol error 1 on object cosmic_corner_radius_toplevel_v1@375
> ```
> 
> at ~20,000/minute for 15 minutes.
> 
> If that reads as a cosmic-panel bug to you, I'll file it there with the journal data. Happy to be pointed elsewhere if it belongs somewhere else.
> 
> Separately, I'll open a small issue for the GetCornerRadiusLayer arm looking up ToplevelHookId instead of LayerHookId, which I mentioned above and left out of this PR.
> 
> Thanks again! :)

Do you perhaps have any custom applets? Potentially one of those is not handling this correctly and cosmic-panel simply forwards the request.

Ping @wash2 just in case.

zyads:
No custom applets here, all 18 are stock (CosmicAppletTime, WorkspacesButton, AppButton, InputSources, StatusArea, A11y, Tiling, Audio, Bluetooth, Network, Battery, Notifications, Power on the panel, plus LauncherButton, WorkspacesButton, AppButton, AppList and Minimize on the dock).

But I think the applet angle is a dead end anyway, and I should have caught this earlier. `cosmic_corner_radius_toplevel_v1` is an xdg toplevel, so those errors can't be coming from the panel or an applet at all, since those are layer surfaces. Splitting my 7 day journal by object type:

```
344,999   cosmic_corner_radius_toplevel_v1   (some app window)
  4,641   cosmic_corner_radius_layer_v1      (xdg-desktop-portal-cosmic)
```

The layer half is the portal, which has its own journal identifier and its own process. The toplevel half is just some ordinary window, and it shows up under cosmic-session because apps launched from the session inherit its stderr.

Which means I also need to correct my earlier comment. I said the client was cosmic-panel and the journal doesn't actually support that. Those lines carry `_COMM=cosmic-session, _EXE=/usr/bin/cosmic-session, PID 3012`, which is the shared stream every session child writes into, so anything would look identical. cosmic-panel has its own identifier and logged nothing during the storm. Sorry for stating it as fact.

I ran WAYLAND_DEBUG=1 against cosmic-term to see whether libcosmic handles the resize case, and it does:

```
-> cosmic_corner_radius_toplevel_v1#132.set_radius(12, 12, 12, 12)
-> xdg_surface#37.ack_configure(284098)
-> xdg_surface#37.ack_configure(284100)
-> cosmic_corner_radius_toplevel_v1#132.set_radius(0, 0, 0, 0)
```

It drops the radius to zero immediately after acking, exactly as you described. So the normal path is fine and whatever hit me is narrower than I made it sound.

I can't reproduce it right now either. Last error was 2026-07-26 00:12:41 in my previous session and the current one has been up about 19 hours with none. That isn't a fix though, corner_radius.rs is byte identical between 693d1ad and 4d370cd, so the trigger just hasn't come back.

One more correction: I wrote 693d1ad in the description but I'm actually on 4d370cd now, 49 commits later. The storm was on the older build.

Closing this since the compositor side is working as designed. If it comes back I'll catch the client properly and open a fresh issue against whatever it turns out to be. I've opened #2651 for the LayerHookId thing separately.

Thanks both for the time.

Drakulix:
> But I think the applet angle is a dead end anyway, and I should have caught this earlier. `cosmic_corner_radius_toplevel_v1` is an xdg toplevel, so those errors can't be coming from the panel or an applet at all, since those are layer surfaces. Splitting my 7 day journal by object type:
> 
> ```
> 344,999   cosmic_corner_radius_toplevel_v1   (some app window)
>   4,641   cosmic_corner_radius_layer_v1      (xdg-desktop-portal-cosmic)
> ```

I know this doesn't change the rest of your analysis, but to correct you here: Popups very much do use `cosmic_corner_radius_toplevel_v1`, even for layer-surfaces.

zyads:
oh good to know, that maybe puts the panel's applet popups back in frame as the suspect

sashok-15:
## Summary
`xdg-desktop-portal-cosmic` crashes (SIGSEGV) with a repeating restart loop when taking a screenshot while a second monitor is connected. Root cause is an unhandled `.unwrap()` on a Wayland protocol error in the corner-radius layer handling, triggered when the compositor rejects a corner radius value as "too large."

## Reproduction steps
1. Connect two monitors (dual-display setup).
2. Attempt to take a screenshot (via `cosmic-screenshot` or the screenshot portal).
3. Observe the portal crash-loop in the journal (see logs below).
4. Disable the second monitor → screenshot succeeds.
5. Re-enable the second monitor → screenshot fails again with the same crash.

This is consistently and reliably reproducible — confirmed across multiple attempts.

## Actual panic / crash
```
2026-07-28T15:27:36.614015Z ERROR panic: thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Backend(Protocol(ProtocolError { code: 1, object_id: 204, object_interface: "cosmic_corner_radius_layer_v1", message: "" }))': src/wayland/mod.rs:281
```

Preceded by:
```
xdg-desktop-portal-cosmic[8248]: cosmic_corner_radius_layer_v1#204: error 1: CosmicCornerRadiusLayerV1 { id: ObjectId(cosmic_corner_radius_layer_v1@204[16], 552), version: 2, data: Some(Any { .. }), handle: WeakHandle { handle: WeakInnerHandle[rs] { .. } } } corner radius too large
xdg-desktop-portal-cosmic[8248]: Protocol error 1 on object cosmic_corner_radius_layer_v1@204
xdg-desktop-portal-cosmic[8248]: 2026-07-28T15:27:36.614049Z ERROR xdg_desktop_portal_cosmic::wayland: wayland thread crash; terminating
```

Followed by systemd repeatedly restarting the crashed service:
```
systemd[2952]: org.freedesktop.impl.portal.desktop.cosmic.service: Main process exited, code=dumped, status=11/SEGV
systemd[2952]: org.freedesktop.impl.portal.desktop.cosmic.service: Failed with result 'core-dump'.
systemd[2952]: Starting org.freedesktop.impl.portal.desktop.cosmic.service - Portal service (COSMIC implementation)...
```

## Root cause (as far as I can tell)
`src/wayland/mod.rs:281` calls `.unwrap()` on the result of a Wayland backend call. When the compositor emits a `cosmic_corner_radius_layer_v1` protocol error ("corner radius too large") — which appears to happen specifically under a dual-monitor configuration — that `.unwrap()` panics instead of handling the protocol error gracefully, taking down the whole Wayland event thread and crashing the portal process (SIGSEGV).

## Environment
- Pop!_OS, COSMIC (versions mixed during troubleshooting: `cosmic-files 1.4.0~1784638526~24.04~9a21542`, please let me know which package/version numbers are most useful and I'll paste `apt list --installed | grep cosmic` in full)
- NVIDIA driver 580.173.02
- Dual monitor setup: `DP-1`, `DP-2`
- Kernel: `7.0.11-76070011-generic`

## Additional (likely unrelated) noise in the same logs
Every COSMIC component also logs a `list_button` theme key error on startup:
```
error loading system dark theme error=GetKey("list_button", Os { code: 2, kind: NotFound, message: "No such file or directory" })
```
This persisted even after fully regenerating the theme config directory (`~/.config/cosmic/com.system76.CosmicTheme.Dark/v2`) and a full `apt full-upgrade`, so it looks like a separate packaging/schema issue (a theme key expected by current code but missing from the default `v2` theme template). Flagging it here in case it's relevant, but the SIGSEGV above is the actual blocker for screenshots.

PR #2624, opened 2026-07-24
backend/kms: Check all connectors for primary gpu

Drakulix:
- [x] I have disclosed use of any AI generated code in my commit messages.
  - If you are using an LLM, and do not fully understand the changes it is making to the code base, do not create a PR.
  - In our experience, AI generated code often results in overly complex code that lacks enough context for a proper fix or feature inclusion. This results in considerably longer code reviews. Due to this, AI authored or partially authored PRs may be closed without comment.
- [x] I understand these changes in full and will be able to respond to review comments.
- [x] My change is accurately described in the commit message.
- [x] My contribution is tested and working as described.
- [x] I have read the [Developer Certificate of Origin](https://developercertificate.org/) and certify my contribution under its conditions.

Alternative to github.com/pop-os/cosmic-comp/pull/2621

PR #2647, opened 2026-07-27
fix(kms): consider all connected connectors when picking primary gpu

techywilbur:
`determine_primary_gpu` looked for a built-in laptop panel (eDP/LVDS/DSI) by scanning only connectors that already had a `Surface` (`dev.inner.surfaces.values()`). At startup a connector may not have been turned into a `Surface` yet, or its `Surface` creation may have transiently failed, so the internal panel could be missed and the wrong render GPU selected on a hybrid (e.g. Intel + NVIDIA) laptop when an external display is connected at boot.

Instead, enumerate every connector via `resource_handles()` and treat a device as holding the built-in panel only when an internal-interface connector reports `ConnectorState::Connected`. A device that fails to report resource handles is skipped rather than aborting selection for all devices. Preserves the existing `COSMIC_RENDER_DEVICE`, boot-gpu and first-device fallbacks unchanged.

Adds trace/info logging of which connectors were considered and why a GPU was chosen, plus a unit test for the internal-panel predicate.

This mirrors the approach of the upstream branch
`fix-check-all-connectors_noble` (9c95ebfc, Victoria Brekenfeld); it is a candidate for pop-os/cosmic-comp issue #3751 and does not claim to resolve every report in that issue (a single-GPU report is unlikely to be affected by primary-gpu selection).

### DIFF
--- a/src/backend/kms/mod.rs
+++ b/src/backend/kms/mod.rs
@@ -29,7 +29,11 @@ use smithay::{
         calloop::{Dispatcher, EventLoop, LoopHandle},
         drm::{
             Device as _,
-            control::{Device as _, connector::Interface, crtc},
+            control::{
+                Device as _,
+                connector::{Interface, State as ConnectorState},
+                crtc,
+            },
         },
         input::{self, Libinput},
         wayland_server::{Client, DisplayHandle},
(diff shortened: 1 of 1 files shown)


Drakulix:
See

https://github.com/pop-os/cosmic-comp/pull/2621
and https://github.com/pop-os/cosmic-comp/pull/2624

PR #2657, opened 2026-07-27
fix: clamp pointer edge after applying the output offset

hojjatabdollahi:
When testing https://github.com/pop-os/cosmic-comp/pull/2442 with `lan-mouse` I noticed if I set the host PC on the right, then it's very tricky to move the cursor between machines, but if I put it on the left, the cursor smoothly moves between machines.

After some digging I noticed that when I move the cursor all the way to the right edge, the cursor ends up exactly on 7949.0 which is my output's exclusive right bound. A pointer parked there is outside every surface.

`lan-mouse` create a 1px wide surface on the right edge, the current calculation allows the cursor to be on 7949.0 instead of 7948.9999999. Which means I have to wiggle the mouse very slowly around the right edge for it be on that 1px surface, send the mouse-enter event and transition to my other machine.

This PR fixes that. The cursor stays within bounds and smoothly moves between machines.

I thought this should be a separate PR since it's fixing a bug in a recently merged PR https://github.com/pop-os/cosmic-comp/pull/2568. If you'd prefer I can close this PR and add this as part of https://github.com/pop-os/cosmic-comp/pull/2442 to allow testing it with `lan-mouse`.
___

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -393,10 +393,12 @@ impl State {
                     // - output_geometry.size so that we don't send leave events to a fullscreen app
                     // - logical size so that the position doesn't end up outside the actual size of the output
                     // See https://github.com/pop-os/cosmic-comp/pull/2568
-                    let max_x = output_geometry_loc.x
-                        + logical.w.min(output_geometry.size.w as f64).next_down();
-                    let max_y = output_geometry_loc.y
-                        + logical.h.min(output_geometry.size.h as f64).next_down();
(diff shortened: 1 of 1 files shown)


RipleyTom (approved):
As I made the original PR, that fix is indeed correct.
next_down() step can be lost if the sum's exponent causes the ULP to be lost.
I honestly didn't even think about it, good catch.

Drakulix (approved):
Thanks!

jacobktm (approved):

PR #2610, opened 2026-07-20
feat: shake cursor to find by magnifying the cursor

hojjatabdollahi:
This adds the feature to shake the cursor to locate it.
The algorithm is based on KDE's implementation. Collect the distance the cursor moved in the past 1 second, and divide by the diagonal of the bounding box, if it's more than a threshold, then the user is shaking their cursor.

https://github.com/user-attachments/assets/11474895-4cd2-4907-9fde-290429b79502

- This is disabled by default, we need to add to add a toggle for this to accessibility page.
___
- [x] I have disclosed use of any AI generated code in my commit messages.
  - If you are using an LLM, and do not fully understand the changes it is making to the code base, do not create a PR.
  - In our experience, AI generated code often results in overly complex code that lacks enough context for a proper fix or feature inclusion. This results in considerably longer code reviews. Due to this, AI authored or partially authored PRs may be closed without comment.
- [x] I understand these changes in full and will be able to respond to review comments.
- [x] My change is accurately described in the commit message.
- [x] My contribution is tested and working as described.
- [x] I have read the [Developer Certificate of Origin](https://developercertificate.org/) and certify my contribution under its conditions.

PR #2659, opened 2026-07-27
Scalable Cursors

### LINKED ISSUE #1370: Support rendering SVG cursors
SVG cursors are currently only supported in KDE plasma I believe, but it's an attempt at standardizing svg-based cursors, which can help with sizing cursors dynamically.

We'd probably want to resolve https://github.com/pop-os/icon-theme/issues/138 in order for there to be a compelling enough reason to support it in cosmic-comp.

https://planet.kde.org/vlad-zahorodnii-2024-10-06-svg-cursors-everything-that-you-need-to-know-about-them/

hojjatabdollahi:
There are two implementations for svg cursors, hyprland and KDE.
xcursor up stream is waiting for other desktops to adopt one, or for KDE to upstream it.

KDE's approach makes more sense. This PR implements that.

This is the first draft, if this is something you want, and you support the approach, I can open a PR to xcursor and work on upstreaming and completing this feature.

Here's how it looks:

https://github.com/user-attachments/assets/e378ba7b-b04c-4061-acd7-18e9fd9a604f

Notes:
- This branch is based on "shake to locate cursor" branch to test
- The cursor theme I'm using to test is extracted from svg file in pop-os/icon-theme repo, and is not clean. We need to add proper cursor_scalable to pop icon theme.
- You can use Breeze for testing:
```bash
sudo mkdir -p /usr/share/icons/breeze_cursors
curl -sL "https://invent.kde.org/plasma/breeze/-/archive/master/breeze-master.tar.gz?path=cursors/Breeze/Breeze" \
  | sudo tar xz --strip-components=4 -C /usr/share/icons/breeze_cursors
echo 'XCURSOR_THEME=breeze_cursors' | sudo tee -a /etc/environment
echo 'XCURSOR_SIZE=24'              | sudo tee -a /etc/environment
```

Closes https://github.com/pop-os/cosmic-comp/issues/1370
____

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -817,6 +817,7 @@ dependencies = [
  "profiling",
  "rand 0.10.0",
  "regex",
+ "resvg 0.45.1",
  "ron 0.12.0",
  "rust-embed",
  "rustix 1.1.4",
@@ -6893,9 +6894,9 @@ checksum = "ea6fc2961e4ef194dcbfe56bb845534d0dc8098940c7e5c012a258bfec6701bd"
 
 [[package]]
 name = "xcursor"
-version = "0.3.10"
+version = "0.3.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bec9e4a500ca8864c5b47b8b482a73d62e4237670e5b5f1d6b9e3cae50f28f2b"
+checksum = "163b33ed8786455e2fa5d72f554057ce3f3182425434f756cd39c99839d88e23"
 
 [[package]]
 name = "xdg"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,7 @@ log-panics = { version = "2", features = ["with-backtrace"] }
 ordered-float = "5.1"
 png = "0.18"
 regex = "1"
+resvg = "0.45"
 ron = "0.12"
 rust-embed = { version = "8.11", features = ["debug-embed"] }
 sanitize-filename = "0.6.0"
--- a/src/backend/render/cursor.rs
+++ b/src/backend/render/cursor.rs
@@ -460,9 +460,7 @@ pub struct CursorStateInner {
 
     cursors: HashMap<CursorIcon, Cursor>,
     current_image: Option<Image>,
-    /// Rasterized cursor frames keyed by `(shape, pixel size, frame index)` so
-    /// each SVG frame is rasterized (and uploaded) at most once per size.
-    image_cache: Vec<((CursorIcon, u32, usize), Image, MemoryRenderBuffer)>,
+    image_cache: Vec<CachedFrame>,
 
     hidden: bool,
     idle_timer: Option<RegistrationToken>,
@@ -476,6 +474,15 @@ pub struct CursorStateInner {
     magnification: f32,
     anim_from: f32,
     anim_start: Option<Instant>,
+    rest_started: Option<Instant>,
+}
+
+/// A rasterized cursor frame, keyed by `(shape, pixel size, frame index)`.
+struct CachedFrame {
+    key: (CursorIcon, u32, usize),
+    image: Image,
+    buffer: MemoryRenderBuffer,
+    unmagnified: bool,
 }
 
 /// One sampled pointer position on the recent motion path.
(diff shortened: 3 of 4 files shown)


Drakulix:
On first glance this looks good, but since this is based on #2610 we need wait for that to be approved first before progressing.

hojjatabdollahi:
I removed the shake-to-locate commit. The upstream xcursor is also merged. So, this is ready to review.

You can use super+scroll to zoom and test.

https://github.com/user-attachments/assets/93e19611-c04c-4dce-a4fb-a8de00cf684e

Drakulix (commented):
Looks mostly good, but I am slightly concerned about the `image_cache` now. I don't think we have any routine to clean it up, which wasn't a big problem, when - at most - we could end up loading the whole active cursor theme with a bunch of 24x24 icons.

But when we cache arbitrary sized svg rasterizations (which we should), this can grow pretty large, especially since we intentionally don't cap the maximum cursor size.

I think this needs a `cleanup` function and call from `State::cleanup`. Easiest would probably be to make the `image_cache` into a `VecDeque`, push new entries to the front and then truncate to `10` in cleanup? (Assuming that newest images are likely re-used soon.)

This would also eliminate long searches for cache entries as the vector grows larger.

hojjatabdollahi:
That makes sense. I will add some telemetry to the compositor and monitor the cache hit/misses and size to see how bad does it get right now.

My concern is:
- VecDeque doesn't address the unbounded cache in bytes (small cursors are <10kb, large ones get to 10s of megabytes).
- Animated cursors would not like a small cache size, our wait cursor has 60 frames.
- Mixed scale multi monitor or screencasting (since it's harcoded to 1 for scale) would also get messed up with a small cache.
- In general the cache does very little for shake-to-find. When we're doing shake-to-find, it will miss on the way up, and miss on the way down if the cache is small. Only useful when we keep the size for 2 seconds at the top.

What if we do this:
- Set a max size for rasterization. After which we're just scaling up, so the max memory usage is limited.
- Maybe only cache on specific sizes and scale down to the target size? Not sure how noticeable it would look, but that would make the ladder much shorter. (I have to eyeball how it looks mid-tween).

hojjatabdollahi:
Here's what I did:
- Added a max size for cursor rasterization (512px). Anything larger than that takes a noticeable amount of time to render and takes a large amount of memory. With this max size, we just scale up after 512px. If you zoom/shake too much the cursor will get a bit blurry, but I'd say that's acceptable.
- Added a ladder with `base * 2^x` for the steps. So if you start with `24px` as soon as you start enlarging the cursor (for zoom or shake-to-find), it will render `24*2^1` and down sample to the target size, so the cursor stays sharp.
- The entries to the cache now are limitted, instead of "all sizes between base size to infinity" now it's "a limited number of x between 24 to 512 based on base*2^x".
- And the cache gets cleared 10 seconds after the cursor rests (going back to the base size). So, if you're actively shaking the mouse, it doesn't clear the cache. The cache is also small and limited to begin with. And the cache only deletes magnified cursors


The only concern is what's the point of SVG then? We could just use pre-rasterize (bitmap) cursors with different sizes!

leviport (approved):
Cursors are much sharper while zooming and/or shaking. I definitely see the upper threshold when it starts scaling the 512 version, but I have to shake/zoom pretty extremely to get to it.

Drakulix:
> The only concern is what's the point of SVG then? We could just use pre-rasterize (bitmap) cursors with different sizes!

I mean, the current code should've already done that. But with svg's we don't need to ship every ever increasing size of cursors in a theme anymore. So it is mostly an improvement for theme builders.

To fully leverage the advantages in the compositor, we would have to upload the path instead and rasterize on the gpu. I believe there is some nvidia-specific extension to do that and potentially more generic svg-render pipelines exist, but that is an optimization that definitely can come later.

Drakulix (approved):
LGTM code-wise now. Thanks!

jinliu:
> The only concern is what's the point of SVG then? We could just use pre-rasterize (bitmap) cursors with different sizes!

Yes we can. But the installed size would bloat quickly, because the XCursor format is uncompressed.

And there are nice things that SVG _could_ enable. E.g., re-color cursors to the theme color.

> But with svg's we don't need to ship every ever increasing size of cursors in a theme anymore.

Yes, but post installation, we still need to generate XCursor files for all cursor sizes we support, for backward compatibility with X11 apps.

E.g., https://github.com/jinliu/svg-cursor/tree/main/svg-theme-to-xcursor

Issue #1692, opened 2025-09-30
mouse starts spinning while playing games with proton


kurai111:
the same thing happens on multiple games but im putting gw2 as an example
https://imgur.com/a/7SS9STI

when i reboot everything is fine but after a few hours this starts happening.the camera will look up and wont let me look down,when i hold right click and try to move it randomly spins. restarting the game/changing proton versions doesnt help

i am using cachyos. i switched to cosmic 1 week ago i was using xfce before and this never happened.

kurai111:
apparently this was an issue in kde too but got fixed https://bugs.kde.org/show_bug.cgi?id=482476

kurai111:
another thing i found.i tried playing dead by daylight and when i tried to move my camera to right,it wouldnt let me move,only 30 degrees or something.then i clicked super button and clicked on game window and it let me do a 360 degree turn then got stuck again.heres the video: https://imgur.com/a/8jbDBpg

very weird

ids1024:
I seem to see similar behavior in some games, but not others. And at least temporarily fixed by restarting the compositor.

Good catch on the KDE issue; that looks very similar.

ids1024:
If it doesn't happen when the compositor initially starts, it would be good to identify exactly what makes the problem start...

kurai111:
i really have no idea because it happens so randomly.sometimes i can play entire day without this happening sometimes it happens twice a day

ids1024:
Yeah. It's annoying. I've seen this, but trying various things, I can't reproduce it. The KDE issue sounds similar, but we don't seem to be generating `relative_motion` events for warps, so a change like https://invent.kde.org/plasma/kwin/-/commit/630ba5fab4b02c2b496553c79d4a315da9271b53 per se isn't needed.

Sirius902:
I've had this issue occur several times when playing FFXIV (usually I notice it when I first open the game after a while of not having it open so presumably something I've done prior to that point triggered it). However, I have found an odd workaround to fix the issue after it happens in FFXIV:
- Ensure that you have the issue triggered first (trying to move the camera by clicking and panning the mouse results in the camera spinning wildly out of control) and that you have a dual monitor setup with the game on the right monitor.
- Next, click into the game window and hold down the left mouse button to allow the camera to be moved (do not release the mouse button).
- Activate the "Focus window left" navigation keyboard shortcut (Super+h by default) which moves the mouse cursor instantly to the center of the left monitor.
[...]

PR #2665, opened 2026-07-29
fix: clamp out-of-bounds cursor position hints instead of dropping them

BrianHotopp:
## What

`apply_cursor_hint` silently discarded cursor position hints even fractionally outside the window bounds (`is_legal` returned false, the caller produced `None`): no `relative_motion`, no `motion`, no `frame`. The client believes its hint was applied, so every drop desyncs the compositor's pointer position from the client's. This restores the pre-296b6be behavior of always producing a correction.

## How

Clamp the hint into `[0, size.next_down()]` — the boundary pattern from #2568 — so a marginal out-of-range hint still produces a best-effort correction. Hints violating an active constraint region are still rejected.

Suspected contributor to the Proton/Xwayland mouselook clamp in #1692, where Xwayland's warp emulator sends a position hint on every motion tick while locked (not confirmed end-to-end with a game).

## Testing

Nested harness (Xvfb + winit backend, output scale 2.0): client locks the pointer, submits an out-of-bounds hint, unlocks. Before: hint dropped, no motion event delivered. After: hint clamped to `(w.next_down(), y)` and motion delivered at the clamped position.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -2377,30 +2377,29 @@ impl State {
                 let pos_in_element = location + surface_offset.to_f64();
                 let window_size = geometry.size.to_f64();
 
-                let is_legal = |p: Point<f64, Logical>| {
-                    let in_window =
-                        p.x >= 0.0 && p.y >= 0.0 && p.x < window_size.w && p.y < window_size.h;
-                    if !in_window {
-                        return false;
-                    }
+                // Clamp marginal out-of-bounds hints instead of dropping them:
+                // a dropped hint desyncs the client's tracked pointer position
+                let clamped = Point::<f64, Logical>::new(
+                    pos_in_element.x.clamp(0.0, window_size.w.next_down()),
+                    pos_in_element.y.clamp(0.0, window_size.h.next_down()),
+                );
 
-                    with_pointer_constraint(surface, pointer, |constraint| {
-                        if let Some(constraint) = constraint
-                            && let Some(region) = constraint.region()
-                        {
-                            let point_in_surface = (p - surface_offset.to_f64()).to_i32_floor();
-                            return region.contains(point_in_surface);
-                        }
-                        true
-                    })
-                };
+                let region_ok = with_pointer_constraint(surface, pointer, |constraint| {
+                    if let Some(constraint) = constraint
+                        && let Some(region) = constraint.region()
+                    {
+                        let point_in_surface = (clamped - surface_offset.to_f64()).to_i32_floor();
+                        return region.contains(point_in_surface);
+                    }
+                    true
+                });
 
                 let workspace_origin = output.geometry().loc.to_f64();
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
Should be fine. Thanks!

skygrango:
> The client believes its hint was applied, so every drop desyncs the compositor's pointer position from the client's

As far as I know, the current behavior is consistent with KDE and GNOME.

If the client performs a pointer unlock, it respects the mouse coordinates provided by the compositor; therefore, there should be no such thing as a desync.

skygrango:
Applying clamping to `apply_cursor_hint()` also alters pointer-warp-v1 behavior; this was discussed [previously](https://github.com/pop-os/cosmic-comp/pull/2432#issuecomment-4633735411), so it is recommended to test both aspects together.

skygrango:
> Suspected contributor to the Proton/Xwayland mouselook clamp in https://github.com/pop-os/cosmic-comp/issues/1692, where Xwayland's warp emulator sends a position hint on every motion tick while locked (not confirmed end-to-end with a game).

If the issue were caused by the cursor position hint being silently discarded, the camera should stop rotating rather than spinning; therefore, this hypothesis does not hold.

I suspect the continuous spinning is actually caused by an issue with XWayland's focus behavior. A fix is ​​already available at https://github.com/pop-os/cosmic-comp/pull/2620; I suggest using the latest version of cosmic-comp to see if the problem has been resolved.

skygrango:
It should also be noted that the current `apply_cursor_hint()` only applies the hint when the pointer is unlocked or pointer warp event; consequently, the cursor hint provided by XWayland clients is not continuously applied while the pointer is in the locked state.

BrianHotopp:
Wrt. stop rotating vs spinning

Stop rotating is exactly what I observed (subnautica)

Can give you any info you need about my system if you want

Iirc in the linked issue some folks were describing halting as well

https://github.com/pop-os/cosmic-comp/issues/1692#issuecomment-3354190659

BrianHotopp:
Good catch — verified against both codebases:

- Mutter stores hints unconditionally and skips the warp at unlock if the hint lands outside the constraint region (`locked_pointer_destroy`, meta-wayland-pointer-constraints.c). No clamping.
- KWin does the same with a per-axis bounds check in `PointerInputRedirection::updatePointerConstraints` (pointer_input.cpp). No clamping.

So the current behavior matches both, the hint spec intentionally leaves this open, and pointer-warp-v1 explicitly wants out-of-bounds warps rejected. Clamping would make cosmic-comp the outlier.

Also a fair point on unlock-time application: only the last hint applies at unlock, so a drop requires the final hint to be the out-of-bounds one — no accumulating desync. And behavior shared with GNOME/KDE wouldn't explain a cosmic-specific symptom anyway. #2620 looks like the better lead for #1692.

Closing, thanks for the review.

PR #2666, opened 2026-07-29
Add a config option to disable server-side titlebars

cramt:
Replaces #1385, which is closed. Part of pop-os/cosmic-epoch#275.

The old attempt made the SSD height configurable, which needed a new `ssd_height` field in `cosmic-theme` (pop-os/libcosmic#865). That got fair pushback:

- @mmstick: libcosmic already has a header size config, and what's actually wanted here is *"a field for disabling window titles in the CosmicCompConfig"*.
- @git-f0x: *"`ssd_enabled` should be a compositor config option, since it can just not use the header bar if false. The compositor sets SSD height on its own, and reducing SSD height would visually break window control buttons"*.

Both are right, so this is the same feature done that way instead: one boolean in the comp config, no libcosmic change. pop-os/libcosmic#865 is closed.

### What it does

`appearance_settings.show_titlebars`, default `true`. When `false`, `CosmicWindowInternal::has_ssd()` returns false, so the compositor doesn't draw a header bar for windows that don't decorate themselves and doesn't reserve `SSD_HEIGHT` for one — the client gets the whole window. Works for Wayland and X11 windows alike, since both already funnel through `has_ssd()`.

Everything that positions, hit-tests, sizes or renders the header already goes through `has_ssd()`, so the gate is one line. The two spots that checked `is_decorated()` directly (`min_size`/`max_size`, `CosmicMapped::ssd_height`) now go through it too.

Stacks keep their tab bar — without it there's no way to switch tabs.

Flipping the option at runtime calls `workspaces.recalculate()`, so open windows immediately gain or lose the header's worth of space instead of waiting for the next resize.

### Notes

- Field lives in `AppearanceConfig` because that struct is already propagated to every mapped window on config change (`update_appearance_conf`), so it's where a per-window appearance flag costs nothing to plumb. Happy to hoist it to a top-level `CosmicCompConfig` key, or rename it to `ssd_enabled` per @git-f0x's suggestion — just say which you prefer.
- No UI yet, on purpose: a cosmic-settings/tiling-applet toggle can follow once the field name is settled.
- `cargo check --all-targets` clean, `cargo fmt --check` clean.

---


maria-komarova:
We should add an option in Settings > Desktop > Window management in the top section under Floating windows gravitate...". 
image
The setting should be a toggle with the following copy:
Body: Use system window headers when possible
Caption: Applications may use their own window headers in some cases

Drakulix:
I am sorry, but I don't think this is an implementation we actually want to have.

Reporting server-side decorations are in use, but then not drawing any is not part of our designs. The only setting we want to expose here is whether we use client-side or server-side decorations by default for applications supporting xdg-decoration but not providing a preference for either.

Not showing any title bar isn't a configuration we intend to support, if I am not mistaken. (Ping @maria-komarova)

Kommynct:
> I am sorry, but I don't think this is an implementation we actually want to have.
> 
> Reporting server-side decorations are in use, but then not drawing any is not part of our designs. The only setting we want to expose here is whether we use client-side or server-side decorations by default for applications supporting xdg-decoration but not providing a preference for either.
> 
> Not showing any title bar isn't a configuration we intend to support, if I am not mistaken. (Ping @maria-komarova)

There are many people who think the title belongs on the panel, is the cosmic desktop not intended to be something you can build any experience you want on? What about a hidden CLI option if you're worried about beginners?

Titlebars are wastes of space that serve no purpose in a tiling keyboard driven environment. Is catering to tiling not important to the cosmic desktop? As a tiler I would never switch to an environment that doesn't allow me to disable them.

Drakulix:
> There are many people who think the title belongs on the panel, is the cosmic desktop not intended to be something you can build any experience you want on? What about a hidden CLI option if you're worried about beginners?

Enabling global menus is a completely different thing, which might then allow us to remove titlebars, as all the options would be provided by the panel.

But at the moment disabling the titlebar would mean you loose access to some functionality. This is a footgun and not something we want users to run into.

As such implementations for all functionality including global menus would need to be implemented first and the latter isn't an easy feat, as support varies widely across toolkits.

"Undecorated" is also not a mode supported by every application. Some terminals allow you to explicitly configure it (e.g. cosmic-term for that matter!), but they do so knowingly, that they have ways for the user to access functionality normally provided by the title bar. This implementation simply tells applications that the compositor will draw decorations but then it doesn't. If we cannot guarantee that all functionality is accessible elsewhere, then we knowingly degrade UX. Most tiling wms do so by offering very dynamic key binds for features often only accessibly via the keyboard in the first place, which isn't exactly what COSMIC is.

cramt:
Fair enough, and I'll wait for @maria-komarova rather than push. Couple of things worth getting straight first though.

The setting you're describing isn't the same feature. Telling an app `ClientSide` just means it draws its own header — the pixels stay, someone else paints them. What #275 is asking for is the space back. I'd happily build the decoration-preference toggle too, I just don't want the two collapsed into one question.

You're right about the protocol bug, and I'll go further: I can't fix it the obvious way. xdg-decoration only has `client_side` and `server_side`, so there's no honest "undecorated" to advertise, and on X11 `try_force_undecorated` is a no-op anyway. This is the compositor overriding the client on the user's say-so, and I'd rather call it that than dress it up. There is precedent — stacking forces `ServerSide` over whatever the client asked for — but I'm not claiming that settles it.

On losing functionality, I undersold it in the PR and you're more right than you realised. It's not the minimize button, it's the whole window menu: `on_right_click(Message::Menu)` on the header is the only compositor-side way in, and there's no shortcut for it anywhere. CSD apps keep a route because their own headers forward `show_window_menu()` and we serve them the same items — a window we force undecorated has nothing, and X11 can't have it at all. That's your cosmic-term point exactly, and it's the real hole.

So I'd rather close it than argue it: a default `Minimize` binding and a bindable action to open the window menu. Both are small, both are worth having on their own, and I'm happy to do them as a separate PR regardless of what happens here.

For what it's worth on the footgun: people are already doing this, worse. My patch in #275 is a one-line `SSD_HEIGHT = 0`, there's a variant from @jcornuz, and folks apply them via overlays. That does exactly the thing you're objecting to, with no default and no way back. I've been running it since May last year and haven't hit a problem with it, which is roughly my point: I'd rather this were served by something that keeps titlebars on by default and can be switched back than by people zeroing constants.

Mostly I just want the scope answered: is it "no undecorated windows, ever", or "not until the panel covers it"? Genuinely either is fine — I'd just like to stop carrying a patch downstream and get onto whatever the supported path is, and I'm happy to build it whichever direction that turns out to be.

radutomy:
My terminal does not need a title bar, unused precious vertical space. I definitely would want this feature.

trust2k:
But it's really a good idea to add a barely noticeable shortcut on top of the application if the user has completely disabled the decorations! Click on this point and the entire list with options appears! And the wolves are fed and the sheep are safe 😉

Kommynct:
> Enabling global menus is a completely different thing, which might then allow us to remove titlebars, as all the options would be provided by the panel.
> 
> But at the moment disabling the titlebar would mean you loose access to some functionality. This is a footgun and not something we want users to run into.

1. I actually don't even want the title in the panel necessarily, not everyone even wants to see window titles at all.
2. I think you could ensure that certain keybinds are bound such that you never end up in a state where the user can't do things, but realistically you can always open the settings menu and undo, no matter the state you're in as long as you know the keyboard shortcut for the launcher?

A hidden option or CLI flag really solves all of this honestly. There's no reason not to give tilers some way to do this, this is standard across every single one i've ever used.

PR #2669, opened 2026-07-29
Use a consistent target for minimize and restore animations

techywilbur:
Related to pop-os/cosmic-epoch#3758

## Summary

The minimise and restore animations already use the same duration, easing, and opacity. However, they can look very different because the target rectangle varies depending on the dock configuration.

Previously:

- With the Minimized Windows applet, the window animated towards the applet.
- With a dock but without that applet, another panel-provided rectangle could be used.
- With no valid target, cosmic-comp used a fixed 200×100 fallback located around two-thirds down the screen.
- When multiple target rectangles existed, the selected target could depend on insertion order.

## Changes

This patch:

- validates minimise target rectangles
- ignores zero-sized targets and targets completely outside the output
- selects valid target rectangles deterministically
- preserves valid applet-provided targets
- uses a consistent 64×64 fallback near the bottom of the output when no valid target exists
- keeps the fallback inside the output and its non-exclusive area
- adds trace logging for target selection and rejected targets
- adds focused tests for candidate validation, deterministic selection, and fallback geometry

This makes the fallback animation resemble the applet-targeted animation more closely instead of using a differently shaped and positioned target.

## Testing

Passed locally:

- `cargo fmt --check`
- `cargo check --all-targets`
- `cargo clippy --all-targets`

The repository test binary could not be linked locally because the required development libraries for GBM, libseat, libinput, and libdisplay-info were unavailable.

The nine pure geometry and target-selection tests were also executed in an isolated test crate:

- 9 passed
- 0 failed

This is a candidate fix for the inconsistency described in issue #3758 and does not change animation duration, easing, opacity, or rendering behaviour.

### DIFF
--- a/src/wayland/handlers/toplevel_management.rs
+++ b/src/wayland/handlers/toplevel_management.rs
@@ -5,9 +5,10 @@ use smithay::{
     input::{Seat, pointer::MotionEvent},
     output::Output,
     reexports::wayland_server::DisplayHandle,
-    utils::{Point, Rectangle, SERIAL_COUNTER, Size},
+    utils::{Logical, Point, Rectangle, SERIAL_COUNTER, Size},
     wayland::seat::WaylandFocus,
 };
+use tracing::trace;
 
 use crate::{
     shell::{CosmicSurface, Shell, WorkspaceDelta, focus::target::KeyboardFocusTarget},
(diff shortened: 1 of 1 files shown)


Drakulix:
The PR template is not optional

mmstick:
Used AI code generation without disclosing and without the required PR template

PR #2670, opened 2026-07-30
fix(kms): retry busy drm devices during startup

FreddyFunk:
#2382 introduced a regression, which causes devices like my Framework 13 with an Ryzen 5 7640U to login into a black screen from cosmic-greeter.

On a greeter to session handoff the outgoing compositor can still hold the dri card when the incoming cosmic-comp starts, so the device open fails with EBUSY and the session dies with "Backend initialized without output" and a black screen. This retries failed devices for up to 5s at 100ms intervals in `init_backend()`, but only while no device came up at all, so a failing secondary GPU never delays startup.

### DIFF
--- a/src/backend/kms/mod.rs
+++ b/src/backend/kms/mod.rs
@@ -52,7 +52,7 @@ use std::{
     collections::{HashMap, HashSet},
     path::Path,
     sync::{Arc, RwLock, atomic::AtomicBool},
-    time::Duration,
+    time::{Duration, Instant},
 };
 
 mod device;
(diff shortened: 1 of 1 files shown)


Drakulix:
> #2382 introduced a regression, which causes devices like my Framework 13 with an Ryzen 5 7640U to login into a black screen from cosmic-greeter.

Can you do regression testing which of the two commits from said PR caused this behavior for you? It was precisely meant to make this more unlikely by optimizing the shutdown procedure of the greeters cosmic-comp instance.

Re-trying is a really ugly workaround for something that should be considered a greeter-bug, so I'd rather revert parts of that PR, while looking into finding a better solution.

FreddyFunk:
Yes, but probably in a week earliest. I'm now traveling without access to the hardware that has the affected AMD GPU. I'm 80% certain that e82a7631643fe1609c4b95522d3734624f68f16b introduced the regression when I debugged this locally while working on this fix.

francoism90:
Any update on this?

PR #2673, opened 2026-07-30
Fix live adaptive-sync configuration changes

AdityaHebballe:
## Summary

Apply the public adaptive-sync output state after backend configuration succeeds.

On the KMS backend, publishing it first caused the backend to compare the requested mode with the already-updated value. It then skipped the `UseAdaptiveSync` command, leaving the rendering surface thread in its previous VRR mode until the compositor restarted.

## Testing

- `cargo check -p cosmic-comp`
- On an AMD GPU with a VRR-capable DisplayPort monitor, changed adaptive sync from Automatic to Off and back while logged in.
- Verified the DRM `VRR_ENABLED` property now follows the new setting immediately; no logout/login is required.

- [x] I have disclosed use of AI assistance in the commit message and reviewed the change.
- [x] I understand these changes and can respond to review comments.
- [x] I have read the Developer Certificate of Origin and certify my contribution under its conditions.

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -524,8 +524,6 @@ impl LockedBackend<'_> {
             )))
             .filter(|x| *x != output.current_location());
             output.change_current_state(mode, transform, scale.map(Scale::Fractional), location);
-
-            output.set_adaptive_sync(final_config.0.vrr);
         }
 
         match self {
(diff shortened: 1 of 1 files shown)


AdityaHebballe:
Tracked in #2674, which includes the reproduction, the observed output-state/DRM-state mismatch, and the KMS surface-thread root cause.

src/state.rs
```diff
@@ -541,6 +539,20 @@ impl LockedBackend<'_> {
             LockedBackend::X11(state) => state.apply_config_for_outputs(test_only),
         }?;
 
+        // The KMS backend compares the requested VRR mode with the value stored on
+        // the output before sending `UseAdaptiveSync` to its surface thread.  Do
+        // not update that value until after the backend has made that comparison,
+        // otherwise a live VRR change is mistaken for a no-op and only takes
+        // effect after the compositor recreates the surface.
```
Drakulix, line 546: Yeah I think the backend is at fault here and needs to be fixed instead. Otherwise we cannot test if the new value actually applies cleanly, which isn't guaranteed for a bunch of configurations.

Drakulix (changes requested):

PR #2683, opened 2026-07-31
fix(shell): take presentation feedback from fullscreen surfaces

### LINKED ISSUE #2677: Chromium YouTube fullscreen exits after ~1.5s; presentation feedback almost entirely discarded while fullscreen
## Summary

On current Pop!_OS Cosmic (Wayland), entering YouTube video fullscreen in Google Chrome briefly succeeds, then Chrome itself exits fullscreen after ~1.5 seconds with no user input. During that fullscreen interval, nearly all `wp_presentation_feedback` events are **discarded**.

This is distinct from the older maximized Chromium “double-fullscreen / configure race” class (#1642 / PR #2099). Cosmic accepts `set_fullscreen` with a correct `1920×1080` configure; Chrome later sends `unset_fullscreen`.

## Environment

- OS: Pop!_OS 24.04 (COSMIC)
- Session: Wayland (`cosmic-comp`)
- `cosmic-comp`: `0.1~1785355703~24.04~091583a` (git `091583ac` — includes wl-dmabuf v6)
- `google-chrome-stable`: `151.0.7922.71-1`
- GPU: Intel Arrow Lake-U iGPU + NVIDIA GeForce RTX 5070 Ti Laptop (hybrid)
- Output used in repro: HDMI `1920×1080` @ 60 Hz (also have laptop eDP `2560×1600` @ 240 Hz, scale 1.5, VRR)
- Regression window (from apt history, 2026-07-30):
  - `cosmic-comp`: `4d370cdf` → `091583ac`
  - Chrome: `150.0.7871.186` → `151.0.7922.71` (same update transaction)

## Steps to reproduce

1. Floating Chrome window (also observed conceptually with prior tiled reports; this capture is floating).
2. Open a YouTube video.
3. Click the player fullscreen control (or press `f`) — not Cosmic/window fullscreen / not F11 first.
4. Do not press Esc or click to exit.

## Actual behavior

1. Video + browser enter true fullscreen (Esc toast appears).
2. After **~1.50 s**, fullscreen ends spontaneously.
3. Window restores to prior floating geometry.

Reproduced twice in one capture with the same timing.

## Expected behavior

Fullscreen remains until the user exits (Esc / player control).

### `WAYLAND_DEBUG=1` (client log)

Full log: https://gist.github.com/jlakin2/6845598cdb65926e2b836dcfb5a53c82

Two cycles:

| Attempt | Client request | Compositor configure | Client exit | Duration |
|--------|----------------|----------------------|-------------|----------|
| 1 | `xdg_toplevel.set_fullscreen` | `configure(1920, 1080, …)` | `unset_fullscreen` | **1.503 s** |
| 2 | `xdg_toplevel.set_fullscreen` | `configure(1920, 1080, …)` (after a brief `823×614` configure with fullscreen state) | `unset_fullscreen` | **1.503 s** |

Immediately before each `unset_fullscreen`, Chrome destroys the fullscreen-exit-bubble subsurface, then unsets fullscreen.

### Presentation feedback while fullscreen

During each ~1.5 s fullscreen window:

- ~90 `wl_surface` commits
- **110** `wp_presentation_feedback.discarded`
- **1** `presented`

Outside fullscreen in the same log, presentation mostly succeeds (log-wide ~1488 presented / 255 discarded).

This strongly suggests a compositor fullscreen scanout / dmabuf-feedback path failure that Chromium eventually abandons, rather than a `#1642`-style missing-Fullscreen configure rejection (those fail immediately, not on a fixed ~1.5 s timer).

## Suspected area (not confirmed)

In the `4d370cdf..091583ac` package range, shell fullscreen enter/restore logic is unchanged. Notable compositor change in that range:

- PR #2513 / `091583ac` — wl-dmabuf v6
  - fullscreen feedback always uses `primary_scanout_feedback` (no `render_feedback` fallback)
  - `new_surface_feedback` keys off `kms.primary_node` instead of the client’s dmabuf global / device

On hybrid Intel+NVIDIA this is a plausible match for “fullscreen briefly works, presentation discarded, client aborts.”

Chrome 151 co-updated in the same apt transaction, so a browser-side change is not fully ruled out without a package bisect.

## Workaround

- `--ozone-platform=x11` (XWayland), or
- Firefox for fullscreen video (not tested in this capture)

## Additional notes

- Screen recording available locally if needed (`wayland_chrome_bug.mp4`); not attached here.
- No existing issue found describing this exact “~1.5 s auto-exit + discarded presentation feedback” regression after the Jul 29/30 dmabuf v6 package.

hojjatabdollahi:
Looks like we don't collect presentation feedback for fullscreen surfaces, and send `discarded` to the client. This has been an issue since adedb705 I think. The latest version of chrome cares about that now, after 1.5s cancels the fullscreen. There may be a cleaner solution to this. But this works.

Fixes #2677 
___

### DIFF
--- a/src/shell/mod.rs
+++ b/src/shell/mod.rs
@@ -5024,6 +5024,20 @@ impl Shell {
         let mut output_presentation_feedback = OutputPresentationFeedback::new(output);
 
         if let Some(active) = self.active_space(output) {
+            for fs in active.get_fullscreen_surfaces() {
+                fs.surface.take_presentation_feedback(
+                    &mut output_presentation_feedback,
+                    surface_primary_scanout_output,
+                    |surface, _| {
+                        surface_presentation_feedback_flags_from_states(
+                            surface,
+                            None,
+                            render_element_states,
+                        )
+                    },
+                );
+            }
+
             active.mapped().for_each(|mapped| {
                 mapped.active_window().take_presentation_feedback(
                     &mut output_presentation_feedback,


atiq-cs:
Yep, since recent update, fullscreen mode is broken on streaming players on the browser:

For example, on youtube or other sites after your enter full screen by clicking full-screen mode it properly enters full screen mode. However, fullscreen mode reverts back within a few seconds. Noticing since last few days after running "apt update". This revert from full-screen mode is happening every time since update.

To re-pro, try youtube, bloomberg.com/live etc. players on a chromium based browser e.g., brave-browser.
(on pop_os 24 on cosmic)

atiq-cs:
Is it possible to also add a test so this is not broken in future? If not how about add a package downgrade option for users as workaround?

hojjatabdollahi:
@atiq-cs you have to downgrade chrome (which is not a good idea)! This existed for more than a year, the new version of chrome now cares about it. It's not something that QA could've caught.

atiq-cs:
> There may be a cleaner solution to this

Why not engage with chromium team for the right solution? Also, make em aware of the most awesome DE :) Gradually, take over Google to adopt pop-os. :D

krakotay:
Same on nvidia 5060, your patch helped me, thank you very much! 
@jacobgkau @mmstick please hurry up

jirmauritz:
I can confirm the fix helped me on nvidia 3050. Thanks!

atiq-cs:
> Same on nvidia 5060, your patch helped me, thank you very much!

Which patch are you referring to? @krakotay 
I see this is still open and nothing is fixed yet.

krakotay:
@atiq-cs i just downloaded the forket repo, cargo build --release and sudo make install

jym272:
tested on Pop!_OS 24.04, kernel 7.0.11, fixes the Chrome 151 fullscreen exit for me

mr-zlaam:
The issue wasn't only for chrome but any chromium based browser. 
Firefox based browser are working perfectly fine

atiq-cs:
> The issue wasn't only for chrome but any chromium based browser

yep, let's hope, a good solution lands for all chromium based browsers including brave-browser

Drakulix (approved):
Thanks!

Lcstyle:
Independent confirmation of the root cause, with two data points that might help this land.

I hit the same bug on ungoogled-chromium 151.0.7922.71 (Fedora 44, NVIDIA, 4 outputs, single GPU — not hybrid) and root-caused it blind before finding this PR; I ended up with essentially the same diff, which I'd call convergent validation of the approach.

**Protocol evidence** (`WAYLAND_DEBUG=1`, isolated browser profile against the live compositor, local `<video>` page with a double-click element-fullscreen handler):

```text
[3782730.836]  -> xdg_toplevel#46.set_fullscreen(nil)
[3782732.965] xdg_toplevel#46.configure(3840, 1600, array[24])   ; granted in 2ms
[3782740.822]  -> xdg_surface#45.ack_configure(958396)
[3782754.957]  -> wl_subcompositor#5.get_subsurface(...)          ; ESC bubble
[3784232.678]  -> wl_subsurface#57.destroy()                      ; bubble torn down
[3784232.760]  -> xdg_toplevel#46.unset_fullscreen()              ; +1502 ms
```

Two attempts, both reverting at **+1502 ms** — consistent with the 1503 ms in #2677 and 1501 ms from the AMD reproduction there, so the Chromium watchdog is a fixed 1.5 s timer. Notably the compositor sent **no further toplevel event** between grant and revert — no intermediate configure, nothing to misread. The revert is driven purely by feedback starvation:

| Window | presented | discarded |
|---|---|---|
| ~1.5 s before fullscreen | 86 | 0 |
| during fullscreen (63 commits) | **0** | **40+** |

**On the regression window**: my compositor is a fork built on the `epoch-1.4.0` tag (it carries my suspend/resume patch set from #2096, which doesn't touch this path) — i.e. it **predates dmabuf-v6 (#2513) entirely** and reproduces the identical signature. That supports the "existed since adedb705" dating over #2677's suspicion of the dmabuf-v6 package: the co-updated Chrome 151 is the trigger, the feedback hole is old. One more repro detail that fits: launching with `--start-fullscreen` does *not* revert on my system — only element/video fullscreen entered at runtime does.

**One suggestion**: the sibling collection has the same hole. `update_primary_output` in `state.rs` walks the active space's fullscreen surface *and* `set.sticky_layer.mapped()`, while `take_presentation_feedback` covers neither — this PR restores parity for fullscreen but sticky (always-on-top) windows still never get their feedback taken, which should produce the same 100%-discarded starvation for any presentation-time-sensitive client pinned sticky. Since the fix shape is identical (one more `for_each` mirroring the mapped() branch, keyed off `self.workspaces.sets.get(output)`), it might be worth folding in here while the walk is being touched.

I've applied the equivalent patch (fullscreen + sticky) to my 1.4.0-based build; happy to report back after it's been through daily use if that's useful.

jacobktm (approved):

PR #2691, opened 2026-08-02
feat(xkb): remember and restore keyboard layout per window

### LINKED ISSUE #3092: Per-window / per-workspace keyboard layout
It would be very useful to have the ability to set a different keyboard layout per window or per workspace in COSMIC.

As a programmer and Linux system administrator, I often switch between tasks that require different layouts. For example, I use a French keyboard when writing emails or documentation, but an English (US) layout when coding or working on Linux system administration tasks — since it makes typing characters like [ ] { } | much easier.

Having the option to assign a keyboard layout per window or workspace would greatly improve productivity and convenience for multilingual users or developers who frequently switch between these contexts.

Thank you

6insanes:
Adds an opt-in `per_window_layout` flag to `XkbConfig`. When enabled, the active xkb layout group is saved for the window losing keyboard focus and restored for the window gaining it, so each window keeps its own layout instead of sharing one global layout across the whole session — matching GNOME's "Use different sources for different windows".

**Behavior**
- Off by default, no behavior change unless explicitly enabled.
- A window that has never had focus before (or was just mapped) simply inherits whichever layout is currently active — no jarring reset to layout 0.
- Only applies to actual windows (`Element`/`Fullscreen` focus targets). Popups, layer-shell surfaces (panel/dock), the window-group focus target, and the lock screen are intentionally excluded, so opening a context menu or tooltip never touches the layout.
- The existing `zcosmic_keyboard_layout_v1` protocol and panel indicator need no changes — they already broadcast whatever the seat's active layout is, which this PR keeps in sync with the focused window.

**Testing**
Manually tested on real hardware (KMS session on a spare VT, and now daily-driven in a real session) with 2 layouts (en/ru) and Alt+Shift switching: opening/switching between multiple windows correctly saves and restores each window's own layout.

Closes pop-os/cosmic-epoch#3092.
Related: pop-os/cosmic-settings#1038 (UI toggle) — companion PR for the settings checkbox will follow once this lands, since it needs this field to exist in a published `cosmic-comp-config` revision.

### DIFF
--- a/cosmic-comp-config/src/lib.rs
+++ b/cosmic-comp-config/src/lib.rs
@@ -162,6 +162,10 @@ pub struct XkbConfig {
     pub repeat_delay: u32,
     #[serde(default = "default_repeat_rate")]
     pub repeat_rate: u32,
+    /// Remember and restore the active layout on a per-window basis, instead of
+    /// sharing a single active layout across all windows.
+    #[serde(default)]
+    pub per_window_layout: bool,
 }
 
 impl Default for XkbConfig {
@@ -174,6 +178,7 @@ impl Default for XkbConfig {
             options: None,
             repeat_delay: default_repeat_delay(),
             repeat_rate: default_repeat_rate(),
+            per_window_layout: false,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


6insanes:
FYI for maintainers: the settings-side UI (a toggle for this in `cosmic-settings`' Keyboard → Input Sources page, closing pop-os/cosmic-settings#1038) is already implemented and tested locally. I'm holding off on opening that PR until this one lands, since it needs `cosmic-comp-config` to have a published revision with the new `per_window_layout` field — happy to open it right away once this merges.

PR #2757, opened 2026-08-18
Fix pointer constraints in all (?) cases

chrisglass:
This is a bit of a tricky PR for me. I've been struggling since the partial (to me at least) fix introduced in https://github.com/pop-os/cosmic-comp/pull/2357 (Thanks @skygrango !) seemed to work fine "sometimes", but I still could reproduce the Helldivers 2 "mouse is stuck" bug somewhat often.

So I took it onto myself to, before attempting this branch, try to come up with a reproducer test. I have produced such a reproducer test in a separate repo (the reproducer test did leverage AI, to hook into the Wayland protocol and analyse what is going on). The reproducer code is: https://github.com/chrisglass/pointer-lock-test

Armed with a reliable reproducer, this is my attempt to solve the issue once and for all. 
Most of the issues stem from mouse window constraints, that get set/unset slightly differently in the existing code, and depend largely on where in the tree the surface was at the particular moment the change of e.g. fullscreen happened.

One thing I am unsure of (honestly): initializing RelativePointerManagerState *seemed* to help, but now removing the lines doesn't make the tests fail, and I can't reproduce the failure/success manually. I didn't measure any noticeable impact... but decided to keep it since I am almost sure it caused a failure before. I've played around with it for so long and the reproducer is so unreliable manually that I am honestly unsure now.

### DIFF
--- a/src/backend/winit.rs
+++ b/src/backend/winit.rs
@@ -30,7 +30,10 @@ use smithay::{
         winit::event_loop::pump_events::PumpStatus,
     },
     utils::Transform,
-    wayland::{dmabuf::DmabufFeedbackBuilder, presentation::Refresh},
+    wayland::{
+        dmabuf::DmabufFeedbackBuilder, presentation::Refresh,
+        relative_pointer::RelativePointerManagerState,
+    },
 };
 use std::{borrow::BorrowMut, cell::RefCell, time::Duration};
 use tracing::{error, info, warn};
@@ -242,6 +245,11 @@ pub fn init_backend(
         }
         state.common.refresh();
     }
+
+    // Create relative pointer global.
+    // Without it Xwayland disables pointer warp emulation entirely, so games
+    // relying on relative motion get none when running nested.
+    RelativePointerManagerState::new::<State>(dh);
 
     if state.common.with_xwayland {
         state.launch_xwayland(None);
--- a/src/backend/x11.rs
+++ b/src/backend/x11.rs
@@ -37,7 +37,10 @@ use smithay::{
         wayland_server::DisplayHandle,
     },
     utils::{DeviceFd, Transform},
-    wayland::{dmabuf::DmabufFeedbackBuilder, presentation::Refresh},
+    wayland::{
+        dmabuf::DmabufFeedbackBuilder, presentation::Refresh,
+        relative_pointer::RelativePointerManagerState,
+    },
 };
 use std::{borrow::BorrowMut, cell::RefCell, os::unix::io::OwnedFd, time::Duration};
 use tracing::{debug, error, info, warn};
@@ -386,6 +389,11 @@ pub fn init_backend(
         }
         state.common.refresh();
     }
+
+    // Create relative pointer global.
+    // Without it Xwayland disables pointer warp emulation entirely, so games
+    // relying on relative motion get none when running nested.
+    RelativePointerManagerState::new::<State>(dh);
 
     if state.common.with_xwayland {
         state.launch_xwayland(None);
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -611,28 +611,7 @@ impl State {
                     if let Some((under, surface_location)) = new_under
                         .and_then(|(target, loc)| Some((target.wl_surface()?.into_owned(), loc)))
                     {
-                        let shell = self.common.shell.read();
-                        let is_focused = seat
-                            .get_keyboard()
-                            .and_then(|k| k.current_focus())
-                            .is_some_and(|f| f.has_surface(&shell, &under));
-
-                        if is_focused {
-                            with_pointer_constraint(&under, &ptr, |constraint| match constraint {
-                                Some(constraint) if !constraint.is_active() => {
-                                    let region = match &*constraint {
-                                        PointerConstraint::Locked(locked) => locked.region(),
-                                        PointerConstraint::Confined(confined) => confined.region(),
-                                    };
-                                    let point =
-                                        (ptr.current_location() - surface_location).to_i32_floor();
-                                    if region.is_none_or(|region| region.contains(point)) {
-                                        constraint.activate();
-                                    }
-                                }
-                                _ => {}
-                            });
-                        }
+                        self.maybe_activate_pointer_constraint(&seat, &under, surface_location);
                     }
 
                     let mut shell = self.common.shell.write();
--- a/src/shell/focus/mod.rs
+++ b/src/shell/focus/mod.rs
@@ -374,11 +374,27 @@ fn update_focus_state(
             && let Some(surface) = old_target.wl_surface()
             && let Some(pointer) = seat.get_pointer()
         {
-            with_pointer_constraint(&surface, &pointer, |constraint| {
-                if let Some(constraint) = constraint {
-                    constraint.deactivate();
-                }
+            // Constraints are oneshot in the common case (Xwayland), so only
+            // deactivate if the surface actually lost focus eligibility. An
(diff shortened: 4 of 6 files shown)


skygrango:
I am not sufficiently familiar with X11's mechanisms, so I couldn't find the correct fix. What puzzles me most is that it wasn't the pointer constraint failed, but rather the relative pointer.

Under certain conditions, X11 can permanently lose responsiveness to relative pointer protocol. Even if pointer lock fails, relative pointer events will still be dispatched.

skygrango:
There was a recent fix (dda2d148328d84fe7c1e0e5c4bf126c11a3d1f1b) for pointer constraint, could you check if it helps resolve this problem

chrisglass:
@skygrango yes, this commit is included in this branch's history already, so it does not fix the problem (at least, not fully).

It seems to me that asking for a pointer lock and keeping it between fullscreen transitions is what ends up "fixing" the helldivers case. But of course, since reproducing it by hand is unreliable, it's hard to make completely definite statements (the human loop to test one branch or another is slow). This branch at least ensures that the cursor lock is acquired, and held.

chrisglass:
Looking at the patch for the linked commit (dda2d14) I can see how it would fix some game problems related to relative mouse motion for sure. But I will only help if cursor lock is acquired properly first (this branch). At least that's my working hypothesis of what is happening :)

src/backend/winit.rs
```diff
@@ -243,6 +246,11 @@ pub fn init_backend(
         state.common.refresh();
     }
 
+    // Create relative pointer global.
+    // Without it Xwayland disables pointer warp emulation entirely, so games
+    // relying on relative motion get none when running nested.
+    RelativePointerManagerState::new::<State>(dh);
+
```
Drakulix, line 253: We don't want to initialize this for either the winit or x11 backend, because we don't have a way to process any relative input events from either winit or x11.

So advertising support is just plain wrong, this is only expected to work in the kms-case.

skygrango:
> But I will only help if cursor lock is acquired properly first (this branch). At least that's my working hypothesis of what is happening :)

This sounds more like a case where a fullscreen transition causes specific resource synchronization issues for X11 window.

skygrango:
I’ve been playing Helldivers 2 exclusively in Wayland mode lately without encountering any issues. So perhaps we should look into X11 state management instead.

Drakulix (commented):
Gonna be brutally honest here, this is tough to review.

I suspect given the very verbose comments, that you used AI to generate at least parts of this patch. And you description does not inspire confidence, that you know why this allegedly fixes some(?) things.

And what I need to have confidence in these fixes is a demonstration of clear intent and a description of the circumstances leading to wrong behavior (or a reliable reproducer). I am not sure I see either:

- `xwayland_constraint_focus_override` is incredibly broad (and the client check does nothing, it will always be Xwayland, which is already checked by `is_xwm`) and thus likely abusable by a malicious client.
- It is unclear to me whether `maybe_activate_pointer_constraints` is just an attempt at code-deduplication or if the additional code it contains in actually needed to fix some bugs in all the cases where it is used now. (It is also thus hard to verify, that it doesn't do anything unintentional or misses some code that was very deliberately put there in previous commits.)
- The changes to `update_focus_state` seem sensible, but we already have code in place, that is supposed to make sure that `KeyboardFocusTarget`s of the same X11 window (no matter if a Toplevel or Fullscreen) compare for equality. So is this new code even hit in practice? If so why do we enter this code path in the first place?

Stuff like this does seem to me, like there wasn't much of an attempt to understand cosmic-comps structure and existing code, but instead makes the PR feel very much like a set of band aids.

EDIT: Please note that this is very much just an expression of what the code looks like to me (which is very much colored by reviewing waaay too much slop in the recent months). I don't mean to insult you or your ability and am in fact happy somebody is trying to investigate this issue. But I am afraid in it's current state this PR definitely needs work.

chrisglass:
> I’ve been playing Helldivers 2 exclusively in Wayland mode lately without encountering any issues. So perhaps we should look into X11 state management instead.

Hmm... would you mind sharing your game's startup parameters exactly? You are running it in wayland mode as I understand, which I assume you mean "PROTON_ENABLE_WAYLAND=1" What Proton version are you using (just in case)? 

If you are running the game fine on cosmic-comp after dda2d14 , then I must be holding it wrong, and my entire assumption (from looking at what the game did with regards to pointer lock) is wrong...

chrisglass:
> Gonna be brutally honest here, this is tough to review.

Nothing brutal about it, I apologize if you feel I wasted your time. I see how this PR might suffer from the overzealous newbie syndrome, it seems my original assumption is wrong in the first place (since apparently others do not suffer from the same symptoms I observe...). Perhaps, my instinct of trying to fix things in a (shoddy, maybe) way was wrong and I should have reached out in an issue first ("is this the right approach" etc..)

> 
> And you description does not expire confidence, that you know why this allegedly fixes some(?) things.

> Stuff like this does seem to me, like there wasn't much of an attempt to understand cosmic-comps structure and existing code, but instead makes the PR feel very much like a set of band aids.

This is probably correct, in that I must have spent too little time on this before sending a branch, or failed to communicate the current state of my understanding well.

Again, I apologize.

Drakulix:
> Nothing brutal about it, I apologize if you feel I wasted your time. I see how this PR might suffer from the overzealous newbie syndrome, it seems my original assumption is wrong in the first place (since apparently others do not suffer from the same symptoms I observe...). Perhaps, my instinct of trying to fix things in a (shoddy, maybe) way was wrong and I should have reached out in an issue first ("is this the right approach" etc..)

Don't worry, as I said, I am glad somebody is trying to debug this as especially the Xwayland code is difficult and takes a lot of developer time. (Especially since the main use-case is games and those are proprietary and wine also is quite a complex project.)

It would have probably helped to simply open this PR as a draft and clarify in some more details what thoughts went into the design process. In that case I wouldn't have done a more in-depth review, while it still gives a way to share WIP code in a way an issue does not.

So if you aren't too discouraged from looking into this some more (if you are, I need to apologize), feel free to re-open the PR at some point and let us continue the discussion here.

chrisglass:
Aww, thanks, that's a great answer. Indeed, not flagging it as WIP was a bad mistake, I see it from your perspective and I'm sorry - it must be quite frustrating.

So, before I waste more of your time, I want to really nail down the "I'm not holding it wrong, it's actually a bug" assumption. If @skygrango doesn't experience the symptoms (because ... well I can definitely see the mouse "not working" when fullscreen in my current setup), I want to understand what we do differently.

skygrango:
>  I want to really nail down the "I'm not holding it wrong, it's actually a bug" assumption. If @skygrango doesn't experience the symptoms (because ... well I can definitely see the mouse "not working" when fullscreen in my current setup), I want to understand what we do differently.

This is indeed a bug, but it should only occur under Xwayland, at least according to my testing.

> Hmm... would you mind sharing your game's startup parameters exactly? You are running it in wayland mode as I understand, which I assume you mean "PROTON_ENABLE_WAYLAND=1" What Proton version are you using (just in case)?

`PROTON_ENABLE_WAYLAND=1` is required, and you also need to use a Proton version that was built with wayland.drv, such as Proton-GE or Proton-CachyOS.

chrisglass:
Ahh then I am holding it slightly wrong indeed :) My Proton-GE install is not selected.

Ok, let's see if this could be fixed for the xwayland case then. Indeed a lot of this branch is useless if that's the root cause.

In a WIP branch, with better comments this time.

skygrango:
steam overlay does not support wayland yet, so this can be a bit of a problem. If we can fix the issue with xwayland, it would be very helpful for players

PR #2759, opened 2026-08-19
chore: update smithay

Drakulix:
This was a bit tricky, since the new pointer constraint handling is a little deadlock prone...

@pop-os/quality-assurance This needs testing for fullscreen games that capture the cursor (most FPS games) and other apps that confine the cursor (e.g. VMs).

- [x] I have disclosed use of any AI generated code in my commit messages.
  - If you are using an LLM, and do not fully understand the changes it is making to the code base, do not create a PR.
  - In our experience, AI generated code often results in overly complex code that lacks enough context for a proper fix or feature inclusion. This results in considerably longer code reviews. Due to this, AI authored or partially authored PRs may be closed without comment.
- [x] I understand these changes in full and will be able to respond to review comments.
- [x] My change is accurately described in the commit message.
- [x] My contribution is tested and working as described.
- [x] I have read the [Developer Certificate of Origin](https://developercertificate.org/) and certify my contribution under its conditions.

PR #2785, opened 2026-08-26
chore: update smithay

skygrango:
need: https://github.com/Smithay/smithay/pull/2144
related: https://github.com/pop-os/cosmic-comp/pull/2759

First of all, I apologize. I'm sorry that the previous approach wasn't good enough. Hopefully, everything goes smoothly this time.

This update introduces `InputTime` and updates Pointer Constraint.

I tested several games, and none of them caused compositor to crash.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -42,9 +42,9 @@ use smithay::{
     backend::input::{
         AbsolutePositionEvent, Axis, AxisRelativeDirection, AxisSource, Device, DeviceCapability,
         GestureBeginEvent, GestureEndEvent, GesturePinchUpdateEvent as _,
-        GestureSwipeUpdateEvent as _, InputBackend, InputEvent, KeyState, PointerAxisEvent,
-        ProximityState, TabletToolButtonEvent, TabletToolEvent, TabletToolProximityEvent,
-        TabletToolTipEvent, TabletToolTipState, TouchEvent,
+        GestureSwipeUpdateEvent as _, InputBackend, InputEvent, InputTime, KeyState,
+        PointerAxisEvent, ProximityState, TabletToolButtonEvent, TabletToolEvent,
+        TabletToolProximityEvent, TabletToolTipEvent, TabletToolTipState, TouchEvent,
     },
     desktop::{PopupKeyboardGrab, WindowSurfaceType, utils::under_from_surface_tree},
     input::{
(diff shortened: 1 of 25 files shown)


src/input/mod.rs
```diff
@@ -2956,23 +2991,14 @@ impl State {
             let serial = SERIAL_COUNTER.next_serial();
             let under = State::surface_under(point, &output, &self.common.shell.write())
                 .map(|(target, pos)| (target, pos.as_logical()));
-            let time = self.common.clock.now();
-            pointer.relative_motion(
```
skygrango, line 2960: @ids1024 I’d like to ask whether removing `relative_motion` could introduce any regressions.

While testing XWayland game(Gray Zone Warfare with GE-Proton 11-5) in the pointer-locked state, I noticed some unusual behavior. When rotating the camera, there was noticeable jitter, and it seemed like something strange was happening on the XWayland.

After removing `relative_motion`, I was no longer able to reproduce the issue during short testing.

Are there any known use cases or potential regressions I should be aware of when removing `relative_motion`?

skygrango: from https://gitlab.winehq.org/wine/wine/-/merge_requests/10830 :
> The existing SetCursorPos implementation abuses zwp_locked_pointer_v1's set_cursor_position_hint side-effect: it briefly locks the pointer, sets the position hint, then unlocks. The compositor clamps position hints to locked-surface bounds, so warps to coordinates outside that area are silently dropped, and the lock/unlock cycle has subtle timing and side-effect issues.

I guess if we add relative_motion here, it could very likely interfere with the calculation of the rotation angle.

Drakulix:
Thanks but I was already half-way done updating https://github.com/pop-os/cosmic-comp/pull/2759 after merging https://github.com/Smithay/smithay/pull/2144.

I might incorporate parts of this PR, but I'll continue with the other one.

skygrango:
> I might incorporate parts of this PR, but I'll continue with the other one.

Feel free to take it. Thank you for the review :)

src/input/mod.rs
Drakulix: Yeah this is here on purpose as Xwayland requires this event to recognize the warp as such. What wine is doing here seems widely out of spec.

skygrango: This may be why the relative_motion event stops responding in XWayland games. It’s possible that Wine’s workaround mechanism requires us to handle the pointer leave surface state immediately.

This improvement may be able to resolve the issue with Xwayland. at least in today’s testing, I haven’t encountered any cases where an XWayland game stopped processing relative_motion events.

PR #2788, opened 2026-08-27
fix screenshot tool sometimes crashing

### LINKED ISSUE #348: Portal SIGSEGV crash-loop taking screenshots (multi-monitor) — unwrapped radius_too_large protocol error
# Portal SIGSEGV crash-loop taking screenshots (multi-monitor) — unwrapped `radius_too_large` protocol error

## Summary

`xdg-desktop-portal-cosmic` crashes (SIGSEGV) in a systemd restart loop when taking a screenshot, reproducibly on multi-monitor setups. The screenshot overlay is a layer surface that can be 1×1 before it is mapped; libcosmic applies a default corner radius to it, and `cosmic-comp` then posts a fatal `cosmic_corner_radius_layer_v1` `radius_too_large` protocol error. The portal's Wayland thread `.unwrap()`s that error instead of handling it, so the whole process aborts and is restarted, breaking screenshots (and, because the portal owns the screenshot `wl_data_offer`, pasting the result).

Original report and reproduction by @sashok-15 on the (closed) pop-os/cosmic-comp#2638; filing here because that is the correct repo — see "Why here" below.

## Crash

```
panic: thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value:
  Backend(Protocol(ProtocolError { code: 1, object_id: 204,
  object_interface: "cosmic_corner_radius_layer_v1", message: "" }))'
```

Preceded by the compositor rejecting the radius, followed by the thread terminating and systemd restarting the service:

```
cosmic_corner_radius_layer_v1#204: error 1: ... corner radius too large
ERROR xdg_desktop_portal_cosmic::wayland: wayland thread crash; terminating
org.freedesktop.impl.portal.desktop.cosmic.service: Main process exited, code=dumped, status=11/SEGV
... Starting org.freedesktop.impl.portal.desktop.cosmic.service ...   (loops)
```

## Reproduction

1. Connect two monitors.
2. Take a screenshot (`cosmic-screenshot` or the screenshot portal).
3. The portal crash-loops (journal above).
4. Disable the second monitor → screenshot succeeds. Re-enable → fails again.

Consistently reproducible.

## Root cause

Two things stack:

1. **The screenshot overlay can carry a corner radius while unmapped/1×1.** In `src/screenshot.rs` the per-output overlay is created as:

   ```rust
   SctkLayerSurfaceSettings {
       layer: Layer::Overlay,
       anchor: Anchor::all(),
       size: Some((None, None)),
       size_limits: Limits::NONE.min_height(1.0).min_width(1.0),
       ..Default::default()
   }
   ```

   The portal does not set a corner radius itself, but libcosmic applies a default one to its layer surfaces. Before the overlay is mapped at full size it can be 1×1, so on the compositor side `half_min_dim` is `0` and any non-zero radius is "too large." Multi-monitor makes this reliable: the per-output overlays are set up before they are all sized.

2. **The portal aborts on the resulting protocol error.** The Wayland thread's dispatch/roundtrip result is `.unwrap()`ed (`src/wayland/mod.rs`), so a fatal protocol error panics the thread ("wayland thread crash; terminating") and the process SIGSEGVs into a restart loop, instead of failing gracefully.

## Why here, not cosmic-comp

The compositor emitting `radius_too_large` for a corner radius set on a surface with no buffer is **intended behaviour**. Per @Drakulix on pop-os/cosmic-comp#2638, under Wayland's atomic-commit model a client must not keep a corner radius set on a commit whose surface has no adequately-sized buffer, and should unset it when the buffer is removed. So the fix belongs on the client (portal/toolkit) side, which is why that compositor PR was closed.

## Suggested fix

Client-side, either or both:

- Have the fullscreen screenshot overlay **opt out of a corner radius** (a screen-covering overlay has no visible rounded corners to gain), or only apply/keep a radius once the surface has a buffer at least `2 × radius` in each dimension — matching the atomic-commit contract above. If the radius comes purely from a libcosmic default, this may need a libcosmic-level opt-out for layer surfaces.
- Independently, **handle the dispatch error** in `src/wayland/mod.rs` rather than `.unwrap()`ing it, so a protocol error can't turn into a SIGSEGV restart loop.

## Environment

- Pop!_OS 24.04, COSMIC
- NVIDIA driver 580.173.02
- Dual monitor (`DP-1`, `DP-2`)
- Kernel `7.0.11-76070011-generic`

Credit: crash, logs, and reproduction from @sashok-15.

hojjatabdollahi:
Fixes https://github.com/pop-os/xdg-desktop-portal-cosmic/issues/348

The layer radius hook validated the *pending* corner radius against `bbox_from_surface_tree()`, the *currently committed* buffer bounding box. This made any non-zero radius fatal for a layer surface that had not yet committed its first buffer.

___

### DIFF
--- a/src/wayland/protocols/corner_radius.rs
+++ b/src/wayland/protocols/corner_radius.rs
@@ -5,16 +5,15 @@ use cosmic_protocols::corner_radius::v1::server::cosmic_corner_radius_toplevel_v
 use cosmic_protocols::corner_radius::v1::server::{
     cosmic_corner_radius_manager_v1, cosmic_corner_radius_toplevel_v1,
 };
-use smithay::desktop::utils::bbox_from_surface_tree;
 use smithay::reexports::wayland_protocols::xdg::shell::server::xdg_popup::XdgPopup;
 use smithay::reexports::wayland_protocols_wlr::layer_shell::v1::server::zwlr_layer_surface_v1::ZwlrLayerSurfaceV1;
 use smithay::reexports::wayland_server::New;
 use smithay::reexports::wayland_server::protocol::wl_surface::WlSurface;
-use smithay::utils::{HookId, Logical, Point, Rectangle};
+use smithay::utils::{HookId, Logical, Rectangle};
 use smithay::wayland::compositor::Cacheable;
 use smithay::wayland::compositor::add_pre_commit_hook;
 use smithay::wayland::compositor::with_states;
-use smithay::wayland::shell::wlr_layer::WlrLayerShellHandler;
+use smithay::wayland::shell::wlr_layer::{LayerSurfaceAttributes, WlrLayerShellHandler};
 use smithay::wayland::shell::xdg::{SurfaceCachedState, XdgShellSurfaceUserData};
 use smithay::{
     reexports::{
@@ -655,7 +654,6 @@ fn pad_rect(
 }
 
 fn layer_radius_hook<D: 'static>(_state: &mut D, _dh: &DisplayHandle, surface: &WlSurface) {
-    let bbox = bbox_from_surface_tree(surface, Point::default());
     with_states(surface, |surface_data| {
         let corners = *surface_data
             .cached_state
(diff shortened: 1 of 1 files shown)


src/wayland/protocols/corner_radius.rs
```diff
@@ -665,8 +663,34 @@ fn layer_radius_hook<D: 'static>(_state: &mut D, _dh: &DisplayHandle, surface: &
             .cached_state
             .get::<CacheablePadding>()
             .pending();
+
+        //  The radius and padding is relative to the size of `zwlr_layer_surface_v1` (according to
```
zyads, line 667: @mmstick  doesnt like claude comments 😋

hojjatabdollahi: What makes this claude comment?

zyads: bad joke on my part. I'm the one who hit the crash in #348, so I'm just glad to see the fix land.

Only real note: "`last_acked` is the becomes current" reads like a leftover word, otherwise LGTM. Thanks for the fix! :)

hojjatabdollahi: If anything this is the opposite of a claude comment! Instead of a "well-crafted hallucinated wall of text", it's a "poorly written and grammatically wrong, but accurate (AFAIK) couple of notes"!

Also the devs have all the rights in the world to reject AI generated walls of text if they think it would make their life harder! So don't take it personally, you got some good feedback on your PR!

Thanks for the note, I will fix the typo, and maybe rewrite the comment tomorrow.

Drakulix (approved):
Looks correct to me. Thanks!

jacobktm (approved):

PR #2792, opened 2026-08-28
Add `cosmic-session-lock-layer-v1` protocol

ids1024:
A very rough draft of an implementation of a protocol to allow cosmic-bg to be shown on the lock-screen.

This could also be used by the panel, but we'll also need another protocol to notify the panel that the session is locked so it can hide applets that shouldn't be shown on the lock screen, and propagate that information to applets. That is more subtle to get right.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -74,7 +74,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ef6978589202a00cd7e118380c448a08b6ed394c3a8df3a430d0898e3a42d046"
 dependencies = [
  "android-properties",
- "bitflags 2.11.0",
+ "bitflags 2.13.2",
  "cc",
  "cesu8",
  "jni",
@@ -390,9 +390,9 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "bitflags"
-version = "2.11.0"
+version = "2.13.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "843867be96c8daad0d758b57df9392b6d8d271134fce549de6ce169ff98a92af"
+checksum = "3ded4057c258ba199e2d26386d3af3780957ecaee6c4ef4041c6b4b8b97c0b06"
 dependencies = [
  "serde_core",
 ]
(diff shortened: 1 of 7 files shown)


hojjatabdollahi:
I tried this with [glowberry](). But it doesn't redraw. Looks like it's short circuiting:
https://github.com/pop-os/cosmic-comp/blob/bb5f8b74fa6ca85a263b4a89b6dd09973a6ebd5a/src/shell/mod.rs#L2049-L2055

It needs to handle show-on-lock layer:
```rust
        if let Some(session_lock) = &self.session_lock {
            if let Some(output) = session_lock
                .surfaces
                .iter()
                .find(|(_, v)| v.wl_surface() == surface)
                .map(|(k, _)| k)
            {
                return Some(output);
            }
            // schedule redraw for commits by show-on-lock layers
            return self.outputs().find(|o| {
                layer_map_for_output(o)
                    .layer_for_surface(surface, WindowSurfaceType::ALL)
                    .is_some_and(|l| layer_show_on_lock(l.wl_surface()))
            });
        }

```


https://github.com/user-attachments/assets/682ec520-ca31-48d7-b154-71af494ace9d

ids1024:
Yeah, there are a couple places that iterate over surfaces that still need to be updated.

ids1024:
Looking at the places surface iteration happens, I think this addresses everything now. `render_input_order_internal` is a little verbose.

I think we can merge this once https://github.com/pop-os/cosmic-protocols/pull/88 is merged, then update cosmic-bg/greeter/osk to use it.

ids1024:
This can be tested with https://github.com/pop-os/cosmic-bg/pull/144 and https://github.com/pop-os/cosmic-greeter/pull/528.

jackpot51 (approved):

jacobgkau (changes requested):
I'm testing the `session-lock-layer` branch of `cosmic-comp`, `cosmic-greeter`, and `cosmic-bg`. (Currently commits e3db65c, f3219a3, and 0946e75.) I'm seeing the panel and dock both still appearing after locking the screen. I'm assuming that's not expected behavior?

The dock is interactable, and I can click an icon to launch an app, although the app isn't visible until after I unlock the screen. I shouldn't be able to do that while it's locked.

jacobgkau:
Oh, https://github.com/pop-os/cosmic-panel/pull/683 is on the same branch and probably got installed at the same time. I'll try without that.

jacobgkau (approved):
Ok, without that cosmic-panel branch installed, this seems to work fine. It gets rid of the flicker when locking the screen. Still appears as expected (other than https://github.com/pop-os/cosmic-greeter/issues/483 still occurring), with or without Frosted Glass enabled. Also tested with multiple displays & on multiple GPU brands, with no issues.

PR #2797, opened 2026-08-31
RFC: implement commit-timing-v1 and fifo-v1

skygrango:
related: https://github.com/pop-os/cosmic-comp/pull/1549

The main purpose of this PR is to verify that the `commit-timing-1` and `fifo-1` functionality work correctly.

Test conditions:
- Vsync ON
- Enable built-in frame limiter
- PROTON_ENABLE_WAYLAND=1
- Disable DLSS, FSR, XeSS (those tech will force Vsync to turn off)
- WAYLAND_DEBUG show `wp_commit_timer_v1` and `wp_fifo_v1`
- no crash, no hang indefinitely and frame limiter work = :white_check_mark:

| Game | Vulkan | DXVK | VKD3D | Note |
|----------|------------|---------|------------| -------|
|   Ready or Not    |        |      |  :white_check_mark:    |  AMD  |
|   Elden Ring    |        |      |  :white_check_mark:    |  AMD  |
|   STAR WARS Zero Company    |        |      |  :white_check_mark:    |  AMD  |
|   MONSTER HUNTER WILD    |        |      |  :white_check_mark:    |  AMD  |
|   ARC Raiders    |        |      |  :white_check_mark:    |  AMD  |
|   PRAGMATA   |        |      |  :white_check_mark:    |  AMD  |
|   Sekiro™: Shadows Die Twice   |        |   :white_check_mark:   |    |  AMD  |

Note: if someone want to test DXVK game, you need custon proton: https://github.com/NelloKudo/proton-dev/actions/runs/33871680219 and `PROTON_PRESENT_TIMING=1 PROTON_ENABLE_WAYLAND=1`

logs sample:

```
[12:38:32.982612] {mesa vk surface 39 swapchain 2 queue} wp_presentation_feedback#60.sync_output(wl_output#22)
[12:38:32.982619] {mesa vk surface 39 swapchain 2 queue} wp_presentation_feedback#60.presented(0, 5433, 958401000, 4166840, 0, 34657727, 7)
[12:38:32.984089] {mesa vk display queue}  -> wp_linux_drm_syncobj_surface_v1#65.set_acquire_point(wp_linux_drm_syncobj_timeline_v1#63, 0, 10)
[12:38:32.984097] {mesa vk display queue}  -> wp_linux_drm_syncobj_surface_v1#65.set_release_point(wp_linux_drm_syncobj_timeline_v1#68, 0, 10)
[12:38:32.984102] {mesa vk display queue}  -> wl_surface#39.attach(wl_buffer#75, 0, 0)
[12:38:32.984104] {mesa vk display queue}  -> wl_surface#39.damage_buffer(0, 0, 2147483647, 2147483647)
[12:38:32.984107] {mesa vk display queue}  -> wp_commit_timer_v1#57.set_timestamp(0, 5433, 970401520)
[12:38:32.984111] {mesa vk surface 39 swapchain 2 queue}  -> wp_presentation#43.feedback(wl_surface#39, new id wp_presentation_feedback#60)
[12:38:32.984121] {mesa vk display queue}  -> wp_fifo_v1#56.set_barrier()
[12:38:32.984129] {mesa vk display queue}  -> wp_fifo_v1#56.wait_barrier()
[12:38:32.984136] {mesa vk display queue}  -> wl_surface#39.commit()
[12:38:32.984144] {mesa vk display queue}  -> wp_fifo_v1#56.wait_barrier()
[12:38:32.984151] {mesa vk display queue}  -> wl_surface#39.commit()
[12:38:32.984174] {mesa vk display queue} discarded wl_buffer#79.release()
[12:38:32.986777] {Display Queue} wl_display#1.delete_id(62)
[12:38:32.986787] {mesa vk surface 39 swapchain 2 queue} wp_presentation_feedback#62.sync_output(wl_output#22)
[12:38:32.986793] {mesa vk surface 39 swapchain 2 queue} wp_presentation_feedback#62.presented(0, 5433, 962567000, 4166840, 0, 34657728, 7)
```

~need: https://github.com/Smithay/smithay/pull/2149~

~The purpose of this PR is to serve as a proof of concept (POC) for validating the `schedule_barrier` handler I added to Smithay.~  No longer need to update smithay.

~The `OutputSurface` introduced in this PR is optional rather than required. It is only used to make it easier for me to iterate on and test the output/window behavior during development, and is not intended to be a fundamental requirement of the approach.~ No longer need to lookup

### DIFF
--- a/src/utils/prelude.rs
+++ b/src/utils/prelude.rs
@@ -16,6 +16,7 @@ use crate::{config::EdidProduct, shell::zoom::OutputZoomState};
 
 use std::{
     cell::{Ref, RefCell, RefMut},
+    collections::HashMap,
     sync::{
         Mutex,
         atomic::{AtomicU8, Ordering},
@@ -203,13 +204,18 @@ impl OutputExt for Output {
             return;
         };
 
-        let dh = &state.common.display_handle.clone();
+        let mut clients = HashMap::new();
         fifo_barriers.0.lock().unwrap().drain(..).for_each(|item| {
             item.barrier.signal();
-            state
-                .client_compositor_state(&item.client)
-                .blocker_cleared(state, dh);
+            clients.insert(item.client.id(), item.client);
         });
+
+        let dh = &state.common.display_handle.clone();
+        for (_id, client) in clients {
+            state
+                .client_compositor_state(&client)
+                .blocker_cleared(state, dh);
+        }
     }
 
     fn set_avg_frametime(&self, duration: Option<Duration>) {
--- a/src/utils/prelude.rs
+++ b/src/utils/prelude.rs
@@ -2,7 +2,9 @@ use cosmic_comp_config::output::comp::{AdaptiveSync, OutputConfig, OutputState};
 use smithay::{
     backend::drm::VrrSupport as Support,
     output::{Output, WeakOutput},
+    reexports::wayland_server::Client,
     utils::Rectangle,
+    wayland::compositor::{Barrier, CompositorHandler},
 };
 
 pub use super::geometry::*;
@@ -36,11 +38,24 @@ pub trait OutputExt {
     fn config_mut(&self) -> RefMut<'_, OutputConfig>;
 
     fn edid(&self) -> Option<&EdidProduct>;
+
+    fn init_fifo(&self);
+    fn fifo_barrier(&self, barrier: Barrier, client: Client);
+    fn signal_fifo(&self, state: &mut State);
 }
 
 struct Vrr(AtomicU8);
 struct VrrSupport(AtomicU8);
 struct Mirroring(Mutex<Option<WeakOutput>>);
+
+#[derive(Debug, Clone)]
+pub struct FifoBarrierItem {
+    pub barrier: Barrier,
+    pub client: Client,
+}
+
+#[derive(Default)]
+struct FifoBarriers(Mutex<Vec<FifoBarrierItem>>);
 
 impl OutputExt for Output {
     fn is_internal(&self) -> bool {
@@ -166,5 +181,34 @@ impl OutputExt for Output {
 
     fn edid(&self) -> Option<&EdidProduct> {
         self.user_data().get()
+    }
+
+    fn init_fifo(&self) {
+        self.user_data()
+            .insert_if_missing_threadsafe(|| FifoBarriers(Mutex::new(Vec::new())));
+    }
+
+    fn fifo_barrier(&self, barrier: Barrier, client: Client) {
+        let user_data = self.user_data();
+        user_data.insert_if_missing_threadsafe(|| FifoBarriers(Mutex::new(Vec::new())));
+        if let Some(barriers) = user_data.get::<FifoBarriers>()
+            && let Ok(mut barriers_vec) = barriers.0.lock()
+        {
+            barriers_vec.push(FifoBarrierItem { barrier, client });
+        }
+    }
+
+    fn signal_fifo(&self, state: &mut State) {
+        let Some(fifo_barriers) = self.user_data().get::<FifoBarriers>() else {
+            return;
+        };
+
+        let dh = &state.common.display_handle.clone();
+        fifo_barriers.0.lock().unwrap().drain(..).for_each(|item| {
+            item.barrier.signal();
+            state
+                .client_compositor_state(&item.client)
+                .blocker_cleared(state, dh);
+        });
     }
 }
--- a/src/utils/prelude.rs
+++ b/src/utils/prelude.rs
@@ -1,4 +1,5 @@
 use cosmic_comp_config::output::comp::{AdaptiveSync, OutputConfig, OutputState};
+use parking_lot::RwLock;
 use smithay::{
     backend::drm::VrrSupport as Support,
     output::{Output, WeakOutput},
@@ -19,6 +20,7 @@ use std::{
         Mutex,
         atomic::{AtomicU8, Ordering},
     },
+    time::Duration,
 };
 
 pub trait OutputExt {
@@ -42,6 +44,9 @@ pub trait OutputExt {
     fn init_fifo(&self);
     fn fifo_barrier(&self, barrier: Barrier, client: Client);
     fn signal_fifo(&self, state: &mut State);
+
+    fn set_avg_frametime(&self, duration: Option<Duration>);
+    fn get_avg_frametime(&self) -> Option<Duration>;
 }
 
 struct Vrr(AtomicU8);
@@ -56,6 +61,8 @@ pub struct FifoBarrierItem {
 
 #[derive(Default)]
 struct FifoBarriers(Mutex<Vec<FifoBarrierItem>>);
+
+struct AvgFrameTime(RwLock<Option<Duration>>);
 
 impl OutputExt for Output {
     fn is_internal(&self) -> bool {
@@ -210,5 +217,17 @@ impl OutputExt for Output {
                 .client_compositor_state(&item.client)
                 .blocker_cleared(state, dh);
         });
+    }
+
+    fn set_avg_frametime(&self, duration: Option<Duration>) {
+        let user_data = self.user_data();
+        user_data.insert_if_missing_threadsafe(|| AvgFrameTime(RwLock::new(None)));
+        if let Some(avg_frametime) = user_data.get::<AvgFrameTime>() {
+            *avg_frametime.0.write() = duration;
+        }
+    }
+
+    fn get_avg_frametime(&self) -> Option<Duration> {
+        *self.user_data().get::<AvgFrameTime>()?.0.read()
     }
 }
--- a/src/utils/prelude.rs
+++ b/src/utils/prelude.rs
@@ -41,7 +41,6 @@ pub trait OutputExt {
 
     fn edid(&self) -> Option<&EdidProduct>;
 
-    fn init_fifo(&self);
     fn fifo_barrier(&self, barrier: Barrier, client: Client);
     fn signal_fifo(&self, state: &mut State);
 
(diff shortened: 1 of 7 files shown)


ids1024:
We definitely want to get fifo and commit-timing working. Though we do have to be careful here since if we fail to signal a fifo barrier for a particular type of surface in any circumstance, it could hang indefinitely.

skygrango:
> We definitely want to get fifo and commit-timing working. Though we do have to be careful here since if we fail to signal a fifo barrier for a particular type of surface in any circumstance, it could hang indefinitely.

yep. I noticed that your original `signal_fifos` seems to hang indefinitely under certain conditions(when switching to full-screen mode), so I replaced it with my version. I’ve tested my version for several months, and it seems to work well. I haven’t investigated yet why the original FIFO signaling mechanism has this issue.

The main purpose of this PR is to investigate whether the commit-timing implementation has reached a sufficient level of correctness, even though its timing precision still needs improvement. It may also provide a good opportunity to use it as a reference when investigating where it differs from your original approach.

skygrango:
Following the [suggestion](https://github.com/Smithay/smithay/pull/2149#issuecomment-5490432536), this no longer needs to depend on updates to smithay.

skygrango:
@ids1024 I’ve updated the `fifo-v1` implementation to use `FifoManagerState::unmanaged` together with `visible_output_for_surface` to keep track of the output associated with the barrier, eliminating the need for a lookup. This should greatly reduce the risk of fifo barrier not being signaled correctly.

skygrango:
commit-timing-v1 requirement:
> content updates should be presented as closely as possible to, but not before, the specified time

I implemented a frame-time-based early commit strategy.

ids1024:
Not having an extra place or two where we iterate over all types of surfaces is definitely good.

src/utils/prelude.rs
```diff
@@ -167,4 +189,45 @@ impl OutputExt for Output {
     fn edid(&self) -> Option<&EdidProduct> {
         self.user_data().get()
     }
+
+    fn init_fifo(&self) {
+        self.user_data()
+            .insert_if_missing_threadsafe(|| FifoBarriers(Mutex::new(Vec::new())));
+    }
```
ids1024: Isn't this redundant if `fifo_barrier()` also calls `insert_if_missing_threadsafe()`?

src/utils/prelude.rs
```diff
@@ -167,4 +189,45 @@ impl OutputExt for Output {
...
+
+    fn fifo_barrier(&self, barrier: Barrier, client: Client) {
+        let user_data = self.user_data();
+        user_data.insert_if_missing_threadsafe(|| FifoBarriers(Mutex::new(Vec::new())));
+        if let Some(barriers) = user_data.get::<FifoBarriers>()
```
ids1024: `insert_if_missing_threadsafe` and `get` can be combining into `get_or_insert_threadsafe`.

src/utils/prelude.rs
```diff
@@ -167,4 +189,45 @@ impl OutputExt for Output {
...
+            state
+                .client_compositor_state(&item.client)
+                .blocker_cleared(state, dh);
+        });
+    }
+
+    fn set_avg_frametime(&self, duration: Option<Duration>) {
+        let user_data = self.user_data();
+        user_data.insert_if_missing_threadsafe(|| AvgFrameTime(RwLock::new(None)));
+        if let Some(avg_frametime) = user_data.get::<AvgFrameTime>() {
```
ids1024: This likewise can use `get_or_insert_threadsafe`.

src/wayland/handlers/compositor.rs
```diff
@@ -235,6 +240,80 @@ impl CompositorHandler for State {
             }
         });
 
+        add_pre_commit_hook::<Self, _>(surface, move |state, _dh, surface| {
```
ids1024, line 182: It's probably unnecessary to add three different pre-commit hooks to each surface in `new_surface()`.

Maybe a single hook that could invoke helper functions for the different things it does?

src/wayland/handlers/compositor.rs
```diff
@@ -378,6 +457,66 @@ impl CompositorHandler for State {
 }
 
 impl State {
+    fn schedule_commit_timing(&self, surface: &WlSurface, deadline: Timestamp) {
+        const SAFE_MARGE: Duration = Duration::from_millis(1);
```
ids1024: Is `MARGE` here a "margin"? If so presumably `SAFE_MARGIN` would be clearer. Unless I misunderstand this.

Is there any particular logic behind the 1ms value, or is it just an extra margin that seemed to work well based on testing?

ids1024:
Storing the fifos to signal with the output seems good. I see it does signal fifos on output removal, which should cover things; either the output will be presented or it will be removed. If a change of primary scanout output happens after setting the barrier it will still wait on the previous output? But that's probably fine.

Commit timing is a protocol I'm a little less familiar with.

> Enable built-in frame limiter

> no crash, no hang indefinitely and frame limiter work 

So is the testing of commit timing mainly just enabling the frame rate limiter and a frame rate overlay and seeing that it seems to report the frame rate it should?

Are there any good clients to test/measure how well it is doing more precisely? I suppose comparing requested commit timings to `presentation-time` feedback.

skygrango:
> Storing the fifos to signal with the output seems good. I see it does signal fifos on output removal, which should cover things; either the output will be presented or it will be removed. If a change of primary scanout output happens after setting the barrier it will still wait on the previous output? But that's probably fine.

work fine in my test, I think the current state of fifo-v1 looks good.

> Commit timing is a protocol I'm a little less familiar with.
> 
> > Enable built-in frame limiter
> 
> > no crash, no hang indefinitely and frame limiter work
> 
> So is the testing of commit timing mainly just enabling the frame rate limiter and frame rate counter and seeing that it seems to report the frame rate it should?

I'm digging deeper into DXVK's design to better understand client-side behavior. Here's how DXVK describes its present timing feature:

> Whether to enable present timing features.
>
> Present timing is used to improve frame time consistency when emulating non-native display modes (e.g. 60 Hz on a high-refresh-rate display), and for the frame rate limiter. It can be disabled for debugging purposes.

FYR: https://github.com/doitsujin/dxvk/pull/5869/changes#diff-51c6436f1bc3ea9e601312655dae5b7534c646e09002ca98467e2758f967014b

To achieve this, DXVK maintains its own internal timeline algorithm and continuously updates it dynamically.

I'm still looking for the most appropriate approach to accommodate this kind of client-side behavior. If you have any suggestions, I'd really appreciate it.

> Are there any good clients to test/measure how well it is doing more precisely? I suppose comparing required commit timings to `presentation-time` feedback.

I recommend: https://github.com/Themaister/Granite/blob/master/tests/present_timing.cpp

the tool is great. It allows you to toggle VSync (FIFO), enable or disable and configure time requests (commit-timing), adjust GPU load, and toggle VRR.

Since this involves frame pacing, it is relevant to https://github.com/pop-os/cosmic-comp/pull/2420

ids1024:
> work fine in my test, I think the current state of fifo-v1 looks good.

Yeah, this version of fifo looks fairly safe.

Commit-timing is a little more complicated. I'm not familiar with the details of optimal frame pacing, but I know it can be a complicated matter.

Probably this version at least a good start and it can be improved later. Though it would be good to see some specific metrics for how well it is working.

skygrango:
> Commit-timing is a little more complicated. I'm not familiar with the details of optimal frame pacing, but I know it can be a complicated matter.

> Probably this version at least a good start and it can be improved later. Though it would be good to see some specific metrics for how well it is working.

After integrating https://github.com/pop-os/cosmic-comp/pull/2420, I think commit-timing performs reasonably well in fullscreen mode. the reason I integrate this is that the PR prevents commits from other non-fullscreen surfaces from interfering with the scheduling, which can affect frame pacing.

here are some screenshots I can use to explain the current situation:
My monitor is running at 240 Hz, which means a refresh interval of 4.167 ms.

The graph below shows the result with commit-timing and FIFO enabled, VRR disabled, and the commit-timing interval set to 10.417 ms. The average frame time is 12.379 ms, which means the presentation is occurring around every third vblank, since 4.167 ms × 3 ≈ 12.5 ms.
Screenshot_2026-09-03_13-29-22

The graph below shows the result with VRR enabled and the commit-timing interval set to 14.063 ms. The average frame time is 14.064 ms.
Screenshot_2026-09-03_13-30-32

If you're interested in testing this as well, I've prepared the branch I used for you to try: https://github.com/skygrango/cosmic-comp/tree/commit-timing_fifo_with_fullscreen_patch

src/wayland/handlers/compositor.rs
skygrango: yeah, it should be `SAFE_MARGIN`.

commit-timing requires us not to present frames early, so this 1 ms is a tolerance designed to respect that requirement. We can improve this part further in the future.

skygrango:
A clarification: `Presentation Error` in the graphs indicates whether the timestamp specified by commit-timing was respected. If it remains as a straight line in the middle, that means it is working as expected.

skygrango:
@ids1024  Do you think I should also include #2420 in this PR, or would you prefer to handle it separately?

ids1024:
Probably best to keep it separate and update https://github.com/pop-os/cosmic-comp/pull/2420 after this one is merged.

skygrango:
Rebase and cleanup.

I don't have an NVIDIA 20–50 series GPU, which supports the open-source driver, so I haven't been able to test this on those GPUs. I've done extensive testing with my AMD GPU, though, and based on those results, I think the current state is good enough for me.

Also, since NVIDIA GPUs are currently forced to use `Duration::Zero` in `next_render_time()`, I'm not sure whether the behavior on NVIDIA hardware might be somewhat unusual.

edit: I noticed something a bit strange. Let me investigate it a little further.

skygrango:
I went back and tested FIFO on its own again, and I realized that FIFO wasn't actually making the window's update rate follow the display refresh rate.

I had forgotten that we only need to process the current barrier, not pending barrier. Otherwise, we would signal all the barriers at once, which defeats the purpose of FIFO.

I've updated the FIFO implementation accordingly. Could you please take another look and verify it?

This time, I'm confident that the window's frame rate is actually being limited by the vblank.

ids1024:
Looks like the Mesa driver packaged on Pop!_OS isn't recent enough to provide support for this in Vulkan. Though `Granite/tests/present_timing.cpp` does detect the necessary functionality on a Debian Sid system.

I'm working on creating my own test client, not involving Vulkan, to get a better sense of this.

mmstick:
The 26.04 base had an update to Mesa 26.2.3 today if that's needed.

skygrango:
There is an issue where enabling FIFO support in Firefox causes severe frame pacing problems during full-screen video playback. A quirk list may be required to circumvent this issue

ids1024:
> The 26.04 base had an update to Mesa 26.2.3 today if that's needed.

Nice. I was already testing 26.04, and that does seem to work now.

> There is an issue where enabling FIFO support in Firefox causes severe frame pacing problems during full-screen video playback. A quirk list may be required to circumvent this issue

Hm. Is that also an issue in Gnome and other compositors with wp-fifo-v1?

git-f0x:
Everything works fine for me on kwin 6.7 on NVIDIA in Firefox (though I don't have a VRR display if that affects anything).

PR #2800, opened 2026-08-31
fix: keep Xwayland frame callbacks progressing

### LINKED ISSUE #2799: Steam overlay updates at ~1 Hz when main Steam window is obscured or minimized in fullscreen Xwayland games
When running a fullscreen Steam game through Xwayland, the in-game Steam overlay (Shift+Tab) becomes extremely sluggish if Steam's main window is fully obscured by the game or minimized. Overlay refresh drops to roughly 1 Hz, making the overlay feel nearly unusable. Leaving any part of the Steam main window visible restores normal responsiveness.

Reproduction steps

1. Start a Steam game through Xwayland in fullscreen.
2. Ensure Steam's main window is fully obscured behind the game or minimized.
3. Open the Steam overlay with Shift+Tab.
4. Move the pointer across overlay controls or interact with its menus.
5. Overlay updates occurring at roughly 1 Hz (extremely slow).
6. Open Steam's main window so part of it is visible.
7. Return to the game. The overlay is now responsive again.

Expected behavior

The Steam overlay should remain responsive whether or not the Steam main window is visible, while native Wayland windows retain the current visibility-based throttling.

Actual behavior

Steam overlay is usable only while some part of the Steam main window is visible. Otherwise it throttles to ~1 Hz.

FreddyFunk:
This fixes sluggish Steam overlays in fullscreen Xwayland games. It does so by letting frame callbacks keep advancing for mapped X11 windows, even ones that are hidden behind other windows or minimized. Native Wayland windows are unaffected and keep their current visibility-based throttling.

fixes #2799

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -139,6 +139,14 @@ struct Localizations;
 pub static LANG_LOADER: LazyLock<FluentLanguageLoader> =
     LazyLock::new(|| fluent_language_loader!());
 
+fn xwayland_frame_throttle(throttle: Option<Duration>, is_xwayland: bool) -> Option<Duration> {
+    if is_xwayland {
+        Some(Duration::ZERO)
+    } else {
+        throttle
+    }
+}
+
 #[macro_export]
 macro_rules! fl {
     ($message_id:literal) => {{
@@ -1365,7 +1373,9 @@ impl Common {
                 && let Some(grab_state) = move_grab.lock().unwrap().as_ref()
             {
                 for (window, _) in grab_state.element().windows() {
-                    window.send_frame(output, time, throttle(&window), should_send);
+                    let throttle =
+                        xwayland_frame_throttle(throttle(&window), window.x11_surface().is_some());
+                    window.send_frame(output, time, throttle, should_send);
                 }
             }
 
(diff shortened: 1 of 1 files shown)


src/state.rs
```diff
@@ -1365,7 +1373,9 @@ impl Common {
                 && let Some(grab_state) = move_grab.lock().unwrap().as_ref()
             {
                 for (window, _) in grab_state.element().windows() {
-                    window.send_frame(output, time, throttle(&window), should_send);
+                    let throttle =
+                        xwayland_frame_throttle(throttle(&window), window.x11_surface().is_some());
```
Drakulix: why does this not simply change the throttle function?

Drakulix (commented):
Thanks for investigating this bug!

So steam will not render the overlay if it's main window doesn't get frequent frame updates? Any further insight you had debugging this? Or was this just a shot in the dark that happened to fix it?

FreddyFunk:
> So steam will not render the overlay if it's main window doesn't get frequent frame updates? Any further insight you had debugging this? Or was this just a shot in the dark that happened to fix it?

From my understanding, the steam overlay is produced by steamwebhelper/CEF and passed through gameoverlayui to gameoverlayrenderer.so which then puts it into the games framebuffer. At least i could not find a separate overlay surface visible to cosmic-comp during debugging. My best explanation is that throttling steams covered X11 window stalls a shared framebuffer injection used by the overlay instances. I cannot prove that proprietary internal edge directly, but at least the A/B was consistent. This also matches Xwaylands ~1 second present fallback. I also found something in KWin that seems to be a related workaround.

https://invent.kde.org/plasma/kwin/-/commit/90161570483026d3a1509ae48029b213d4bb9551

src/state.rs
FreddyFunk: Good question and I have no argument against it. I added a commit that refactors that.

Drakulix (approved):
Code looks good now and indeed the analysis holds up in my testing. That plus the referenced KDE commit, definitely convinces me, that this is suitable solution. Thanks again!

jacobktm (approved):

PR #2801, opened 2026-08-31
fix(xwayland): Properly hide KDE video bridge

### LINKED ISSUE #1204: "Wayland to X Recording bridge" is shown as a white tiled window
https://invent.kde.org/system/xwaylandvideobridge autostarts through `/etc/xdg/autostart`. But on Cosmic-comp, instead of being hidden it shows up as a normal window, all white.

This may reflect a few Xwayland issues. So it would be good to fix this specifically, but it also seems like a nice small test case for some Xwayland things that may impact other applications.

https://invent.kde.org/system/xwaylandvideobridge/-/blob/master/src/contentswindow.cpp

Not sure why the `setOpacity(0);` doesn't work. I would have though XWayland just turns that into a transparent Wayland surface without the compositor needing to do anything.

`KX11Extras::setState(winId(), NET::SkipTaskbar | NET::SkipPager | NET::SkipSwitcher);`

This shouldn't show in things like cosmic-workspaces. Either the workspaces protocol needs to pass through these flags, or with an X window ID, clients could fetch that and other things (like the window icon) from the X server themselves. The former is probably better, though it does mean more protocol bloat.

Or the compositor could have some heuristic to decide which X windows should just be omitted from the top-level info protocol altogether, which would probably make sense at least in this case.

Presumably things like i3 and sway have some heuristic that recognizes that a window like this should not be tiled, even though it's technically visible and has a size. Maybe that's related to `setFlag(Qt::Dialog);`.

FreddyFunk:
I installed KDE Plasma alongside my COSMIC desktop installation, so I can use KDE when gaming. After logging in to COSMIC after the KDE installation, I ran into #1204. This properly fixes #1204 

Without these fixes:

[2026-08-31 20-39-56-firefox-hq (Copy 1).webm](https://github.com/user-attachments/assets/7741ed17-f1f6-494b-90bd-2f572f5071ef)

With these fixes:

[2026-08-31 20-28-33-firefox-hq.webm](https://github.com/user-attachments/assets/999eb900-ae33-4d16-8541-15f8fc2b0809)

### DIFF
--- a/src/shell/element/surface.rs
+++ b/src/shell/element/surface.rs
@@ -211,6 +211,15 @@ struct Sticky(AtomicBool);
 struct GlobalGeometry(Mutex<Option<Rectangle<i32, Global>>>);
 
 impl CosmicSurface {
+    pub fn effective_render_alpha(&self, alpha: f32) -> f32 {
+        match self.0.underlying_surface() {
+            WindowSurface::Wayland(_) => alpha,
+            WindowSurface::X11(surface) => {
+                alpha * surface.opacity().unwrap_or(u32::MAX) as f32 / u32::MAX as f32
+            }
+        }
+    }
+
     pub fn title(&self) -> String {
         match self.0.underlying_surface() {
             WindowSurface::Wayland(toplevel) => with_states(toplevel.wl_surface(), |states| {
--- a/src/shell/element/mod.rs
+++ b/src/shell/element/mod.rs
@@ -653,6 +653,8 @@ impl CosmicMapped {
             return None;
         }
 
+        let alpha = self.active_window().effective_render_alpha(alpha);
+
         match &self.element {
             CosmicMappedInternal::Stack(s) => s
                 .shadow_render_element::<R, CosmicMappedRenderElement<R>>(
@@ -694,6 +696,8 @@ impl CosmicMapped {
         R::TextureId: Send + Clone + 'static,
         CosmicMappedRenderElement<R>: RenderElement<R>,
     {
+        let alpha = self.active_window().effective_render_alpha(alpha);
+
         #[cfg(feature = "debug")]
         if let Some(debug) = self.debug.lock().unwrap().as_mut() {
             let window = self.active_window();
--- a/src/shell/grabs/moving.rs
+++ b/src/shell/grabs/moving.rs
@@ -405,7 +405,10 @@ impl MoveGrab {
             let mut window_geo = self.window.geometry();
             window_geo.loc += location.to_i32_round() + grab_state.window_offset;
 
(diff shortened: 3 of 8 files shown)


git-f0x:
Just a note that newer versions of KDE don't use the bridge anymore, afaik (I don't see it on COSMIC on Arch). But this is likely still a correctness improvement for the Xwayland impl (as mentioned in the issue).

src/shell/element/surface.rs
```diff
@@ -211,6 +211,15 @@ struct Sticky(AtomicBool);
...
 impl CosmicSurface {
+    pub fn effective_render_alpha(&self, alpha: f32) -> f32 {
+        match self.0.underlying_surface() {
+            WindowSurface::Wayland(_) => alpha,
+            WindowSurface::X11(surface) => {
+                alpha * surface.opacity().unwrap_or(u32::MAX) as f32 / u32::MAX as f32
+            }
+        }
+    }
+
```
Drakulix, line 222: I don't think I like the effective_alpha approach here. It is error prone, as you need to make sure to use it every time any window that is potentially from Xwayland is rendered. Imo this should simply be part of `CosmicSurface::push_render_elements`. There is no case where we don't want to respect the windows opacity. This is identical to AlphaModifier for wayland windows imo.

Drakulix (commented):
Thanks, this is an improvement, but I don't think I like how the below layer is implemented here.

First of all, this is very targeted at the video bridge, which we don't officially support in the first place and thus only implements the Below layer. But proper support would also care about the above layer. Imo those should be treated as conceptionally being on the same layer as layer-shell `Bottom` and `Top` elements.

Second of all, I am not sure we want to implement the `Below` hint. Different to layer-shell, these are workspace specific, can be grabbed and the window can request on it's own to be placed below everything or above everything without any user intervention. And thus they suddenly can also not be tiled anymore?

I don't think any of this is good UX. Yes it fixes the Xwayland video bridge, but this gives the same ability to any X11 app and I question that we need to support this mess.

**If** we want to support it, then this needs to go through @pop-os/ux first and we need to discuss in detail how it shall work.

FreddyFunk:
> Just a note that newer versions of KDE don't use the bridge anymore, afaik (I don't see it on COSMIC on Arch). But this is likely still a correctness improvement for the Xwayland impl (as mentioned in the issue).

I did a fresh install of the Fedora 44 COSMIC spin, updated everything and then added KDE Plasma and the KDE video bridge was installed 🤷‍♀️

FreddyFunk:
@Drakulix yeah, that's fair. I converted this pull request to a draft while it goes through @pop-os/ux first

Issue #2706, opened 2026-08-07
Regression: cosmic-comp panics in focus handling when last enabled monitor disconnects


HYP3R00T:
## Description

`cosmic-comp` crashes when the last enabled external monitor disappears and COSMIC switches to the disabled internal laptop panel.

This appears to be a regression of #906 / commit `7ac204ee799224de389ed988ee96a79411adc695`, which fixed a panic when the last output was disconnected. The current panic is now in `src/shell/focus/mod.rs:286`.

My laptop's internal display is physically broken, so I normally disable `eDP-1` and use an external HDMI monitor as the only enabled output. When electricity went out and the external monitor immediately lost power, COSMIC attempted to activate the internal display. The compositor panicked and restarted, taking all Wayland GUI applications with it. The machine itself did not reboot.

## Steps to reproduce

1. Connect an external monitor to a laptop.
2. In COSMIC Settings, disable the internal laptop display (`eDP-1`), leaving the external display as the only enabled output.
3. Open several GUI applications.
4. Disconnect the external display or abruptly remove its power.

## Expected behavior

COSMIC should activate the internal panel and migrate focus/windows without restarting the compositor or applications.

## Actual behavior

The last-output transition causes `cosmic-comp` to panic. COSMIC starts a replacement compositor/session a few seconds later, but GUI applications connected to the previous Wayland compositor are lost.

## Timeline and logs

The incident occurred on 2026-08-07 (Asia/Kolkata / IST):

```text
Aug 07 13:34:34.659165 cosmic-comp[1092224]: Failed to destroy old mode property blob: No such file or directory (os error 2)
Aug 07 13:34:35.154533 cosmic-comp[1092224]: thread 'main' panicked at 'called `Option::unwrap()` on a `None` value': src/shell/focus/mod.rs:286
Aug 07 13:34:35.457936 xdg-desktop-portal-cosmic[1092976]: Io error: Connection reset by peer (os error 104)
```

The replacement compositor and panel started at:

```text
cosmic-comp   Fri Aug 7 13:34:37 2026
cosmic-panel  Fri Aug 7 13:34:39 2026
```

When monitor power returned at 13:44, the replacement compositor survived, but emitted these related output/rendering errors and restarted its panel applets:

```text
Aug 07 13:44:24.859249 cosmic-comp[1263837]: Failed to destroy old mode property blob: No such file or directory (os error 2)
[...]

PR #2811, opened 2026-09-02
shell: Tolerate a removed workspace set in active_num, send_frames and send_dmabuf_feedback

### LINKED ISSUE #2727: Turning off a monitor can kill the whole session (unwrap on removed workspace set, src/shell/mod.rs:1359)
> **Update 2026-09-01: root cause found.** Regression from `eb41b076` (2026-03-06, first shipped in 1.3.0), which added an `active_num` call inside `take_presentation_feedback`'s layer-surface loop, outside the `active_space` guard, while the layer-map mutex is held. That is why the session dies at the *next* output-configuration request rather than immediately, and why it only fails when the removed output carries a layer surface. Four-line fix and full mechanism in [the comment below](https://github.com/pop-os/cosmic-comp/issues/2727#issuecomment-5499597764). Reproduces with a **single** output disabled, on Intel/NVIDIA hybrid, NVIDIA-only, and AMD (@oppsig). **Fix proposed in #2811**; two further unwrap sites, a deterministic enable-plus-mirror trigger and a hotplug failure are in the last comment.

### TL;DR

**Turning off a monitor while another stays on can kill the whole COSMIC session.** Every open application is lost. ~~Turning off one monitor at a time is fine.~~ *(Wrong, see the update below: it is a coin flip.)*

Why: when a display is switched off, cosmic-comp forgets its workspaces immediately, but that display's rendering thread is still running for a moment and asks about those workspaces. The lookup fails, and the code calls `.unwrap()` instead of handling it, so the compositor panics.

The nasty part is that **the session does not die right away.** The compositor keeps running and looks completely healthy for anywhere from 8 seconds to over an hour, then dies at the next display-related request — which makes the crash look like it was caused by whatever you did *next*, not by the disable that actually broke it.

Unlike the other reports of this shape, **this one does not need any physical unplugging** — it reproduces on demand from a normal client request, which should make it much easier to debug.

---

### The panic

```rust
// src/shell/mod.rs:1359
pub fn active_num(&self, output: &Output) -> (Option<usize>, usize) {
    let set = self.sets.get(output).or(self.backup_set.as_ref()).unwrap();
    (set.previously_active.map(|(idx, _)| idx), set.active)
}
```

`remove_output` does `self.sets.shift_remove(output)`, and `backup_set` is only populated when **no** outputs remain. So while at least one display survives, `sets.get()` is `None`, `backup_set` is `None`, and the unwrap panics.

The two accessors immediately above it, `active` and `active_mut`, both return `Option` and already handle this case — only `active_num` unwraps.

The panicking thread is the removed output's own render thread (`surface-DP-4`), so this is a race between the main thread's bookkeeping and a render thread that has not finished shutting down.

### Steps to reproduce

With three displays connected, send **one** output configuration that disables two of them:

```
create_configuration(serial)
  enable_head(eDP-1)
  disable_head(DP-4)
  disable_head(HDMI-A-1)
apply()
```

`apply()` returns `succeeded`. The panic appears within about a second; the session dies some time later.

Disabling **one** output per transaction does not reproduce it: 13+ disable/restore cycles across two- and three-display layouts, on both GPUs of this hybrid machine, zero errors — including operations that fail reliably when batched.

### Logs

```
cosmic-comp[432130]: Failed to submit rendering: Failed to submit result for display
  0: The underlying drm surface encountered an error: DRM access error:
     Page flip commit failed on device `Some("/dev/dri/card1")` (Invalid argument (os error 22))
  1: DRM access error: Page flip commit failed on device `Some("/dev/dri/card1")` (Invalid argument (os error 22))
  2: Invalid argument (os error 22)

cosmic-comp[432130]: thread 'surface-DP-4' panicked at
  'called `Option::unwrap()` on a `None` value': src/shell/mod.rs:1359

# ...8 seconds later, at the next output-configuration request:
cosmic-comp[432130]: thread 'main' panicked at
  'called `Result::unwrap()` on an `Err` value: PoisonError { .. }':
  smithay/src/desktop/wayland/layer.rs:70
cosmic-comp[432130]: thread 'surface-eDP-1' panicked at
  'called `Result::unwrap()` on an `Err` value: PoisonError { .. }':
  smithay/src/backend/drm/output.rs:811
```

Observed five times. Once the machine froze hard enough to need a power cycle.

### A workaround that does not work

Splitting the change into two transactions, each disabling one display, is **not** reliable. Tested with gaps from 0 (protocol handshake only) to 10 seconds, including a sysfs-verified settle between stages: clean four times, fatal twice. It narrows the race window without closing it.

### Suggested fix

Make `active_num` handle the missing key the way `active` and `active_mut` already do — return `Option`, or fall back — instead of unwrapping. A surface on its way out asking about workspaces that are already gone should be a no-op, not a session-ending abort.

To be explicit about what this is *not*: it is not about choosing which display the workspaces move to. `remove_output` has a separate `TODO` about that heuristic ("supposed to be the *most* internal, we just pick the first one for now"), and improving it would not help here — the panic happens whether or not the destination is sensible, because the problem is a stale lookup racing a teardown, not a bad destination.

### Possibly related

All share the shape "an output goes away and something still referencing it panics":

- pop-os/cosmic-epoch#2715 — cosmic-comp `Option::unwrap()` on `None` at `src/shell/layout/floating/mod.rs:308` when displays return via a KVM. Different call site, same class.
- pop-os/cosmic-epoch#3770 — `cosmic-app-library` panics on external-display hot-unplug, sometimes taking the session with it.
- pop-os/cosmic-comp#2474 — hybrid-graphics HDMI unplug freezes the machine; page-flip failure reported as `EPERM` rather than `EINVAL`, no `active_num` panic.
- pop-os/cosmic-comp#2495 — AMD page-flip `EINVAL` stall, no panic or crash.

### Environment

| | |
|---|---|
| OS | Pop!_OS 24.04 LTS |
| Kernel | 7.0.11-76070011-generic |
| cosmic-comp | `0.1~1785355703~24.04~091583a` |
| GPUs | `card0` Intel (i915), `card1` NVIDIA — hybrid |
| Displays | `card0-eDP-1` (internal), `card1-DP-4`, `card1-HDMI-A-1` |
| Mesa | 25.2.8-0ubuntu0.24.04.2 |
| libdrm | 2.4.125-1ubuntu0.1~24.04.2 |

Found while writing a Win+P-style display-mode switcher that speaks `wlr-output-management` directly. It now refuses any configuration that would switch off more than one display at a time, which is the only reliable mitigation I found from the client side.

Grenghis-Khan:
Fixes #2727 ("Turning off a monitor can kill the whole session").

When an output is switched off or made a mirror while another output stays on, `remove_output` drops its workspace set while that output's surface thread may still be mid-frame. Three places then look the set up and unwrap:

- `Workspaces::active_num`, called from `take_presentation_feedback` on the surface thread inside the layer-surface loop, **while the output's layer-map mutex is held**. The panic poisons the mutex and the compositor aborts at the next output-configuration request (`layer_map_for_output(output).arrange()` over all outputs in `apply_config_for_outputs`), anywhere from seconds to over an hour later, taking the session with it. This is the `shell/mod.rs:1359` panic in the issue.
- `Common::send_dmabuf_feedback` and `Common::send_frames`, on the main thread, from a `RenderStates` / `SendFrames` message the surface queued before the set was removed. These abort immediately (`state.rs:1179`, `state.rs:1362`). They are reachable when the set is removed but the surface stays, i.e. a mirror; the issue has a deterministic trigger for it (enable a head and make another head mirror it in one transaction — `set_mirroring` resolves the source against the pre-apply live set, so the follower is recorded as not mirroring).

`backup_set` is only populated when no output remains, so with a survivor all three unwrap `None`. `active` and `active_mut` beside `active_num` already return `Option` for this case (db13eea9); `active_num` was left unwrapping because every caller at the time checked `active()` first. eb41b076 added the call in the layer-surface loop that is not behind that guard.

This makes all three treat a missing set as "nothing to do". Every remaining caller of `active_num` either checks `active()` first on the same guard or uses the index only as a label, so `(None, 0)` is inert.

Tested on Pop!_OS 24.04 (cosmic-comp `314fc67`, NVIDIA-only): before, 4 of 6 single-output disables killed the session; with this change, 25+ mode switches in every direction with zero panics. Details, journal excerpts and the analysis are in #2727.


Grenghis-Khan:
Superseded by #2839, which covers the same three sites plus the seat/last-output case (#2706) that the build carrying this patch still died to on 2026-09-05. Review data posted there.

PR #2817, opened 2026-09-02
fix: use existing order of corner radii

wash2:

The radii don't need to be reordered as far as I can tell, but maybe I have missed something? This gets blurred corners back on the right spot for a rounded panel without gaps. The non-zero radii seem to still somehow be slightly smaller than what the panel set via the protocol, but I assume that is a separate issue.

### DIFF
--- a/src/wayland/handlers/corner_radius.rs
+++ b/src/wayland/handlers/corner_radius.rs
@@ -30,16 +30,16 @@ pub fn surface_corners(states: &SurfaceData, size: Size<i32, Logical>) -> Option
     let corners = guard.current().0?;
 
     Some([
-        u8::try_from(corners.top_left)
+        u8::try_from(corners.bottom_right)
             .unwrap_or(u8::MAX)
             .min(half_min_dim),
         u8::try_from(corners.top_right)
             .unwrap_or(u8::MAX)
             .min(half_min_dim),
-        u8::try_from(corners.bottom_right)
+        u8::try_from(corners.bottom_left)
             .unwrap_or(u8::MAX)
             .min(half_min_dim),
-        u8::try_from(corners.bottom_left)
+        u8::try_from(corners.top_left)
             .unwrap_or(u8::MAX)
             .min(half_min_dim),
     ])


Drakulix (dismissed):

hojjatabdollahi:
Isn't the compositor's internal order for the corner radii [BR, TR, BL, TL] everywhere? Would changing it only in these two spots enough?

wash2:
> Isn't the compositor's internal order for the corner radii [BR, TR, BL, TL] everywhere? Would changing it only in these two spots enough?

As far as I could tell, it seemed to be [TL, TR, BR, BL] for [corner radius protocol](https://github.com/pop-os/cosmic-comp/blob/0d7592857964fe1061440b2b4504dbc4d3ec826d/src/wayland/handlers/corner_radius.rs#L25) and also Iced. Maybe that order could be changed instead. I guess using just one format across the compositor would be ideal, I can go through and change all the orderings in cosmic-comp, but it might be a bit error prone considering I am less familiar with the code base.

hojjatabdollahi:
I think it's [BR, TR, BL, TL] everywhere in the compositor, except the boundary for the corner radius. For example  [here](https://github.com/pop-os/cosmic-comp/blob/0d7592857964fe1061440b2b4504dbc4d3ec826d/src/backend/render/wayland/clipped_surface.rs#L148-L151) in the same file, or for [shadows](https://github.com/pop-os/cosmic-comp/blob/0d7592857964fe1061440b2b4504dbc4d3ec826d/src/backend/render/shadow.rs#L89-L92), or [here](https://github.com/pop-os/cosmic-comp/blob/0d7592857964fe1061440b2b4504dbc4d3ec826d/src/shell/element/window.rs#L431-L437).

It would probably be a good idea to change the undocumented internal order to what the theme, and the spec says.

If not, maybe we only change the boundary [here](https://github.com/pop-os/cosmic-comp/blob/86919566440fffbb437aff9c000774a7c32ce6b4/src/wayland/handlers/corner_radius.rs#L32-L45).

hojjatabdollahi:
So, it seems that the inconsistency between the theme corner radii order, and the internal compositor order is already an issue on master, if you set the corner radii to [0,x,x,x] in theme, the apps correctly will have the top left as a square, so all the widgets are correct, but the window itself has the bottom right corner as a right angle.

With This PR, the mismatch becomes more pronounced, now the top left of a window that used to be curved, is both curved and right angle! and the bottom right, is still right angle, but missing the window content!

image

hojjatabdollahi:
I tried to find all the places that we interact with corner radius: https://github.com/pop-os/cosmic-comp/pull/2819

But I'm not 100% sure why the internal convention was different to begin with.

wash2:
closing for #2819

PR #2834, opened 2026-09-08
fix: set keyboard focus on touch target (similar to tablet/pointer)

### LINKED ISSUE #840: Touchscreen doesn't change focus
When using a touchscreen device, I can interact with windows, however windows don't change focus. this means that unless the window is already focused, you can click buttons, but cannot enable stuff like text boxes for typing.

It would be nice if this was configurable as behavior is actually somewhat desirable as it works really well with waydroid.

As a side question, would it be possible to have a separate focus just for touch input? The idea would be to have `keyboard + mouse` as one input group, then use `touch + <something> (ex. OSK)` as a separate group, so the inputs could interact with applications independently of each other.

My sister often uses a graphics tablet with her PC and would love a workflow like this and has asked me about it in the past. The general idea would be that her mouse and keyboard are free, and uses touch + pen + macropad with multiple apps on the graphics tablet.

map_to_output working for keyboards and mice would work for this, but still would need multiple applications being focused.

hojjatabdollahi:
Currently if you touch a window using your touchscreen, the event is sent to that window, but the window doesn't get brought to the front, or get the focus hint around it.
This fixes it.

fixes https://github.com/pop-os/cosmic-comp/issues/840
____

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -1462,11 +1462,18 @@ impl State {
                     };
                     let under = State::surface_under(position, &output, &shell)
                         .map(|(target, pos)| (target, pos.as_logical()));
+                    let focus_target = State::element_under(position, &output, &shell, &seat);
 
                     std::mem::drop(shell);
 
                     let serial = SERIAL_COUNTER.next_serial();
                     let touch = seat.get_touch().unwrap();
+                    // change the keyboard focus unless the touch is grabbed, like pointer buttons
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
LGTM. Thanks!

leviport (approved):

hojjatabdollahi:
Can we get this also merged please since it's ready?

PR #2839, opened 2026-09-09
Don't lose the session when an output is missing from the workspace sets

### LINKED ISSUE #2706: Regression: cosmic-comp panics in focus handling when last enabled monitor disconnects

### LINKED ISSUE #2727: Turning off a monitor can kill the whole session (unwrap on removed workspace set, src/shell/mod.rs:1359)

dev-goalz:
Switching a monitor off, or a dock dropping its link, can currently take down the
whole COSMIC session and every open window with it. The worst part is the delay. The
session usually keeps running for minutes or hours after the display change that
broke it, so the crash looks like it was caused by whatever you happened to be doing
at the time. On one machine here the gap between cause and crash reached 23 hours.

What happens is that an output's render thread asks about workspaces for an output
the main thread has just removed. On a laptop the common case is the built in panel
being switched off because the lid is closed, while its own render thread is still
finishing a frame. `active_num` unwraps that lookup instead of handling the miss, so
the thread panics. The panic poisons a lock on its way out, and the compositor then
aborts at the next display operation, whenever that happens to be.

This makes the three `Workspaces` lookups total instead of partial, the way `active`
and `active_mut` already are. `idx_for_handle` and `len` had no `backup_set` fallback
at all. It also moves one call out from under the layer map guard so that a panic
there cannot poison it, and fixes a double unwrap on the same accessor in the image
capture path.

Two further sites reach the same lookups through a seat rather than through the
render loop, and neither is covered by the `backup_set` fallback. A seat keeps its
own idea of which output is focused or active, and that field is never checked
against `sets`. When the last output goes away the fallback stores the workspace set
aside and the seat fix ups are skipped, so every seat still names an output that no
longer exists. `update_active` then unwraps `active_space` on it and aborts on the
main thread, with no delay and no poisoned lock, which is a different failure from
the one above but the same root cause.

Rather than harden each consumer, `add_output` now restores the invariant. When it
takes the stored set back it re points any seat that still names an output which is
no longer present. That covers the sites in this PR and any future caller that
trusts a seat's output, which is what the rest of the code already assumes it can do.

A surface on its way out asking about workspaces that are already gone is now a no-op
instead of a session ending abort. The worst case after this is one frame of layer
shell surfaces tagged with workspace zero, on an output that is being torn down
anyway.

### How it was checked

Running on one laptop with an external display on USB-C since 2026-08-31. In that
time the guarded lookup absorbed four faults that would each have aborted the
compositor on an unpatched build, the most recent on 2026-09-07, and the session
survived all of them with the same process still running. Before the patches the
same machine lost its session ten times.

The second commit also covers `send_frames` and `send_dmabuf_feedback`, which unwrap
the same missing set on the main thread when an output becomes a mirror. #2811 found
those two and fixes them the same way; it overlaps with the `active_num` change here
too, so the two will conflict and either can be folded into the other.

Fixes #2727, fixes #2706.

---

### DIFF
--- a/src/shell/focus/mod.rs
+++ b/src/shell/focus/mod.rs
@@ -284,7 +284,25 @@ impl Shell {
                 }
 
                 let output = seat.focused_or_active_output();
-                let space = self.active_space(&output).unwrap();
+                // `focused_or_active_output()` is seat-stored state that is never
+                // validated against `sets`. `Common::refresh_focus` re-points stale
+                // seats before it calls `update_active`, but the other caller,
+                // `Shell::set_focus`, does not - and it fires as a consequence of an
+                // output change, with no user input needed.
+                //
+                // The fallback inside `active_space` is no help on its own. Removing
+                // the last output parks its set in `backup_set` and skips the seat
+                // fix-ups; the next `add_output` then *takes* that backup for the new
+                // output. `sets` is non-empty and `backup_set` is `None` while the
+                // seat still names the output that went away.
+                let Some(space) = self.active_space(&output) else {
+                    tracing::warn!(
+                        target: "cosmic_comp::wsdiag",
+                        output = %output.name(),
+                        "update_active: output absent from workspace sets (upstream unwrap would abort here)"
+                    );
+                    return None;
+                };
                 let stack = space.focus_stack.get(seat);
                 stack.last().and_then(|target| match target {
                     FocusTarget::Window(window) => Some(window.clone()),
--- a/src/shell/mod.rs
+++ b/src/shell/mod.rs
@@ -857,6 +857,7 @@ impl Workspaces {
     pub fn add_output(
         &mut self,
         output: &Output,
+        seats: &Seats,
         workspace_state: &mut WorkspaceUpdateGuard<'_, State>,
     ) {
         if self.sets.contains_key(output) {
@@ -922,6 +923,26 @@ impl Workspaces {
             }
         }
         self.sets.insert(output.clone(), set);
+
+        // Taking `backup_set` above is the point where the fallback stops covering
+        // for stale seats. `remove_output` re-points seats only inside its
+        // `if let Some(new_output)` branch, which is skipped when the *last* output
+        // goes away: it parks the set in `backup_set` and leaves every seat naming
+        // an output that is no longer in `sets`. Bring them back in line here, so
+        // that an output-keyed lookup cannot miss on a seat nothing ever fixed up.
+        //
+        // Only the insert path needs this - it is the one that consumes the backup.
+        for seat in seats.iter() {
+            if !self.sets.contains_key(&seat.active_output()) {
+                seat.set_active_output(output);
+            }
+            if seat
+                .focused_output()
+                .is_some_and(|focused| !self.sets.contains_key(&focused))
+            {
+                seat.set_focused_output(None);
+            }
+        }
     }
 
     pub fn remove_output<'a>(
@@ -1362,12 +1383,14 @@ impl Workspaces {
     }
 
     pub fn active_num(&self, output: &Output) -> (Option<usize>, usize) {
-        let set = self.sets.get(output).or(self.backup_set.as_ref()).unwrap();
+        let Some(set) = self.sets.get(output).or(self.backup_set.as_ref()) else {
+            return (None, 0);
+        };
         (set.previously_active.map(|(idx, _)| idx), set.active)
     }
 
     pub fn idx_for_handle(&self, output: &Output, handle: &WorkspaceHandle) -> Option<usize> {
-        let set = self.sets.get(output).unwrap();
+        let set = self.sets.get(output).or(self.backup_set.as_ref())?;
         set.workspaces
             .iter()
             .enumerate()
@@ -1375,8 +1398,10 @@ impl Workspaces {
     }
 
     pub fn len(&self, output: &Output) -> usize {
-        let set = self.sets.get(output).unwrap();
-        set.workspaces.len()
+        self.sets
+            .get(output)
+            .or(self.backup_set.as_ref())
+            .map_or(0, |set| set.workspaces.len())
     }
 
     pub fn iter(&self) -> impl Iterator<Item = (&Output, &WorkspaceSet)> {
@@ -1535,9 +1560,12 @@ impl Drop for OutputId {
 impl Common {
     pub fn add_output(&mut self, output: &Output) {
         let mut shell = self.shell.write();
-        shell
-            .workspaces
-            .add_output(output, &mut self.workspace_state.update());
+        let shell_ref = &mut *shell;
+        shell_ref.workspaces.add_output(
+            output,
+            &shell_ref.seats,
+            &mut self.workspace_state.update(),
+        );
 
         output
             .user_data()
@@ -2690,12 +2718,22 @@ impl Shell {
                 .upgrade()
                 .unwrap_or_else(|| self.seats.last_active().active_output());
             toplevel_enter_output(&window.active_window(), &output);
-            let set = self
+            // Same window as `update_active`: `active_output()` can name an output
+            // that is in neither `sets` nor `backup_set`. Carrying the fallback is
+            // not sufficient - it has to be allowed to miss.
+            let Some(set) = self
                 .workspaces
                 .sets
                 .get_mut(&output)
                 .or(self.workspaces.backup_set.as_mut())
-                .unwrap();
+            else {
+                tracing::warn!(
+                    target: "cosmic_comp::wsdiag",
+                    output = %output.name(),
+                    "remap_unfullscreened_window: output absent from workspace sets (upstream unwrap would abort here)"
+                );
+                return window;
+            };
             set.sticky_layer.map_internal(
                 window.clone(),
                 Some(state.geometry.loc),
@@ -4065,7 +4103,18 @@ impl Shell {
             return FocusResult::None;
         }
 
-        let set = self.workspaces.sets.get(&output).unwrap();
+        // `seat.active_output()` can name an output that is no longer in `sets`:
+        // `remove_output` only re-points seats when another output remains, so
+        // removing the last one leaves every seat pointing at it while its set
+        // moves to `backup_set`. Fall back the way the other accessors do.
+        let Some(set) = self
+            .workspaces
+            .sets
+            .get(&output)
+            .or(self.workspaces.backup_set.as_ref())
+        else {
+            return FocusResult::None;
+        };
         let sticky_layer = &set.sticky_layer;
         let workspace = &set.workspaces[set.active];
 
@@ -4699,7 +4748,15 @@ impl Shell {
         loop_handle: &LoopHandle<'static, State>,
     ) -> Option<KeyboardFocusTarget> {
         let focused_output = seat.focused_output()?;
-        let set = self.workspaces.sets.get_mut(&focused_output).unwrap();
+        // Same stale-output case as `next_focus`. Note the `?` above does not
+        // cover it: the `set_focused_output(None)` that would make it `None`
+        // lives in the same branch of `remove_output` that is skipped when the
+        // last output goes away.
+        let set = self
+            .workspaces
+            .sets
+            .get_mut(&focused_output)
+            .or(self.workspaces.backup_set.as_mut())?;
         let workspace = &mut set.workspaces[set.active];
 
         if matches!(
@@ -5113,9 +5170,9 @@ impl Shell {
             }
         });
 
+        let namespace = self.workspaces.active_num(output).1;
         let map = smithay::desktop::layer_map_for_output(output);
         for layer_surface in map.layers() {
-            let namespace = self.workspaces.active_num(output).1;
             layer_surface.take_presentation_feedback(
                 &mut output_presentation_feedback,
                 surface_primary_scanout_output,
--- a/src/state.rs
+++ b/src/state.rs
@@ -581,7 +581,12 @@ impl LockedBackend<'_> {
             });
 
             match final_config.enabled {
-                OutputState::Enabled => shell_ref.workspaces.add_output(output, workspace_state),
+                OutputState::Enabled => {
+                    let shell = &mut *shell_ref;
+                    shell
+                        .workspaces
+                        .add_output(output, &shell.seats, workspace_state)
+                }
                 _ => {
                     let shell = &mut *shell_ref;
                     shell.workspaces.remove_output(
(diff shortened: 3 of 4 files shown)


leviport:
The PR template is required. Please add it back and complete the checklist.

dev-goalz:
Adding the capture behind the "How it was checked" note, since a reasoned argument
and a recorded event are not the same thing.

The build under test carried this patch plus a diagnostic that logs the miss instead
of taking it, so the branch that would have been an abort is observable. Stimulus was
the reproducer: dock attached with the lid open, lid shut, dock unplugged, dock
reconnected with the lid still shut.

```
19:33:50  kernel  DP-1 disconnected -> connected      dock attached, lid open
19:33:51  comp    add_output eDP-1, add_output DP-1   sets settle to ["eDP-1","DP-1"]
19:35:01  logind  lid closed
19:35:01  comp    remove_output eDP-1                 sets=["eDP-1","DP-1"]
19:35:05  kernel  DP-1 connected -> disconnected      dock unplugged, lid shut
19:35:06  comp    remove_output DP-1, remove_output eDP-1, then add_output eDP-1
19:35:16  kernel  DP-1 disconnected -> connected      dock reconnected, lid shut
19:35:17  comp    add_output eDP-1, add_output DP-1
19:35:17  comp    remove_output eDP-1                 sets=["eDP-1","DP-1"]
19:35:17  comp    the lookup misses
```

The recorded line:

```
active_num: output absent from workspace sets (upstream unwrap would abort here)
caller = src/shell/mod.rs:5149:41
output = eDP-1
sets   = ["DP-1"]
```

That is the unguarded call inside the layer-map guard in `take_presentation_feedback`.
`eDP-1` is out of `sets` because the lid path removed it, and its own render thread
reaches the lookup before the removal settles. On an unpatched build the `unwrap`
takes `None` here and the compositor aborts. With the lookup made total it returns
instead.

Immediately after: same process, every surface thread alive, no panic anywhere in the
log for it, nothing lost. Before the patch the same machine lost its session ten
times under the same class of stimulus, nine of them at this exact line.

It has kept happening since, and kept being absorbed:

```
2026-08-30  output=eDP-1  sets=["DP-1"]
2026-09-01  output=eDP-1  sets=["DP-1"]
2026-09-02  output=eDP-1  sets=["DP-1"]
2026-09-07  output=eDP-1  sets=["DP-1"]
```

Four in five and a half weeks of ordinary use, with the session surviving each one.

Two caveats so this is not read as more than it is. It is one machine, and the
diagnostic fires only when the race actually lands, so it says nothing useful about
how often the window is open. And it does not address the device-wide lock poisoning
that follows any other surface-thread panic, which is a separate problem this patch
does not claim to close.

dev-goalz:
> The PR template is required. Please add it back and complete the checklist.

Done

Grenghis-Khan:
Author of #2811 here. This is the better fix and should be the one merged; I'll close #2811 in favour of it.

**Independent confirmation, second machine.** I have run the #2811 subset (`active_num` + the two `state.rs` sites) on a NVIDIA-only laptop with a USB-C display since 2026-09-02: 41 mode switches, zero of the patched panics. On 2026-09-05 the session was lost anyway, and it was exactly the case this PR covers and mine does not:

```
12:36:39  kernel   usb 3-4: New USB device found ... USB 2.0 BILLBOARD     (dock re-enumerated)
12:36:42  comp     thread 'main' panicked at 'called `Option::unwrap()` on a `None` value': src/shell/focus/mod.rs:286
```

No display change was requested; the dock hotplug alone did it. So the seat re-pointing in `add_output` is not theoretical on this hardware either. Hoisting `active_num` out from under the layer-map guard is also the right call, and the `flat_map` shape at the two `state.rs` sites is more precise than the early return in #2811, which also skipped the layer-surface frames and feedback that follow.

**One nit.** `Workspaces::len` now returns 0 for a missing set, and `shell/mod.rs:3177` still computes `self.workspaces.len(to_output) - 1`, a `usize` underflow (panic in debug, wrap in release). The other callers already use `saturating_sub(1)`; that one probably should too.

**Two things neither PR addresses**, noted so they are not assumed fixed, both from #2727:

- `apply_config_for_outputs` records a mirror source by looking it up among outputs that currently *have* a set (`state.rs`, the `set_mirroring(match …)` block), so enabling a head and making another head mirror it in one transaction records the follower as not mirroring. After this PR that is no longer fatal, but the follower keeps wrong output-level state until the next config change.
- A saved mirror config that fails to re-apply on a hotplug reload (`Unrecoverable config error: Failed to reset config`) leaves the follower's surface thread spinning on `OutputNoMode` with no back-off: frozen screen, live machine. Separate bug; happy to file it on its own if useful.

dev-goalz:
Thanks, and thanks for closing #2811 in favour of this. The trace from your machine is useful: a dock re-enumerating on its own is the seat case in `add_output`, and it is good to see it on different hardware.

You are right about the underflow. Making `len` total is what opened it, since the old unwrap would have panicked one call earlier. It is at `move_current` on this branch and now uses `saturating_sub(1)` like the other callers, pushed as a separate commit.

Agreed that the two mirror problems are separate from this PR. Please do file the `OutputNoMode` one on its own, a frozen screen on a live machine deserves its own issue. The mirror source lookup in `apply_config_for_outputs` probably wants one too, unless you would rather keep it tracked under #2727.

Issue #2761, opened 2026-08-19
[Bug] Laptop monitor stays active when the laptop is booted with its lid closed


bobcat141:
**Steps to reproduce**
1. Connect a shut down laptop to an external USB-C monitor.
2. Turn on the laptop and close its lid right away.
3. Login using the external monitor and peripherals.
4. Open the display settings.
5. Notice the laptop's screen is active.
6. Move the cursor to the edge that connects to the laptop's screen.
7. Notice the cursor disappears, indicating it's now in the laptop's screen.
8. Quickly open and close the lid of the laptop.
9. Notice the laptop's monitor is not disabled.

**System information**
OS: Pop!_OS 24.04
Package version: `cosmic-comp/noble,now 0.1~1786994555~24.04~8806436`

leviport:
duplicate of https://github.com/pop-os/cosmic-epoch/issues/1532

PR #2840, opened 2026-09-09
Power the built-in panel down when the lid closes instead of leaving it lit

dev-goalz:
Close the lid on a laptop and the built-in panel can stay backlit inside it. It goes
dark only when the idle timeout expires, ten minutes by default. On one machine here
the panel was still lit almost four minutes after the lid shut.

This happens when the internal panel is the only enabled output. `should_handle_lid`
requires two or more outputs, so with one the compositor releases its logind
inhibitor and ignores the lid entirely, and the lid switch handler is gated on
holding that inhibitor. logind policy then applies, and none of its options means
stay awake but blank. With `HandleLidSwitch=lock` the session locks and nothing ever
powers the panel down.

The `>= 2` guard is right as far as it goes. Disabling the only output would leave
the session with nowhere to render. But powering a connector down is not the same as
disabling an output. The connector can go dark while the output stays configured and
present, which is what this does. When no inhibitor is held, the lid switch now sets
DPMS on the internal output instead of being ignored.

One more change is needed for that to hold. Every input event powers all surfaces
back on, including the lid switch itself, so internal outputs are exempted while the
lid is shut. That needs the lid state on hand, so it is tracked as a field on
`Common` and seeded from logind when outputs change, which also makes a compositor
started with the lid already closed behave correctly.

The second patch covers the case where the lid is shut and the external display goes
away, for instance when a dock is unplugged. Re-enabling the internal connector at
that point is not optional, since it is the only output left and a session with none
has nowhere to render, but it must not light up inside a closed lid. It is powered
down instead, the same trade the single-output path makes.

The lid state is a plain field rather than a logind query on purpose. Asking the bus
from the output configuration path deadlocked an earlier version of this change and
hung the greeter, so that call is not made there.

### How it was checked

Running on one laptop since 2026-08-31 across daily dock and lid cycles, with no
recurrence of either the lit panel or the greeter hang that the first version caused.

This is only visible when `HandleLidSwitch` has been changed from the default
`suspend`, since suspending blanks the panel as a side effect. That limits who runs
into it, but a panel inside a closed lid arguably should not be lit whatever logind
is configured to do.

Related: pop-os/cosmic-epoch#1532, #2761.

---

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -1918,14 +1918,40 @@ impl State {
             #[allow(unused_variables)]
             InputEvent::SwitchToggle { event } => {
                 #[cfg(feature = "logind")]
-                if event.switch() == Some(Switch::Lid) && self.common.inhibit_lid_fd.is_some() {
+                if event.switch() == Some(Switch::Lid) {
+                    let closed = event.state() == SwitchState::On;
+                    self.common.lid_closed = closed;
+
+                    if self.common.inhibit_lid_fd.is_none() {
+                        // Fewer than two outputs, so the internal panel is the only
+                        // one. Disabling it would leave the session with nowhere to
+                        // render, which is why the lid is otherwise ignored here.
+                        // Powering the connector down instead keeps the output
+                        // configured while the panel goes dark.
+                        let internal = {
+                            let backend = self.backend.lock();
+                            backend
+                                .all_outputs()
+                                .iter()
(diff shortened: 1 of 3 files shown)


leviport:
The PR template is required. Please add it back and complete the checklist.

dev-goalz:
> The PR template is required. Please add it back and complete the checklist.

Done

PR #2845, opened 2026-09-10
shell/focus: Don't dismiss grabbing popups of exclusive layer surfaces

hojjatabdollahi:
I added context menu to all input boxes for copy/cut/paste in libcosmic. It works in all libcosmic apps, except for `app-library` and `cosmic-launcher`.
 
Turns out `cosmic-comp`'s focus refresh decides that a context menu is not allowed to have focus while `app-library` is open. Since `app-library` opens with `KeyboardInteractivity::Exclusive` and refresh-focus has a rule that if a surface has  exclusive keyboard interactivity then no pop up is a valid focus target, it ungrabs the pop up in the next tick.

The solution would be to check if the parent of the pop up is the surface with exclusive keyboard interactivity and allow it.

[Our discussion in mattermost](https://chat.pop-os.org/pop-os/pl/83frxitawibuimt3xbay71hiyo).

___


ids1024 (approved):
It seems correct to treat popups on an exclusive layer as valid keyboard targets.

hojjatabdollahi:
For QA, this can be tested with https://github.com/pop-os/cosmic-app-library/pull/408.

That PR without this fix:

https://github.com/user-attachments/assets/aff6be78-f53d-4b68-9baa-9b9f3593d1d9

With this fix:

https://github.com/user-attachments/assets/130ef687-b456-401b-8de1-b6e0235d02ca

jacobgkau (approved):

Issue #1495, opened 2025-07-02
Graphic Tablet - Inconsistent  Behavior between Laptop and Desktop


camden-bock:
I know that tablet functionality is still in development - wondering if anyone has suggestions on how I might be able to resolve this on my own system. I am aware that general tablet support is tracked in cosmic settings (https://github.com/pop-os/cosmic-settings/issues/141) and #313   - but this seemed most related to the compositor.

I have a wacom intuous tablet connected via USB to two laptops (Intel S76 Lemur Pros) and a desktop. Each computer is using a Fedora Atomic Cosmic Spin, with an ansible script for identical configuration. However, the wacom tablet controls the cursor on both laptops but not the desktop. Running a live gnome wayland session (also fedora) does enable the tablet, and it is exposed in gnome settings.

The wacom tablet doesn't work on the desktop with PopOS, but the version of libwacom is too old for this device to be supported. On live images of both a clean Fedora Atomic Cosmic Spin and Fedora Cosmic (non-atomic) the wacom tablet does not work.

The wacom tablet appears identically in libinput list-devices across all systems, and libinput debug-events shows appropriate buttons and tracking on all systems. The wacom kernel module is present by default on all systems.

```
Device:                  Wacom Intuos BT M Pen
Kernel:                  /dev/input/event12
Id:                      usb:056a:03c7
Group:                   4
Seat:                    seat0, default
Size:                    216x135mm
Capabilities:            tablet
Tap-to-click:            n/a
Tap-and-drag:            n/a
Tap button map:          n/a
Tap drag lock:           n/a
Left-handed:             n/a
Nat.scrolling:           n/a
Middle emulation:        n/a
Calibration:             n/a
Scroll methods:          none
Scroll button:           n/a
Scroll button lock:      n/a
Click methods:           none
Clickfinger button map:  n/a
Disable-w-typing:        n/a
Disable-w-trackpointing: n/a
Accel profiles:          none
Rotation:                n/a
Area rectangle:          (0.00, 0.00) - (1.00, 1.00)

Device:                  Wacom Intuos BT M Pad
Kernel:                  /dev/input/event13
Id:                      usb:056a:03c7
Group:                   4
Seat:                    seat0, default
Capabilities:            tablet-pad
Tap-to-click:            n/a
Tap-and-drag:            n/a
Tap button map:          n/a
Tap drag lock:           n/a
Left-handed:             n/a
Nat.scrolling:           n/a
Middle emulation:        n/a
[...]

Issue #313, opened 2024-02-17
Support for Graphic & Display Drawing Tablets


senntore:
Currently in COSMIC DE, I can connect the tablet(using bluetooth) & move the pointer with pen but the input is not registering on the screen. 
So in Krita, while I can move the brushes, I can't actually paint using the tablet.
I'm testing with Wacom Intuos Pro & it works in Pop OS 22.04.
This is quite important for me as I use the tablet almost all the time, so I'm hoping there can be basic support for it in alpha & full support in the final release with settings & everything.

Drakulix:
Current support is very rudimentary, however what you are describing sounds like a bug of the current implementation. I'll see what we can do for the alpha release.

However fear not what you describe as "full support" is certainly on the TODO-list for the final release.

senntore:
Just wanted to let you know & anyone else that might come to this issue, I got my tablet working using Open Tablet Driver(https://opentabletdriver.net/).
I just had to install the driver & start the service. It will make cosmic more usable to me before we get the support out of the box, so you guys don't have to worry about this issue for now.

Izaird:
I got one tablet working with OTP(open tablet driver) a wacom pth-451, but my main tablet is not working, a xp-pen deco pro medium

Izaird:
I tested two tablets with OTD and so far only absolute mode and relative mode works, artist mode does not work on any of the tablets I tested (wacom PTH-451, XP-Pen Deco Pro Medium)

senntore:
Hello @Drakulix, thanks for working on this issue. After the last few updates the tablet input seems to work now, even the pen pressure & tilt. Now it seems like I will be able to use cosmic without much trouble.
Though I did ran across some bugs. When we click on a window with the pen, the focus(& active hint) does not switch to that window. 
And the pen input does not work on any libcosmic related stuff like cosmic-panel, applibrary, launcher, dock & other cosmic applications.

Karvie0:
For anyone finding this in the future, it looks like drawing tablet support is being tracked in [pop-os/cosmic-settings/141](https://github.com/pop-os/cosmic-settings/issues/141)

HeitorAugustoLN:
> I tested two tablets with OTD and so far only absolute mode and relative mode works, artist mode does not work on any of the tablets I tested (wacom PTH-451, XP-Pen Deco Pro Medium)

Same experience for me

jonmatthis:
Tablets are broken for me too. 

---

[...]

Issue #141, opened 2023-08-01
Mix Interface dual monitors(HDMI and DP) crashes Cosmic


Vixea:
When running one monitor with HDMI and one with DP this happens.
```
cosmic-comp[2051]: [EGL] 0x3004 (BAD_ATTRIBUTE) eglGetPlatformDisplay: eglGetPlatformDisplay
cosmic-comp[2051]: [EGL] 0x3004 (BAD_ATTRIBUTE) eglGetPlatformDisplayEXT: Could not create EGLDisplay
cosmic-comp[2051]: [EGL] 0x3004 (BAD_ATTRIBUTE) eglGetPlatformDisplay: eglGetPlatformDisplay
cosmic-comp[2051]: [EGL] 0x3004 (BAD_ATTRIBUTE) eglGetPlatformDisplayEXT: Could not create EGLDisplay
cosmic-comp[2051]: Unable to find suitable EGL platform
cosmic-comp[2051]: Failed to restore previous state. Error: invalid argument
```

Vixea:
1. RX 6700XT
2. Sure thing
3. huh I did a search for this to make sure I had a fresh one, couldn't find it but it does exist so let me try it

Vixea:
Bad news running a fast debug build guys me no new information

Vixea:
Do I need to pull out some drm debugging?

Drakulix:
> Bad news running a fast debug build gives me no new information

You can try setting this to debug: https://github.com/pop-os/cosmic-comp/blob/master_jammy/src/logger/mod.rs#L20

It might log a lot in that case though.

> Do I need to pull out some drm debugging?

Wouldn't hurt either.

Vixea:
Cosmic with debug build:
https://pastebin.com/CdPUwv6A
DRM Debugging:
https://drive.google.com/file/d/13dZ0nU2hAx8rmJ8etFxkUf0856U1fBPj/view?usp=sharing

Vixea:
Using my DP monitor I get the same thing, using the HDMI monitor I get the sddm issue a had in #130 keep in mind this is before I get into GDM(which I'm using now) so this is a kernel-level bug.

DRM INFO:
https://drive.google.com/file/d/1kA60ZHJUVbQQbDk3_uzVIQ6rqxMPTada/view?usp=sharing

Vixea:
[...]

PR #2287, opened 2026-04-13
fix(input): fix tablet focus and output routing

danth3b0t:
## Summary

- Fixes the remaining compositor-side issues in cosmic-comp's existing `zwp_tablet_manager_v2` / `zwp_tablet_v2` integration so stylus input behaves correctly in real applications.
- Builds on the smithay tablet support already wired through `TabletManagerState`, `delegate_tablet_manager!`, `TabletSeatTrait`, and the existing `TabletToolAxis`, `TabletToolProximity`, `TabletToolTip`, and `TabletToolButton` event paths.
- Emits pointer frames for tablet axis and proximity updates, focuses the target surface on stylus tip-down, and defaults tablet output routing to the seat's focused or active output so tablets work on external displays instead of always mapping to the internal panel.
- Addresses #313.
- No Wayland tablet protocol code was written from scratch here; smithay already implements `zwp_tablet_v2`, and this change wires the remaining compositor behavior around that existing implementation.

## Testing

- Hardware: Wacom Intuos Pro S (PTH-460) connected over Bluetooth.
- Kernel/input stack was verified before the patch with `libinput debug-events`, which reported `TABLET_TOOL_PROXIMITY` and `TABLET_TOOL_AXIS` events for the pen on `/dev/input/event23`.
- `libwacom` 2.18.0 recognizes the device and its stylus capabilities.
- After the patch, pen input was confirmed working in Krita on both the laptop display and an external monitor.
- Verified proximity detection, pressure-sensitive drawing, tilt, and pen buttons with the current build.
- Tablet pad functionality and settings UI were not exercised as part of this test pass.

## Checklist

- [x] I have disclosed use of any AI generated code in my commit messages.
- [x] I understand these changes in full and will be able to respond to review comments.
- [x] My change is accurately described in the commit message.
- [x] My contribution is tested and working as described.
- [x] I have read the [Developer Certificate of Origin](https://developercertificate.org/) and certify my contribution under its conditions.

PR #2848, opened 2026-09-11
input: map tablets to an output on systems without an internal panel

MarkWind85:
My Huion Drawing tablet does not move the cursor on my desktop, and it turns out the events never
make it out of the input handlers.

## Problem

`mapped_output_for_device` decides which output an absolute device maps onto.
With no per-device `map_to_output` set it falls back to `Shell::builtin_output()`,
which only matches `eDP-`/`LVDS-`/`DSI-` connectors. A desktop has none of those,
so it returns `None` and the `else { return }` arms in `TabletToolAxis` and
`TabletToolProximity` drop every event.

What makes this a pain to track down is that nothing else is broken. The
compositor opens the evdev node and emits `zwp_tablet_v2` perfectly well —
XWayland even builds stylus/eraser/cursor devices out of it — so there's no error
message anywhere, the tablet is just silently dead.

## Fix

Fall back to the first output for tablet tools. Touch keeps what it does now:
assuming the built-in panel is fair for a laptop touchscreen, and working out
which output an *external* touchscreen belongs to is a separate problem — the
TODO above the function already asks about it.

This is a no-op on any machine that has an enabled internal panel, since
`builtin_output()` returns `Some` there and the new fallback never runs. It also
covers a docked laptop with the lid closed, where the `eDP-` output has been
removed from the set and tablets are dead for the same reason.

## What this doesn't do

It doesn't try to answer which output a tablet *should* map to. That needs the
settings work in #141, and libwacom so a display tablet can map to itself —
nothing available here tells an opaque tablet apart from a Kamvas. All this does
is stop the events being dropped when there's no answer at all. Whether that
partial step is worth taking is your call.

## AI/LLM disclosure

Done with the help of an LLM coding agent, directed and reviewed by me, and
disclosed in the commit message. I understand the change in full — it's one
`.or_else` behind a two-line wrapper, with two call sites pointed at it — and
I'll handle review comments myself.

## Testing

Huion H1060P on Pop!_OS 24.04 (COSMIC), two external outputs (3440x1440 and a
rotated 2560x1440), no internal panel, built at master (a557859). With no
`input_devices` config the cursor doesn't move at all. Running this branch on a
bare VT with that same empty config, it does.

Context: #313 and pop-os/cosmic-epoch#2826. #1495 landed on this same line and was
closed with the config workaround — this makes the unconfigured case work.

## Checklist

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -1594,7 +1594,7 @@ impl State {
                     self.common.idle_notifier_state.notify_activity(&seat);
                     notify_cursor_activity(self, &seat);
                     let Some(output) =
-                        mapped_output_for_device(&self.common.config, &shell, &event.device())
+                        mapped_output_for_tablet(&self.common.config, &shell, &event.device())
                             .cloned()
                     else {
                         return;
@@ -1698,7 +1698,7 @@ impl State {
                     self.common.idle_notifier_state.notify_activity(&seat);
                     notify_cursor_activity(self, &seat);
                     let Some(output) =
-                        mapped_output_for_device(&self.common.config, &shell, &event.device())
+                        mapped_output_for_tablet(&self.common.config, &shell, &event.device())
                             .cloned()
                     else {
                         return;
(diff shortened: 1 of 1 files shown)


MarkWind85:
Closing this — #2287 already covers the same routing problem and more (focused/active output rather than first, plus pointer frames and tip-down focus), and it's approved. I should have checked open PRs before opening this one; apologies for the noise.

For what it's worth, #2287 currently shows as conflicting and hasn't been touched since May. Happy to take the rebase if that would help.

PR #2849, opened 2026-09-12
cursor: Replace the global idle-hide timeout with contextual hiding

herder:
Replaces the global `cursor_hide_timeout` with contextual cursor hiding, per the UX feedback on [pop-os/cosmic-settings#2088](https://github.com/pop-os/cosmic-settings/pull/2088): each trigger has an unambiguous restore gesture, so hiding reads as intentional rather than as losing the cursor.

These are not three features. All of them share one mechanism — the cursor is hidden, pointer input reveals it — and differ only in what *arms* the hide and after how long. So the existing state machine gained more arming sources and a context-dependent delay rather than three parallel code paths.

| trigger | arms on | delay | reveals on |
| --- | --- | --- | --- |
| idle | no pointer activity | `idle_timeout` | any pointer event |
| fullscreen idle | no pointer activity, fullscreen on the seat's active output | `fullscreen_idle_timeout`, which only ever *shortens* `idle_timeout` | any pointer event |
| while typing | a real keystroke | immediate | any pointer event |
| after touch | touch down | immediate | **pointer motion only** |

**Config.** `cursor_hide_timeout: Option<u32>` becomes `cursor_hide: CursorHideConfig { idle_timeout, fullscreen_idle_timeout, while_typing, after_touch }`. Grouped rather than four flat keys because cosmic-config stores one file per top-level key, so flat keys would put four interacting settings in four unrelated-looking files.

**Migration.** The rename orphans anyone's existing `cursor_hide_timeout` file, and the recipe for setting it by hand circulated in [pop-os/cosmic-settings#603](https://github.com/pop-os/cosmic-settings/issues/603). A startup shim reads the legacy key once when the new one is absent and seeds `idle_timeout` from it; the old file is left on disk so a downgrade still finds it.

**Touch reveals only on motion.** A cursor hidden by touch waits for real pointer movement; one hidden by idle or typing is also revealed by a click or scroll, so the user is never forced to click blind. That is why `hidden` carries a reason rather than being a bool.

**Typing excludes bare modifiers.** COSMIC binds Super+drag to move windows, so hiding on Super-down would take the cursor away exactly as the user reaches for it. A bare modifier press changes the modifier state; a real keystroke does not. Super+1 still counts as typing.

**The arming rule, which is the subtle part.** `InputEvent::PointerMotion` calls into the cursor code while holding a `shell.write()` guard, so arm-time code cannot read fullscreen state without self-deadlocking. The timer is therefore armed pessimistically — the shortest timeout that could apply in any context, minus the time already elapsed since the last *pointer* event — and the context is evaluated inside the calloop callback, which holds no lock. An early fire simply re-arms for the remainder. This is also why no hook is needed in any of the six fullscreen transition sites in `shell/mod.rs`.

## Questions for reviewers

**Defaults are a UX decision, not mine.** Currently `after_touch: true` and `fullscreen_idle_timeout: Some(3)`, on the reading that @maria-komarova described touch and fullscreen as behaviours and only typing as a toggle. That turns on new behaviour for every user. `while_typing: false` and `idle_timeout: None`. Happy to ship whatever UX prefers — the implementation is identical either way.

**A known gap, stated rather than hidden.** An application that enters fullscreen with no input at all, while the pointer is already idle and `idle_timeout` is `None`, keeps its cursor until the next input of any kind. Entering fullscreen by clicking a player's button or by pressing F11 both arm correctly. The alternative was either a repeating idle wakeup or hooking six fullscreen transition sites that a seventh could silently miss. Happy to revisit if you would rather have the hooks.

**A small cost worth naming.** Under the default config on a windowed desktop, a keystroke after the pointer has been still arms a zero-delay timer that resolves to "nothing applies" on the next loop iteration. Bounded by the key-repeat rate. It is not pure waste — it is the same mechanism that makes F11-into-fullscreen hide immediately rather than a timeout late — but it is new work on the input path, so it should be a decision rather than a surprise.

## Notes

- Renaming the public `cursor_hide_timeout` field is a breaking change to `cosmic-comp-config`, which cosmic-settings consumes. Nothing in-tree references it and the settings UI PR was declined, so the blast radius is nil.
- The migration writes to the user's config during `Config::load()`, which re-enters through the `ConfigWatchSource` inserted a few lines earlier. Benign: the in-memory value is assigned before the write, so the reload's inequality check makes it a no-op.
- Two dev-dependencies are declared (`tempfile`, `ron`). Both crates were already in `Cargo.lock` — `ron` is already an optional dependency of `cosmic-comp-config` itself — so no new crates enter the dependency graph.

## Testing

19 unit tests across the two crates, covering the timeout resolution and arming arithmetic, the reveal rule, the typing predicate, config round-tripping with missing fields, and all four migration cases.

Verified on a live KMS session (a nested backend does not exercise the cursor plane):

- **Idle** hides at the configured delay and any pointer event reveals.
- **Fullscreen** hides at its own, shorter delay. Tested with the two timeouts set to different values, because with both at the same value the two triggers are indistinguishable by observation.
- **Typing** hides on a keystroke; a bare modifier press does not, so Super+drag keeps its cursor.
- **Touch** hides, and — the part that matters — a scroll afterwards does *not* reveal while pointer motion does, so the hide reason is genuinely honoured rather than the cursor being hidden generically. **Verified with a synthetic `uinput` touchscreen** (`INPUT_PROP_DIRECT`, tagged `ID_INPUT_TOUCHSCREEN` by udev), not real touch hardware, which this machine lacks. Worth a second pair of eyes from anyone with a panel.
- **The arming rule**, which the unit tests cannot pin end to end: with `idle_timeout: Some(10)` and `while_typing: false`, typing continuously hid the cursor 10 s after the last *pointer* event rather than being postponed indefinitely by the keystrokes.
- **The migration**, on a config that genuinely predated the rename: a hand-set `cursor_hide_timeout: Some(3)` was carried into `cursor_hide.idle_timeout` at login, with the legacy file left in place.

---

### DIFF
--- a/cosmic-comp-config/src/lib.rs
+++ b/cosmic-comp-config/src/lib.rs
@@ -3,6 +3,7 @@
 use cosmic_config::{CosmicConfigEntry, cosmic_config_derive::CosmicConfigEntry};
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
+use std::time::Duration;
 
 use crate::input::TouchpadOverride;
 
@@ -106,8 +107,8 @@ pub struct CosmicCompConfig {
     pub edge_snap_threshold: u32,
     pub accessibility_zoom: ZoomConfig,
     pub appearance_settings: AppearanceConfig,
-    /// Hide the cursor after this many seconds of pointer inactivity (None disables)
-    pub cursor_hide_timeout: Option<u32>,
+    /// When the cursor hides itself: idle, fullscreen idle, typing, touch
+    pub cursor_hide: CursorHideConfig,
     /// Briefly magnify the cursor when the pointer is shaken, to help locate it
     pub cursor_shake_to_find: bool,
     pub activation_policy: ActivationPolicy,
@@ -148,7 +149,7 @@ impl Default for CosmicCompConfig {
             edge_snap_threshold: 0,
             accessibility_zoom: ZoomConfig::default(),
             appearance_settings: AppearanceConfig::default(),
-            cursor_hide_timeout: None,
+            cursor_hide: CursorHideConfig::default(),
             cursor_shake_to_find: true,
             activation_policy: ActivationPolicy::default(),
             decoration_preference: DecorationPreference::default(),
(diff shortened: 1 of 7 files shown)


cosmic-comp-config/src/lib.rs
```diff
@@ -263,3 +337,103 @@ pub enum XwaylandDescaling {
...
+    }
+
+    #[test]
+    fn defaults_match_the_spec() {
+        let cfg = CursorHideConfig::default();
+        assert_eq!(cfg.idle_timeout, None);
+        assert_eq!(cfg.fullscreen_idle_timeout, Some(3));
+        assert!(!cfg.while_typing);
+        assert!(cfg.after_touch);
+    }
```
Drakulix, line 446: Please drop this test. A default implementation is good enough, this is needlessly verbose.

src/backend/render/cursor.rs
```diff
@@ -837,54 +866,129 @@ pub fn draw_cursor<R>(
...
+    }
+}
+
+/// Whether the seat's active output is showing a fullscreen surface. Only ever
+/// called from the timer callback, which holds no shell lock.
+fn fullscreen_focused(state: &State, seat: &Seat<State>) -> bool {
+    let shell = state.common.shell.read();
+    shell
+        .active_space(&seat.active_output())
+        .is_some_and(|workspace| workspace.get_fullscreen_surfaces().next().is_some())
```
Drakulix: Hmm, I don't think this gives us quite what we want. Fullscreen surfaces can be renderered behind windows in which case we probably don't want to logic to be active. We should probably check if the currently focus target of the seat is a fullscreen surface instead.

cosmic-comp-config/src/lib.rs
```diff
@@ -436,4 +439,27 @@ mod test {
...
+        // A file written before a field existed must keep the rest of the user's
+        // settings rather than resetting the whole key.
+        let partial: CursorHideConfig = ron::from_str("(idle_timeout: Some(5))").unwrap();
+        assert_eq!(
+            partial,
+            CursorHideConfig {
+                idle_timeout: Some(5),
+                ..CursorHideConfig::default()
+            }
+        );
```
Drakulix: Please also drop this test. We don't need to test functionality of `#[serde(default)]` and any config struct changes (not replacements like in this PR) are reviewed very carefully.

(please don't forget to drop the dev-dependency alongside it.)

src/backend/render/cursor.rs
```diff
@@ -914,32 +918,35 @@ pub fn refresh_idle_timer(state: &State, seat: &Seat<State>) {
...
+        }
+        // `idle_timer` stays `None`, so the next activity is unthrottled and retries.
+        Err(err) => warn!(?err, "Failed to arm the cursor idle-hide timer"),
     }
 }
 
 /// Whether the seat's active output is showing a fullscreen surface. Only ever
 /// called from the timer callback, which holds no shell lock.
-fn fullscreen_focused(state: &State, seat: &Seat<State>) -> bool {
+fn fullscreen_on_active_output(state: &State, seat: &Seat<State>) -> bool {
```
Drakulix: So other review comment on earlier commit, we probably want to remove the rename as well.

Drakulix (commented):
Thanks, this was surprisingly reviewable for being largely generated. (Which I am sure says more about the person submitting this, than the model used.)

Still a bit more verbose in places than I would like, see comments. (Which I hope are actually displayed in the right spots, given I reviewed this commit-by-commit. They certainly aren't in github's review drafting interface...)

Also please just drop aa2c496 altogether, the tests are painfully obvious and the code extractions needlessly verbose in my opinion.

In a complex and large codebase like this, which needs to be understandable and maintainable by humans brevity and locality of information is king. The unit tests are welcome, but can only ever serve as smoke tests and thus should be kept for actual error cases.

Functionality-wise this looks good. Addressing your questions:

> Defaults are a UX decision, not mine. [...]

Seems good to me. They can always be tweaked after, but to be on the safe side: Ping @maria-komarova.

> A known gap, stated rather than hidden. [...]

This is fine and a sufficiently unlikely cause of events.

> A small cost worth naming. [...]

Likewise fine. We already incur a small code on pointer-events for the idle-timeout and now that we handle keyboard events as well, this was to be expected.

cosmic-comp-config/src/lib.rs
herder: Not sure how many of these tests you wanted to drop - should I drop only the last one here, or all of them?

herder:
Thanks for the review @Drakulix!

I addressed your changes and dropped the code extraction commit. Not sure how much of the test code you wanted dropped - happy to drop more of them if that's what you meant :)

cosmic-comp-config/src/lib.rs
Drakulix: I mean the last one, thx!

Drakulix (approved):
LGTM now. Much simpler in a lot of places, thanks for taking the time!

PR #2858, opened 2026-09-14
Don't wake displays on input device removal, and count input wakes as activity

### LINKED ISSUE #2854: Displays turn back on after idle blanking when an input device disconnects, and never blank again
After the idle daemon blanks the screen, an external monitor sometimes comes back on by itself and then stays on until someone touches the mouse or keyboard. On my setup this happens about 30 minutes after the last input, which turns out to be a Bluetooth keyboard going to sleep and dropping its link. Its input device is removed, and that alone is enough to light the monitor.

`process_input_event` in `src/input/mod.rs` starts with

```rust
crate::wayland::handlers::output_power::set_all_surfaces_dpms_on(self);
```

That runs for every `InputEvent`, including `DeviceAdded`, `DeviceRemoved`, `SwitchToggle` and `Special`. Those arms never call `idle_notifier_state.notify_activity`, so the outputs come on while the ext-idle-notify notification is still idled. cosmic-idle never receives `resumed`, so it never arms again and never sends Off. It also ignores the `zwlr_output_power_v1.mode` event that `OutputPowerState::refresh` sends. The screen stays lit until real input arrives.

It is easy to miss in logs, because removing an input device leaves no kernel log line. The only trace is the modeset that relights the CRTC, which then looks like it came from nowhere. An idle-notify client registered next to cosmic-idle stays idle the whole time, which shows the compositor never counted it as activity.

The same thing happens with a dock re-enumerating its USB hubs, a receiver resetting, or anything else that adds or removes an input device after the screen has blanked. It is related to pop-os/cosmic-greeter#19 and is the same stranded idle state that #2148 works around.

### Steps to reproduce

1. Let the idle timeout blank an external display, or run `wlopm --off <output>`.
2. Remove an input device without touching it. The simplest way is `echo 0 | sudo tee /sys/bus/usb/devices/<mouse>/authorized`, or wait for a Bluetooth keyboard to go to sleep.
3. The display turns on within a second. Run `wlopm` and it reports `on`. The display does not blank again until real input.

### Suggested fix

Don't wake displays for hotplug, and make any wake that does happen count as activity:

```diff
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
-        crate::wayland::handlers::output_power::set_all_surfaces_dpms_on(self);
+        if !matches!(
+            &event,
+            InputEvent::DeviceAdded { .. }
+                | InputEvent::DeviceRemoved { .. }
+                | InputEvent::Special(_)
+        ) {
+            crate::wayland::handlers::output_power::set_all_surfaces_dpms_on(self);
+        }
--- a/src/wayland/handlers/output_power.rs
+++ b/src/wayland/handlers/output_power.rs
     if changed {
+        let seat = state.common.shell.read().seats.last_active().clone();
+        state.common.idle_notifier_state.notify_activity(&seat);
         OutputPowerState::refresh(state);
     }
```

`SwitchToggle` keeps waking on purpose, so opening the lid still lights the screen. The `notify_activity` call only runs when an output actually went from off to on, so the input path costs nothing extra. A client calling `set_mode(On)` through wlr-output-power is left alone, since cosmic-idle does that itself whenever it rebuilds its timers.

Compared to #2148, this does not count a device being plugged or unplugged as the user being present, so a dock or a sleeping keyboard can't keep the session from idling or locking.

Seen on `5c93094`, and the code is unchanged on current master (`a5578599`). Intel xe, external display over USB-C, COSMIC on Pop!_OS 24.04.

dev-goalz:
Fixes #2854.

After the idle daemon blanks the screen, a monitor could come back on by itself and then stay on until someone touched the mouse or keyboard. The common trigger is a Bluetooth keyboard going to sleep and dropping its link, which removes its input device. This issue keeps the monitor active for full nights when I'm a a sleep.

Two things combine here. `process_input_event` powers all outputs on for every input event, including device removal, which is not user activity. And for device events, and for the lid switch, nothing tells the idle notifier, so the ext-idle-notify notification stays idled. cosmic-idle never gets `resumed`, never arms its timer again, and the screen stays lit.

This change skips the wake for `DeviceRemoved` and `Special`. `set_all_surfaces_dpms_on` now calls `notify_activity` on the last active seat when an output actually went from off to on, so any wake that does happen resets idle and the screen blanks again after the normal timeout. That call only runs on a real off to on transition, so input on a lit screen does not reach it. A client setting the mode through wlr-output-power does not go through this function and is left alone.

`DeviceAdded` and the lid switch still wake on purpose. Opening the lid lights the screen as before, and after a system resume the re-enumerated devices can be the only input the compositor sees, so skipping additions could leave the screens dark. A stray addition, such as a dock re-enumerating, now lights the screen for one idle timeout instead of indefinitely.

### How it was checked

Running on a laptop with an external display since 2026-09-13. With the screen blanked by idle, a Bluetooth keyboard going to sleep used to relight it within a second, with an idle-notify client next to cosmic-idle staying idle the whole time. With the change the display stays off. Moving the mouse still wakes it, the idle client goes active in the same second, and the display blanks again one timeout later without further input. Opening the lid on a blanked display wakes it and resets idle the same way, and closing the lid again on a lit display does not count as activity.

Unplugging a USB mouse while the screen is blanked leaves it off. Plugging it back in wakes the screen and resets idle, and the screen blanks again one timeout later. The keyboard and lid runs above were done with an earlier version that also ignored device additions. That part of the code path is unchanged.

---


leviport:
AI disclosure needs to be in commit messages, as the checklist states.

dev-goalz:
@leviport I'm not sure if you're a bot, copy-pasting, or just a bit short on social skills, so I'm happy to help you rephrase your comment:

"Hi @dev-goalz, thanks for submitting a fix that will clearly help other users too! This is a long-standing issue that nobody has been able to reproduce so far, so it's great that you were able to pinpoint it. Before we can review or merge this, could you please update the commit message to mention that you used AI, as you noted in the PR description? We'll consider your PR once that's done. Thanks!

dev-goalz:
Sure @leviport, sorry this is my first contribution (out of 3 open PR's). Was not aware this is a requirement. Will fix this message and force-push the fix. 

Hope this PR is to your quality standards and can be merged :-)

dev-goalz:
Done, the disclosure is in the commit message now. Force-pushed, the diff is unchanged.

Drakulix (approved):

jackpot51:
> @leviport I'm not sure if you're a bot, copy-pasting, or just a bit short on social skills, so I'm happy to help you rephrase your comment:

This is an unacceptable comment. You are not free to ignore our requirements or to insult others when you do not like them. These requirements are mandatory, as is following our code of conduct.

dev-goalz:
> > @leviport I'm not sure if you're a bot, copy-pasting, or just a bit short on social skills, so I'm happy to help you rephrase your comment:
> 
> This is an unacceptable comment. You are not free to ignore our requirements or to insult others when you do not like them. These requirements are mandatory, as is following our code of conduct.

I'm not trying to insult anyone, I tried to make a small statement with humor that the communication style here is not community friendly.  The way you communicate as follow up does not make it better.

Maybe we are from a very different culture and we therefore don't understand eachother correctly. I'm sorry if that is the case, was not intentionally to be rude.

Let's keep the comments from here just about the code 🙏

leviport (approved):
Beep boop. This does indeed fix https://github.com/pop-os/cosmic-comp/issues/2854, and I'm not seeing any regressions. Thanks for the fix!

PR #2859, opened 2026-09-14
only cancel grabs the compositor started for its own interactions

hojjatabdollahi:
Currently if you press `esc` while a popup is open, the popup loses grab silently, but it stays open.

For example:
- open cosmic-files
- right click on a directory
- press `esc` once, the popup loses keyboard focus, but stays open
- now use the arrow keys to change the selected file
- if you press `esc` again, that event is sent to cosmic-files that handles it and closes the popup.

This PR sends `ungrab` to the popups with grab when `esc` is pressed, so they're properly dismissed.

We already do this for clicking outside the popup in Smithay: https://github.com/Smithay/smithay/blob/22571baa20d34d71092942dbb520c4e3fbbd6263/src/desktop/wayland/popup/grab.rs#L613

___


hojjatabdollahi:
Thinking about it, the popup itself should handle its own `esc`. I don't understand why cosmic-comp even calling `unset_grab` for the pointer and the keyboard whenever there is a popup.

Wouldn't that cause issues for any program that grabs the cursor for example? Like in an app that needs you to press and hold the mouse 1 down or virtual machines? It looks like that we only call this if swap mode is on, or a popup is open.

I don't think popups should be on the allow list. Looks like they were added by accident?! https://github.com/pop-os/cosmic-comp/pull/1126

Also there is an allow list for keyboard, but not the pointer. If pointer grab is on, then esc is swallowed.

hojjatabdollahi:
I think https://github.com/pop-os/cosmic-comp/pull/2859/commits/ecffa6c10317db9323a0f5a1d5931f7f44c8eef6 makes more sense.
Previously pointer didn't have any condition, now it only unsets grab if it's one of the 3 interactions it starts (`Resize`, `Move`, and `Menu` grab) and keyboard used to also apply to popups, but now it's only `SwapWindowGrab`.

I kept the previous commit, just in case that's preferred.

wash2:
https://github.com/pop-os/cosmic-comp/commit/ecffa6c10317db9323a0f5a1d5931f7f44c8eef6 makes most sense to me as well.

Drakulix:
I agree, could you remove the first two commits from the PR and only keep the new fix?

hojjatabdollahi:
Done.

Drakulix (approved):
Thx!

jklgrasso (commented):
Pressing `esc` after opening a context menu for a folder now causes `cosmic-files` to close.

https://github.com/user-attachments/assets/062ce7af-d75c-4fa2-809a-bf40bb1f892b

hojjatabdollahi:
@jklgrasso I'm guessing cosmic-files is panicking and not closing. Let me see if I can repro the issue.

hojjatabdollahi:
@jklgrasso Cosmic-files was crashing, but that should already be fixed since https://github.com/pop-os/cosmic-files/pull/2071 got merged.

Now if you press esc it shouldn't crash, but it does change the menu, which will be fixed in https://github.com/pop-os/cosmic-files/pull/2078.

jklgrasso (approved):

PR #2866, opened 2026-09-17
fix: unmaximize size after moving maximized window between workspaces

### LINKED ISSUE #2865: Unmaximizing maximized windows that have been moved to a different workspace or output
When moving maximized windows between workspaces or outputs in floating mode, the original size of the window before maximizing is lost. This means that if you unmaximize the window, it gets its pre-maximized position, but the size stays the full maximized size instead of the original size. It should return to the size it was before it was maximized like it does if the window isn't moved to a different workspace or output first.

jasonrhansen:
Fixes bug where in floating mode, when a window was moved to another workspace or output, then unmaximized, the window wouldn't return to its original size that it was before being maximized. Instead it would return to its original position, but keep its maximized size.

Closes #2865


src/shell/mod.rs
```diff
@@ -3560,12 +3560,8 @@ impl Shell {
                 ),
                 _ => (None, false, None),
             };
-            to_workspace.floating_layer.map(mapped.clone(), position);
+            let geometry = to_workspace.floating_layer.map(mapped.clone(), position);
             if was_maximized {
-                let geometry = to_workspace
-                    .floating_layer
-                    .element_geometry(mapped)
-                    .unwrap();
```
Drakulix, line 3568: I fail to see how this behaves any different from the old code. `element_geometry` should return `win_geo` as long as it is called before `map_maximized`, no?

Drakulix (commented):
Thanks for looking into this, in the future I would really appreciate if you tried to separate refactorings into separate commits for easier review. I unfortunately fail to understand how this fix works.

src/shell/mod.rs
jasonrhansen: For some reason the old code gets the old maximimized geometry that it had before it was unmapped to move to another workspace. I'm not sure why that is the case, but it means when you unmaximize on the new workspace it keeps the maximized size.

Drakulix: Well, I guess then I'll have to attach a debugger here at some time. I am not going to merge this until we figured out the root cause and thus are able to author a more targeted patch.

jasonrhansen: That's completely understandable if you feel like this solution is just band-aiding over an issue that should be fixed elsewhere. But I'd really like to get this fixed because it's been bothering me for quite some time now. I'd like to spend some more time with it to really try to understand why it's not behaving like one would expect.

